Issue #29, opened 2016-06-15
replace timeout and timeoutUnit with Duration


helmbold:
breaking change, needs to got to 2.0.0

sksamuel:
I didn't know there was a kotlin equiv of scala's duration. Or is it a java thing ?

helmbold:
Wait for my pull request ... ;-)

PR #30, opened 2016-06-15
Replace timeout + timeUnit with Duration

helmbold:
The changelog style follows https://github.com/olivierlacan/keep-a-changelog . #29

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,32 @@
+Changelog
+=========
+
+This project follows [semantic versioning](http://semver.org/).
+
+Version 2.0.0, released YYYY-MM-DD TODO
+----------------------------------
+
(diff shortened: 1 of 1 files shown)


sksamuel:
Awesome. I might change the release version to 1.3 rather than 2.0 as that's still a major version. I know that's not strictly semver but it seems to be what most JVM projects do, eg Scala, Spark, Akka etc. What do you think?

helmbold:
I think that many projects don't use versioning properly. The version numbers lose there expressiveness. And today version numbers are actually pretty cheap ;-) Let's follow semantic versioning.

Have you noticed that the build fails? I try to fix it (what is not optimal without beeing able to compile it under windows).

sksamuel:
Yes I noticed, I will attempt a fix in a minute. I'm on windows atm and
need to reboot into linux.

helmbold:
I tried to fix it and created another pull request. #31

sksamuel:
Can't merge it - conflicts.

PR #31, opened 2016-06-15
Replace timeout + timeUnit with Durationl, fix code

helmbold:


helmbold:
It should work now.

helmbold:
Everything should work now!

Issue #52, opened 2016-07-02
FreeSpec incorrectly labelled as FlatSpec in README.md


Hikaru755:
What should be the FreeSpec example in the README.md is incorrectly titled as FlatSpec, and the description as well as the code sample both show FlatSpec as a name.

As a newcomer to this library this gave me a hell of a time figuring out why the IDE would flag the minus sign even when I just copied and pasted the example code, until I finally noticed that FlatSpec had two conflicting sample sections and FreeSpec had none.

PR #53, opened 2016-07-02
Update README.md

Hikaru755:
This would fix the incorrect labeling of Free Spec as Flat Spec, as reported in issue #52.


sksamuel:
:+1:

PR #55, opened 2016-07-03
Always call fireTestFinished and notify from main thread only.

mfriedenhagen:
* fireTestFinished needs to be called always.
* Call all notifier events from the main thread. This
  is needed because build tools might depend on `ThreadLocals`.
* Addition to kotlintest/kotlintest#42.


sksamuel:
Great thanks.

PR #57, opened 2016-07-03
Depend on mockito-core only.

mfriedenhagen:
`kotlintest` only uses stuff from `mockito-core`. Otherwise e.g.
when you use `maven-enforcer-plugin` and define a rule which
forbids duplicate classes there is a clash because `mockito-all`
defines some matchers itself.


sksamuel:
:+1:

PR #61, opened 2016-07-05
Allow to write tests directly in the class constructors.

hastebrot:
It allows to write tests into the class constructor. Writing tests in the `init()` block is still supported.

~~~kotlin
class StringSpecExample : StringSpec({

    "strings.size should return size of string" {
        "hello" should haveLength(5)
    }
})
~~~

### DIFF
--- a/src/main/kotlin/io/kotlintest/specs/BehaviorSpec.kt
+++ b/src/main/kotlin/io/kotlintest/specs/BehaviorSpec.kt
@@ -5,7 +5,8 @@ import io.kotlintest.TestCase
 import io.kotlintest.TestSuite
 import java.util.*
 
-abstract class BehaviorSpec : TestBase() {
+abstract class BehaviorSpec(body: BehaviorSpec.() -> Unit = {}) : TestBase() {
+  init { body(this) }
 
   var current = root
 
--- a/src/main/kotlin/io/kotlintest/specs/FeatureSpec.kt
+++ b/src/main/kotlin/io/kotlintest/specs/FeatureSpec.kt
@@ -5,7 +5,8 @@ import io.kotlintest.TestCase
 import io.kotlintest.TestSuite
 import java.util.*
 
-abstract class FeatureSpec : TestBase() {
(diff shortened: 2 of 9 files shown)


hastebrot:
Tests are incomplete for this. I've added `StringSpecConstructorTest`, which introduces an additional test class to `StringSpecTest.kt`. Constructor tests for other spec classes also need to be written. Maybe there is a better way to write the internal tests without introducing additional classes.

helmbold:
It looks nice, but I wonder whether it is a good idea to introduce another way to write it and to introduce some variance for it. I think the intended way is the init block which we already have. It's a matter of taste but I'm a fan of the python mantra "there should be one, and preferably only one, obvious way to do it.

What do you think, @sksamuel? 

But in any case, thanks for the effort.

sksamuel:
I probably wouldn't use this style, but it does look nicer than the horrible init block (https://discuss.kotlinlang.org/t/init-block-for-constructor-body/1703/8). I don't see any harm in having the option? If you're a fan of pythons mantra, then we should remove all *Spec except one :)

helmbold:
Maybe you're right: compared to the nearby testing styles it is not a big difference to introduce this syntactical variant. And if it helps to make some more developers to like Kotlintest, why not? 

And by the way: string spec would be enough for me ;-)

hastebrot:
Maybe this test is better:

~~~kotlin
"constructor should receive tests" {
  val spec = object : StringSpec({
    this should be a StringSpec::class
  }){}

  spec.run(RunNotifier())
}
~~~

sksamuel:
Lets merge. @hastebrot if you want to push new test and we can merge.

helmbold:
Let's merge it to 1.4, since it is a new feature and not a bug fix.

sksamuel:
Agreed, but do we want to go to 2.0 now since you've persuaded me to change the one instance per test option

sksamuel:
@hastebrot please reopen this PR against the release/2.0 branch

helmbold:
@hastebrot could please mention the new syntax in the readme?

sksamuel:
And changelog

PR #64, opened 2016-07-06
added Tests for Generators

reckter:
added Tests for Generators and implemented a fix for the choose Generator, that didn't fully comply with the given ranges.

Gen's didn't have any test for them, so a few bugs slipped through.
We Implemented tests for gens (We're not claiming they're complete, but better then nothing.

Also applied a few fixes for the implemented tests.


sksamuel:
Awesome

PR #65, opened 2016-07-06
Readme: Add constructor tests example.

hastebrot:


hastebrot:
Changelog entry would be

~~~
### Added

* You can write tests alternatively into a lambda parameter in the class constructor, eg:

```kotlin
class StringSpecExample : StringSpec({
  "strings.size should return size of string" {
    "hello".length shouldBe 5
    "hello" should haveLength(5)
  }

  "strings should support config" {
    "hello".length shouldBe 5
  }.config(invocations = 5)
})
```
~~~

sksamuel:
:+1:

PR #68, opened 2016-07-07
Switched to 'java.util.Random'

superbobry:
My original intention was to modify `Gen.string` to support arbitrary alphabet, but this required modifying a `Random` wrapper from `koors`. After some investigation I came to the conclusion that `com.sksamuel.koors.Random` does not provide any additional benefit compared to `java.util.Random`, thus this PR.

The only downside to the private-global random is that the user has no way of seeding it. I think the ultimate solution would be to expose the random generator as part of `TestConfig`, but this is better done in a separate PR.

`Gen.string` modification is to follow.

Quick question: Kotlin has a concise [style guide](https://kotlinlang.org/docs/reference/coding-conventions.html) which among other things mentions 4 spaces for indentation. Is there a reason for kotlintest using 2 spaces?

### DIFF
--- a/src/main/kotlin/io/kotlintest/properties/Gen.kt
+++ b/src/main/kotlin/io/kotlintest/properties/Gen.kt
@@ -1,9 +1,12 @@
 package io.kotlintest.properties
 
-import com.sksamuel.koors.Random
 import java.lang.reflect.ParameterizedType
 import java.lang.reflect.Type
 import java.lang.reflect.WildcardType
+import java.util.*
+
+/** A shared random number generator. */
+private val RANDOM = Random()
 
 interface Gen<T> {
   fun generate(): T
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/kotlintest/properties/Gen.kt
```diff
@@ -1,67 +1,70 @@
...
+          rand += max - min
         }
-        return rand + min;
+        return rand + min
       }
     }
 
     fun <T> oneOf(values: List<T>): Gen<T> = object : Gen<T> {
-      override fun generate(): T = Random.default.shuffle(values).first()
+      override fun generate(): T = values[RANDOM.nextInt(values.size - 1)]
```
superbobry: Using `shuffle` seemed a bit redundant for generating a single item, so I've replaced it with a simpler version.

helmbold:
To give your quick question a quick answer: @sksamuel decide to use only 2 spaces to avoid deep indentations with multiple levels of nesting.

sksamuel:
4 spaces is a Java thing that kotlin has inherited. 2 spaces is used in
Scala and makes much more sense when you have nested functions.

russel:
4 spaces is also a Python thing. 2 spaces is a C, C++ thing for many styles. Personally I am going with the Go thing of tabs for indents, 1 tab per indent and some number of spaces for rendering the tab. This actually is a way of pleasing all the people all the time, except you can't.

reckter:
This btw fixed a bug with Gen.oneOf(List) where it would only select the first value in the list. I added test for this in #81

helmbold:
Could this be closed?

superbobry:
@reckter this might indicate a bug in `shuffle`, because in theory `shuffle(...).first()` should produce a uniform distribution over the element of a given list.

PR #81, opened 2016-07-19
fixed that Gen.oneOf would not choose the last element and added a test for it.

reckter:
(if someone knows how to update my fork without creating a pull request every time, I would be flattered!)


reckter:
Maybee a release of this should be considered, as of right now the release version is broken. (it only ever chooses the first element of the given list)

sksamuel:
Create a remote back to the original repo, and then you can fetch that, `git fetch original` and then you can merge that into your fork master, `git checkout master; git merge original/master`

helmbold:
Could this be closed?

sksamuel:
It's a PR, its already closed :)

PR #87, opened 2016-08-05
Update Eventually section of README.md

dbeckwith:
Made it clear that the test class needs to extend the `Eventually` interface in order to use `eventually`.


sksamuel:
:+1:

PR #89, opened 2016-08-24
remove redundant reference to object instance when calling ext. function

memoizr:
Hey, loving your work on this lib. This is just a minor tweak which I consider to be more idiomatic Kotlin: passing the object instance as a parameter to the function is a bit util-function like, which is helpful when calling it from Java, but from Kotlin we can embrace the fact that it's an extension function to the full extent, and treat it as if it were a member function, hence we can omit `this`.

However, as this is just a style issue open to interpretation I understand if you'd rather keep it as it is. Hope this helps.


sksamuel:
:+1:

PR #93, opened 2016-09-22
Allow returning null from a function parameter of shouldThrow()

### LINKED ISSUE #90: shouldThrow should be able to accept a function that returns a nullable reference
Consider the following example:

```
fun toTest(string: String?): String? {
  return string?.substring(1)
}
```

The function above cannot be tested with `shouldThrow()` because it returns a nullable `String`.
```
shouldThrow<IndexOutOfBoundsException> { toTest("") }
// Type mismatch. Required: Any. Found: String?.
```

The compilation error can be fixed with `!!` but I think it would be better if kotlintest allowed such expressions.

MichaelRocks:
Fixes #90


sksamuel:
Great thanks. Will do a release with this.

Issue #99, opened 2016-10-02
Easy way to make a custom dsl for matchers


sksamuel:
I wanted some way of allowing people to easily add new keywords for the matcher dsl. 
Eg, if you wanted to do

`myValue should explode before 12`

Which translates to something like

`assert(myValue.getExplosionHour() < 12)`

I've come up with this:

- Add a keyword that comes after `should`. In this case `explode`
```kotlin
object explode : ShouldKeyword<explode>
```
- Add an infix extension method on `ShouldBuilder` which contains the next word and the matcher 
```kotlin
infix fun ShouldBuilder<explode, Int>.before(hour:Int) :Unit {
 if (value >= expected)
    throw AssertionError("$value is not less than $expected")
}
```

Import the extension methods, the voila,

`myValue should explode before 12`

Works on ints, but

`myValue should explode before "12"`

Fails because we defined it on int only.

This can be extended to multiple words.

```kotlin
object distance

infix fun ShouldBuilder<have, String>.levenshtein(distance: distance) = ShouldBuilder<distance, String>(this.value)

infix fun ShouldBuilder<distance, String>.of(d: Int): Unit {
  // test impl
}

"hello" should have levenshtein distance of 13
```

What do you think @helmbold

helmbold:
The result looks good, but I would use this feature very rarely. I would just use default matchers, because I'd be too lazy to build my own matcher DSL. But for developers with the desire to build their own DSL it is really helpful.

sksamuel:
I didn't think the average person would use it, but it is a nice way for people to add matchers back into KotlinTest, and for me and you, etc, to add them.

sksamuel:
I'm starting to think we should not bother with this and deprecate the existing 'string should start with "substring" syntax. Because Kotlin doesn't have infix precedence its not possible to support what we need in a way that's compatible with our other features.

If we say that every matcher is of the form

`value should matcher`

Then we can do

`"hello" should startWith("h")`
`"hello" shouldNot startWith("h")`
`"hello" should (startWith("h") or endWith("o")`
`"hello" should (startWith("h") and endWith("o")`
`"hello" shouldNot (startWith("h") or endWith("s")`

You can always do _fancier_ matchers with parenthesis.

`"hello" should (always startWith("h"))`
`list should (always be empty)`

wdyt

helmbold:
I'm a big proponent of simplification through unification. So, let's use `value should matcher`.

Issue #96, opened 2016-09-30
shouldNot


jmfayard:
A `shouldNot` keyword would be handy

Examples

```kotlin
class NotTests : FeatureSpec() { init {
    feature("collections") {
        val data = (1..10).toList()
        scenario("not empty") {
            data shouldNot beEmpty()

        }
        scenario("does not contain") {
            val filtered = data.filter { it % 2 == 0 }
            filtered shouldNot contain(3)
        }
    }
    feature("strings") {
        scenario("endsWith") {
            val url = "http://example.com/api"
            url shouldNot endWith("/")
        }
    }


}
}

// make sure the tests compile & fail
infix fun <T> T.shouldNot(matcher: (T) -> Unit): Unit = matcher(this)
infix fun <T> T.shouldNot(matcher: Matcher<T>) = matcher.test(this)

```

sksamuel:
This is now part of the 2.0 branch.

PR #101, opened 2016-10-02
Feature/matcherdsl

sksamuel:
Implementation for #99


sksamuel:
I'm going to have to change the implementation to support #96 so I'll do more commits before this is ready.

sksamuel:
@helmbold docs added, `shouldNot` supported.

PR #107, opened 2016-10-12
Feature/#77 extensions

helmbold:
I decided to leave the `ProjectExtension` interface as is (and not split it), because the user had to take care of the order of separately registered `beforeAll` and `afterAll` methods otherwise, while the framework can easily manage this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,13 +338,33 @@ exception.message should start with "Something went wrong"
 Interceptors
 ------------
 
-TODO explain execution order or interceptors
-
 If you need to execute some logic before and/or after each test case, then you can use an interceptor. This is for example useful to cleanup a database after the test have run. 
+
+In the `ProjectConfig` (see below) you can override `beforeAll` and `afterAll` or add extentions with such methods to the ProjectConfig. Logic in theses methods will be executed before and/or after the first/last test of the project.
+
+In a spec class you can intercept the spec execution by overriding the `interceptors` property and providing a list of interceptors or by overriding `interceptSpec`.
+
+A single test case can be intercepted by overriding `interceptTestCase` or by providing a list of interceptors in the `defaultTestCaseConfig` or in the `config` of a test case.
 
 Interceptors replace `beforeEach`, `afterEach`, `beforeAll`, and `afterAll` functions from KotlinTest 1.x.
 
-### Intercepting a test case
+### Interceptor Execution Order
+
+There are several points where you can hook in the test execution. 
+
+* ProjectConfig.extentions beforeAll
+  * ProjectConfig.beforeAll
+    * Spec.interceptors
+      * Spec.interceptSpec
+        * test case
+      * Spec.interceptSpec (interceptor from above continued)
+    * Spec.interceptors (interceptors from above continued)
+  * ProjectConfig.afterAll
+* ProjectConfig.extensions afterAll
+
+The general philoshopy is here that the closer an interceptor is to a test case, the closer it is to the test case in the execution order. 
+
+### Intercepting a Test Case
 
 Override `interceptTestCase` in a spec class to provide logic that should be called before and after each test case. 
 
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -319,38 +321,200 @@ val exception = shouldThrow<IllegalAccessException> {
...
 
-If you need to run a method before each test and/or after each test (perhaps to reset some values to defaults etc), then simply override the `beforeEach` and `afterEach` methods in your test class, eg:
+Interceptors are just functions and can be reused between specs or even between projects. Just pass interceptors to the `config` on test case or spec level.
 
 ```kotlin
-override fun beforeEach() {
-  println("Test starting")
+"should do it correctly" {
+  ...
+}.config(interceptors = listof(myTestCaseInterceptor))
```
martin-g: s/listof/list**O**f/

src/main/kotlin/io/kotlintest/InvalidConfigException.kt
```diff
@@ -0,0 +1,6 @@
+package io.kotlintest
+
+/**
+ * Signals an invalid configuration
+ */
+class InvalidConfigException(message: String) : Exception(message)
```
martin-g: Extend from RuntimeException for better Java interop ?!

src/main/kotlin/io/kotlintest/Project.kt
```diff
@@ -0,0 +1,39 @@
...
+    val configClasses = Reflections("io.kotlintest").getSubTypesOf(ProjectConfig::class.java)
+    if (configClasses.size > 1)  {
+      val configClassNames = configClasses.map { config -> config.simpleName }
+      throw InvalidConfigException("Duplicate GlobalConfig found: $configClassNames")
+    }
+    projectConfig = configClasses.firstOrNull()?.kotlin?.objectInstance
+  }
+
+  fun beforeAll() {
+    synchronized(executedBefore) {
```
martin-g: Better use `AtomicBoolean#compareAndSwap()`

src/main/kotlin/io/kotlintest/TestBase.kt
```diff
@@ -30,19 +36,33 @@ abstract class TestBase : PropertyTesting(), Matchers, TableTesting {
...
+  protected open val defaultTestCaseConfig: TestConfig = TestConfig()
 
-  // the root test suite which uses the simple name of the class as the name of the suite
-  // spec implementations will add their tests to this suite
-  internal val root = TestSuite(javaClass.simpleName, ArrayList<TestSuite>(), ArrayList<TestCase>())
+  /**
+   * Interceptors that intercepts the execution of the whole spec. Interceptors are executed from
+   * left to right.
+   */
+  protected open val specInterceptors: Iterable<(TestBase, () -> Unit) -> Unit> = listOf()
```
martin-g: Since it is `protected` and `open` and the type is `Iterable` I believe a subclass can assign a Set as a value and then the promise to execute from left to right is broken.

README.md
helmbold: fixed

src/main/kotlin/io/kotlintest/Project.kt
helmbold: I think you meant `compareAndSet`? But since the result of the comparison is not useful here, I go with just `set(true)`.

src/main/kotlin/io/kotlintest/TestBase.kt
helmbold: You're right! I've changed it to `List`.

helmbold:
Thanks for the good review @martin-g !

martin-g:
Keep the good work, @helmbold & @sksamuel !

sksamuel:
Thank you @martin-g

helmbold:
@sksamuel Do you have any objections or can we merge this branch?

sksamuel:
Looks good. Merge away.

PR #123, opened 2016-12-18
Improvements to the Gen class

ColinHebert:
- Make `create()` a first class citizen for `Gen` creation
- Add the `nullable()` gen that allows the creation of nullable values
- Inline the `create()` function

### DIFF
--- a/src/main/kotlin/io/kotlintest/properties/Gen.kt
+++ b/src/main/kotlin/io/kotlintest/properties/Gen.kt
@@ -12,65 +12,30 @@ interface Gen<T> {
   fun generate(): T
 
   companion object {
-
-    fun choose(min: Int, max: Int): Gen<Int> = object : Gen<Int> {
-      override fun generate(): Int = RANDOM.nextInt((max.toLong() - min.toLong()).toInt()) + min
-    }
-
-    fun choose(min: Long, max: Long): Gen<Long> = object : Gen<Long> {
-      override fun generate(): Long {
-        var rand = (RANDOM.nextLong() % (max - min))
-        if (rand < 0) {
-          rand += max - min
-        }
-        return rand + min
-      }
-    }
-
-    fun <T> oneOf(vararg generators: Gen<T>): Gen<T> = object : Gen<T> {
-      override fun generate(): T = Gen.oneOf(generators.toList()).generate().generate()
-
-    }
-
-    fun <T> oneOf(values: List<T>): Gen<T> = object : Gen<T> {
-      override fun generate(): T = values[RANDOM.nextInt(values.size)]
-    }
-
-    fun string(): Gen<String> = object : Gen<String> {
-      override fun generate(): String = nextPrintableString(RANDOM.nextInt(100))
-    }
-
-    fun int() = object : Gen<Int> {
-      override fun generate(): Int = RANDOM.nextInt()
-    }
-
-    fun long() = object : Gen<Long> {
-      override fun generate(): Long = RANDOM.nextLong()
-    }
-
-    fun bool() = object : Gen<Boolean> {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/kotlintest/properties/Gen.kt
```diff
@@ -8,69 +8,34 @@ import java.util.*
 /** A shared random number generator. */
 private val RANDOM = Random()
 
-interface Gen<T> {
+interface Gen<out T> {
```
ColinHebert: This is always used as an `out` might as well make it explicit.

src/main/kotlin/io/kotlintest/properties/Gen.kt
```diff
@@ -8,69 +8,34 @@ import java.util.*
...
-    fun double() = object : Gen<Double> {
-      override fun generate(): Double = RANDOM.nextDouble()
-    }
-
-    fun float() = object : Gen<Float> {
-      override fun generate(): Float = RANDOM.nextFloat()
-    }
-
-    fun <T> create(fn: () -> T): Gen<T> = object : Gen<T> {
+    inline fun <T> create(crossinline fn: () -> T): Gen<T> = object : Gen<T> {
```
ColinHebert: Inlining the `create` function (but the `fn` one is executed in a different context, make it crossinline)

src/main/kotlin/io/kotlintest/properties/Gen.kt
```diff
@@ -8,69 +8,35 @@ import java.util.*
...
+    fun <T> oneOf(values: List<T>) = create { values[RANDOM.nextInt(values.size)] }
+    fun <T> oneOf(values: Array<T>) = oneOf(values.toList())
+    fun string() = create { nextPrintableString(RANDOM.nextInt(100)) }
+    fun int() = create { RANDOM.nextInt() }
+    fun long() = create { RANDOM.nextLong() }
+    fun bool() = create { RANDOM.nextBoolean() }
+    fun double() = create { RANDOM.nextDouble() }
+    fun float() = create { RANDOM.nextFloat() }
+    fun <T> list(gen: Gen<T>) = create { (0..RANDOM.nextInt(100)).map { gen.generate() } }
+    fun <T> set(gen: Gen<T>) = create { list(gen).generate().toSet() }
```
ColinHebert: `set()` uses `list()` as a base to generate the values.

src/main/kotlin/io/kotlintest/properties/Gen.kt
```diff
@@ -8,69 +8,35 @@ import java.util.*
...
+      }
+      rand + min
     }
 
-    fun <T> list(gen: Gen<T>): Gen<List<T>> = object : Gen<List<T>> {
-      override fun generate(): List<T> = (0..RANDOM.nextInt(100)).map { gen.generate() }.toList()
-    }
+    fun <T> oneOf(vararg generators: Gen<T>) = create { Gen.oneOf(generators.toList()).generate().generate() }
+    fun <T> oneOf(values: List<T>) = create { values[RANDOM.nextInt(values.size)] }
+    fun <T> oneOf(values: Array<T>) = oneOf(values.toList())
```
ColinHebert: Interact specifically with Arrays converting to a list automatically

helmbold:
Thanks for your effort. I think we should merge this better into 2.0 branch. See [branching model](https://github.com/kotlintest/kotlintest/wiki/contribute#branching-model).

sksamuel:
I don't like it when we ignore pull requests so my apologies for that @ColinHebert. This is certainly a long time outstanding. Now that 3.0.0 RC1 is out, we can look at your PR and see what we need to apply (the generators have changed somewhat in 3.0.0).

PR #122, opened 2016-12-16
Matcher support for any objects inheriting `Comparable<T>`

hhandoko:
Added support for matchers comparing two objects inheriting `Comparable<T>`.

Test conditions reads closer to natural language, e.g. `value1 should be gt value2` as opposed to `(value1 > value2) shouldBe true`.

PR #126, opened 2016-12-19
Matcher support for any objects inheriting `Comparable<T>`

hhandoko:
**NOTE:** A reimplementation for `release/2.0` branch from #122.

Added support for matchers comparing two objects inheriting Comparable<T>.

Test conditions reads closer to natural language, e.g. `value1 shouldBe gt(value2)` as opposed to `(value1 > value2) shouldBe true`.


src/main/kotlin/io/kotlintest/matchers/ComparableMatchers.kt
```diff
@@ -0,0 +1,49 @@
+package io.kotlintest.matchers
+
+@Deprecated("Use `value shouldBe lt(x)` or `value should beLessThan(x)`")
```
martin-g: What is the idea behind introducing a new class which methods are born deprecated ?

hhandoko: Primarily keeping to conventions (as like `Int` or `Long` matchers), as they have the same surface behaviours. Also with the added benefit of helping users coming from other libraries and/or `v1`.

sksamuel:
:+1:

hhandoko:
Thanks!

helmbold:
Do we really want to introduce *new* code annotated with `@Deprecated`?

hhandoko:
@helmbold I'm not fussed if it was removed. I just thought it would provide some context to new KotlinTest users since the surface behaviours are the same with existing matchers.

helmbold:
What do you think @sksamuel : should we relase a new version with deprecated methods? If I would be a solo developer I would remove them ...

sksamuel:
Probably take them out.

hhandoko:
OK, do you guys want me to remove it? If so, from which branch?

sksamuel:
Release 2.

hhandoko:
@sksamuel I just had a look at `release/2.0` branch and the `@Deprecated` methods for `Comparable<T>` matchers are not there. Did you already remove it?

helmbold:
The deprecated methods were removed in commit [41854941f095e95fcc7f2bb92dc665260fe8d4a8](https://github.com/kotlintest/kotlintest/commit/41854941f095e95fcc7f2bb92dc665260fe8d4a8) by Aleix Morgadas.

hhandoko:
OK, cool 👌

PR #127, opened 2017-01-01
Remove overhead in Gen.oneOf(vararg Gen)

Glease:
Previous implementation creates too much junk too easily. I just bumped into this piece of code when learning this framework. The newline at the EOF is added by Github web editor. I can't remove it without cloning the entire repo.


Glease:
My fault

PR #141, opened 2017-01-31
Added ability to pass annotations to JUnit`s description

saksmt:
For motivation see updates in README.md

Also as an example of test reporting framework you can see [allure](https://github.com/allure-framework/allure1)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ target
 
 # Gradle wrapper
 !gradle/wrapper/gradle-wrapper.jar
+classes/
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ is the same as the length of `a` plus the length of `b`. In this example KotlinT
 execute the test 100 tests for random String combinations.
 
 ```kotlin
-class PropertyExample: StringSpec() {
+class PropertyExample: StringSpec() { init {
 
   "String size" {
     forAll({ a: String, b: String ->
@@ -222,7 +222,7 @@ class PropertyExample: StringSpec() {
     })
   }
 
-}
+} }
 ```
 
 There are generators for all the common types - String, Ints, Sets, etc. If you need to generate custom types
@@ -230,7 +230,7 @@ then you can simply specify the generator manually (and write your own). For exa
 with the generators specified.
 
 ```kotlin
-class PropertyExample: StringSpec() {
+class PropertyExample: StringSpec() { init {
 
   "String size" {
     forAll(Gen.string(), Gen.string(), { a: String, b: String ->
@@ -238,7 +238,7 @@ class PropertyExample: StringSpec() {
     })
   }
 
-}
+} }
 ```
 
 To write your own generator for a type T, you just implement the interface `Gen<T>`. For example you could write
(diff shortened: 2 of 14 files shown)


helmbold:
Thanks for your contribution.

I have mixed feelings about bringing annotations to KotlinTest. The syntax is annotation free at the moment what is clean and simple. Adding annotations brings a bunch of new code only for this purpose.

Have you tried to add annotations to the existing config method?

saksmt:
@helmbold,
I've tried to add them in the most natural way with backward-compatibility in mind (that's why there is so much code). Also code will remain annotation free on side of end-user of library if they want so. There are not so much cases where this PR could be useful until you try to write integration tests with KotlinTest.

Well, they could be added to config, but adding them as optional parameter seems more natural approach, at least because in other code you see annotations before body of the method. Also it can't be fully achieved if you want to add some annotations on `TestSuite`

helmbold:
Thanks for your contribution.

I have mixed feelings about bringing annotations to KotlinTest. The syntax is annotation free at the moment what is clean and simple. Adding annotations brings a bunch of new code only for this purpose.

Have you tried to add annotations to the existing config method?

helmbold:
What concerns me here is that we would introduce second way to add metadata to tests. That would make the programming model less consistent. And since annotations in you approach don't look like usual annotations (`a<MyAnnotation>`) it doesn't resemble the classic annotation approach either.

I agree with you that it might be useful in some cases to have annotations, but I'd like to have a way that better fits the existing programming model. Since we are using the `config` function to add meta-data that would be the obvious place to add annotations. Per-spec metadata is currently added by overriding properties of the Spec class (e.g. `override val oneInstancePerTest = false`). I think we could have something similar for Spec (and pass to `TestSuite` internally).

Would this approach be ok for you, @saksmt ? And what do you think @sksamuel ?

saksmt:
@helmbold,
So you're suggesting definition of specification like following:
```kotlin
class MySpec : SomeSpec() { init {
  annotations = arrayOf(
    a<SomeClassLevelAnnotation>(),
    a<AnotherClassLevelAnnotation>()
  )

  `when`("using annotations") {
    annotations = arrayOf( // using another `SomeSpec` instance as receiver
      a<SuiteLevelAnnotation>()
    )
    then("they must be passed to config") {
      // some code
    }.config(annotations = arrayOf(a<TestCaseLevelAnnotation>(
      "hugeAmountOfAnnotationParameters" to arrayOf(/* ... */)
    )))
  }
} }
```

or even

```kotlin
class MySpec : SomeSpec() {
  override val annotations = arrayOf(
      a<SomeClassLevelAnnotation>(),
      a<AnotherClassLevelAnnotation>()
  )

  init {
    `when`("using annotations") {
      then("they must be passed to config") {
        // some code
      }.config(annotations = arrayOf(a<TestCaseLevelAnnotation>(
        "hugeAmountOfAnnotationParameters" to arrayOf(/* ... */)
      )))
    }.config(annotations = arrayOf(
        a<SuiteLevelAnnotation>()
    ))
  }
}
```
?

If so, I agree that from library developer point of view it would be easier to implement and maintain, but from end-user point of view it might look a little bit messy...

sksamuel:
We should provide an adapter for Allure documentation.

I don't have a problem with having annotations, if the thing being annotated is an annotation to start with. It seems a bit silly to say annotation types can't be used as annotations but instead parameters.

saksmt:
@sksamuel,
Adapter for allure is good idea, but for other cases, which not mentioned there (like attachements and steps). Also current allure adapter is written for junit which is used inside kotlintest, so it is applicable, except missing ability to pass annotations to description.

The only way to use real annotations is to make their target to be expression which is not good idea, because they are hard to extract (at least because it is poorly documented).

If we forget about annotations for a moment and think of it as of ability to change some property of `Description` which consists of `testClass`, `annotations`, `uniqueId`, `displayName` and `children` we get following: only one property should be user-configurable - `annotations` while others are purely internal and current problem is that there is no way of providing that list of annotations for particular test case, because test-case is "virtual" (it is not real method). So we need to either provide a way to pass that annotations or we end up writing adapters for any junit-based event listeners.

sksamuel:
Allure extension has been added to 3.0.0-RC3.

PR #147, opened 2017-02-09
Minor testfix for "should endWith"-tests

danielspaniol:
Really just a small fix.
There were two cases in which the "should endWith"-test used `startWith` instead of `endWith`


sksamuel:
Thanks.

PR #150, opened 2017-02-25
FIX #146 - Update mockito dependency to 2.X.X

### LINKED ISSUE #146: Upgrade "org.mockito:mockito-core:1.10.19" dependency to "org.mockito:mockito-core:2.X.X"

aleixmorgadas:
* updated gitignore
* upgraded kotlin-gradle-plugin into 1.1.0-rc-91
* upgraded kotlin-stdlib and kotlin-reflect from 1.0.3 to 1.0.6
* upgraded mokito-core from 1.10.19 to 2.7.11

Created `MockFinalClassTest` as sample test case. Mockito reference [here](https://github.com/mockito/mockito/wiki/What's-new-in-Mockito-2#unmockable)


sksamuel:
Thanks for the PR. However I think the gradle wrapper script should always be checked into source. That's the point of the wrapper - to ensure you can always bootstrap with the version of gradle the project was built with.

aleixmorgadas:
The reason was that in branch release/2.0 there was not the gradle wrapper, so I thought that better do not push my own gradle version until you decide.

I revert the commit where the gitignore is modified and I push again.

sksamuel:
I see. If you want to add the gradle wrapper that would be the best thing I think (its the same wrapper regardless of who adds it)

Issue #111, opened 2016-10-25
shouldEqual defines equality differently than kotlin


SilverFishCat:
According to [the kotlin reference](https://kotlinlang.org/docs/reference/operator-overloading.html) the equality operator `==` is translated to `a?.equals(b) ?: b === null`. (`!=` is similarly defined)
Additionally, `equals` receives a nullable parameter.

The `shouldBe` (and `shouldEqual`) functions are defined as 

```
if (this == null && any != null)
  throw AssertionError(this.toString() + " did not equal $any")
if (this != null && any == null)
  throw AssertionError(this.toString() + " did not equal $any")
if (this != any)
  throw AssertionError(this.toString() + " did not equal $any")
```

This means that if `a == null` returns `true` (whether by design or by mistake), `a shouldBe null` will still throw. Checking `this == null` is also not needed. 
This is also in contrast to the new `equalityMatcher`, which only checks `this == value`.

SilverFishCat:
There maybe there should be a way to check for [Referential equality](https://kotlinlang.org/docs/reference/equality.html#referential-equality).

sksamuel:
Is this the case you're referring to or am I missing something subtle? https://github.com/kotlintest/kotlintest/blob/master/src/test/kotlin/io/kotlintest/NullTests.kt#L20

SilverFishCat:
No, it's referring to https://github.com/kotlintest/kotlintest/blob/master/src/test/kotlin/io/kotlintest/NullTests.kt#L29

Actually, after doing some tests I have noticed a weird and inconsistent behavior in kotlin:
```kotlin
class A(var i: Int) {
  override fun equals(other: Any?): Boolean = other == null && i==0
}

var g:A? = null
println(A(0) == g) // true
println(A(0) == null) //false
```

This might be an issue in the kotlin compiler. I'm going to open an issue with JetBrains and will update this issue once they fix it/create a consistent definition for the `==` operator.
Regardless, the expected behavior in my opinion is ```println(A(0) == g) // true``` (it fits with the kotlin reference and the code behind the `==` operator). `A(0) shouldBe(g)` will still throw an AssertionError.

sksamuel:
Did you get anywhere with raising this with Jetbrains?

aleixmorgadas:
Resolving the inconsistent behavior.
`A(0) == null ` uses the equality of null which resolves the operation returning `false`
`A(0) == g) uses the `A.equals` method

[...]

Issue #155, opened 2017-03-05
shouldBe should give more helpful failures when comparing Strings


xenomachina:
When comparing strings, the `AssertionError` produced by JUnit's `assertEquals` will highlight the difference. In IntelliJ (and possibly other IDEs) there's even a "<Click to see difference>" link that brings up a side-by-side diff.

kotlintest's `shouldBe` just shows the two strings one after the other with "did not equal" in between. There aren't even any delimiters around the two strings. When dealing with large strings this can make finding the cause of the failure more difficult than it has to be, especially if the difference is something like white space.

PR #158, opened 2017-03-12
Feature/155

aleixmorgadas:


helmbold:
We should close #111 and #155 then, right? I would, by the way, prefer separate pull requests for each issue.

sksamuel:
I've closed them. I think we should be grateful that someone has taken the time to contribute :)

helmbold:
Yes, I'm really happy that we have some additional contributors!

PR #162, opened 2017-03-19
FIX #144 - eventually returns the computation inside f()

### LINKED ISSUE #144: eventually should be able to return a value that has been computed inside it
It would be useful to be able to use the `eventually` construct to get a value that has been computed inside of the `f` function.

aleixmorgadas:


sksamuel:
:+1:

PR #165, opened 2017-03-24
Updated Kotlin to stable version 1.1.1 removing unneeded eap reposito…

jrgonzalezg:
…ries. Also removed unneeded sourceSets declaration.

Using kotlintest is requiring to use eap repositories which should not be needed now that Kotlin 1.1 is stable. Could you generate a new version of kotlintest with this so it can be used without issues in Kotlin 1.1 projects?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,7 @@
 buildscript {
-    ext.kotlin_version = '1.1.0-rc-91'
+    ext.kotlin_version = '1.1.1'
     repositories {
         mavenCentral()
-        maven {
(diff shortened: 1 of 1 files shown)


sksamuel:
Can you PR this against the 2.0 branch please.

jrgonzalezg:
Done on #166 although there the eap repositories where already gone so that was not needed. Does this mean that the 2.0 branch is going to produce a release soon? If not, i still think it would be good to release a 1.x version with Kotlin 1.1.1

sksamuel:
2.0 will be released this weekend.

jrgonzalezg:
Great!

PR #170, opened 2017-04-02
Improvement for TypeMatchers.beOfType

sandwwraith:
In current implementation, given the following hierarchy of classes:

```kotlin
open class Base

class Derived : Base()
```

It's possible to write such code:

```kotlin
val d = Derived() 
d should beOfType<Derived>()
d shouldNot beOfType<Base>()
```

but compiler would complain, if you write:

```kotlin
val d : Base = Derived()
d should beOfType<Derived>()
```

Because it expects `Matcher<Base>` on the right-hand, or Matcher for any supertype to make automatic up-cast for `d`. It makes impossible to test, for example, factory methods. This PR addresses this issue.

Also, I've noticed that `"TypeMatchers.theSameInstanceAs"` and `"beTheSameInstanceAs"` are the same tests. Was it made intentionally or by mistake?

### DIFF
--- a/src/main/kotlin/io/kotlintest/matchers/TypeMatchers.kt
+++ b/src/main/kotlin/io/kotlintest/matchers/TypeMatchers.kt
@@ -13,10 +13,10 @@ fun <T> beTheSameInstanceAs(ref: T): Matcher<T> = object : Matcher<T> {
   override fun test(value: T) = Result(value === ref, "$value should be the same reference as $ref")
 }
 
-inline fun <reified T : Any> beOfType() = object : Matcher<T> {
+inline fun <reified T : Any> beOfType() = object : Matcher<Any> {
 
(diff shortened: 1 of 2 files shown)


sksamuel:
Thanks for the PR. Could this also be fixed by making the matcher covariant in T ?

sandwwraith:
I've tried this approach first, but compiler complained about functions `and` and `or` saying that `Type parameter T is declared as 'in' but occurs in 'out' position in type Matcher<T>`. To be honest, I don't know why it thinks that  it has out variance.

Anyway, it seems more logical to have `Matcher<Any>`, just like `beInstanceOf` - IMHO, they have very similar use cases.

sksamuel:
I think the and and or methods would need a type projection for them to work. However, I think Any is probably fine since it is an instance of check anyway.

PR #173, opened 2017-04-23
Fix the document of Matchers

koko-u:
Since there is an absent matcher 'containInAnyOrder' in the collections matcher section, fix it to 'containsAll'.


sksamuel:
thanks

PR #176, opened 2017-05-03
fix(lambda) #171: fix lambda npes: initialize attribute before init block

### LINKED ISSUE #171: NullPointerException at ShouldSpec using lambda parameter
My test:
```kotlin
// version 2.0.0
class ExampleSpec : ShouldSpec({
    "test" {
        should("succeed") {
            0 shouldBe 0
        }
    }
})
```
The code compiled, but throws an exception.
```
java.lang.NullPointerException
    at io.kotlintest.specs.ShouldSpec.invoke(ShouldSpec.kt:20)
```
However, when I use `init` block, the test succeeded.
```kotlin
class ExampleSpec : ShouldSpec() {
    init {
        "test" {
            should("succeed") {
                0 shouldBe 0
            }
        }
    }
}
```

neyb:
fix #171


sksamuel:
Thanks!

PR #197, opened 2017-07-29
Several useful Gen extensions

adelnizamutdinov:
I found myself copying these over and over from project to project. Might as well contribute them


sksamuel:
Awesome

PR #200, opened 2017-08-07
Generators for maps and pairs

sandwwraith:
Motivation: 
Currently, one can simply create generators for standard collections like lists and sets, but there are no default generator for map, which is also important standard collection.

Changes:
* Function for creating `Gen` object for pairs and maps (map generator uses pair generator)
* Inferring for default types also supported


sksamuel:
Lovely thanks!

PR #215, opened 2017-09-29
Improve "eventually" syntax to allow passing custom Exception type

vkolomeyko:
This is useful when it is know that the block of code can fail with "AssertionError" for some time.
AssertionError is not of type Exception so old implementation is limiting in this regard.

### DIFF
--- a/src/main/kotlin/io/kotlintest/Eventually.kt
+++ b/src/main/kotlin/io/kotlintest/Eventually.kt
@@ -1,13 +1,21 @@
 package io.kotlintest
 
-fun <T>eventually(duration: Duration, f: () -> T): T {
+import java.lang.Exception
+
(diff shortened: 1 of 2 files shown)


sksamuel:
You might have a test that throws a file not found exception for a period of time and then it would now fail. Perhaps we should just swallow exceptions plus AssertionErrors ?

vkolomeyko:
This change is not a breaking API/behaviour change - it just adds ability to specify an **optional** exception parameter.
Just as before by default `Exception` is used in the catch block.
I have added an additional unit test to show that in case of `FileNotFoundException` it works as before.

sksamuel:
Ok I was confused :)

sksamuel:
:+1:

Issue #41, opened 2016-06-21
Add configuration parameters to property testing


sksamuel:
- invocation count - the number of different values that should be fed in 
- minsuccessful - the mininum number of attempts that must pass
- threads - number of threads to use in the executor pool

sksamuel:
If you want to pick this up @helmbold feel free, otherwise I'll do it after my other tickets.

helmbold:
Yes, I would do it next week.

sksamuel:
Is this still outstanding?

helmbold:
Yes, it is still open. This was not on my radar ...

sksamuel:
Check three comments up :p

helmbold:
Yes, I know :-/ But completely forgotten this issue, though.

sksamuel:
Ok I'll do it.

sksamuel:
I'm closing this as since #222 we can specify the number of attempts, and the number of threads can be specified by using any standard test block.

PR #222, opened 2017-10-07
Improvements in property-based testing.

psa-anddev:
I was thinking that it should be possible to specify the number of attempts a piece of property-based testing is run and I was about to open an issue for it when I found issue #41 which I think is relevant. 
This PR adds an attempts parameter that can be introduce as the first parameter. When that is done, the test will only be run the number of times specified in the parameter. I have also updated the documentation to point this fact out.

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -233,6 +233,25 @@ class PropertyExample: StringSpec() {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


sksamuel:
Brilliant.

PR #224, opened 2017-10-09
FloatMatchers

elect86:


sksamuel:
Thanks for the PR, however at the moment it breaks the build.

PR #228, opened 2017-10-18
Fixed lambda in reference.md

ConradoMateu:


sksamuel:
That's a bit of Scala-ism creeping in there :) Thanks for fixing.

PR #229, opened 2017-10-19
MockK mocking framework

oleksiyp:
Hi guys,

I've spent a few weeks coding new Kotlin mocking framework. It's called MockK. I doubt I have enough time to push it myself, but still believe it worth your review and possibly merge to "experimental" kotlin-test package. With enough joint effort this can be killer feature of kotlin-test

(don't forget to Load diff on MockK.kt)

### Nice features

- [x] pure kotlin mocking DSL
- [x] signature based partial arguments matching (innovation)
- [x] chained calls by default aka mockito deep stubs (but better)
- [x] removing finals
- [x] instrumenting no args constructor
- [x] matcher expressions
- [x] mocking coroutines
- [x] capturing lambdas
- [x] bunch of matchers, verification modes and useful DSL constructs

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,6 +59,11 @@ dependencies {
     compile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
     compile 'org.mockito:mockito-core:2.7.11'
     compile 'org.reflections:reflections:0.9.11'
+    compile 'org.jetbrains.kotlinx:kotlinx-coroutines-core:0.19'
+    compileOnly 'org.slf4j:slf4j-api:1.7.25'
+    testCompile 'org.slf4j:slf4j-api:1.7.25'
+    testCompile 'ch.qos.logback:logback-classic:1.+'
+
 }
 
 uploadArchives {
--- a/build.gradle
+++ b/build.gradle
@@ -118,3 +118,8 @@ jacocoTestReport {
         html.destination "${buildDir}/jacocoHtml"
     }
 }
+kotlin {
+    experimental {
+        coroutines "enable"
+    }
+}
--- a/doc/mockk.md
+++ b/doc/mockk.md
@@ -4,15 +4,15 @@ pure kotlin mocking framework
 
 ### Nice features
 
- [x] pure kotlin mocking DSL
- [x] signature based partial arguments matchers (innovation)
- [x] chained calls by default aka mockito deep stubs (but better)
- [x] removing finals
- [x] instrumenting no args constructor
- [x] matcher expressions
- [x] mocking coroutines
- [x] capturing lambdas
- [x] bunch of matchers, verification modes and useful DSL constructs
+ - [x] pure kotlin mocking DSL
+ - [x] signature based partial arguments matchers (innovation)
+ - [x] chained calls by default aka mockito deep stubs (but better)
+ - [x] removing finals
+ - [x] instrumenting no args constructor
+ - [x] matcher expressions
+ - [x] mocking coroutines
+ - [x] capturing lambdas
+ - [x] bunch of matchers, verification modes and useful DSL constructs
 
 ### DSL
 
--- a/doc/mockk.md
+++ b/doc/mockk.md
@@ -1,0 +1,73 @@
+MockK
+=====================================
+pure kotlin mocking framework
+
+### Nice features
+
+ [x] pure kotlin mocking DSL
+ [x] signature based partial arguments matchers (innovation)
+ [x] chained calls by default aka mockito deep stubs (but better)
+ [x] removing finals
+ [x] instrumenting no args constructor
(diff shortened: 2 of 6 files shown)


sksamuel:
Hi.

What's the advantage of this library over say https://github.com/nhaarman/mockito-kotlin

oleksiyp:
It's written from scratch, so main advantage is close integration with Kotlin.

Partial argument matching for example is very useful because allows to match only few arguments(so don't need to match all, and thus functions with defaults benefit a lot)

Chained calls allow to keep DSL style expressions in "when" ("every") clause

Lambda capture is useful to mock DSL constructs

e.t.c. 

In general there is a lot more flexibility to write required features instead of trying to match Kotlin with Mockito. It would take ages to request changes in Mockito if they are required

Here is simple example of tests written in MockK DSL:
https://gist.github.com/oleksiyp/289c2a49cabe38023a05ea0e9335d22f

sksamuel:
I see. Very interesting. Is there no other Kotlin based mocking library atm?

oleksiyp:
Yep, I am not aware of any of libraries except Mockito wrappers

mfriedenhagen:
Hello, why do you need MockKJUnitRunner in RunWith. AFAIK this annotation is deprecated with junit5 e.g
.

oleksiyp:
Basically for class transformation. (to remove finals and introduce "no-arg" constructor)
After setting it up it just passes control to the runner on the first found super-class. 

Anything that is setting up somehow class-loader should be okay.
I'll check JUnit 5 docs on how to replace it

sksamuel:
@oleksiyp why are you not creating a project for this, rather than a PR here? I don't think KotlinTest should supply it's own mocking framework, let the users decide which one they want to use.

oleksiyp:
For me personally it's a matter of involvement. I don't want to create new project and perform all startup activities. I just want to be a dev. 

Second thing it's not a matter of strong suggestion on "what to use" for user. It's about delivering one framework with all the functionality. I don't want any ownership privileges "to develop my framework", but   rather want to have integrated solution for users (and myself as well). User can still easily use Mockito wrapper. There is even no function name clashes and just a few public functions to introduce.

Third thing is that I find KotlinTest really cool framework with all the matchers DSLs and generators. I am going to use it at work. But it's almost impossible to push own framework for use. So having all-in-one is better for me.

sksamuel:
> For me personally it's a matter of involvement. I don't want to create new project and perform all startup activities. I just want to be a dev.

Right, and for the same reason I don't want to take on a new (sub) project that involves more of my time maintaining it, as I already have 5 projects. If this code was merged, and then people request features or report bugs, and you've decided you're bored of it, it's this project left to do the work.  I don't know much about mocking frameworks as I don't use them, but if yours is genuinely better, than start it up as it's own project and people can choose to use it.

oleksiyp:
Great, no worries. Can you give some basic hints from your experience creating OS projects?

oleksiyp:
I am pushing my own project here http://mockk.io

oleksiyp:
@Kerooker did you know that MockK started as [one 1.7K LOC file](https://github.com/kotlintest/kotlintest/pull/229/files#diff-99c5b5cb266e214ce84ca4e4e3257481R1) project and was submitted as a PR to kotlintest? Now MockK is 42K LOC and zillion of small files :smile:

LeoColman:
I've seen this issue in the past :P

I think Mockk ended up much better than as part of KotlinTest

oleksiyp:
Never know where you lose and where you gain... (we have such Russian idiom) 
Maybe it would be even more successful :smile: 
After all, it is was just the right separation. 
Mocking is mostly about dirty reflection/bytecode hacking and packaging it in a nice package.

The support for me is the biggest problem now. I return back to it every 1-2 months to release a new version and feel guilty for each bug and time it was in an opened state :smile: From another side learning curve for the project itself is very steep. Many tried so far, none survived. Partly due to not so good codebase, partly due to complex implementation details and approaches.

LeoColman:
Perhaps you could create a guide on how to navigate through stuff and how everything work? I would be glad to help with some issues if I could understand everything

oleksiyp:
I started doing that: https://github.com/mockk/mockk/wiki
Even recorded a few videos. But now think that videos are not good enough.
They just go through source code, not always describing the whole idea.

Don't even know how to approach it. Maybe some guide explaining internals(with nice pictures :smile:)

LeoColman:
Never seen that!

> But now think that videos are not good enough.

Anything is better than nothing

sksamuel:
I remember declining the PR @oleksiyp :)
Not because I didn't think it was a good project, but because it was better as a stand alone project that people could use outside of KotlinTest. And it's proven to be a good decision as MockK is very very popular.

Issue #174, opened 2017-04-28
lateinit variable doesn't work with interceptTestCase


lenguyenthanh:
I have a problem with init `lateinit var` property in `interceptTestCase` method.
Here is my example:
```
class ExampleTest : StringSpec() {
    private lateinit var someString: String

    override fun interceptTestCase(context: TestCaseContext, test: () -> Unit) {
        println("Start Testcase")
        someString = "Hello"
        test()
        println("End Testcase")
    }

    init {
        "Hello should equal to Hello" {
            someString shouldEqual "Hello"
        }
    }
}
```

And here is out put:
```
Start Testcase

kotlin.UninitializedPropertyAccessException: lateinit property someString has not been initialized

	at com.locata.myapplication.ExampleTest.access$getSomeString$p(ExampleTest.kt:7)
	at com.locata.myapplication.ExampleTest$1.invoke(ExampleTest.kt:18)
	at com.locata.myapplication.ExampleTest$1.invoke(ExampleTest.kt:7)
	at io.kotlintest.Spec$runTest$callable$1$1.invoke(Spec.kt:124)
	at io.kotlintest.Spec$runTest$callable$1$1.invoke(Spec.kt:15)
	at io.kotlintest.Spec$runTest$initialInterceptor$1$1.invoke(Spec.kt:116)
	at io.kotlintest.Spec$runTest$initialInterceptor$1$1.invoke(Spec.kt:15)
	at com.locata.myapplication.ExampleTest.interceptTestCase(ExampleTest.kt:12)
	at io.kotlintest.Spec$runTest$initialInterceptor$1.invoke(Spec.kt:116)
	at io.kotlintest.Spec$runTest$initialInterceptor$1.invoke(Spec.kt:15)
	at io.kotlintest.Spec$runTest$callable$1.call(Spec.kt:124)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)


Process finished with exit code 255
```

It's happen for both 2.0.0 and 2.0.1.

lenguyenthanh:
Just investigated this issue. In example above, We use oneInstancePerTest so the `someString` in test is in different instance with `someString` we initialized in `interceptTestCase`

[...]

PR #230, opened 2017-10-20
FIX: For one instance per test, interceptTestCase is called on that given instance #174

michalbrz:
I believe this didn't work as intended as `interceptTestCase` was called on one 'main' instance of `Spec`. 
Now it is called on the same instance of `Spec` that will be used for test. Therefore, test and `interceptTestCase` can reference class fields and it works as users coming from JUnit would expect - changing field value in `interceptTestCase` changes field value in test.

### DIFF
--- a/src/main/kotlin/io/kotlintest/Spec.kt
+++ b/src/main/kotlin/io/kotlintest/Spec.kt
@@ -113,7 +113,7 @@ abstract class Spec {
           else Executors.newFixedThreadPool(testCase.config.threads)
       notifier.fireTestStarted(testCase.description)
       val initialInterceptor = { context: TestCaseContext, testCase: () -> Unit ->
-        interceptTestCase(context, { testCase() })
+        spec.interceptTestCase(context, { testCase() })
       }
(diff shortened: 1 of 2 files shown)


sksamuel:
Thanks for the PR. I will need to incorporate this into the latest changes (junit 5).

michalbrz:
Thanks @sksamuel but does it mean that this change will not be introduced until 3.0 release?

tonilopezmr:
@sksamuel When are you going to merge this?

sksamuel:
I thinks this superseeds it https://github.com/kotlintest/kotlintest/pull/231

Ben-Woolley:
Hi @sksamuel , now that  #231 is merged will this be included in an upcoming release?

sksamuel:
I'm working on a release at the moment but it won't be for a week or so.

sksamuel:
This is now fixed in 3.0.0-RC1. I didn't merge this as the code has changed significantly but your PR was a good source of information so thank you.

PR #231, opened 2017-10-21
Feature/junit5

sksamuel:
Upgrade to jUnit 5, which involved rewriting much of the infrastructure around tests. It does mean we get a much nicer implementation of project config - afterall is now elegant - and we can support better naming in the specs (parenthesis for example). One instance is now off by default.


sksamuel:
@all All reviewers welcome.

sksamuel:
I will merge this if you have no feedback or objections @helmbold

helmbold:
I'll have a look at it at the weekend. I don't have as much time as I'd like to have at the moment ...

src/main/kotlin/io/kotlintest/KotlinTestEngine.kt
```diff
@@ -0,0 +1,220 @@
...
+
+import org.junit.platform.commons.JUnitException
+import org.junit.platform.commons.util.ReflectionUtils
+import org.junit.platform.engine.*
+import org.junit.platform.engine.discovery.PackageSelector
+import org.junit.platform.engine.support.hierarchical.EngineExecutionContext
+import java.lang.reflect.Modifier
+import java.util.*
+
+class KotlinTestContext(val executionRequest: ExecutionRequest) : EngineExecutionContext
```
helmbold: It seems that KotlinTestContext is never used. Is it used via reflection? If so, it should be documented.

src/main/kotlin/io/kotlintest/KotlinTestEngine.kt
```diff
@@ -0,0 +1,220 @@
...
+
+class KotlinTestEngine : TestEngine {
+
+  override fun execute(request: ExecutionRequest) {
+    println("Executing test request $request")
+    Project.beforeAll()
```
helmbold: No that we have an object which is a container with all specs, we should be able to harmonize the global interceptors with spec and test cases interceptors and use functions to wrap the whole project as with the interceptor chain in `SpecExecutor`. This would lift the whole interceptor mechanism on a new level of elegance and conceptual simplicity (through uniformity).

src/main/kotlin/io/kotlintest/KotlinTestEngine.kt
```diff
@@ -0,0 +1,220 @@
...
+        else -> false
+      }
+    }
+
+  private fun readProperty(name: String): List<String> =
+      (System.getProperty(name) ?: "").split(',').map { it.trim() }
+}
+
+// the root descriptor
+class EngineTestDescriptor : TestDescriptor {
```
helmbold: This class seems to be unused? If it is used reflectively, this should be documented.

src/main/kotlin/io/kotlintest/Project.kt
```diff
@@ -1,59 +1,33 @@
...
-      ReflectionsHelper.registerUrlTypes()
-      val configClasses = Reflections().getSubTypesOf(ProjectConfig::class.java)
-      when {
-        configClasses.size == 0 -> null
-        configClasses.size > 1 -> {
-          val configClassNames = configClasses.map { config -> config.simpleName }
-          throw InvalidConfigException("Multiple ProjectConfigs found: $configClassNames")
-        }
-        else -> configClasses.firstOrNull()?.kotlin?.objectInstance
+  private val ProjectConfigFQN = "io.kotlintest.provided.ProjectConfig"
```
helmbold: Lets start fields with a lower case letter. What is FQN? Better use a self-explanatory name.

src/main/kotlin/io/kotlintest/Project.kt
```diff
@@ -1,59 +1,33 @@
...
+
+  private fun discoverProjectConfig(): AbstractProjectConfig? {
```
helmbold: Why not simply use a `val` and initialize with the body of this function?

src/main/kotlin/io/kotlintest/Spec.kt
```diff
@@ -71,99 +39,27 @@ abstract class Spec {
...
-
-  private fun <CONTEXT> createInterceptorChain(
-      interceptors: Iterable<(CONTEXT, () -> Unit) -> Unit>,
-      initialInterceptor: (CONTEXT, () -> Unit) -> Unit): (CONTEXT, () -> Unit) -> Unit {
-    return interceptors.reversed().fold(initialInterceptor) { a, b ->
-      {
-        context: CONTEXT, testCase: () -> Unit -> b(context, { a.invoke(context, { testCase() }) })
-      }
-    }
+  internal open fun interceptSpec(chain: () -> Unit) {
```
helmbold: I think `chain` is misleading here, since the intercepted thing is primarily the spec itself. The fact that the spec itself could be decorated with other interceptors is not as important here. From the callers point of view, it is the spec. --> rename to `spec`.

src/main/kotlin/io/kotlintest/SpecExecutor.kt
```diff
@@ -0,0 +1,80 @@
...
+      interceptors: Iterable<(CONTEXT, () -> Unit) -> Unit>,
+      initialInterceptor: (CONTEXT, () -> Unit) -> Unit): (CONTEXT, () -> Unit) -> Unit {
+    return interceptors.reversed().fold(initialInterceptor) { a, b ->
+      { context: CONTEXT, testCase: () -> Unit ->
+        b(context, { a.invoke(context, { testCase() }) })
+      }
+    }
+  }
+
+  private fun run(descriptor: TestDescriptor, listener: EngineExecutionListener) {
```
helmbold: `run` and `execute` sound very similar without explaining the difference. More describing names would be helpful.

src/main/kotlin/io/kotlintest/TestRunner.kt
```diff
@@ -0,0 +1,62 @@
+package io.kotlintest
+
+import org.junit.platform.engine.EngineExecutionListener
+import org.junit.platform.engine.TestExecutionResult
+import org.junit.runners.model.TestTimedOutException
+import java.util.concurrent.Executors
+
+class TestRunner(val listener: EngineExecutionListener) {
```
helmbold: `listener` could be private.

src/main/kotlin/io/kotlintest/TestRunner.kt
```diff
@@ -0,0 +1,62 @@
...
+
+  private fun <CONTEXT> createInterceptorChain(
```
helmbold: Maybe we can make this a top level function and reuse it for test cases, specs and the whole project.

src/main/kotlin/io/kotlintest/TestRunner.kt
```diff
@@ -0,0 +1,62 @@
...
+      val testCaseContext = TestCaseContext(spec, testCase)
+
+      val errors = mutableListOf<Throwable>()
+
+      for (j in 1..testCase.config.invocations) {
+        executor.execute {
+          try {
+            testInterceptorChain(testCaseContext, { testCase.test() })
+          } catch (t: Throwable) {
+            errors.add(t)
```
helmbold: Shouldn't this be synchronized? Most of the time it will probably work with multiple threads, but there is no guarantee.

src/test/kotlin/io/kotlintest/KotlinTestJunitRunnerTest.kt
```diff
@@ -1,83 +1,83 @@
...
-}
-
-class HasIgnoredTest : StringSpec() {
-  init {
-    "ignored test" {
-    }.config(enabled = false)
-  }
-}
\ No newline at end of file
+//package io.kotlintest
```
helmbold: make this test work or remove it, but commented code shouldn't be in the main branch.

sksamuel:
Other than some minor code style issues you pointed out, and changing project interceptors to use the chained pattern, you think the PR is ok.

helmbold:
Yes, but I think that the interceptor design would be a relevant change. I wanted to implement it that way, but it was not possible due to limitations of JUnit 4. So actually the current implementation with `beforeAll` and `afterAll` is a workaround.

PR #233, opened 2017-11-01
Added any() to return non-null objects

tonilopezmr:
When we use mockito with kotlintest:

```
val myMock = mock()
Mockito.verify(myMock, never()).myAction(Mockito.any())
```

We have this issue:

`
java.lang.IllegalStateException: anyObject() must not be null
`

This is because `Mockito.any()` always return null, with this simple commit just call to kotlintest `any()` for solve this issue 👍


sksamuel:
:+1:

PR #234, opened 2017-11-14
Fix code bock formatting in the reference

xrigau:
Before | After
--- | ---
screen shot 2017-11-14 at 4 22 38 pm | screen shot 2017-11-14 at 4 24 36 pm


sksamuel:
Nice spot.

PR #237, opened 2017-11-21
Clarify how includeTags/excludeTags interact with untagged tests

sschuberth:
In particular, `includeTags` does not include that tag to the set of untagged tests, but it limits the run to exactly the tag.


sksamuel:
👍

PR #245, opened 2018-01-03
Added double matching to Matchers.md (documentation only)

Kerooker:
Not having these in the documentation might make the user believe they don't exist. As any comparable may use these assertions, it's fair to add them to the double type


sksamuel:
Excellent thanks.

PR #250, opened 2018-01-31
Bumped kotlin version to 1.2.21

Arneball:


hastebrot:
What happens when we increment to Kotlin version, e.g. from Kotlin 1.1.x to 1.2.x? Will the artifact still be compatible to Kotlin 1.1? Are artifacts compiled with Kotlin 1.1.x compatible to Kotlin 1.2?

hastebrot:
So, does it make sense to update the Kotlin version, if we don't need its features?

The other day I wanted to use "kotlinx.serialization" but noticed it still defines Kotlin 1.1 as dependency. Turned out it works pretty well with Kotlin 1.2.

sksamuel:
You make a good point. Probably best not to change it unless theres another
reason.

Arneball:
Should be BC according to https://kotlinlang.org/docs/reference/compatibility.html

sksamuel:
What about if by upgrading to 1.2, someone adds a feature only available in 1.2 ?

Arneball:
What do you mean? Which feature would someone add?

tors 1 feb. 2018 kl 04:07 skrev Stephen Samuel <notifications@github.com>:

elect86:
It's always suggested to update to the last kotlin and has nothing to do with features.

For example, it's been a while since I read always warning in console during builds about potential incompatibilities with multiple kotlin reflection lib compiled with different kotlin version.

elect86:
Ps: you need to upgrade also gradle or it won't work with java 9

NersesAM:
> What about if by upgrading to 1.2, someone adds a feature only available in 1.2 ?

you can use `kotlin.compiler.apiVersion` property to tell compiler you want to stick to 1.1 api. https://kotlinlang.org/docs/reference/using-maven.html#attributes-common-for-jvm-and-js

PR #253, opened 2018-02-15
Allow Eventually to use a predicate for additional flexibility

philipsdoctor:
If you're testing code that fails but doesn't throw exceptions, for example checking an HTTP status code, you might want to just pass an arbitrary predicate instead of relying on an Exception to signal a failure.

As an aside, I have no idea how to run tests in this project, 
```
./gradlew test
```

compiles tests but per the generated report at 
kotlintest/build/reports/tests/test/index.html

Nothing is actually running.  I also tried to write some failing tests, but again, nothing fails from the gradle test task. 

Am I running this the wrong way?  I checked travis, but travis seems to use the test task as well.


sksamuel:
Thanks for the pr.
I'll investigate while the tests are not running as well.

PR #254, opened 2018-02-18
Update: minor optimization for `Gen::nats` and `Gen::negativeIntegers…

outis:
Here's a minor optimization for Gen's `nats()` and `negativeIntegers()` that uses `abs` to reduce the number of out-of-range values.


sksamuel:
:+1:

PR #256, opened 2018-03-01
Allowing supertypes in shouldThrow matcher

NersesAM:
While introducing my change the following test started failing. https://github.com/kotlintest/kotlintest/blob/c2c1e925ece52e94bc95fc7ba2e176d6297142ab/src/test/kotlin/io/kotlintest/matchers/ExceptionMatchersTest.kt#L23-L31

Throwing Throwable and expecting Throwable is currently causing AssertionError. I do not understand how is that test valid to be honest. Am I missing something?

I have replaced that test case with the one that is working for Supertypes.

WDYT?

### DIFF
--- a/src/main/kotlin/io/kotlintest/matchers/Matcher.kt
+++ b/src/main/kotlin/io/kotlintest/matchers/Matcher.kt
@@ -1,5 +1,7 @@
 package io.kotlintest.matchers
 
+import kotlin.reflect.full.isSubclassOf
+
 interface Matcher<T> {
 
   fun test(value: T): Result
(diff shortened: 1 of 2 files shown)


NersesAM:
By the way seems Gradle is not running the tests at all. Probably same reason Intellij is not understanding those as tests. Haven't used gradle for ages, not sure how to fix it :)

sksamuel:
Without this you can't test for Exception or Throwable. Although I think some people might argue that you should be testing for a specific exception only.

What about introducing, `shouldThrowSuperOf<SuperType>` or some variation on the name (something that doesn't sound as bad as that). Or making a new one that's specific to the type only, `shouldThrowExactly`

sksamuel:
I'm working on a new release 3.0 so I'm happy to introduce some breaking changes if needed into that.

sksamuel:
See https://github.com/kotlintest/kotlintest/issues/247

NersesAM:
> Without this you can't test for Exception or Throwable. Although I think some people might argue that you should be testing for a specific exception only.

Sorry I didn't get what you mean by this? That test above looks wrong to me. Throwing Throwable and expecting it should be true, isn't it? 

Adding `shouldThrowExactly` and changing the behaviour of `shouldThrow` sounds great to me.

sksamuel:
I've implemented `shouldThrowExactly` in master.

sksamuel:
This is now part of 3.0.0 RC1.
If you have the time, please give it a try. There are some breaking changes that should be aware of by reading the changelog here: https://github.com/kotlintest/kotlintest/blob/master/CHANGELOG.md

PR #281, opened 2018-04-11
Simplify how Travis calls Gradle

sschuberth:
There is no need to call the wrapper task with a version here as we
already use the wrapper script, and that one will bootstrap the Gradle
version as specified in gradle/wrapper/gradle-wrapper.properties.

Also, there is no need to call the clean task as each Travis build
starts from scratch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,4 @@ cache:
 
 install: /bin/true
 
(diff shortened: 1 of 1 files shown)


sksamuel:
:+1:

PR #282, opened 2018-04-11
Enable the Gradle build cache to reduce build time

sschuberth:
This is a cache that is persistent across clean builds, see

https://docs.gradle.org/current/userguide/build_cache.html

Adjust cached directories on Travis accordingly.


sksamuel:
Nice!

PR #284, opened 2018-04-11
Add an AppVeyor configuration for Windows CI

sschuberth:


sksamuel:
wow quick

sksamuel:
I think they are running because this output only comes from tests:

```
52.20% odd number
47.80% even number
0.10% zero
50.80% odd number
49.20% even number
0.10% zero
61.00% no whitespace
39.00% has whitespace
50.30% odd number
49.70% even number
0.10% zero
[WARN] When comparing doubles consider using tolerance, eg: a shouldBe (b plusOrMinus c)
[WARN] When comparing doubles consider using tolerance, eg: a shouldBe (b plusOrMinus c)
:kotlintest-tests:test SKIPPED
:kotlintest-extensions:kotlintest-extensions-allure:compileKotlin
```

I think that SKIPPED line is a misnomer because gradle creates the junitTest scope?

sschuberth:
On the other hand, I do see test result artifacts for that build:

https://ci.appveyor.com/project/sksamuel/kotlintest/build/1.0.3/tests

Need to look into this a bit more.

sksamuel:
Easist way to test it is to break a test.

sksamuel:
Like this one failed (not on purpose!)
https://ci.appveyor.com/project/sksamuel/kotlintest/build/1.0.4

Maybe something to do with the gradle cache?

sschuberth:
Nope, that's probably a Java 9 issue which I'm using on AppVeyor. Does kotlintest support Java 9?

sksamuel:
I've never tried. I assumed it would just work - it's rare for JDK's to not be backwards compatible. I'll add it to the travis grid.

sschuberth:
Anyway, let me revert to Java 8 on AppVeyor for now.

sksamuel:
Ok, I've added it to travis.

sksamuel:
https://travis-ci.org/kotlintest/kotlintest/builds/365333173

sschuberth:
Yeah, same error as on AppVeyor with Java 9: https://travis-ci.org/kotlintest/kotlintest/jobs/365333175#L776

sksamuel:
Looks like they removed javax/xml/bind/annotation/XmlElement from Java9.

sschuberth:
With Java 8 AppVeyor works again: https://ci.appveyor.com/project/sksamuel/kotlintest/build/1.0.11

sksamuel:
Ok cool. If my fix works for java 9 on linux I'll try java 9 again on windows.

sksamuel:
Although I suppose it doesn't matter a great deal which JDK to use. JDK 8 is the long term support release.

sschuberth:
Actually, let me either add a matrix build for Java 8 and Java 9 on AppVeyor similar to Travis then, or we simply test Java 9 only on Linux, which probably is enough.

sksamuel:
Failed anyway. https://travis-ci.org/kotlintest/kotlintest/builds/365337808

sksamuel:
That was my fault actually. Trying again.

sksamuel:
Ok green on both.

PR #287, opened 2018-04-13
Add org.junit.platform.gradle.plugin in README

tgirard12:
When I switch from 2.x to 3.x Gradle Tests won't start because I haven't had the junit plugin.
I think it should be in the README

#288


sksamuel:
Thanks!

PR #292, opened 2018-04-15
AppVeyor: Only create %HOME%\.gradle if it does not exist

sschuberth:
This avoids a bogus build failure if it already exists.


sksamuel:
Awesome

PR #315, opened 2018-04-30
Change jUnit 4 Runner to show nested tests.

psa-anddev:
This version of the jUnit 4 runner should display test nested. Some caveats:
- I have only tested success and error cases. I believe the behaviour for ignore should work too but it would be worth writing some unit tests to make sure this is the case.
- As I've seen in the jUnit reports, only the leaf is marked as failed. This differs from the 2.x version which would mark the intermediate nodes as failed too. It would be worth looking at it.

Finally, this PR contains a Spec which contains a failing Spec so that the error case can be tested through the runner. The Gradle tests for this module, i.e. the jUnit 4 runner has been configured to ignore that Spec in order to avoid false positives.


sksamuel:
:+1:

PR #334, opened 2018-05-17
[Feature] Inspectors: inspecting & extracting

Tallic:
Thought I provide you with two operators I used in my project.

**Context**
I testest a graph like structure and made up a small test graph in which I wanted to check whether properties were correctly set. Testing graphs is always annoying and a lot of work.

**Problem**
I needed to check several properties, leading to a lot of code similar to this one:
```kotlin
node1.prop1 shouldBe xxx
node1.prop2 shouldBe yyy
node1.prop3 shouldBe zzz

forOne(node1.prop4){
  it.name shouldBe "name1"
}
forOne(node1.prop4){
  it.name shouldBe "name2"
}
```
**Solution/Helpers**
I defined two very simple methods that allow a simpler to read and comprehend version.
**inspecting**
*inspect* exposes the properties and provides an assertion/mental context
```kotlin
inspecting(node1){
  prop1 shouldBe xxx
  prop2 shouldBe yyy
  prop3 shouldBe zzz
}
```

**extracting**
Extracting allows to pull out properties from a collection to make *or* assertions. *forOne* only allows *and* assertions, however, i might have missed something in your documentation.
```kotlin
extracting(node1.prop4){ name }
  .shouldContainAll("name1", "name2")
```

In sum, I think these two fit well in the kotlintest dsl and are a simple but useful addition.
Furthermore, I think all inspectors should work similar to *inspecting*, i.e., providing the context relative to the inspected object; however, this is another topic.

### DIFF
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/Inspectors.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/Inspectors.kt
@@ -82,6 +82,70 @@ fun <T> forNone(col: Collection<T>, f: (T) -> Unit) {
   }
 }
 
+/**
+ * Inspecting allows to assert the properties of an object in a typed fashion providing a proper testing context.
+ *
+ * The **simple example** shows how inspecting can build up a assertion context making the tests more readable.
+ * ```
+ * inspecting(person){
+ *  name shouldBe "John Doe"
+ *  age shouldBe 20
+ * }
+ * ```
+ *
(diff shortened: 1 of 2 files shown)


sksamuel:
Simple and effective - great!

sksamuel:
I moved them to their own files and fixed it for you.
https://github.com/kotlintest/kotlintest/commit/611eee578e931e0b8c075743339e279c5f3be7f2

PR #338, opened 2018-05-21
fix URI matchers

frozenice:
This fixes some errors in the URI matchers (wrong comparisons).

### DIFF
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/uri/matchers.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/uri/matchers.kt
@@ -1,4 +1,4 @@
-package  io.kotlintest.matchers.uri
+package io.kotlintest.matchers.uri
 
 import io.kotlintest.Matcher
 import io.kotlintest.Result
@@ -36,8 +36,8 @@ fun haveHost(host: String) = object : Matcher<URI> {
   )
 }
 
-fun URI.shouldHavePath(path: String) = this should io.kotlintest.matchers.uri.haveScheme(path)
-fun URI.shouldNotHavePath(path: String) = this shouldNot io.kotlintest.matchers.uri.haveScheme(path)
+fun URI.shouldHavePath(path: String) = this should io.kotlintest.matchers.uri.havePath(path)
+fun URI.shouldNotHavePath(path: String) = this shouldNot io.kotlintest.matchers.uri.havePath(path)
 fun havePath(path: String) = object : Matcher<URI> {
   override fun test(value: URI) = Result(
       value.path == path,
@@ -47,7 +47,7 @@ fun havePath(path: String) = object : Matcher<URI> {
 }
 
 fun URI.shouldHaveParameter(key: String) = this should io.kotlintest.matchers.uri.haveParameter(key)
-fun URI.shouldNotHaveParameter(key: String) = this shouldNot io.kotlintest.matchers.uri.haveScheme(key)
+fun URI.shouldNotHaveParameter(key: String) = this shouldNot io.kotlintest.matchers.uri.haveParameter(key)
 fun haveParameter(key: String) = object : Matcher<URI> {
   override fun test(value: URI) = Result(
       value.query.split("&").any { it.split("=").first() == key },


sksamuel:
Thank you!

sksamuel:
I've released version 3.1.3 with this fix.

frozenice:
While I have you @sksamuel, something off-topic. :) I noticed [FreeSpecLambdaTest.kt](https://github.com/kotlintest/kotlintest/blob/newconfig/kotlintest-tests/src/test/kotlin/com/sksamuel/kotlintest/tests/specs/FreeSpecLambdaTest.kt) in the newconfig branch (you were linking to it in the Gradle issue or somewhere).
When will this be available? I had the desire to structure my tests like this a couple of times in the past, but something wasn't quite right every time with all the `*Spec`s (IIRC e.g. the naming).
What are the plans for this branch? In my quick search, I couldn't find any notes about it.

sksamuel:
It was introduced in 3.1.0. https://github.com/kotlintest/kotlintest/blob/master/doc/styles.md#free-spec

So you should be good to go.

frozenice:
Yeah I know about FreeSpec for a long time, but note that there is a minus required in this example (which was also true in my recent tests, it wouldn't work otherwise). The FreeSpecLambdaTest doesn't have the minus before the closures.

sksamuel:
Ah I see. I thought about introducing it, and there's nothing stopping it really. It's just once it's introduced there's no going back and at the moment we can use the non-minus cases to determine leaf nodes.

What's the reason for wanting it, just cosmetic ?

frozenice:
Not having to think about where to put a minus and where not. :)
I'll check out the current state when I'm at home this evening and report back what doesn't work for me.

sksamuel:
Ok, I think the current state would be pretty much as it is now :)

frozenice:
Okay, this is getting pretty off-topic and nothing you can fix, but something to keep in mind (see note at the end).

So, what I want is this:
image
Which is what I get with JUnit5 (Jupiter 5.2.0), the IntelliJ JUnit runner and this simple java test-suite:
```java
class MyTest {
  @Test
  void testA() {
  }

  @Nested
  class NestedClass {
    @Test
    void testB() {
    }

    @Nested
    class AnotherTestClass {
      @Test
      void testC() {
      }
    }
  }
}
```

Running the same thing through Gradle's (4.7) `test` task (the same Gradle project), I get:
image
No nesting (also none in the generated HTML report, the nested classes show up as separate classes). :(

I found this [note in the Gradle docs](https://docs.gradle.org/current/userguide/java_testing.html#enablingJUnitPlatform):
> There are some known limitations of using JUnit 5 with Gradle, for example that tests in static nested classes won’t be discovered and classes are still displayed by their class name instead of `@DisplayName`. These will be fixed in future version of Gradle.

That doesn't really apply to my code (the nested classes aren't static, per the [JUnit docs](https://junit.org/junit5/docs/current/user-guide/#writing-tests-nested)), but it seems to show that Gradle isn't fully supporting JUnit 5 features (nesting, in my case).
So, nothing you can fix, but Gradle must implement. I also tried Gradle 4.8 RC and 4.9 nightly, but they produced the same output, which leads me to believe this won't really happen in the near future.

**The note**: I have no deep knowledge of JUnit's inner workings and haven't dived into kotlintest's code, but it would be nice if kotlintest with nested closures would produce the same output as plain JUnit Java code does with nested classes, as soon as Gradle supports showing it (that is, if kotlintest doesn't already do this).

Also: Comparing the nested Java classes, I would find it nice, if the `FreeSpec` (and maybe `StringSpec`? I don't really know what the intended differences between those two are) would support nesting arbitrarily without the minus - it just looks cleaner. :)

Sorry for the long post, have a good night!

sksamuel:
Hi,

In reverse order :) StringSpec is just a single level of nested strings. FreeSpec will do this and more, but StringSpec is there if you want to enforce one level only. I think with the number of Specs - a bit like ScalaTest - there's a bit of duplication and it might be  overkill, but it means everyone is happy. I know when I do Scala for my day job, that sometimes I use WordSpec, sometimes FlatSpec (one we can't do in Kotlin), sometimes FunSuite. Depends on what I'm testing.

As for the gradle stuff. You're right, and it's a gradle issue. This thread kind of covers it. https://github.com/kotlintest/kotlintest/issues/336

Gradle at the moment makes lots of (incorrect) assumptions about the tests. They're making moves to fix it which is good. Intellij does it correctly. If you are seeing a difference between JUnit @nested and KotlinTest when running via _intellij_ that's a bug. Gradle, that will be "flat" until 4.9 is out.

PR #339, opened 2018-05-23
Improve shouldBe array overloads

ajalt:
This PR contains three changes to the `shouldBe` overloads for arrays:

* It adds overloads for the remaining array types (float, char, short, and byte)
* It changes the types to be nullable
* It uses `asList` instead of `toList` to avoid copying the arrays.


sksamuel:
Awesome thanks!

PR #340, opened 2018-05-23
fix eventually nanos conversion

dtanner:
After upgrading from kotlintest 2->3, some of our tests failed because eventually wasn't waiting nearly long enough.  Looks like as part of the migration to use Duration, there was some bad math in converting millis to nanos.  There are `1_000_000` nanos in a millisecond, not `1_000` (I had to look this up - I would've thought it was a thousand too!).

Adjusting the test as I did below without modifying Eventually.kt shows the issue, and fails with an error like this:
`java.lang.AssertionError: Test failed after 3 seconds; attempted 942 times`


sksamuel:
Great spot thanks!

sksamuel:
I've released 3.1.4 with this fix. Thanks again.

dtanner:
👍 - thanks for the quick release.

PR #341, opened 2018-05-23
Fixing argument names in ReplaceWith of deprecated matchers

douevencode:
Currently arguments name in 'ReplaceWith' don't match so Intellij substitutes new method incorrecly. This commit fixes that and Intellij can do automatic substitution - which does pretty much nothing in this case, but it's better than doing it incorrectly. Still it can't substitute imports, but I'm afraid there is nothing we can do about it.


sksamuel:
:+1:

PR #344, opened 2018-05-25
Fixing typos in reference.md

michalbrz:


sksamuel:
Wow so many typos. I must have been under the influence of [malbec](https://en.wikipedia.org/wiki/Malbec) when writing those docs.

Thanks for the PR.

PR #345, opened 2018-05-28
Fixing function return type in documentation for forAll()

michalbrz:
forAll() expects that passed lambda returns Boolean, not Unit


sksamuel:
Yes good spot. I am encouraging people to use assertAll now as well.

PR #348, opened 2018-05-31
Fix deprecation warning

ajalt:
We don't need to import the depreacted `shouldBe`, since it just forwards back to io.kotlintest.shouldBe.


sksamuel:
:+1:

PR #349, opened 2018-06-01
Add extension functions for table assertions.

ajalt:
These extensions simplify the common case of creating a table and
immediately asserting on its contents. They let you write code like
this:

```kotlin
table(...).forAll { a, b -> ...}
```

instead of needing an intermediate variable:

```kotlin
val myTable = table(...)
forAll(myTable) { a, b -> ... }
```

I deliberately did not change any documentation in this PR. If it's
accepted, documentation can be updated in another commit.


sksamuel:
Excellent!

PR #350, opened 2018-06-01
Allow nullable arguments to more matcher functions.

ajalt:
In the case of the Double and String matchers, nullable arguments would
use the generic overload in some cases. This would compile, but would not produce the
expected results.


sksamuel:
:+1:

PR #351, opened 2018-06-03
Run tests for all rows in a table, even after errors.

ajalt:
This commit updates kotlintest's table testing to run all rows in the table and report any failures at the end of the test. This behavior is similar to most other testing libraries that offer a way to parameterize tests.

The error reporting algorithm looks like this:
If there is exactly one failure, throw it directly. Otherwise, throw a new error with a message that contains the message of each individual error and the stack frame that it occurred on.

For example:
```
The following 3 assertions failed:
1) Test failed for (a, none), (b, null) with error kotlin.KotlinNullPointerException
    at com.example.ExampleTest$test_function$1.invoke(ExampleTest.kt:28)
2) Test failed for (a, foo), (b, bar) with error expected:<[bar]> but was:<[foo]>
    at com.example.ExampleTest$test_function$1.invoke(ExampleTest.kt:31)
3) Test failed for (a, baz), (b, qux) with error expected:<[qux]> but was:<[baz]>
    at com.example.ExampleTest$test_function$1.invoke(ExampleTest.kt:31)
```

IntelliJ will allow you to click the line number of each error.

### DIFF
--- a/kotlintest-core/src/main/kotlin/io/kotlintest/tables/MultiAssertionError.kt
+++ b/kotlintest-core/src/main/kotlin/io/kotlintest/tables/MultiAssertionError.kt
@@ -1,0 +1,27 @@
+package io.kotlintest.tables
+
+/** An error that bundles multiple other [Throwable]s together */
+class MultiAssertionError(errors: List<Throwable>) : AssertionError(createMessage(errors)) {
+  companion object {
+    private fun createMessage(errors: List<Throwable>) = buildString {
+      append("\nThe following ")
+
+      if (errors.size == 1) {
+        append("assertion")
+      } else {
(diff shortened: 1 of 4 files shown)


sksamuel:
Great addition. The build is red so I cannot merge at the moment if you want to check that.

Also, take a look at my commits from last night where I added a new style of table driven testing that automatically infers the parameter name. I believe this should supercede the current code as it's less boilerplate.

ajalt:
Sorry about the build failures. It should be fixed now. I think the new name inference is excellent! This change will work with both `forAll` and the new `forall`, since `forall` delegates to `forAll`.

sksamuel:
:+1:

PR #352, opened 2018-06-04
Allow nullable receivers for string matchers.

ajalt:
Allow nullable receivers for string matchers.

This PR allows receivers for string matchers to be of type `String?`. All matchers will fail if the actual string is null. Their inverses will also fail in the string is null. This matches assertj's behavior, and is, in my opinion, the least surprising behavior. (e.g. [1](https://github.com/joel-costigliola/assertj-core/blob/1bd10d9ee3845167c708cd2b984668f38da45544/src/main/java/org/assertj/core/internal/Strings.java#L624), [2](https://github.com/joel-costigliola/assertj-core/blob/1bd10d9ee3845167c708cd2b984668f38da45544/src/main/java/org/assertj/core/internal/Strings.java#L641))

This also adds at least one test for each string matcher and matcher extension, and fixes a couple of bugs in the matchers that didn't previously have coverage (e.g. `haveMaxLength` was checking `value.length >= length`, and `haveLineCount` would fail if the string used line separators that didn't match the current system))

Finally, this PR removes `!!` from tests that were using string assertions on nullable strings.

### DIFF
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/Matcher.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/Matcher.kt
@@ -49,6 +49,36 @@ interface Matcher<T> {
 }
 
 /**
+ * A [Matcher] that asserts that the value is not `null` before performing the test.
+ *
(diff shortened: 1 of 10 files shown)


sksamuel:
Superb stuff.

PR #357, opened 2018-06-07
Remove duplicate kotlin gradle dependencies.

ajalt:
A number of submodules declared the kotlin buildscript dependency. This
is unnecessary, but not harmful. Additionally, some of the submodules
redeclared `ext.kotlin_version` to an older version of kotlin, which caused the 
reflect jar to be older than the stdlib in some cases.

### DIFF
--- a/kotlintest-samples/kotlintest-samples-gradle/build.gradle
+++ b/kotlintest-samples/kotlintest-samples-gradle/build.gradle
@@ -1,5 +1,15 @@
 apply plugin: 'org.junit.platform.gradle.plugin'
 
+buildscript {
+    ext.kotlin_version = '1.2.41'
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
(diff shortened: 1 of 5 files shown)


kotlintest-samples/kotlintest-samples-gradle/build.gradle
```diff
@@ -1,15 +1,5 @@
...
-    ext.kotlin_version = '1.2.31'
-    repositories {
-        mavenCentral()
-    }
-    dependencies {
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-    }
-}
-
 dependencies {
```
sksamuel: Won't this project need the kotlin gradle plugin for the non test dependencies?

ajalt: Good catch. Sorry about that, I didn't realize that the samples weren't submodules.

PR #359, opened 2018-06-09
Optimize sorted matcher

ajalt:
Instead of sorting the list and comparing every value, we can iterate
through the list once and stop when we find an element that's larger
than the following element. This also allows us to add the out-of-order
element and its index to the error message.

A similar optimization was added to the `containsInOrder` matcher, with the additional benefit that it will now work on unsorted collections.


ajalt:
The build failure is the same as the master branch: the `TestBoom` test added in 0df6552

sksamuel:
Yep, fixing the master build now as part of that issue.

sksamuel:
This is a great pr, thanks.

PR #360, opened 2018-06-09
Add matchers containsExactly and containsExactlyInAnyOrder

ajalt:
These are equivalent to the similarly named matchers in assertj.


sksamuel:
Excellent.

PR #368, opened 2018-06-25
Use less ambiguous string representations in equality errors

ajalt:
`toString` produces ambiguous results for a number of common types. For example:

```kotlin
"1" shouldBe 1L
```

```
expected: 1 but was: 1
```

or

```kotlin
listOf(1f) shouldBe listOf(1.0)
```

```
expected: [1.0] but was: [1.0]
```

etc.

This PR formats objects the way their literal values appear in code.

e.g.

```
expected: "1" but was: 1L
```

Now that the representations are less ambiguous, we can use the JUnit assertions for all types.


sksamuel:
Awesome!

PR #371, opened 2018-06-26
Fix error when running single test in kotlintest-tests

ajalt:
Currently, `AutoCloseListener` is registered project wide, but closables it checks are only closed when `AutoCloseTest` is run. This is fine if you run the entire test suite, but when running an individual test, it causes the listener's assertions to fail.

This PR changes `AutoCloseTest` so that the closables are considered closed unless that test has actually run.


sksamuel:
Great. This was bugging me actually.

I'll release 3.1.8 tomorrow or so.

PR #372, opened 2018-06-26
Fix table testing forNone and Double between matcher

ajalt:
Currently, `forNone` will never fail. If the callback succeeds, an `AssertionError` is thrown, but it's immediately caught by the try/catch that's meant to catch failures in the callback.

This PR also fixes the `between` matcher for Doubles, which was incorrectly passing its tests due to `forNone` not working.


sksamuel:
:+1:

PR #373, opened 2018-07-01
Add soft assertions

ajalt:
This PR adds the ability to perform multiple assertions in a block without the test stopping on the first failed assertion. The `verifyAll` function takes a lambda, and any assertion failures inside the block will be gathered, then thrown when the block exits.

Many other testing framework have some version of this functionality:

* [AssertJ](https://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#soft-assertions)
* [TestNG](https://jitpack.io/com/github/cbeust/testng/master/javadoc/org/testng/asserts/SoftAssert.html)
* [Spock](http://spockframework.org/spock/docs/1.1/all_in_one.html#_what_s_new_in_this_release_4)

The implementation works by changing the `should*` dsl functions to send their failures to an `ErrorCollector`, which will gather the errors when inside a `verifyAll` block, or throw the error immediately otherwise. The collected errors are held in thread-local storage to prevent race conditions in the unlikely case that multiple threads invoke `verifyAll` concurrently.

Existing assertion functions that take a lambda often rely on throwing an exception for control flow, and so will continue to throw immediately instead of collecting their errors. This includes `eventually`, `shouldThrow*`, `shouldTimeout`, and property and table testing.

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -178,10 +178,10 @@ Soft Assertions
 Normally, assertions like `shouldBe` throw an exception when they fail.
 But sometimes you want to perform multiple assertions in a test, and
 would like to see all of the assertions that failed. KotlinTest provides
-the `verifyAll` function for this purpose.
+the `assertSoftly` function for this purpose.
 
 ```kotlin
-verifyAll {
+assertSoftly {
   foo shouldBe bar
   foo should contain(baz)
 }
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -172,6 +172,33 @@ fun String.shouldNotContainFoo() = this shouldNot containFoo()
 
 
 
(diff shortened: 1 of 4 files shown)


ajalt:
`verifyAll` is not the most descriptive name. Let me know if you think of a better one.

sksamuel:
verifyAll is currently used by property testing as well. I'll try to think of a different name.

ajalt:
`assertSoftly`, `assertBlock`, `assertGroup`?

sksamuel:
`soft`, as daft as the name is, is what assertJ uses so why not stick with that. So I'm happy with `softAssert { } `

sksamuel:
:+1:

PR #374, opened 2018-07-02
Fix beLeft is comparing against Either.Right instead of Either.Left

logistark:
This PR should fix a bug where is comparing against Either.Right instead of comparing to Either.Left type.


sksamuel:
Oops. I wrote it wrong and then wrote the test wrong!

sksamuel:
Thanks for the fix.

PR #375, opened 2018-07-03
Update testCompile to prevent copy-paste errors

stephenjfox:
Didn't make a branch because it was a two-character change.
Checked against the [Contribution guidelines](https://github.com/kotlintest/kotlintest/wiki/contribute) and I think I'm okay.

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -19,7 +19,7 @@ test {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


sksamuel:
:+1:

stephenjfox:
Should we be concerned about the failed checks?

sksamuel:
AppVeyor is always a bit odd.
Seems like it passed.
https://ci.appveyor.com/project/sksamuel/kotlintest/history

PR #383, opened 2018-07-17
`sortedWith` (and related) matchers.

outis:
Here are some matchers like `sorted()` (and those based on it), but that take custom orderings.

The commit also includes a new test subproject to test assertions.


sksamuel:
Awesome! but can I ask for a couple of things before merging.

1. That the matchers are placed in io.kotlintest.matchers.collections/matchers.kt
2. That we add extension function versions of them too to keep it consistent.

sksamuel:
1. Only the new ones, the old ones can stay where they are for now, to avoid people changing imports for existing ones. I may migrate them slowly across as we bump versions.
2. Yes sorry, I missed that first time around.
3. As for naming, I think either sounds alright, so whatever you prefer.

outis:
If it were just one of the two suffixes, "by" would have been my choice (it's also slightly more common). However, since "by" is used for methods that take selectors, it's probably least surprising to stick with the convention established by library classes. Speaking of "by", would it be worth it to add sort matchers taking selectors? They're not necessary, as the matchers with comparators are more general, but might be of use.

sksamuel:
I think it's fine as it is, but if you think otherwise I'd be happy to merge another PR.

PR #389, opened 2018-08-02
Removed unnecessary `Comparable<T>` upper-bound from `sortedWith` matchers.

outis:
This update allows `sortedWith` (&c.) to be used more generally. Since a comparator is passed to the `sortedWith` matcher family, they don't require that they're called on `Comparable` instances.

With this change, a comparator-taking matcher can be used wherever a (currently non-existent) selector-taking matcher is needed.


sksamuel:
:+1:

PR #391, opened 2018-08-07
`Sequence` matchers.

outis:
This adds matchers for `Sequence`, based on the `Collection` matchers.

Github lists 11 commits, but only the last isn't already on kotlintest:master.


sksamuel:
Awesome.

PR #392, opened 2018-08-08
Use unambiguous representations in collection matchers

ajalt:
I recently ran into some confusing output when comparing a list of `Int`s with a list of `Long`s. This PR uses the existing `stringRepr` function (used in equality comparisons) in the collection matchers in order to print values unambiguously.


sksamuel:
Great. Maybe worth putting in 3.1.x branch too.

PR #397, opened 2018-08-26
Fix shouldBe on arrays without static types

ajalt:
Currently, there are a number of `shouldBe` overloads for specialized types. If you call `shouldBe` on an `Any` that is actually one of these types, the overloads can't be used, and the comparisons fail incorrectly. 

This PR uses runtime inspections rather than the `shouldBe` overloads. This has the additional positive effect of fixing all the other `shouldBe`-equivalent functions like `should be(expected)` and `shouldNotBe`.


sksamuel:
Excellent.

Issue #402, opened 2018-09-11
Better support for comparing multi-line strings


sschuberth:
When doing

    actualResult shouldBe expectedResult

where both `actualResult` and `expectedResult` are (long) multi-line strings, console output in case of a mismatch could be improved. Currently, as can be seen e.g. [here](https://ci.appveyor.com/project/heremaps/oss-review-toolkit/build/1.0.3355#L2861), console output starts with the expected text. However, as that text is long and console output is limited, it does not even come as far as also showing the actual output. To address this, it would be nice if long multi-line output was condensed by stripping matching lines, and only keeping "diff hunks" similar like Git does to display only the areas of text where there actually is a difference.

sksamuel:
I've added my attempt.

You end up with a sequence of diffs. Each diff has a type (delete, change, addition), the line where it started, and the snippet that changed (with a line before/after for context).

The raw exception message is like this:

```
java.lang.AssertionError: 
expected: "[Change at line 1] Our neural pathways have become accustomed to your sensory input patterns.
      Mr. Crusher, ready a collision course with the Borg ship.
      A lot of things can change in twelve years, Admiral.

[Deletion at line 4] A lot of things can change in twelve years, Admiral.
      Take the ship into the Neutral Zone"

 but: was "[Change at line 1] Our neural pathways have become accustomed to your sensory input patterns.
      Mr. Crusher, ready a collision course with the Klingon ship.
      A lot of things can change in ten years, Admiral.

[Deletion at line 4] A lot of things can change in ten years, Admiral."
```

Which intellij picks up nicely and shows the diff window.

PR #403, opened 2018-09-11
WIP: Multi line string

sschuberth:
This is just to outline the idea behind issue #402.

### DIFF
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/dsl.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/dsl.kt
@@ -126,7 +126,32 @@ private fun equalsError(expected: Any?, actual: Any?): Throwable {
   return throwable
 }
 
-private fun equalsErrorMessage(expected: String, actual: String) = "expected: $expected but was: $actual"
+private fun equalsErrorMessage(expected: String, actual: String): String {
+  val multiLineThreshold = 50
+
+  val expectedLines = expected.lines()
+  val actualLines = actual.lines()
+
+  return if (expectedLines.count() > multiLineThreshold && actualLines.count() > multiLineThreshold) {
+    var commonLineCount = 0
+    val maxCommonLines = minOf(expectedLines.count(), actualLines.count())
+
+    while (commonLineCount < maxCommonLines && expectedLines[commonLineCount] == actualLines[commonLineCount]) {
+      ++commonLineCount
+    }
+
+    """
+    expected:
+    [...]
+    ${expectedLines.drop(commonLineCount).joinToString("\n")}
+    but was:
+    [...]
+    ${actualLines.drop(commonLineCount).joinToString("\n")}
+    """.trimIndent()
+  } else {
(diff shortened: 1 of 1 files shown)


sschuberth:
Another idea would be to use more sophisticated libraries like https://github.com/google/diff-match-patch or https://github.com/wumpz/java-diff-utils to generate the difference lines.

sksamuel:
I like the idea of using the google diff match patch. Interestingly, many years ago, I released it onto maven as at the time, google hadn't released it.

https://github.com/sksamuel/google-diff-match-patch

sschuberth:
Great, would you have the time and will to look into the implementation?

sksamuel:
Yes I think its a great idea, I'll add it to 3.2.

sschuberth:
Great! But I think the milestone should go to the issue, not this PR.

sksamuel:
Yes of course.

ajalt:
I agree that truncating error output is not ideal, but I'm :-1: on this. We used to do message compaction ourselves, but it was removed in #366 because it breaks IDE integration and makes it impossible to see the context when you do need it.

JUnit 5 removed message compaction for the same reasons, although if you use the JUnit 4 test runner, you'll still get the old behavior.

In my opinion, it would be worth seeing if it's possible to get the full strings to print on the command line rather than going back to mangled messages.

sschuberth:
When looking at the context of how `equalsErrorMessage()` is called e.g. at https://github.com/kotlintest/kotlintest/blob/47b7a428d056fcac7d9abb2cb44ab4953fad0f70/kotlintest-assertions/src/main/kotlin/io/kotlintest/dsl.kt#L119 I was under the impression that this is really just the custom user-facing message, and that JUnit would construct its own error from `expectedRepr` and `actualRepr` that it gets passed, so I was assuming this would not break anything.

Also, compared to #366 I wasn't thinking of a word-diff feature within a line, but only removing / compacting whole equal lines. I currently don't see how that would break IDE integration.

ajalt:
So the way that the IDE integration works depends on the runner you're using. 

When available, we use the JUnit assertions, which IntelliJ treats specially.
In JUnit5, the IDE ignores the message entirely, and uses the expected and actual values for its diff.
In JUnit4, we don't even pass the message to the exception, since the exception doesn't use it.
On other runners like TestNG, we throw a generic `kotlin.AssertionError` with the message.

It's that last case that would affect IDEs if you change the message. IDEs create a diff based on [pattern matching](https://github.com/JetBrains/intellij-community/blob/99614a63425774df58eea3ac92e2b20af1633663/plugins/junit_rt/src/com/intellij/junit4/ExpectedPatterns.java#L27), and right now your proposed change doesn't look like it matches any of them. However, that's easy to fix. We'd still lose the ability for IDEs to diff the entire file, which can definitely be confusing, since there's no way to see what the actual values were any more.

Now that I've given it some more thought, I agree with you that the negative impact is not likely to be significant. So I'll change my vote to +1, but let's make sure the compacted message matches one of the patterns the IDE looks for.

Issue #404, opened 2018-09-12
Add infix modifier to collections matchers


jensim:
```
val found = repo.findAll()
found shouldContainExactly listOf("one", "two", "three")
```

```
//Before
fun <T, C : Collection<T>> C?.shouldContainExactly(expected: C) = this should containExactly(expected)

//After
infix fun <T, C : Collection<T>> C?.shouldContainExactly(expected: C) = this should containExactly(expected)
```

This is not applicable to vararg param functions due to limitation in language, but it would look better in for all collections matchers where it is possible.

sksamuel:
Yes good idea.

jensim:
I forked master and created a branch with the `infix` added

sksamuel:
Excellent. Please make a PR and I'll merge.

sksamuel:
I'll include this in 3.1.10

sksamuel:
3.1.10 is released.

sksamuel:
You'll be waiting hours for that UI to update.
Once it's in the repo you're good to go
http://repo1.maven.org/maven2/io/kotlintest/kotlintest-assertions/

jensim:
Thanks @sksamuel !

PR #405, opened 2018-09-12
#404 Added infix keyword to collections matchers where it is applicable

jensim:
Added `infix` for applicable functions in collections matchers


sksamuel:
Excellent thanks.

PR #409, opened 2018-09-13
Adds the infix operator to functions

Kerooker:
To add a more idiomatic and kotlinistic way to match dates, this commit adds the infix notation to all Date Matchers that are infixable (e.g. has a single parameter and makes sense to be infix, such as `date1 shouldHaveSameYearAs date2`).

Takes care of https://github.com/kotlintest/kotlintest/issues/408

More about infix: https://kotlinlang.org/docs/reference/functions.html#infix-notation


sksamuel:
Great.

PR #414, opened 2018-09-14
Add documentation to Date matchers

Kerooker:
Currently a work in progress, don't merge yet.

This PR is for code review to be made while changes happen, as they can happen individually.


Kerooker:
I recommend viewing commit by commit while reviewing the docs

Kerooker:
@ajalt 
First things first: thanks for the review and the comment!

But for me, Intellij resolves Dokka correctly. What might be going on?
image

ajalt:
The screenshot I posted was from another project that I added the dependency to. I don't believe that the samples are included in sources jars, but please correct me if I'm wrong.

Kerooker:
Understood. Should I include the samples in the jar or is it better to just do what you said and add samples to the docs (not referencing at all)?

ajalt:
If I was writing it, I would add the samples to the docs, but if you can get the samples to resolve in the jar, that would work, too. I'll let @sksamuel weigh in if he prefers one over the other.

sksamuel:
I have no preference either way.

Kerooker:
Then to avoid having troubles with a bigger jar (not that bigger, but bigger anyways) and maintainability of the code (people might forget to update the samples when they update the code), I'll refactor the samples to be in-docs

Kerooker:
@ajalt Please review this again and see if this matches what we talked about.

Kerooker:
Don't merge just yet. There are still some docs in this file to handle. I should finish soon

Kerooker:
@ajalt and @sksamuel 
I believe it's all done.

Please review one more time before mergind. There was a lot of copy pasting stuff, and I might have forgotten to replace an instance or two. I don't believe it happened, but just to be sure..

PR #420, opened 2018-10-01
Add documentation to Date matchers

Kerooker:
Solves https://github.com/kotlintest/kotlintest/issues/410

PR #423, opened 2018-10-02
Matchers for shouldBeWithin duration

Kerooker:
Attention:
This was branched from https://github.com/kotlintest/kotlintest/pull/420 and should be merged if and only if #420 is approved. 

Solves https://github.com/kotlintest/kotlintest/issues/416


sksamuel:
Will merge but needs conflict resolving first.

Kerooker:
Will fix now

sksamuel:
Excellent yes.

PR #426, opened 2018-10-03
Add arrow-assertion Failure matcher that checks underlying throwable equality

osheari1:
Ability to check not just whether a Try returned a Failure, but whether the Failure contained a specific Throwable would be very useful. (https://github.com/kotlintest/kotlintest/issues/422)

Currently to check the underlying exception type, one needs to do something similar to:
```
val failure = Try.Failure(Exception("foo"))
failure should beFailure() // passes
shouldThrow<Exception> {
    failure.getOrElse { fail("Could not get exception") }
}
```
```
val failure = Try.Failure(Exception("foo"))
failure should beFailureWithThrowable<Exception>() // Passes test
failure should beFailureWithThrowable<SomeOtherException>() // Does not pass test
```

This was accomplished relatively easily with Kotlins inline function + reified generic types.

### DIFF
--- a/kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/try/matchers.kt
+++ b/kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/try/matchers.kt
@@ -39,9 +39,9 @@ fun beFailure() = object : Matcher<Try<Any>> {
   }
 }
 
-inline fun <reified A : Throwable> Try<Any>.shouldBeFailureWithThrowable() = this should beFailureWithThrowable<A>()
-inline fun <reified A : Throwable> Try<Any>.shouldNotBeFailureWithThrowable() = this shouldNot beFailureWithThrowable<A>()
-inline fun <reified A : Throwable> beFailureWithThrowable() = object : Matcher<Try<Any>> {
+inline fun <reified A : Throwable> Try<Any>.shouldBeFailureOfType() = this should beFailureOfType<A>()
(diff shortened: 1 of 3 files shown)


kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/try/matchers.kt
```diff
@@ -37,4 +37,20 @@ fun beFailure() = object : Matcher<Try<Any>> {
       is Try.Failure<*> -> Result(true, "Try should be a Failure", "Try should not be Failure")
     }
   }
-}
\ No newline at end of file
+}
+
+inline fun <reified A : Throwable> Try<Any>.shouldBeFailureWithThrowable() = this should beFailureWithThrowable<A>()
+inline fun <reified A : Throwable> Try<Any>.shouldNotBeFailureWithThrowable() = this shouldNot beFailureWithThrowable<A>()
+inline fun <reified A : Throwable> beFailureWithThrowable() = object : Matcher<Try<Any>> {
```
sksamuel: What do you think about the more concise name shouldBeFailureOfType ?

osheari1: Yeah that makes sense to me. I'll update the name.

sksamuel:
Excellent :+1:

sksamuel:
We could do this on master as well.

osheari1:
Yeah that should be straightforward. I'll make a new PR for master as well.

PR #427, opened 2018-10-03
Add arrow-assertion Failure matcher that checks underlying throwable equality

osheari1:
Same as PR (https://github.com/kotlintest/kotlintest/pull/426)
References (https://github.com/kotlintest/kotlintest/issues/422)

-----
Ability to check not just whether a Try returned a Failure, but whether the Failure contained a specific Throwable would be very useful. (https://github.com/kotlintest/kotlintest/issues/422)

Currently to check the underlying exception type, one needs to do something similar to:
```
val failure = Try.Failure(Exception("foo"))
failure should beFailure() // passes
shouldThrow<Exception> {
    failure.getOrElse { fail("Could not get exception") }
}
```
```
val failure = Try.Failure(Exception("foo"))
failure should beFailureOfType<Exception>() // Passes test
failure should beFailureOfType<SomeOtherException>() // Does not pass test
```

This was accomplished relatively easily with Kotlins inline function + reified generic types.


sksamuel:
Excellent thanks.

PR #429, opened 2018-10-04
Refactor double matchers to include an infix modifier

Kerooker:
Some of the double matchers can be used with an infix modifier, following the standards for all matchers in the lib (for instance, X shouldBe Y), and enables developers for a more idiomatic way to match double values.
This commit adds these infix modifiers to applicable functions

Solves https://github.com/kotlintest/kotlintest/issues/428


sksamuel:
Sensible addition :+1:

PR #432, opened 2018-10-04
BugFix: Double matcher shouldNotBeExactly

Kerooker:
The matcher shouldNotBeExactly is currently a copy of Double.shouldBeExactly, but it's a different method, causing a bug.
This commit fixes it and adds a test to make sure it doesn't happen again.

Solves https://github.com/kotlintest/kotlintest/issues/431


sksamuel:
Good spot.

PR #434, opened 2018-10-09
Moving Double Matchers to the same file; Deprecated DoubleMatchers.kt

Kerooker:
Double matchers are maintained in two different files. This commit moves all the matchers to the same file, refactoring the old matchers (DoubleMatchers.kt) to simply delegate to the correct file. These old matchers were also deprecated. In a future release, it's intended to remove them altogether.

Solves https://github.com/kotlintest/kotlintest/issues/433


sksamuel:
:+1:

Issue #435, opened 2018-10-09
Double should have oppostive matchers for shouldBePositive and shouldBeNegative


Kerooker:
Most of the matchers have a opposite matcher. For example
```
Double.shouldBeLessThan
Double.shouldNotBeLessThan
```

This behavior should be added to the `shouldBePositive` and `shouldBeNegative` matchers.

A use case for this, is for example matching the number 0.0;

Zero is not positive, and it should match `shouldNotBePositive`. However, the opposite of this is not `shouldBeNegative`, as it won't match 0.0 (zero is also not negative).

----

I propose that we add the following matchers to the Double Matchers:

`Double.shouldNotBePositive`
`Double.shouldNotBeNegative`

**ATTENTION:** These matchers might also be useful for Int, Float, Short.... But I didn't take a look at them for now. It's a point to be kept in the radar!

raverona:
I created a pull request to fix this issue, it also creates a `DoubleMatchersTest.kt` on a new `com.sksamuel.kotlintest.matchers.double` package inside the `kotlintest-tests/kotlintest-tests-core` module as I talked with @Kerooker beforehand and it's the aproach he is taking for these tests

raverona:
Let me know if you guys think more test cases are necessary for this feature

PR #437, opened 2018-10-10
Add unit tests for new matchers 'shouldNotBePositive', 'shouldNotBeNe…

raverona:
…gative' and unit tests for both

Fixes: #435

### DIFF
--- a/doc/matchers.md
+++ b/doc/matchers.md
@@ -90,6 +90,10 @@ For the extension function style, each function has an equivalent negated versio
 | `double.shouldBeLessThanOrEqual(n)` | Asserts that the double is less or equal to than the given value n |
 | `double.shouldBeGreaterThan(n)` | Asserts that the double is greater than the given value n |
 | `double.shouldBeGreaterThanOrEqual(n)` | Asserts that the double is greater than or equal to the given value n |
+| `double.shouldBePositive()` | Asserts that the double is positive |
+| `double.shouldBeNotPositive()` | Asserts that the double is not positive|
+| `double.shouldBeNegative()` | Asserts that the double is negative |
+| `double.shouldBeNotNegative()` | Asserts that the double is not negative |
 
 | Collections ||
 | -------- | ---- |
--- a/doc/matchers.md
+++ b/doc/matchers.md
@@ -91,9 +91,9 @@ For the extension function style, each function has an equivalent negated versio
 | `double.shouldBeGreaterThan(n)` | Asserts that the double is greater than the given value n |
 | `double.shouldBeGreaterThanOrEqual(n)` | Asserts that the double is greater than or equal to the given value n |
 | `double.shouldBePositive()` | Asserts that the double is positive |
-| `double.shouldBeNotPositive()` | Asserts that the double is not positive|
+| `double.shouldNotBePositive()` | Asserts that the double is not positive|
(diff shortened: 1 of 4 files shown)


kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -0,0 +1,89 @@
...
+                    row(Double.NEGATIVE_INFINITY)
+            )
+
+            forAll(table) {
+                it.shouldNotBePositive()
+                it shouldNotBe positive()
+            }
+
+            shouldThrow<AssertionError> {
+                0.1.shouldNotBePositive()
```
Kerooker: This will stop tests

raverona: ok, will change it

raverona:
Actually I discovered a bigger issue about this PR then the one with the tests. The thing is: should `Double.NaN` be considered not positive? (and not negative?). There are two possible interpretations: 

1. no, because NaN is not a number, so it does not make sense to say it's positive or negative or not positive or not negative.
2. yes, because NaN > 0 is false and not positive is the negation of that

if we follow the current implementation the interpretation in this case would be the second one, but which one is the correct or most accurate?

Kerooker:
A 3rd option to this would be "Don't consider anything, throw an exception."

If an user tries to assert that NaN is either positive or negative, don't let them do this. It shouldn't be an assertion at all.

raverona:
I pushed a commit that correct the tests according to the current implementation of the `positive` and `negative` functions, so this PR can be merged if its ok to expect this behavior (answering yes for `Double.NaN.shouldNotBePositive()`) when asking if `Double.NaN` is not positive and not negative

doc/matchers.md
```diff
@@ -90,6 +90,10 @@ For the extension function style, each function has an equivalent negated versio
 | `double.shouldBeLessThanOrEqual(n)` | Asserts that the double is less or equal to than the given value n |
 | `double.shouldBeGreaterThan(n)` | Asserts that the double is greater than the given value n |
 | `double.shouldBeGreaterThanOrEqual(n)` | Asserts that the double is greater than or equal to the given value n |
+| `double.shouldBePositive()` | Asserts that the double is positive |
+| `double.shouldBeNotPositive()` | Asserts that the double is not positive|
```
Kerooker: The documentation here is wrong. I believe you're aiming for `double.shouldNotBePositive`

doc/matchers.md
```diff
@@ -90,6 +90,10 @@ For the extension function style, each function has an equivalent negated versio
...
+| `double.shouldBeNegative()` | Asserts that the double is negative |
+| `double.shouldBeNotNegative()` | Asserts that the double is not negative |
```
Kerooker: Same as the positive case, I believe you're going for `double.shouldNotBeNegative`

raverona: oh that's correct, my bad

sksamuel:
:+1:

PR #442, opened 2018-10-12
Fixed message in Double between matcher

Kerooker:
The message was pretty bizarre due to a small typo. This commit fixes that message and makes it clearer.

Solves https://github.com/kotlintest/kotlintest/issues/441


sksamuel:
Good spot.

PR #444, opened 2018-10-12
Remove matchers "shouldNotBePositive" and "shouldNotBeNegative" from …

raverona:
…the documentation

These matchers do not belong in the documentation because it already states that every extension function matcher has a negated version, so only "shouldBePositive" and "shouldBeNegative" need to be documented


sksamuel:
:+1:

Issue #448, opened 2018-10-16
Spec and TestDescriptor modifications


lsrom:
Is it possible to change the `io.kotlintest.Spec` class to use *canonicalName* instead of *simpleName* as default? I'm talking about the `Class<out Spec>.displayName()` extension. If there is no issue with that breaking something, I'll gladly submit a merge request.

Also, can source class be included in TestDescriptor created in method `fun createTestCaseDescriptor(description: Description, type: TestType): TestDescriptor` from class `JUnitTestRunnerListener`. TestDescriptor from this method doesn't have set source which I need.

Can I submit these changes or are there any issues I'm not aware of? Thanks.

sksamuel:
I'm not sure about changing the display name, it's used by gradle and other things. 

As for the source class, yes I think it could be included, I don't see why not.

lsrom:
So can I add another method that will return the *canonicalName*? I really need it in my use case.

sksamuel:
Absolutely.

PR #449, opened 2018-10-22
Feature/add source to test descriptor

lsrom:
Adding class source to test descriptor  together with new Spec method `canonicalName`.

Solves #448


lsrom:
What can I do about the failing CI? It seems that `:kotlintest-extensions:kotlintest-extensions-allure:test` is stuck but I don't know why. If I run it, it runs fine and finishes in 58 ms. Any idea what should I try? Thanks

sksamuel:
It's something I did so don't worry about that.

PR Merged, thanks.

Issue #450, opened 2018-10-26
Extract double matchers each to it's own file


Kerooker:
The file `io.kotlintest.matchers.double.matchers.kt` has a lot of functions. Each of them brings a completely different scope.

All single responsibilities should be refactores to it's own file, avoiding confusion and avoiding a god-class, which it is right now.

Every one of these responsibilities should be extracted to it's own file. *however*, the package should not change. Any clients already using this function shouldn't have to change any import, and the functions should work the same way. For this, they must remain top-level functions.

Kerooker:
Merged.

PR #451, opened 2018-10-26
Implements tests for Double Matchers

Kerooker:
Given the necessity to continually improve the code, having unit tests guarantees that nothing will break when a change is made. This commit focuses on all Double Matchers, refactoring their tests to cover a lot more scenarios and corner cases.

One problem that this commit introduces is that there are way too many lines for a single class. This exposes a code smell on the matchers.kt file, in which ALL the double matchers are placed. This is bad, as each responsibility should be refactored to it's own class. This correction is possible when issue #450 is solved.

Solves https://github.com/kotlintest/kotlintest/issues/430

### DIFF
--- a/kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
+++ b/kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
@@ -1,72 +1,1061 @@
 package com.sksamuel.kotlintest.matchers.doubles
 
-import io.kotlintest.matchers.doubles.*
+import io.kotlintest.matchers.doubles.beGreaterThan
+import io.kotlintest.matchers.doubles.beGreaterThanOrEqualTo
+import io.kotlintest.matchers.doubles.beLessThan
+import io.kotlintest.matchers.doubles.beLessThanOrEqualTo
+import io.kotlintest.matchers.doubles.between
+import io.kotlintest.matchers.doubles.exactly
+import io.kotlintest.matchers.doubles.gt
+import io.kotlintest.matchers.doubles.gte
+import io.kotlintest.matchers.doubles.lt
+import io.kotlintest.matchers.doubles.lte
+import io.kotlintest.matchers.doubles.negative
+import io.kotlintest.matchers.doubles.positive
+import io.kotlintest.matchers.doubles.shouldBeBetween
+import io.kotlintest.matchers.doubles.shouldBeExactly
+import io.kotlintest.matchers.doubles.shouldBeGreaterThan
+import io.kotlintest.matchers.doubles.shouldBeGreaterThanOrEqual
+import io.kotlintest.matchers.doubles.shouldBeLessThan
+import io.kotlintest.matchers.doubles.shouldBeLessThanOrEqual
+import io.kotlintest.matchers.doubles.shouldBeNegative
+import io.kotlintest.matchers.doubles.shouldBePositive
+import io.kotlintest.matchers.doubles.shouldNotBeBetween
+import io.kotlintest.matchers.doubles.shouldNotBeExactly
+import io.kotlintest.matchers.doubles.shouldNotBeGreaterThan
+import io.kotlintest.matchers.doubles.shouldNotBeGreaterThanOrEqual
+import io.kotlintest.matchers.doubles.shouldNotBeLessThan
+import io.kotlintest.matchers.doubles.shouldNotBeLessThanOrEqual
+import io.kotlintest.matchers.doubles.shouldNotBeNegative
+import io.kotlintest.matchers.doubles.shouldNotBePositive
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.assertAll
+import io.kotlintest.should
+import io.kotlintest.shouldBe
+import io.kotlintest.shouldNot
 import io.kotlintest.shouldNotBe
-import io.kotlintest.specs.ShouldSpec
-import io.kotlintest.tables.*
+import io.kotlintest.shouldThrow
+import io.kotlintest.specs.FreeSpec
+import kotlin.Double.Companion.MAX_VALUE
+import kotlin.Double.Companion.NEGATIVE_INFINITY
+import kotlin.Double.Companion.NaN
+import kotlin.Double.Companion.POSITIVE_INFINITY
+import kotlin.math.absoluteValue
 
-class DoubleMatchersTest : ShouldSpec() {
-    init {
-
-        should("Not be positive") {
-            val acceptedValues = table(
-                    headers("a"),
-                    row(-1.0),
-                    row(-0.00001),
-                    row(-42.0),
-                    row(0.0),
-                    row(Double.NaN),
-                    row(Double.NEGATIVE_INFINITY)
-            )
-
-            forAll(acceptedValues) {
-                it.shouldNotBePositive()
-                it shouldNotBe positive()
+class DoubleMatchersTest : FreeSpec() {
+  
+  private val nonNumericDoubles = listOf(NaN, POSITIVE_INFINITY, NEGATIVE_INFINITY)
+  
+  private val numericDoubles = Gen.double().filterNot { it in nonNumericDoubles }
+  private val nonMaxValueDoubles = numericDoubles.filter { it != MAX_VALUE }
+  
+  private val delta = 0.0001
+  
+  init {
+    "Exactly Matcher" - {
+      
+      "Every numeric Double" - {
+        
+        "Should be exactly" - {
+          
+          "Itself" {
+            assertAll(numericDoubles) {
+              it shouldExactlyMatch it
             }
-
-            val notAcceptedValues = table(
-                    headers("a"),
-                    row(0.1),
-                    row(1.0),
-                    row(Double.MAX_VALUE),
-                    row(Double.MIN_VALUE),
-                    row(Double.POSITIVE_INFINITY)
-            )
-
-            forNone(notAcceptedValues) {
-                it.shouldNotBePositive()
-                it shouldNotBe positive()
-            }
+          }
         }
-
-        should("Not be negative") {
-            val acceptedValues = table(
-                    headers("a"),
-                    row(1.0),
-                    row(Double.MIN_VALUE),
-                    row(Double.MAX_VALUE),
-                    row(0.001),
-                    row(0.0),
-                    row(Double.NaN),
-                    row(Double.POSITIVE_INFINITY)
-            )
-
-            forAll(acceptedValues) {
-                it.shouldNotBeNegative()
-                it shouldNotBe negative()
+        
+        
+        "Should not be exactly" - {
+          
+          "Any number smaller than itself" {
+            assertAll(nonMaxValueDoubles) {
+              it shouldNotMatchExactly it - delta
             }
-
-            val notAcceptedValues = table(
-                    headers("a"),
-                    row(-0.1),
-                    row(-1.0),
-                    row(Double.NEGATIVE_INFINITY)
-            )
-
-            forNone(notAcceptedValues) {
-                it.shouldNotBeNegative()
-                it shouldNotBe negative()
+          }
+          
+          "Any number bigger than itself" {
+            assertAll(nonMaxValueDoubles) {
+              it shouldNotMatchExactly it + delta
             }
+          }
+          
+          "Anything that's not numeric" {
+            assertAll(numericDoubles) {
+              nonNumericDoubles.forEach { nonNumeric ->
+                it shouldNotMatchExactly nonNumeric
+              }
+            }
+          }
         }
+        
+      }
+      
+      "The non-numeric double" - {
+        "NaN" - {
+          
+          "Should not be exactly" - {
+            
+            "Any number" {
+              assertAll(numericDoubles) {
+                NaN shouldNotMatchExactly it
+              }
+            }
+            
+            "NaN" {
+              NaN shouldNotMatchExactly NaN
+            }
+            
+            "The infinities" {
+              NaN shouldNotMatchExactly POSITIVE_INFINITY
+              NaN shouldNotMatchExactly NEGATIVE_INFINITY
+            }
+          }
+        }
+        
+        "Positive Infinity" - {
+          
+          "Should be exactly" - {
+            "Itself" {
+              POSITIVE_INFINITY shouldExactlyMatch POSITIVE_INFINITY
+            }
+          }
+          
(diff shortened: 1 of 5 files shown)


Kerooker:
@sksamuel and @ajalt 

Could you help me with this by code-reviewing?

I know this class is massive right now, but it will test every possible scenario. With a possible refactor from #450, we will get this into a smaller size.

However, #450 is only possible with correct tests for the classes, which are in this PR.

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -1,72 +1,1061 @@
...
-                it shouldNotBe positive()
+          }
+        }
+        
+        
+        "Should not be exactly" - {
+          
+          "Any number smaller than itself" {
+            assertAll(nonMaxValueDoubles) {
+              it shouldNotMatchExactly it - delta
```
ajalt: Tests like this will only pass for values with an exponent relatively close to 1. For example:
`1e300 + 0.0001 === 1e300`
`1e300 + 0.0001 === 0.0001`

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -1,72 +1,1061 @@
...
+  }
+  
+  private fun shouldThrowAssertionError(message: String, vararg expression: () -> Any?) {
+    expression.forEach {
+      val exception = shouldThrow<AssertionError>(it)
+      exception.message shouldBe message
+    }
+  }
+  
+  private infix fun Double.shouldExactlyMatch(other: Double) {
```
ajalt: I like these functions that increase coverage without duplication 👍

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -1,72 +1,1061 @@
...
+    "Between matcher" - {
+      
+      "Every numeric double that is not Double.MAX_VALUE" - {
+        
+        "Should match between" - {
+          
+          "When it's equal to the first number of the range" - {
+            
+            "With tolerance" {
+              onDoubleRange { double, _ ->
```
ajalt: If you aren't using `valueToVariate`, then this is taking 10x the time without really helping. I would suggest useing `assertAll` directly in these cases.

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -1,72 +1,1061 @@
...
+          "Itself" {
+            onDoubleRange { double, _ ->
+              double shouldNotMatchLessThan double
+            }
+          }
+          
+          "Numbers smaller than itself" {
+            onDoubleRange { double, valueToVariate ->
+              double shouldNotMatchLessThan double - delta
+              double shouldNotMatchLessThan double - valueToVariate
```
ajalt: I have the same comment about `valueToVariate` that I did about `delta`: it's not valid if `double` has a large positive or negative exponent.

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
Kerooker: I didn't quite understand what you meant.

Do you believe this test may sometimes fail?

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
Kerooker: Good call

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
Kerooker: I'm not quite sure if I understood this. Do you have suggestions on how to deal with this?

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
Kerooker: They're good because we expect the exact same behavior now. But this is dangerous, because if I change any behavior in the future, the old tests wouldn't cover them. This way we can cover on all fronts, including trying the exception ones

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
ajalt: It looks like you're trying to create a number `it - 0.0001` that is different from `it`. But for most `Double` values, `it - 0.0001` is exactly equal to `it`.

[Double precision floating point](https://en.wikipedia.org/wiki/Double-precision_floating-point_format) can only represent about 16 significant decimal digits. `1e300 - 0.0001` would require 300 significant decimal digits, so the result just rounds back to the original `1e300`.

So your delta needs to be relative to the value you're trying to change. For example, rather than `it - delta`, you could do something like `it * .9`, which should work for all numbers with a magnitude larger than ~1e-323 (the smallest positive denormal number).

Kerooker: That's very nice to know! Learn something every day, I guess.

I'll add these corner case numbers to the generated double lists

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -1,72 +1,1061 @@
...
+                                                      tolerance: Double,
+                                                      message: String = "$this should not be smaller than $b and should not be bigger than $a"
+  ) {
+    
+    shouldThrowAssertionError(message,
+                              { this.shouldNotBeBetween(a, b, tolerance) },
+                              { this shouldNotBe between(a, b, tolerance) })
+  }
+  
+  private fun onDoubleRange(block: (double: Double, valueToVariate: Double) -> Unit) {
```
sksamuel: Are you using the valueToVariate parameter here ?

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -1,72 +1,1061 @@
...
+    (0.1..1.0 step 0.2).forEach { valueToVariate ->
+      assertAll(300, nonMaxValueDoubles) {
+        block(it, valueToVariate)
+      }
+    }
+  }
+  
+  private infix fun ClosedRange<Double>.step(step: Double): Iterable<Double> {
+    val sequence = generateSequence(start) { previous ->
+      if (previous == Double.POSITIVE_INFINITY) return@generateSequence null
```
sksamuel: I think return@ is a code smell. It's more functional to have the return be the last value all the time. I'd simply change this to have the next two lines inside an else {}

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -1,72 +1,1061 @@
...
+  private infix fun Double.shouldThrowExceptionOnNotLessThan(x: Double) {
+    shouldThrowAssertionError("$this should not be < $x",
+                              { this shouldNotBe lt(x) },
+                              { this shouldNotBeLessThan x },
+                              { this shouldNot beLessThan(x) })
+  }
+  
+  private infix fun Double.shouldNotMatchLessThan(x: Double) {
+    this shouldNotBe lt(x)
+    this shouldNotBeLessThan x
```
sksamuel: If this fails, won't the next two lines not be executed?

Kerooker: That's correct, but I don't think it's a bad thing. We want the test to fail when this fails

Kerooker:
@sksamuel and @ajalt 
Take a look now, I addressed your issues.

However, @sksamuel, Trevis is unable to test anything for quite some time. I think something broke ./gradlew test after [this commit](https://github.com/kotlintest/kotlintest/commit/74c132ebdb72b99a93f647cc1fbb8214f01eb4c6)

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -1,72 +1,1085 @@
...
+            }
+          }
+        }
+      }
+    }
+    
+    
+  }
+  
+  private fun Double.smallDelta(): Double {
```
ajalt: We're getting there! These extensions work in more cases, but still fail for values near the representable limits. For example, `smallDelta` and the `slighty*` extensions fail because `Double.MIN_VALUE * 0.01 == Double.MIN_VALUE`. You're looking for the smallest number you can add to a double that will change its value. The term for this is [Unit of Least Precision (ULP)](https://en.wikipedia.org/wiki/Unit_in_the_last_place). The JDK has a built in function for this: [`Math.ulp`](https://docs.oracle.com/javase/10/docs/api/java/lang/Math.html#ulp(double)).

For the `muchGreater` and `muchSmaller` extensions, many large double values will end up as an infinity (e.g. `1.5e308 * 1.25 == POSITIVE_INFINITY`), and many small values will produce incorrect results (e.g. `3e-324.muchGreater() == Double.MIN_VALUE`, which is _smaller_ than `3e-324`). I suggest using a multiple of the ULP like `this + 2 * Math.ulp(this)`.

Kerooker: Incredible!! I'm learning a lot with this! Will fix this by tomorrow.

Thanks a lot for the insight!

Kerooker: I think this is fixed now.

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -1,72 +1,1065 @@
...
+        }
+      }
+    }
+    
+    
+  }
+  
+  
+  private fun Double.toleranceValue(): Double {
+    return 5 * ulp
```
ajalt: This is fine, but I would use `ulp, `2 * ulp`, and `3 * ulp` rather that 5, 10, and 100. The larger values don't catch any extra errors, and will cause more values to be rounded to infinity.

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -1,72 +1,1065 @@
...
+  }
+  
+  private fun Double.slightlyGreater(): Double {
+    return this + (10 *ulp)
```
ajalt: Can you reformat the file? The spacing is a little messed up here.

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
```diff
@@ -1,72 +1,1065 @@
...
+    shouldThrowExceptionOnBetween(a, b, tolerance, message)
+  }
+  
+  private fun Double.shouldThrowMaximumExceptionOnBetween(a: Double, b: Double, tolerance: Double) {
+    val message = "$this should be smaller than $b"
+    shouldThrowExceptionOnBetween(a, b, tolerance, message)
+  }
+  
+  
+  private fun Double.shouldThrowExceptionOnBetween(a: Double,
```
ajalt: This is an extremely minor nit, but the kotlin style guide recommends formatting multi-line parameters like this:

```kotlin
private fun Double.shouldThrowExceptionOnBetween(
    a: Double,
    b: Double,
    tolerance: Double,
    message: String = "$this should be smaller than $b and bigger than $a"
) {
```

Which is nice in this case to avoid the long line due to the default parameter.

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/doubles/DoubleMatchersTest.kt
ajalt: We have `assertSoftly` for this case.

Kerooker:
I agree to wait for CI. The issue for it is https://github.com/kotlintest/kotlintest/issues/457, if you'd like to follow it

PR #462, opened 2018-11-05
Make kotlin compiler happy ;)

nikos:


doc/reference.md
```diff
@@ -301,7 +301,7 @@ object TimerListener : TestListener {
   }
 
   override fun afterTest(description: Description, result: TestResult): Unit {
-    println("Duration of $description = " + System.currentTimeMillis() - started)
+    println("Duration of $description = " + (System.currentTimeMillis() - started))
```
Kerooker: Would it be even better to use string templates for everything?

```suggestion
    println("Duration of $description = ${System.currentTimeMillis() - started}")
```

Issue #457, opened 2018-10-31
Fix CI


Kerooker:
Currently, w hen deploying KotlinTest, the CI is broken. Whenever it tries to execute `./gradlew test`, it will become stuck at the first test it tries to execute, and never finishes.

I've done some digging, and I found that the class `JUnitTestRunnerListener` had an addition from PR #449 that makes it kinda messy:

Line 222:
```
descriptor = object : AbstractTestDescriptor(id, description.name, source)
```

When creating this descriptor with a `source`, the test never completes. I'm not sure what's causing this, but if we create the descriptor with

```
descriptor = object : AbstractTestDescriptor(id, description.name)
```

It will work again.

I wasn't able to investigate further, but giving this information may be helpful to solve it. I'll keep digging, but I'm not sure how tests are actually executed

Kerooker:
@sksamuel

Kerooker:
Ok, I have some more information on this issue:

Currently, source is created using

```
    val source = ClassSource.from(klazz?.java) ?: null
```

For some reason, this source only fails if TestType is TestType.Test. It'll never complete the test. However, if we change it to null on TestType.Test, something interesting happen, because it works:

```
  val source = if (type == TestType.Test) {
      null
    } else {
      ClassSource.from(klazz?.java) ?: null
    }
```

Kerooker:
I have no clue on why this works. But using this way, we're able to goto source on any descriptor that's not the final one (test one).

I'm still trying to find a real reason for this, but having this documented may help any of you to find it.

@sksamuel I also believe that JUnitRunnerListener needs a refactoring. It's pretty hard to understand right now :x

sksamuel:
We are green again!

Issue #447, opened 2018-10-13
BeforeTest/AfterTest use different threads than the actual test


Burtan:
Hi,
it seems that BeforeTest and AfterTest functions are running on different threads than the actual test. This is problematic for some frameworks like GAE and Objectify. Is it possible to make them all run on the same thread?

Burtan:
Any release date?

sksamuel:
I'll merge this back into 3.1.x along with the fix for #413 and release 3.1.11.

Burtan:
Any estimation for a new release version?

PR #463, opened 2018-11-05
Fixes CI #457

Kerooker:
For some odd reason, the pull request #449 introduces a bug in execution via `./gradlew test`. This small bug will happen for any project, and not only KotlinTest. The gradlew test executor seems to get stuck when trying to resolve the source, and won't execute tests correctly, leading to a permanent freeze.

This commit reverts this problem, but issue #448 will be reopened, as it's not completely fixed for now. The canonical name change won't be reverted, as it didn't cause the bug.

Looking at some other projects, such as Spek, the executor doesn't add a source to the TestDescriptor (see [this line in Spek's TestDescriptor](https://github.com/spekframework/spek/blob/31cad369dbb410a35f686100dea0488a36d505ef/spek-runner/junit5/src/main/kotlin/org/spekframework/spek2/junit/SpekTestDescriptor.kt#L43). I'd assume the problem is similar to this one, and they weren't able to add the source there.

A issue should be opened to Gradle, but I couldn't pinpoint the root cause of this issue, and this will be left as a TODO.

This commit also finds a bug introduced by #447, that makes FailuresTest to fail. This will be fixed in a further commit.

Part of #457

### DIFF
--- a/kotlintest-runner/kotlintest-runner-junit5/src/main/kotlin/io/kotlintest/runner/junit5/JUnitTestRunnerListener.kt
+++ b/kotlintest-runner/kotlintest-runner-junit5/src/main/kotlin/io/kotlintest/runner/junit5/JUnitTestRunnerListener.kt
@@ -123,8 +123,7 @@ class JUnitTestRunnerListener(private val listener: EngineExecutionListener,
     // that we do not "start" a test that is later marked as skipped.
     if (!started.contains(set.testCase.description)) {
       started.add(set.testCase.description)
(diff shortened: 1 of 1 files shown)


Kerooker:
Now that you approved this, we'll have to fix what was mentioned from issue #447.

Any suggestions?

sksamuel:
447 is not the issue with the source parameter? Did you mean 448

sksamuel:
I've merged this and fixed the remaining errors.

PR #465, opened 2018-11-07
Improves Failures implementation

Kerooker:
As it was, the Failures class would fail to filter classes that are neither from `sun.reflect.*` nor from `java.*`. This commit aims to create a generic implementation that correctly filters stacktraces.

The implemented behavior observes a pattern to clean the stacktrace:

- There may be elements that are not from Kotlintest and are not from User Classes, such as Reflection elements or Java elements (but may be others). These elements are removed.
- Then the code tries to find elements that are indeed from Kotlintest, such as runners and other internals
- After this, it's expected for user classes to be in the stacktrace, so the algorithm stops and returns the stacktrace

By observation of the FailuresTest class, one may believe that only classes from `io.kotlintest` and `sun.reflect` would be in the way of user classes, but that's not the case, and this commit fixes it.

Very related to [this commit](https://github.com/kotlintest/kotlintest/commit/2a2cf22cc6ded49e616f72fc7018c45027a2d719)

### DIFF
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/Failures.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/Failures.kt
@@ -5,28 +5,31 @@ object Failures {
   /**
    * Whether KotlinTest-related frames will be removed from the stack traces of thrown [AssertionError]s.
    *
-   * This defaults to `true`. You can change it at runtime, or set it with the system property
-   * `kotlintest.failures.stacktrace.clean`.
+   * This defaults to `true`. You can change it by setting the system property `kotlintest.failures.stacktrace.clean`.
    *
-   * e.g.
+   * E.g.:
    *
    * ```
    * -Dkotlintest.failures.stacktrace.clean=false
    * ```
    */
-  val shouldRemoveKotlintestElementsFromStacktrace: Boolean = readSystemProperty()
+  val shouldRemoveKotlintestElementsFromStacktrace by lazy { readSystemProperty() }
+  
+  private fun readSystemProperty(): Boolean {
+    return System.getProperty("kotlintest.failures.stacktrace.clean")?.toBoolean() ?: true
+  }
 
   /**
    * Creates an [AssertionError] with the given [message] and [cause]
    *
-   * If [shouldRemoveKotlintestElementsFromStacktrace] is `true`, [removeKotlintestElementsFromStacktrace] will be
-   * called on the returned error.
+   * If [shouldRemoveKotlintestElementsFromStacktrace] is `true`, the stacktrace will be reduced to the user-code
+   * StackTrace only.
    */
   fun failure(message: String, cause: Throwable? = null): AssertionError = AssertionError(message).apply {
     if (shouldRemoveKotlintestElementsFromStacktrace) {
       removeKotlintestElementsFromStacktrace(this)
     }
-    this.initCause(cause)
+    initCause(cause)
   }
 
   /**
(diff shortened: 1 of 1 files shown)


Kerooker:
@ajalt 
Requested your review because you originally implemented the Failures algorithm. Would you mind taking a look?

sksamuel:
Some context here - the reason for my change overnight was because when you have threads > 1, we execute these on a thread pool, and if threads == 1 we execute in the main thread. So the stack trace can vary. It won't always be the case that the "last" io.kotlintest stack element is directly before the user's own code.

Kerooker:
That's correct.
However, as I see, it won't always be `sun.reflect` or `java.*`, so I changed it to a way to find the user classes every time

kotlintest-assertions/src/main/kotlin/io/kotlintest/Failures.kt
```diff
@@ -5,28 +5,31 @@ object Failures {
...
    *
-   * e.g.
+   * E.g.:
    *
    * ```
    * -Dkotlintest.failures.stacktrace.clean=false
    * ```
    */
-  val shouldRemoveKotlintestElementsFromStacktrace: Boolean = readSystemProperty()
+  val shouldRemoveKotlintestElementsFromStacktrace by lazy { readSystemProperty() }
```
ajalt: This needs to be a `var` so that it can be configured by the user at runtime

Kerooker:
@ajalt 
The `FailuresTest` itself started failing when we made the change in the aforementioned commit. It was failing with the stacktrace:

```
java.lang.AssertionError: msg
	at io.kotlintest.runner.jvm.TestSetExecutor.execute(TestSetExecutor.kt:40)
	at io.kotlintest.runner.jvm.TestCaseExecutor.executeTestIfActive(TestCaseExecutor.kt:69)
	at io.kotlintest.runner.jvm.TestCaseExecutor.access$executeTestIfActive(TestCaseExecutor.kt:14)
	at io.kotlintest.runner.jvm.TestCaseExecutor$execute$3.invoke(TestCaseExecutor.kt:43)
	at io.kotlintest.runner.jvm.TestCaseExecutor.execute(TestCaseExecutor.kt:53)
	at io.kotlintest.runner.jvm.spec.SharedInstanceSpecRunner$execute$1.invoke(SharedInstanceSpecRunner.kt:16)
	at io.kotlintest.runner.jvm.spec.SharedInstanceSpecRunner$execute$1.invoke(SharedInstanceSpecRunner.kt:10)
	at io.kotlintest.runner.jvm.spec.SpecRunner$interceptSpec$chain$1.invoke(SpecRunner.kt:37)
	at io.kotlintest.runner.jvm.spec.SpecRunner$interceptSpec$chain$1.invoke(SpecRunner.kt:12)
	at io.kotlintest.runner.jvm.spec.SpecRunner.interceptSpec(SpecRunner.kt:38)
	at io.kotlintest.runner.jvm.spec.SharedInstanceSpecRunner.execute(SharedInstanceSpecRunner.kt:13)
	at io.kotlintest.runner.jvm.TestEngine.executeSpec(TestEngine.kt:118)
	at io.kotlintest.runner.jvm.TestEngine.access$executeSpec(TestEngine.kt:20)
	at io.kotlintest.runner.jvm.TestEngine$submitSpec$1.run(TestEngine.kt:97)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.RuntimeException
	at com.sksamuel.kotlintest.FailuresTest$1$1$1.invoke(FailuresTest.kt:15)
	at com.sksamuel.kotlintest.FailuresTest$1$1$1.invoke(FailuresTest.kt:10)
	at io.kotlintest.runner.jvm.TestSetExecutor.execute(TestSetExecutor.kt:40)
	at io.kotlintest.runner.jvm.TestCaseExecutor.executeTestIfActive(TestCaseExecutor.kt:69)
	at io.kotlintest.runner.jvm.TestCaseExecutor.access$executeTestIfActive(TestCaseExecutor.kt:14)
	at io.kotlintest.runner.jvm.TestCaseExecutor$execute$3.invoke(TestCaseExecutor.kt:43)
	at io.kotlintest.runner.jvm.TestCaseExecutor.execute(TestCaseExecutor.kt:53)
	at io.kotlintest.runner.jvm.spec.SharedInstanceSpecRunner$callingThreadContext$1.registerTestCase(SharedInstanceSpecRunner.kt:31)
	at io.kotlintest.TestContext.registerTestCase(TestContext.kt:41)
	at io.kotlintest.specs.AbstractFreeSpec$FreeSpecScope.invoke(AbstractFreeSpec.kt:47)
	at com.sksamuel.kotlintest.FailuresTest$1$1.invoke(FailuresTest.kt:14)
	at com.sksamuel.kotlintest.FailuresTest$1$1.invoke(FailuresTest.kt:10)
	at io.kotlintest.specs.AbstractFreeSpec$minus$1.invoke(AbstractFreeSpec.kt:18)
	at io.kotlintest.specs.AbstractFreeSpec$minus$1.invoke(AbstractFreeSpec.kt:11)
	... 19 more
```

With the fix the stacktrace is correct

sksamuel:
@Kerooker @ajalt Why would it be a var ?

It's detected from the system properties so is still settable.
Or are you expecting people to do `Failures.shouldRemoveKotlintestElementsFromStacktrace = false` in code?

If so, this is better off being part of ProjectConfig which contains all the other "settings" that uses can override ?

Kerooker:
From what @ajalt said, I expected users to use `Failures.shouldRemoveKotlintestElementsFromStacktrace = false`. However, I cannot picture anyone actually doing this. I can revert the commit if necessary

sksamuel:
Having the programmatic version is useful, but I think it should be controlled from `ProjectConfig` that's all. That's where all the other settings are held.

ajalt:
We had the same discussion when I implemented this. Quoting from [that PR](https://github.com/kotlintest/kotlintest/pull/369#issuecomment-400108793):

> Failures is part of kotlintest-assertions, and ProjectConfig is part of kotlintest-core, and neither depends on the other.

Kerooker:
Makes sense to me to keep them separated. But my concern is another one: Why is this customizable?
Do we have any user that alters this behavior at runtime? 

Do we have any use for displaying Kotlintest internal stacktraces ever? I mean as an user. As a KT developer it may be interesting, but as an user I don't see use for this. What do you guys think?

ajalt:
Kotlintest is generally a very customizable framework. There isn't any cost to allowing it to be configured, so why not? I'm sure that someone will want to turn off the magic at some point to help them debug.

sksamuel:
> We had the same discussion when I implemented this. Quoting from [that PR](https://github.com/kotlintest/kotlintest/pull/369#issuecomment-400108793):
> 
> > Failures is part of kotlintest-assertions, and ProjectConfig is part of kotlintest-core, and neither depends on the other.

Yes of course. Sorry I've blinked since we had that convo :fearful:

sksamuel:
> Kotlintest is generally a very customizable framework. There isn't any cost to allowing this functionality to be configured, so why not? I'm sure that someone will want to turn off the magic at some point to help them debug.

Agreed. Val back to a var and I think we're good.

Kerooker:
Alright. Can we merge this then?

sksamuel:
I'm happy.

PR #466, opened 2018-11-09
Fix function name in property docs examples

dhoepelman:
Property-based testing docs should use `assertAll`

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -485,7 +485,7 @@ class PropertyExample: StringSpec() {
   init {
 
     "String size" {
-      forAll(2300) { a: String, b: String ->
+      assertAll(2300) { a: String, b: String ->
         (a + b).length == a.length + b.length
       }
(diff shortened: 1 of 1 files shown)


sksamuel:
:+1:

Kerooker:
Hey, @sksamuel 
```
forAll(Gen.string()) {
...
}
```

Is also a valid call. Is it actually better to use `assertAll`?

Kerooker:
Maybe we should improve the documentation to say that forAll and assertAll are different in that sense?

sksamuel:
That would be better.

dhoepelman:
> Maybe we should improve the documentation to say that forAll and assertAll are different in that sense?

Would be a good addition! My confusion was caused by multiple `forAll` operators existing, causing me to import the incorrect one, and `forAll` not being mentioned previously in the property-drive testing docs.

Issue #475, opened 2018-11-20
Add 'as' support to assertions


checketts:
AssertJ has a ['describe' feature](http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#describe-assertion) for assertions. 

```
assertThat(frodo.getAge()).as("check %s's age", frodo.getName()).isEqualTo(100);
//output: [check Frodo's age] expected:<100> but was:<33>
```

The current equivalent in kotlintest is:
```
frodo.age shouldBe 100
//output: expected:<100> but was:<33>
```

I would love to have the ability to add a context to a failure.  Perhaps like so:

```
"check ${frodo.name}'s age" { frodo.age shouldBe 100 }
// or to avoid adding invoke to string:
describe("check ${frodo.name}'s age") { frodo.age shouldBe 100 }
```

I have my own current extension functions that catch the `AssertionError` and wraps the context, but I would love to see it supported as a first class citizen in kotlintest. 

Is this a welcome feature? I would love to begin a PR.

sksamuel:
There's something similar already called `withClue`.
If that works for you then great. If it's not powerful enough then a PR is most welcome to replace it / improve it.

checketts:
`withClue` is a great start!

With `withClue("id") { actual.id shouldBe expected.id }`
```
java.lang.AssertionError: id org.opentest4j.AssertionFailedError: expected: "123" but was: "124"
Expected :"123" 
Actual   :"124"
```

I'll look into how hard it would be to get it to be:
```
java.lang.AssertionError: id expected: "123" but was: "124"
Expected :"123" 
Actual   :"124"
```

Kerooker:
Closed as per #477

PR #477, opened 2018-11-20
Simplify `withClue` and add to readme

checketts:
Addresses #475

### DIFF
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/matchers.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/matchers.kt
@@ -2,15 +2,17 @@
 
 package io.kotlintest.matchers
 
-import io.kotlintest.Failures
+import io.kotlintest.ErrorCollector
 import io.kotlintest.Matcher
 import io.kotlintest.Result
 
 fun withClue(clue: String, thunk: () -> Any) {
+  val currentClue = ErrorCollector.clueContext.get()
   try {
+    ErrorCollector.clueContext.set("$clue ")
     thunk()
-  } catch (e: AssertionError) {
-    throw Failures.failure("$clue $e")
+  } finally {
+    ErrorCollector.clueContext.set(currentClue)
   }
 }
 
--- a/kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/MatchersTest.kt
+++ b/kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/MatchersTest.kt
@@ -1,9 +1,16 @@
 package com.sksamuel.kotlintest.matchers
 
+import io.kotlintest.Matcher
+import io.kotlintest.Result
+import io.kotlintest.assertSoftly
+import io.kotlintest.fail
 import io.kotlintest.matchers.beInstanceOf
+import io.kotlintest.matchers.collections.shouldHaveSize
 import io.kotlintest.matchers.haveLength
 import io.kotlintest.matchers.haveSameHashCodeAs
 import io.kotlintest.matchers.haveSize
+import io.kotlintest.matchers.string.shouldContain
+import io.kotlintest.matchers.withClue
 import io.kotlintest.specs.FreeSpec
 import io.kotlintest.should
 import io.kotlintest.shouldBe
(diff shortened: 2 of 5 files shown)


kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/matchers.kt
```diff
@@ -10,7 +10,7 @@ fun withClue(clue: String, thunk: () -> Any) {
   try {
     thunk()
   } catch (e: AssertionError) {
-    throw Failures.failure("$clue $e")
+    throw Failures.failure("$clue ${e.message}")
```
Kerooker: I suppose there is no test for this at the moment. Maybe we should create one for this behavior as well?

checketts: Strangely there are no tests in the `kotlintest-assertions` module at all (not even a test source path). I'm happy creating a test, any preference on where I should create it? Should I just add the test sources to the assertions module?

Kerooker: There's a special source-set for testing at the moment: `kotlintest-tests`. They're all there. I'm not sure why all tests are in the same place at the moment, but that's what we have atm :P

sksamuel: The reason is because in order to test kotlintest as a genuine junit-platform plugin, it needs to be tested as a jar on the dependency path. In order to do this, kotlintest-core must be built before the tests, hence why the test are in a separate package.

The ones for assertions _could_ live in assertions though, it's just the "core" tests that _need_ be separate. Arrow tests for example are in the arrow module.

sksamuel:
Happy to merge if you are @Kerooker

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/matchers.kt
checketts: OK I just added tests. As an aside, the tests were in the package `com.sksamuel.kotlintest.matchers` which struck me as odd (I'm guessing due to historical reasons)

sksamuel: The reason is that there are certain tests that only work if they're not in the io.kotlintest package. So I just put them all in my private domain.

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/MatchersTest.kt
```diff
@@ -15,6 +19,33 @@ import kotlin.collections.HashMap
 
 class MatchersTest : FreeSpec({
 
+  "withClue()" - {
+
+    "should prepend clue to message" {
+      try{
+        withClue("a clue:") { listOf(1) shouldHaveSize 2 }
```
Kerooker: I'm a little bit afraid of "hardcoding" the exception this way. What if the `shouldHaveSize` message changes? We'll have another point of code to take care of, while `withClue` will keep working.

Technically, this is testing two things: That Clue works and that Collection `shouldHaveSize` message is x

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/MatchersTest.kt
```diff
@@ -15,6 +19,33 @@ import kotlin.collections.HashMap
...
+    "should play nice with multiple/softAssert" {
+      try{
+        withClue("multi clue:") { assertSoftly {
+          "kotlintest" shouldContain "magic"
+          listOf(1) shouldHaveSize 2
+        } }
+      } catch(e: AssertionError) {
+        e.message.apply {
+          shouldContain("multi clue:")
+          shouldContain("kotlintest should include substring magic")
```
Kerooker: Same point as above

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/MatchersTest.kt
checketts: Fixed. I created a custom matcher just for the test

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/MatchersTest.kt
```diff
@@ -15,6 +22,36 @@ import kotlin.collections.HashMap
...
+
+    "should play nice with multiple/softAssert" {
+      val ex = shouldThrow<AssertionError> {
+        withClue("multi clue:") { assertSoftly {
+          "1" shouldBe withClueEcho("the details!")
+          "2" shouldBe withClueEcho("the details!")
+        } }
+      }
+      ex.message.apply {
+        shouldContain("multi clue:")
```
Kerooker: Maybe it's better if we explicitliy test the message, similarly to the first test case. Testing for an exact value using `shouldBe`.

This test case doesn't cover a possible scenario of adding more than one clue, for example (I'm not even sure if this is the intended behavior, just as an example):
```
multi clue: Should have the details of '1' and the details!
multi clue: Should have the details of '2' and the details!
```

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/MatchersTest.kt
```diff
@@ -15,6 +22,36 @@ import kotlin.collections.HashMap
...
+    }
+
+    "should prepend clue to message" {
+      val ex = shouldThrow<AssertionError> {
+        withClue("a clue:") { "1" shouldBe withClueEcho("here are the details!") }
+      }
+      ex.message shouldBe "a clue: Should have the details of '1' and here are the details!"
+    }
+
+    "should play nice with multiple/softAssert" {
```
Kerooker:
```suggestion
    "should add clues correctly with multiple/softAssert" {
```

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/MatchersTest.kt
checketts: Gah! You are correct. This doesn't work:

```
withClue("multi clue:") { assertSoftly {
          withClue("inner clue1") {"1" shouldBe withClueEcho("the details!") }
          withClue("inner clue2") {"2" shouldBe withClueEcho("the details!")}
        } }
```

The inner clues are ignored because the exceptions aren't thrown (and thus not caught). I'm going to investigate how the soft assertions work to see if I can integrate with it.

Also I used the `shouldContain` because the assertSoftly includes call site info:

```
java.lang.AssertionError: multi clue: 
1) Should have the details of '1' and the details!
	at com.sksamuel.kotlintest.matchers.MatchersTest$1$1$3$ex$1$1$1$1.invoke(MatchersTest.kt:42)
2) Should have the details of '2' and the details!
	at com.sksamuel.kotlintest.matchers.MatchersTest$1$1$3$ex$1$1$1$2.invoke(MatchersTest.kt:43)
 should include substring inner clue1: Should have the details of '1' and the details!
```

Kerooker: Interesting.
It's OK to use `shouldContain`, but verify a specific amount is better IMO.

As for the test case itself using `assertSoftly`, I'm not sure what behavior is intended here. And I believe that this discussion should be in another ticket, as this one is not for this specific bug-report.

What do you think?

Kerooker:
It's ok for me. What do you think, @sksamuel

PR #481, opened 2018-11-21
Add asserts that the either is of type Left with a specified type A 

daneko:
Current e.x

```kotlin
sealed class MyError() {
  class MyError1(): MyError()
  ...
}

// in test code

either.fold(
  ifLeft = { it should beOfType<MyError1>() },
  ifRight = { fail() }
)

```

```
either.shouldBeLeftOfType<MyError1>()
```

like #427

### DIFF
--- a/doc/arrow-matchers.md
+++ b/doc/arrow-matchers.md
@@ -15,6 +15,7 @@ This page lists all current matchers in the KotlinTest arrow matchers extension 
 | `either.shouldBeRight(v)` | Asserts that the either is of type Right with specified value v |
 | `either.shouldBeLeft()` | Asserts that the either is of type Left |
 | `either.shouldBeLeft(v)` | Asserts that the either is of type Left with specific value v |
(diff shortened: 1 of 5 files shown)


sksamuel:
Should we do the same for shouldBeRightOfType ?

daneko:
In my case, I rarely use it.
Because Left type are usually Throwable or sealed class,
But Right type compare with Left is usually a (data) class.

e.x
```kotlin
sealed class MyError {
  object Error1: MyError()
  class Error2(val detailMessageEtc:String, ...): MyError()
  ...
}
data class MySuccess(val a:A, val b:B, ...)

fun method(...): Either<MyError, MySuccess>

method(...).fold(
  ifError = {
    when(it){
      is MyError1 ->
      is MyError2 ->
      ...
    }
  },
  ifSuccess -> {...}
)
```
(In my opinion) For test case, 
I would like to check Right value and Left pattern.

of course, Someday I need shouldBeRightOfType, I make PR it :D .

PR #484, opened 2018-11-27
Fix 479 - wrong assertion on shouldThrow

### LINKED ISSUE #479: Why shouldThrow<> {} always pass this check?
In documentation you said:
```
Exceptions
To assert that a given block of code throws an exception, one can use the shouldThrow function. Eg,

shouldThrow<IllegalAccessException> {
  // code in here that you expect to throw an IllegalAccessException
}
```

But in reality it green always

Context: just for fun - green
    Scenario: throw exception - green
    Scenario: no exception - green

```
        context("just for fun") {
            // pass ??? why  
            //  even I set wrong Exception type -   there should be AssertionFailedError
            it("throw exception") {
                shouldThrow<java.lang.AssertionError> { 1 shouldBe 2 }
            }
            // pass ??? why
            it("no exception") {
                shouldThrow<java.lang.AssertionError> { 1 shouldBe 1 }
            }
        }
```

====
If I change  **shouldThrow**<java.lang.AssertionError> to **shouldThrowExactly**<java.lang.AssertionError>  it works well now:

```
 context("just for fun") {
            //  pass
            it("throw exception") { 
                shouldThrowExactly<org.opentest4j.AssertionFailedError> { 1 shouldBe 2 }
            }
            // failed
            it("no exception") {
                shouldThrowExactly<org.opentest4j.AssertionFailedError> { 1 shouldBe 1 }
            }
        }
```

Kerooker:
I recommend reviewing commit-by-commit, as one commit makes the changes, while the other just fixes imports

Refactors Throwable Handling to it's own package, fixing bugs  …
This commit moves ThrowableHandling to their own package, as it's a pattern KT is following for a while. While moving the classes and updating the package, this commit also adds documentations to all use cases, and add some extra cases that `shouldThrowUnit`, such as `shouldThrowAnyUnit`.

This commit deprecates usage, that should be removed from the library in a future. For now, code will still compile.

By changing the order in which verifications are made, this commit also fixes #479, and adds a test for every possible scenario, stabilizing and reviewing these methods.


kotlintest-assertions/src/main/kotlin/io/kotlintest/Failures.kt
```diff
@@ -1,5 +1,23 @@
...
+ * Verifies that [block] throws an [AssertionError]
+ *
+ * If [block] throws an [AssertionError], this method will pass. Otherwise, it will throw an error, as a failure was
+ * expected
+ */
+fun shouldFail(block: () -> Any?) {
+  val passed = try {
+    block()
+    true
+  } catch (_: Throwable) {
```
sksamuel: This will catch any exception or error. 
I think if it throws something that is not AssertionError we should just let the error propagate? Otherwise the test case will not have the correct exception attached to it.

kotlintest-assertions/src/main/kotlin/io/kotlintest/throwablehandling/AnyThrowableHandling.kt
```diff
@@ -0,0 +1,49 @@
...
+ * Use function to wrap a block of code that you want to verify that throws any kind of [Throwable], where using
+ * [shouldThrowAny] can't be used for any reason, such as an assignment of a variable.
+ *
+ * If you want to verify a specific [Throwable], use [shouldThrowExactlyUnit].
+ *
+ * If you want to verify a [Throwable] and any subclass, use [shouldThrowUnit].
+ *
+ * @see [shouldThrowAny]
+ */
+inline fun shouldThrowAnyUnit(block: () -> Unit) = shouldThrowAny(block)
```
sksamuel: Docs are great, but I think the double spacing is hard to read.

kotlintest-assertions/src/main/kotlin/io/kotlintest/throwablehandling/AnyThrowableHandling.kt
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * If you want to verify a specific [Throwable], use [shouldThrowExactly].
+ *
+ * If you want to verify a [Throwable] and any subclasses, use [shouldThrow]
+ *
+ *
+ * **Attention to assignment operations:**
+ *
+ * When doing an assignment to a variable, the code won't compile, because an assignment is not of type [Any], as required
+ * by [block]. If you need to test that an assignment throws a [Throwable], use [shouldThrowAnyUnit] or it's variations.
```
sksamuel: Nice. I'd slightly reword to say that an assigment is not an expression but a statement, and therefore has no return value.

kotlintest-assertions/src/main/kotlin/io/kotlintest/throwablehandling/StrictThrowableHandling.kt
```diff
@@ -0,0 +1,74 @@
...
+    block()
+    null  // Can't throw Failures.failure here directly, as it would be caught by the catch clause, and it's an AssertionError, which is a special case
+  } catch (thrown: Throwable) { thrown  }
+
+  return when {
+    thrownThrowable == null -> throw Failures.failure("Expected exception ${T::class.qualifiedName} but no exception was thrown.")
+    thrownThrowable::class == expectedExceptionClass -> thrownThrowable as T  // This should be before `is AssertionError`. If the user is purposefully trying to verify `shouldThrow<AssertionError>{}` this will take priority
+    thrownThrowable is AssertionError -> throw thrownThrowable
+    else -> throw Failures.failure("Expected exception ${expectedExceptionClass.qualifiedName} but a ${thrownThrowable::class.simpleName} was thrown instead.", thrownThrowable)
+  }
```
sksamuel: :+1:

kotlintest-assertions/src/main/kotlin/io/kotlintest/Failures.kt
Kerooker: I was thinking that maybe users that use this method wouldn't care about anything, just that this code fails. Do you think it's better with just AssertionError?

kotlintest-assertions/src/main/kotlin/io/kotlintest/throwablehandling/AnyThrowableHandling.kt
Kerooker: What do you mean with double spacing? The space between lines?

Kerooker:
@sksamuel 
One other thing that worries me a bit:
This is the second time these matchers are changing packages. I thought of some options:

- Leave them at the ExceptionHandlers class anyway, and add all the matchers there (i'm not a big fan of this one. Separation of responsibilities is better, IMO). This is an option to remove this burden from final user
- Move them to individual files, in the `throwablehandling` directory, but leave them at `io.kotlintest` package (compiler will complain, but it will allow). This option enables no deprecation and no changes to real user, but better organization of code
- Leave it this way, and force the user to change implementation some time in the future. This isn't very good, as users shouldn't need to do this only because I find it better

- Create them in `io.kotlintest`, in the wrong directory (kotlin allows), but some point in the f uture create a breaking change, that removes all deprecated features, and force users to change all imports at once, but never do it again (never is way too strong, but for a long time at least).

sksamuel:
Option 5 - leave them in the io.kotlintest package for now, in the correct directory, but spilt up into seperate files. This would be my preferred option, no burden for the user.

kotlintest-assertions/src/main/kotlin/io/kotlintest/throwablehandling/AnyThrowableHandling.kt
sksamuel: Never mind, it was just my screen.

kotlintest-assertions/src/main/kotlin/io/kotlintest/Failures.kt
sksamuel: I think it's better with just AssertionError.

You have two possibilities.

1. I want to check that my code throws an exception - shouldThrow and it's variants
2. I want to check that my code fails an assertion - shouldFail

And the downside to catching everything with 2 is that if your test fails for some other reason - FileNotFoundException say - you won't see _that_ exception, but _our_ error message. Exceptions are properly propagated out to the junit platform as it stands - they have different statues for "failed" and "errored".

sksamuel:
Well we could do that. What's worse, having the files in the wrong place or having a few extra files in that directory for a couple of release?

Kerooker:
I don't think there's a problem in having the files in the wrong place. Kotlin compiler will give us a warning, but it will be ok with it. But it will increase the file organisation without messing the user code

sksamuel:
I think the opposite :) That directory is hardly overflowing with files a compiler warning is annoying.

Up to you.

Kerooker:
We can suppress the compile warning, and leave a comment that the package is purposefully left wrong. IDK, but i feel that 3 files that are very related should be in a concise scope. We can also leave them all in one single file, the ExceptionHandlers one that already exists.

Do you think it's better?

kotlintest-assertions/src/main/kotlin/io/kotlintest/Failures.kt
Kerooker: Makes sense to me. I'll change shouldFail to AssertionError again. Good one!

kotlintest-assertions/src/main/kotlin/io/kotlintest/throwablehandling/AnyThrowableHandling.kt
Kerooker: Good one.

kotlintest-assertions/src/main/kotlin/io/kotlintest/throwablehandling/StrictThrowableHandling.kt
Kerooker: :+1:

sksamuel:
> We can suppress the compile warning, and leave a comment that the package is purposefully left wrong. IDK, but i feel that 3 files that are very related should be in a concise scope. We can also leave them all in one single file, the ExceptionHandlers one that already exists.
> 
> Do you think it's better?

Personally my choice is just to leave them in the same package, in the same directory. It avoids compiler "hacks" and it means no changes for users. I think moving them is not gaining us much, these packages are not that busy anyway. But I'll not object if you still move them.

sksamuel:
The standard is to put them in a package like io.kotlintest.xxx
It's only these old functions (dating back to KotlinTest 1.x) that are in the top level io.kotlintest package.
Back then there was only about 15 files :)

kotlintest-assertions/src/main/kotlin/io/kotlintest/Failures.kt
Kerooker: Fixed.

kotlintest-assertions/src/main/kotlin/io/kotlintest/throwablehandling/AnyThrowableHandling.kt
Kerooker: done.

Kerooker:
@sksamuel Please verify your points were addressed correctly

sksamuel:
:+1:

Kerooker:
Will merge when CI does it's thing

sksamuel:
Without looking at your code.

shouldThrow should pass if an exception is thrown, and fail otherwise (AssertionError is an Error not an exception)
shouldFail should pass if an AssertionError is thrown, and fail for any other exception or no exception.

Kerooker:
`AssertionError` is a `Throwable`, and `shouldThrow` verifies `Throwables`. Therefore, 

`fun shouldFail(block: () -> Any?) = shouldThrow<AssertionError>(block)`
Works correctly.

What do you think?

sksamuel:
Yes

PR #488, opened 2018-11-29
Improve `assertSoftly` documentation

Kerooker:
As it was, the documentation wasn't precise about the use of `assertSoftly`. Its example as for another method, and not `assertSoftly`.

This commit fixes this error in the example, and make some other changes to the documentation, improving understandability.


sksamuel:
:+1:

Issue #461, opened 2018-11-01
Support package selectors in discovery


sksamuel:
See https://stackoverflow.com/questions/52780940/kotlintestengine-discover-doesnt-return-any-tests/53110162#53110162

PR #491, opened 2018-12-02
Added support for package selectors in test discovery; split construc…

sksamuel:
…tion extensions from discovery extensions #461


Kerooker:
Took a little bit longer to review this one. Sorry

Issue #494, opened 2018-12-02
Test Listener / Extension rework


sksamuel:
There's been some requests recently for functionality the current test listener/extension framework cannot handle.

Notably:

Listeners run even when tests are disabled #468
Listener to run before/after a set of tests #458
Extensions should be able to find out whether test case is enabled #440 
Why beforeTest and afterTest parameter is not a TestCase #469

We should improve the listener support for 3.2 to cover all these cases.

Issue #469, opened 2018-11-11
Why beforeTest and afterTest parameter is not a TestCase


alisabzevari:
I am trying to write a listener which runs some initialization code before running some specific test cases (distinguishable by their tags). But `beforeTest` function only receives `Description`. Is there any reason that this function doesn't receive a `TestCase` instance?

Issue #468, opened 2018-11-10
Listeners run even when tests are disabled


ScottPeterJohnson:
I have a few integration tests I want to run separately from all others; these are generally slow and expensive and set up shared database resources.

I tried setting up these resources via TestListeners, both beforeSpec and beforeTest. But the problem is, even when the tests in question are disabled or excluded by tag, the setup code is still run!

Example reproducing the problem:

`class ExampleTest : StringSpec({
    "this shouldn't run".config(enabled = false) {
        1 shouldBe 2
    }
}) {
    override fun listeners(): List<TestListener> = listOf(object : TestListener {
        override fun beforeSpec(description: Description, spec: Spec) {
            println("beforeSpec")
        }

        override fun beforeTest(description: Description) {
            println("beforeTest")
        }
    })
}`

The test does not run, but both setup lines print.

Is there a better way I should be doing this? Or could the behavior be changed?

sksamuel:
We should probably add more listener methods that include the enabled status of a test. Would that work?

Like `beforeTest(description: Description, enabled: Boolean)`

`beforeSpec` must always run even if all the tests are disabled inside the spec, because in order to find out if tests are enabled requires actually creating the spec.

ScottPeterJohnson:
More listeners would probably help! Even just `beforeTest` with an enabled flag, although it might be confusing how you'd get Spec-level resource initialization from there.

Naively coming to the library, I'd still be a little confused at the default behavior though. I also don't currently understand why finding out if tests are enabled requires running `beforeSpec()`. My impression for kotlintest's initialization is that it goes:
1. Instantiate every Spec, the constructors of which will generate root test cases to run
2. `beforeSpec()`
3. For each test, `beforeTest()` then `test()`

It seems like you'd have enough information to decide whether any root test cases are enabled and whether to call `beforeSpec()` at step 1. Am I missing something?

sksamuel:
In that case it could work but not when some tests are diabled and some
enabled in the same spec.

Issue #458, opened 2018-11-01
Listener to run before/after a set of tests


alisabzevari:
I want to have a different set of tests (probably specified by different tags) and have hooks to run before and after running those set of tests. Is it possible to do this in kotlintest?

One use case for this feature is separating integration tests from unit tests. Usually, integration tests need to set up a testing environment which takes time and it is not desired to run this set up process before all the tests (`beforeProject` listener).

sksamuel:
Can you put your integration tests into separate files, and have a "before file" "after file" setup methods? We call those `beforeSpec` and `afterSpec`.

alisabzevari:
I want to have a piece of code running before and after **a set of tests**.

One realworld example would be preparing the database schema (run migration scripts) for testing. After running migration script I want to only clean the tables (but not the whole database). Another one would be starting the mocked dependent servers in integration tests.

sksamuel:
Yep, I'm just asking how you intend to make a "set". Because one way is to put each set into it's own file, in which case `beforeSpec` and `afterSpec` can work, and those can be implemented in an extension like class called a `TestListener`. If that doesn't work we can think of something else?

alisabzevari:
If there was a way to not limit to one file, that would be great.

sksamuel:
You can make a test listener class, and then apply it to every test file that you need it in. Does work?

sksamuel:
Ok I see what you want now. The listener should be "shared" between all the classes that use it.
I'll have a think about how we can support this for 3.2

sksamuel:
You could do this by having a var inside the listener,

`var initialized:Boolean = false`

And each time `beforeSpec` is called, only initialise if the field is false.
And then to clean up you could use afterProject`.

Would that work ?

sksamuel:
Another idea, is we can introduce "groups" and allow each spec to have a single group defined (or null the default). Then we can have two callbacks:

`beforeGroup(group:GroupName)`
`afterGroup(group:GroupName)`

Which would be executed only once for any given group name.

sksamuel:
@Kerooker

Raibaz:
Another possible way to do this would be to have something like [tags and categories](https://www.baeldung.com/junit-filtering-tests) as in JUnit.

LeoColman:
We do have tags and theoretically that enables categories to place tests in. However, I don't think that's a solution to the issue of grouping tests and executing something before and after without boilerplate.

[...]

PR #497, opened 2018-12-03
WIP: Test listener improvements #494

sksamuel:
This is a wip, but see what you think. There's a few things left to implement, so don't worry about that, this PR is so I can get early feedback.

Changed the signatures of some listener functions to be more powerful, including the status of the tests. This will  address #468 #440 #469

I've also included the concept of a Suite (which isn't implemented yet) but will address #458

I'm fairly happen with it, except for the naming of the callback functions. I can't think of good names for "before the very first spec is executed" and "just before each spec is executed" and similarly for the after versions of those. So the names I've used here and totally subject to change.


kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/TestTimeoutTest.kt
```diff
@@ -0,0 +1,24 @@
+//package com.sksamuel.kotlintest
```
Kerooker: Should this really be commented?

Kerooker:
Making listener functions more powerful is pretty interesting!

The Suite concept is somewhat interesting, but wouldn't another level of abstraction be too much? I'm afraid that new users will have too much to grasp at once. Or are these "advanced features" not very available unless you want it?

I liked the `beforeSuite` and `afterSuite` names. If it's clear what a Suite is, it will be clear what these methods do, and I really liked their names!

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/TestTimeoutTest.kt
sksamuel: I didn't mean to check it in.

sksamuel:
You would only have to use Suite if you wanted, you have to go looking for it.

sksamuel:
I've removed suites for now, I'll work that into 3.3

Issue #498, opened 2018-12-05
[3.4] Assertion improvements


Kerooker:
The module `kotlintest-assertions` has some room for improvements. There are many matchers/assertions that are deprecated and were moved, and many matchers have no documentation and not enough tests.

## Breaking change: Remove deprecated methods
This issue aims to mark as error all deprecated methods, forcing users to change their imports to the new ones. The deprecated methods will be moved to a `legacy` directory, that will be removed

**This will be an issue to current users**, but will save new and old users of the following situation:
![image](https://user-images.githubusercontent.com/1577251/49532571-f5ae5200-f8a3-11e8-84ad-abbd395d1996.png)
"Which import should I use?"
"Oh, this one is deprecated, then I believe it's the other one"

I've noticed that this happens a lot to new users. The confusion is more of an annoying thing than a big problem, but it's bad for user experience. Marking these deprecations as an Error will provide users a quick way to move to the new versions, and will remove this problem.

## Code improvement: All tests, in assertions module itself
Some matchers aren't tested for corner cases. Some also don't have any code coverage. To allow new matchers to be developed, and an improvement to assertion quality assurance, all matchers should be covered by unit tests, guaranteeing their functionality.

To further improve the assertions module, their tests can be moved to its module, and not the global `kotlintest-tests` module. In the near past, 2 contributors were asked to create tests to their new features, and neither of them knew at first where to place the tests. **If possible**, moving all tests to be near their test-targets will encourage new contributors to add more tests.

However, I remember @sksamuel commenting that due to module building order, it wouldn't be possible to do this. If it is, it would be better for contributors to understand what to do directly

## Documentation improvements: Add KDocs to all matchers

Good documentation in-code will allow users to know what a matcher/assertion does without opening Github and scanning through the documentation there. It improves the user experience and the resilience of matchers, as we'll have the documentation of exactly how it does (our current docs have short sentences. Maybe they're not explicit enough to some users).

# Contributing to this
I created the branch feature/assertions-improvements. If you have any contribution related to the above comments, feel free to open a Pull Request to that branch!

Kerooker:
@sksamuel I know that forcing users to change their packages due to deprecation is very bad, but I don't know if we should keep the deprecated methods.

What do you think about this?

sksamuel:
I am happy for the deprecated methods to be removed in 3.3

Kerooker:
And then we do all the changes at once, to avoid users having to change everything everytime

[...]

PR #500, opened 2018-12-06
Moves all assertions tests to the assertions module

Kerooker:
As part of https://github.com/kotlintest/kotlintest/issues/498, all tests that test the `kotlintest-assertions` module were moved to it, bringing the tests closer to their targets, and decreasing confusion to future contributors.

An important point to be noted: the root package chosen is `assertions.io.kotlintest`. This was done because putting files in `io.kotlintest` package make them subject to stacktrace cleaning, by the Failures class. The Failures class removes all Kotlintest stacktraces from throwables, so that the user only sees the real problem, and nothing internal. However. as we're using `io.kotlintest` for all the tests, the user stacktraces would be removed (we are the users), and that's not good when a test fails, as no reason would be given.

Fixes a part of #498 - Moves tests closer to their targets


Kerooker:
@sksamuel 
Could you tell appveyor to rerun tests, please?

sksamuel:
> @sksamuel
> Could you tell appveyor to rerun tests, please?

Yep

sksamuel:
> An important point to be noted: the root package chosen is `assertions.io.kotlintest`. This was done because putting files in `io.kotlintest` package make them subject to stacktrace cleaning, by the Failures class. The Failures class removes all Kotlintest stacktraces from throwables, so that the user only sees the real problem, and nothing internal. However. as we're using `io.kotlintest` for all the tests, the user stacktraces would be removed (we are the users), and that's not good when a test fails, as no reason would be given.

That's why I use com.sksamuel for tests in kotlintest core :)
I don't want to use 'kotlintest' anywhere in the package name to totally remove anything like this.

Kerooker:
Should e use com.sksamuel? Maybe io.kt?

sksamuel:
I own com.sksamuel io.kotlintest, anything else is fake.
But up to you.

Kerooker:
com.sksamuel it is then

sksamuel:
Looks good when green.

Kerooker:
Should be ok with last commit

PR #502, opened 2018-12-09
Update README.md

elect86:
- updated kotlintest version to the last one
- added an hint to clean up configuration by old dependencies


sksamuel:
:+1:

PR #509, opened 2018-12-13
Eventually should report what's going wrong together with the attempt…

### LINKED ISSUE #508: Eventually should report what's going wrong together with the attempts
Given the following piece of code:

```
init {
        "Foo" {
            eventually(3.seconds, AssertionError::class.java) {
                3 shouldBe 3
                1 shouldBe 2
                4 shouldBe 4
            }
        }
    }
```

It's a code that will test a lot of assertions. In this specific case, I mocked some assertions with one of them being wrong, but escalate that to a real world scenario.

This test obviously fails, as `1 shouldBe 2` will throw an AssertionError. However, when we take a look at the failure message, it doesn't tell us what failed:

` java.lang.AssertionError: Test failed after 3 seconds; attempted 168628 times`

With this message, I'm not sure if the non-deterministic didn't complete (for example, an API didn't return), or the expected return didn't work.

----

To solve this issue, the only workaround is to debug the code line by line to try to find what happened for it to fail.

It would be better if `eventually` included the thrown exception message with it, including the stacktrace of that exception, so that the developer could pinpoint the error.

sksamuel:
…s (#508)

Fixes #508

### DIFF
--- a/kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/EventuallyTest.kt
+++ b/kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/EventuallyTest.kt
@@ -74,7 +74,7 @@ class EventuallyTest : WordSpec() {
       }
       "display the underlying assertion failure" {
         shouldThrow<AssertionError> {
-          eventually(Duration.ofSeconds(2), AssertionError::class.java) {
+          eventually(Duration.ofMillis(10), AssertionError::class.java) {
             1 shouldBe 2
           }
         }.message.shouldEndWith("; underlying cause was expected: 2 but was: 1")
--- a/kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/EventuallyTest.kt
+++ b/kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/EventuallyTest.kt
@@ -1,6 +1,7 @@
 package com.sksamuel.kotlintest
 
 import io.kotlintest.eventually
+import io.kotlintest.matchers.string.shouldEndWith
 import io.kotlintest.shouldBe
 import io.kotlintest.shouldThrow
 import io.kotlintest.specs.WordSpec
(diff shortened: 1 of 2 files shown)


kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/EventuallyTest.kt
```diff
@@ -71,6 +72,13 @@ class EventuallyTest : WordSpec() {
         }
         result shouldBe 1
       }
+      "display the underlying assertion failure" {
+        shouldThrow<AssertionError> {
+          eventually(Duration.ofSeconds(2), AssertionError::class.java) {
+            1 shouldBe 2
+          }
+        }.message.shouldEndWith("; underlying cause was expected: 2 but was: 1")
```
Kerooker: What would be in this stacktrace?
I think that line 78 should be the one with the stacktrace, but it may be ok to display line 77.

What do you think?

sksamuel: I've added 78 to the stack.

Kerooker: Seems good

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/EventuallyTest.kt
```diff
@@ -71,6 +72,13 @@ class EventuallyTest : WordSpec() {
         }
         result shouldBe 1
       }
+      "display the underlying assertion failure" {
+        shouldThrow<AssertionError> {
+          eventually(Duration.ofSeconds(2), AssertionError::class.java) {
```
Kerooker: I don't think this Duration must be that great. The assertion is so simple, it can be reduced to `Duration.ofMillis(10)` and it will still work correctly

Kerooker: Having a big duration on this test will add more time to the test suite that we don't really need. If we try to reduce this time everywhere, we may have a much faster test suite!

sksamuel: I have changed the timeout and will now merge.

PR #514, opened 2018-12-26
Implements annotations to execute code in AnnotationSpec

Kerooker:
This commit implements some Annotations and annotation processing to make the transition between JUnit4/5 to Kotlintest easier, giving users some annotations they're familiar with.

This commit also fixes a small bug in test case executor, in which `before` and `after` would be executed even if the test is ignored (bug found in class `AnnotationSpecAnnotationsTest`).

The documentation was updated accordingly.

This should be further reviewed to decide if implementation will be expanded to other specs, or only to `AnnotationSpec`

Fixes https://github.com/kotlintest/kotlintest/issues/513

### DIFF
--- a/kotlintest-core/src/main/kotlin/io/kotlintest/specs/AbstractAnnotationSpec.kt
+++ b/kotlintest-core/src/main/kotlin/io/kotlintest/specs/AbstractAnnotationSpec.kt
@@ -1,10 +1,26 @@
 package io.kotlintest.specs
 
 import io.kotlintest.AbstractSpec
+import io.kotlintest.Description
+import io.kotlintest.Spec
 import io.kotlintest.TestCase
+import io.kotlintest.TestCaseConfig
+import io.kotlintest.TestResult
 import io.kotlintest.TestType
+import io.kotlintest.specs.AbstractAnnotationSpec.After
+import io.kotlintest.specs.AbstractAnnotationSpec.AfterAll
+import io.kotlintest.specs.AbstractAnnotationSpec.AfterClass
+import io.kotlintest.specs.AbstractAnnotationSpec.AfterEach
+import io.kotlintest.specs.AbstractAnnotationSpec.Before
+import io.kotlintest.specs.AbstractAnnotationSpec.BeforeAll
+import io.kotlintest.specs.AbstractAnnotationSpec.BeforeClass
+import io.kotlintest.specs.AbstractAnnotationSpec.BeforeEach
+import io.kotlintest.specs.AbstractAnnotationSpec.Ignore
+import kotlin.reflect.KClass
+import kotlin.reflect.KFunction
+import kotlin.reflect.full.memberFunctions
 
-annotation class Test
+typealias Test = AbstractAnnotationSpec.Test
 
 abstract class AbstractAnnotationSpec(body: AbstractAnnotationSpec.() -> Unit = {}) : AbstractSpec() {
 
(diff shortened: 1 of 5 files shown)


Kerooker:
The goal here is to ease transition from JUnit 5 to KotlinTest. For instance, slowly changing classes to use `kotlintest-assertions` instead of `Assert.assertTrue` and other KT's feature without having to change the TestBase (keep using AnnotationSpec if desired).

Kerooker:
What should be discussed further:
If we want to allow the `@XXXTest` to other Specs. For now, I limited it to AnnotationSpec only

sksamuel:
I don't want to support this outside of AnnotationSpec.

Kerooker:
@sksamuel, I believe this is ready for revision

kotlintest-core/src/main/kotlin/io/kotlintest/specs/AbstractAnnotationSpec.kt
```diff
@@ -13,8 +30,247 @@ abstract class AbstractAnnotationSpec(body: AbstractAnnotationSpec.() -> Unit =
...
+   * Marks a Test to be ignored
+   *
+   * This can be used in AnnotationSpec to mark a Test as Ignored.
+   */
+  annotation class Ignore
+
+}
+
+@Suppress("DEPRECATION")
+fun KClass<out AbstractAnnotationSpec>.findBeforeTestFunctions(): List<KFunction<*>> {
```
Kerooker: I thought of marking these functions as internal, but they wouldn't be available in `SpecRunner`

Kerooker:
I'm not sure if we want to endorse JUnit's annotations. Besides, not all of them will work all the time, as Gradle + IntelliJ will see the JUnit-specific annotation and will try to override it, depending of the annotation.

There's nothing we can do with it, as it's a hardcoded annotation. I manually added the support, but it's not possible to test within KotlinTest.

JUnit 5 syntax isn't t preferred. We discussed that a KotlinTest syntax would be better, but a JUnit-like syntax would be supported. They're deprecated in their JUnit form because if users want to just change from JUnit to KT changing annotations only, they'll keep working, but they're not the proposed solution

sksamuel:
I haven't gone through the entire PR yet as I've been busy elsewhere, but the annotations should just be placed in a package called org.junit (or whatever the equivalent junit class uses).
This means no changes for the user, and we don't need junit on the classpath. It's a win win.
See what I do with the test factory annotation for the same effect.

Kerooker:
We can't reutilize JUnit's package. The problem here is that Gradle will try to execute any `@BeforeClass` method and will warn when it's not static. It's hard-coded to work that way when using JUnitPlatform. I couldn't think of any alternative to this, that's why I kinda gave up on JUnit's annotation. Right now I'm getting them through reflection

We're not using JUnit in our classpath, btw

sksamuel:
Ok I see. If we are looking for junit annotations via reflection then do we even need our own?

Kerooker:
I don't see why not.
People can slowly move out of JUnit this way, step by step.

The truth is: I'm not really sure if JUnit's Annotation are going to work. I tried to design their implementation through reflection, but when running tests in KT, all of them failed because gradle tried to execute them as JUnit, and not as KotlinTest. That's why I added all annotations from JUnit.

I also added our own, with `BeforeSpec` and `BeforeTest`, for users to get used to KotlinTest with ease (changing just an annotation a time, for example)

sksamuel:
Sorry for the delay in reviewing properly.
Two comments

1. I don't think we should add an annotation with the same name as junit, deprecate it, and then add another one of our own. The reason is that annotations are a stop gap - the idea in this PR is that we have them to make migration easy, but I don't think we should be encouraging people to use them for new tests. So for that reason, we should have one set of annotations - the same name as junit so it's just a matter of changing imports, and then if you want to migrate, you can change

```kotlin
@BeforeEach
fun myBeforeEachFunc() { }
```

to

```kotlin
override fun beforeTest(....) { }
```

No point having an extra step in the middle.

2. Rather than add support for the annotations into spec runner, let's keep them isolated in the annotation spec itself. You just need to do something like

```kotlin
override fun beforeTest(...) {
  // find annotated method if exists and invoke
}
```

In other words, the spec runner just invokes the normal listener methods, but in annotation spec they delegate to an annotation lookup.

Kerooker:
Your comments make a lot of sense. I agree 100%, and will change accordingly.

However, which annotations should we keep? Only ones with JUnit 5, JUnit 4 or both?

Also, should we really port JUnit's annotations then? Or should we not try to do that and prevent some bugs with JUnitPlatform execution

Kerooker:
`@Ignore` must still be in SpecRunner, I think

sksamuel:
Looks good but the renaming of the tests in ProjectConfigTest is a mistake
I assume.

Kerooker:
You're correct, but this force-push removed the renaming.

PR #520, opened 2019-01-03
Fixes documentation to the right annotation

Kerooker:
As of PR #514, a small bug passed through the documentation. We renamed `@BeforeTest` to `@BeforeEach`, but we forgot to change the documentation


Kerooker:
@sksamuel 
Just FYI. Merging this

PR #521, opened 2019-01-05
Test listener improvements #494 #497

sksamuel:
I've closed the other test listener branch and re-done the work here bringing it into line with the coroutine changes.

Once this is approved and merged, I think we're good to release RC1 for 3.2.


Kerooker:
What is appveyor complaining about?

sksamuel:
Haven't looked as its always complaining.

kotlintest-core/src/main/kotlin/io/kotlintest/extensions/TestListener.kt
```diff
@@ -4,26 +4,126 @@ import io.kotlintest.Description
...
+  /**
+   * Is invoked once the Test Engine has finished.
+   */
+  fun afterProject(): Unit = Unit
+
+  /**
+   * This callback will be invoked before a [TestCase] is executed.
    *
-   * @param description the [Description] for the [TestCase] instance.
+   * If a test case is inactive (disabled), then this method will not
```
Kerooker: Wouldn't this work similarly to `afterTest`? If yes, then the description should be similar

kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
```diff
@@ -163,25 +167,25 @@ class TestCaseExecutor(private val listener: TestEngineListener,
...
-    userListeners.forEach { it.beforeTest(testCase.description) }
-
-    if (testCase.config.enabled) {
-      // Only execute before test from the spec if the test is enabled
-      testCase.spec.beforeTest(testCase.description)
+    val userListeners = testCase.spec.listeners() + testCase.spec + Project.listeners()
+    userListeners.forEach {
+      it.beforeTest(testCase.description)
+      if (isActive(testCase)) {
+        it.beforeTest(testCase)
```
Kerooker: Wouldn't there be scenarios in which we still want to listen to a test even if it's inactive? For instance, if I want to override the testcase execution.

I set it to only ignore if it was a `spec` because by default we don't want to execute an ignored testcase when we're on a spec.

I think that `isActive` should only be verified if we're running from a spec.
Also, by executing the deprecated method before we verify if the test is active, we aren't allowing some of our own tests to fail if they should, as the scenario would be hidden inside this if

kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
```diff
@@ -163,25 +167,25 @@ class TestCaseExecutor(private val listener: TestEngineListener,
...
   private fun after(testCase: TestCase, result: TestResult) {
-    val userListeners = testCase.spec.listeners() + Project.listeners()
-    userListeners.reversed().forEach { it.afterTest(testCase.description, result) }
-
-    if(testCase.config.enabled) {
-      // Only execute after test from spec if the test is enabled
-      testCase.spec.afterTest(testCase.description, result)
+    val userListeners = testCase.spec.listeners() + testCase.spec + Project.listeners()
+    userListeners.reversed().forEach {
+      it.afterTest(testCase.description, result)
```
Kerooker: similar comments of `before`

kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/internal/topLevelTests.kt
```diff
@@ -0,0 +1,33 @@
+package io.kotlintest.runner.jvm.internal
```
Kerooker: I feel that this function could be placed near `Spec`, as it's completely related to `Spec`.
I would also add it as a member function of the `Spec` class ,with `Spec` as receiver, but I don't think that's needed now.

kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/spec/SpecExecutor.kt
```diff
@@ -0,0 +1,87 @@
...
+class SpecExecutor(private val engineListener: TestEngineListener,
+                   private val listenerExecutors: ConcurrentLinkedQueue<ExecutorService>,
+                   private val scheduler: ScheduledExecutorService) {
+
+  private val logger = LoggerFactory.getLogger(this.javaClass)
+
+  private fun withListenerExecutor(thunk: (ExecutorService) -> Unit) {
+    val listenerExecutor = listenerExecutors.poll() ?: Executors.newSingleThreadExecutor()
+    thunk(listenerExecutor)
+    listenerExecutors.add(listenerExecutor)
```
Kerooker: Should this be before thunk?
I'm thinking that if `thunk` fails, we'll have a dangling executor that might not be garbage collected if it's executing something

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/runner/junit5/JUnitTestRunnerListenerTest.kt
```diff
@@ -32,8 +32,8 @@ class JUnitTestRunnerListenerTest : WordSpec({
       val listener = JUnitTestRunnerListener(mock, rootDescriptor)
 
       val spec = JUnitTestRunnerListenerTest()
-      listener.prepareSpec(spec.description(), spec::class)
-      listener.completeSpec(spec.description(), spec::class, null)
+      listener.beforeSpecClass(spec.description(), spec::class)
```
Kerooker: If we do things like this, we're not guaranteeing that our deprecated methods work anymore. Is that ok?

Kerooker:
I approved the changes, but I don't think we should merge before fixing Appveyor.

It's complaining for a reason, and we should pinpoint this reason or stop using Appveyor. We should never ignore its broken build

kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
Kerooker: My first point is: We should never execute `beforeTest` if the test is inactive, unless an Extension of some sort wants to do it. There is a test (I don't remember which, but it broke recently) that exercises this behavior.

My second point, and the last paragraph: In line 172, we're using a deprecated method, right? If we do that for deprecations, some of our tests won't test the new behavior, they'll keep testing the old one

Kerooker:
I just have my concerns on executing `before` and `after` if the test is disabled. I don't think it maked sense to do. 

I do believe that extensions might do something with a test even if it's ignored, that's why we shouldn't deny this extensions to override `before` and `after`. But I believe that they're `intercepted`, not ran with `before` and `after`.

If my concern here is cleared, I'm ok to merge this

Kerooker:
And my point was if this is ok.
I think we're fine to merge and release

sksamuel:
I see.

Well I don't know. You can either call them all, and have a flag, or skip
the disabled ones.
I did it this way as it was easier to implement.
The beforeSpec listener does get a list of all tests, disabled or nt.

Kerooker:
Let's go with this for snapshot 1 then. If anything seems weird we can fix

PR #534, opened 2019-01-09
Adds documentation on test ordering

### LINKED ISSUE #439: Documentation on test ordering
The note at the end of the chapter **Parallelism** (https://github.com/kotlintest/kotlintest/blob/master/doc/reference.md#parallelism) states:

> Note: Test cases inside each spec will always run sequentially (either in definition order, or in a random order, see documentation on test ordering).

However I can't find this documentation on test ordering. Where can I find it?

At the moment, my tests rely heavily on their definition order in their spec and it looks like they are actually running in this way. But I don't know if, in the future, a spec "decided" to run its tests in random order.
Should I avoid to depend on test ordering inside each spec? In this way I can save some computational time, recycling resources.

Kerooker:
Fixes #439

### DIFF
--- a/doc/test_ordering.md
+++ b/doc/test_ordering.md
@@ -1,0 +1,44 @@
+                                                         
+### Test Ordering
+
+When running multiple tests from a Spec, there's a certain order on how to execute them.
+ 
+ By default, a **sequential** order is used (order that tests are defined in the spec), but it's also possible to configure them to be executed in a **random** order.
+
+This setting can be configured in either a `Spec` or in [ProjectConfig](/doc/reference.md#project-config), by overriding the `testCaseOrder` function. If both exist, the `Spec`'s configuration will have priority.
+
+**Note**: Nested tests will always run in discovery order (sequential)
+
+```kotlin
+class SequentialSpec : StringSpec() {
+
+    override fun testCaseOrder(): TestCaseOrder? = TestCaseOrder.Sequential
+    
+    init {
+      "This test runs first" {
+        // Test
+      }
+      
+      "This test runs second" {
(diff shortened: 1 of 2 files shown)


Kerooker:
Just need to check if it's ok to be in another file, or if we want to add it at `reference.md` itself. I don't see a very big difference, and I'm ok either way.

doc/test_ordering.md
```diff
@@ -0,0 +1,45 @@
+                                                         
+### Test Ordering
+
+When running multiple tests from a Spec, there's a certain order on how to execute them.
+ 
+ By default, a **sequential** order is used (order that tests are defined), but it's also possible to configure them to be executed in a **random** order.
```
sksamuel: I'd say (the order that tests are defined in the spec).

doc/test_ordering.md
```diff
@@ -0,0 +1,45 @@
...
+
+This setting can be configured in either a `Spec` or in [ProjectConfig](/doc/reference.md#project-config), by overriding the `testCaseOrder` function. If both exist, `Spec`'s configuration will have priority.
```
sksamuel: If both exist, the Spec's...

doc/test_ordering.md
```diff
@@ -0,0 +1,45 @@
...
+      "This test may run first or second" {
+        // Test
+      }
+    }
+
+}
+
+```
+
+**Note**: Nested tests will always run in discovery order (sequential)
```
sksamuel: I would move this above the example as people may not notice it.

Kerooker:
Resolved your three comments.

PR #559, opened 2019-01-14
Implement expected exception for AnnotationSpec

### LINKED ISSUE #527: ExpectedException for AnnotationSpec
In the same line of https://github.com/kotlintest/kotlintest/issues/522, JUnit has a parameter in `@Test` called `expected`
This parameter verifies that the test threw an Exception or a subclass of it.

For example:

```
@Test(expected = FooException::class)
fun foo() {
    throw FooException() // Test passes
}

@Test(expected = BarException::class)
fun bar() {
    println("Bar?")  // Test fails, BarException wasn't thrown
}
```

Should we also support it?

LeoColman:
JUnit's tests can test an exception is thrown by using a parameter in the `@Test` annotation. https://github.com/junit-team/junit4/wiki/exception-testing

Although we support `shouldThrow` and it's variants, we're trying to improve the likelihood of people using KotlinTest instead of JUnit, by making the migration as easy as possible. Because of this objective, this commit implements a parameter in `@Test`, to enable users to configure an expected exception just like JUnit's.

To do this, this commit creates the parameter and annotation processing for it.

To reuse code, this commit moves some internal classes around, to enable `kotlintest-core` module to use some of the classes from `kotlintest-assertions` and `kotlintest-runner-jvm`. This doesn't impact any user, as the moved classes and functions are internal/private.

To test that the new behavior works, this commit uses `SpecLevelExtension` to map specific tests with specific behaviors. It's not very pretty, but it works.

Fixes #527

### DIFF
--- a/kotlintest-core/src/main/kotlin/io/kotlintest/internal/ThrowableUnwrapping.kt
+++ b/kotlintest-core/src/main/kotlin/io/kotlintest/internal/ThrowableUnwrapping.kt
@@ -1,0 +1,16 @@
+package io.kotlintest.internal
+
+import java.lang.reflect.InvocationTargetException
+
+/**
+ * In some particular cases, such as AnnotationSpec, a call will be made using Reflection.
+ * When using reflection, any error will be wrapped around a InvocationTargetException, as explained
+ * in https://stackoverflow.com/questions/6020719/what-could-cause-java-lang-reflect-invocationtargetexception
+ * By verifying if this is an InvocationTargetException, we can unwrap it and throw the cause instead
+ */
+fun Throwable.unwrapIfReflectionCall(): Throwable {
+  return when (this) {
+    is InvocationTargetException -> cause ?: this
+    else -> this
+  }
+}
--- a/kotlintest-core/src/main/kotlin/io/kotlintest/specs/AbstractAnnotationSpec.kt
+++ b/kotlintest-core/src/main/kotlin/io/kotlintest/specs/AbstractAnnotationSpec.kt
@@ -5,8 +5,10 @@ import io.kotlintest.Description
 import io.kotlintest.Spec
 import io.kotlintest.TestCase
 import io.kotlintest.TestCaseConfig
+import io.kotlintest.TestContext
 import io.kotlintest.TestResult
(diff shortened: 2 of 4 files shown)


kotlintest-core/build.gradle
```diff
@@ -2,4 +2,6 @@ dependencies {
     compile 'org.slf4j:slf4j-api:1.7.25'
     compile "org.jetbrains.kotlin:kotlin-reflect"
     compile 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.1.0'
+
+    compile project(":kotlintest-assertions")
```
sksamuel: Don't include assertions with core, not everyone wants them.

Kerooker: I thought that everyone always had all of the code anyways, so this wouldn't make a difference. Assertions and Core don't share anything, then? Should I duplicate the Failures class? This doesn't look very good

sksamuel: They typically do, but some people only use assertions, and some people only use the framework.

I suppose it's not too bad for core to use assertions actually, but definitely assertions should stand alone.

Perhaps core should be renamed to framework or something.

Kerooker: Should we keep this import, then?
Or should I find another way?

sksamuel: What do you need from assertions for it to work ?

Kerooker: Just using Failures for now... It may be overkill to compile project to this one, but it's also not good to copy Failures or any implementation from there to over here, thus duplicating code

sksamuel: Let's (for now) copy failures (at least just the bits you need).
Then for 3.3 we'll make core depend on assertions and maybe rename core to -test-framework or something.

Kerooker: Ok, I did it with the latest force push.

Kerooker:
If we're ok with this, we should keep this PR opened until we release 3.2.0.

But is everything fine now?

sksamuel:
Yep merge after 3.2.0

sksamuel:
Approved

PR #590, opened 2019-01-18
Add cancel of TestCaseExecutor interruption job if test succeeds

### LINKED ISSUE #588: Starting to see InterruptedExceptions with kotlintest 3.2.0
While running tests on the [PR that migrates our app to kotlintest 3.2.0](https://github.com/heremaps/oss-review-toolkit/pull/1195) I start to see `InterruptedExceptions` where previously (with kotlintest 3.1.11) everything was fine. See e.g.

https://travis-ci.com/heremaps/oss-review-toolkit/builds/97502719#L1263
https://ci.appveyor.com/project/heremaps/oss-review-toolkit/builds/21656801#L3532

I'm wondering whether that's related to the [fix for test timeouts](https://github.com/kotlintest/kotlintest/issues/476) that was merged as part of #518.

LeoColman:
As it is, the current TestCaseExecutor will by mistake recycle a thread from a threadpool between multiple test executions. That shouldn't be a problem. However, the scheduler will cancel the executor after a timeout has passed, and if another thread is running there, it will be interrupted.

This interruption is undesired, as it had nothing to do with the thread that successfully passed without a timeout.

This commit cancels the interruption job if the test ran successfully, avoiding this issue.

Fixes #588

### DIFF
--- a/kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
+++ b/kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
@@ -138,7 +138,7 @@ class TestCaseExecutor(private val listener: TestEngineListener,
       }
 
       // we need to interrupt the threads in the executor in order to effect the timeout
-      scheduler.schedule({
+      val scheduledInterrupt = scheduler.schedule({
         error.compareAndSet(null, TimeoutException("Execution of test took longer than ${testCase.config.timeout}"))
         // this will ruin the listener executor so after() won't run if the test times out but I don't
         // know how else to interupt the coroutine context effectively. job.cancel() won't cut the mustard here.
@@ -150,6 +150,10 @@ class TestCaseExecutor(private val listener: TestEngineListener,
       }
 
       supervisorJob.join()
+
+      // This cancel is necessary, or else the scheduler is reutilized by another test, and the shutdown will happen
+      // in the other test. This leads to tests "sharing" a timeout, and another test being timed out due to this task.
+      scheduledInterrupt.cancel(false)
       val result = buildTestResult(error.get(), context.metaData())
 
       listener.afterTestCaseExecution(testCase, result)
--- a/kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
+++ b/kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
@@ -39,7 +39,7 @@ import java.util.concurrent.atomic.AtomicReference
  * The executor can be shared between multiple tests as it is thread safe.
  */
 class TestCaseExecutor(private val listener: TestEngineListener,
-                       private val listenerExecutor: ExecutorService,
+                       private val executor: ExecutorService,
                        private val scheduler: ScheduledExecutorService) {
 
   private val logger = LoggerFactory.getLogger(this.javaClass)
@@ -48,7 +48,8 @@ class TestCaseExecutor(private val listener: TestEngineListener,
 
     try {
 
-      context.launch(listenerExecutor.asCoroutineDispatcher()) {
+      // invoke the "before" callbacks here on the main executor
+      context.launch(executor.asCoroutineDispatcher()) {
         before(testCase)
       }.join()
 
@@ -58,20 +59,10 @@ class TestCaseExecutor(private val listener: TestEngineListener,
 
       // get active status here in case calling this function is expensive (eg
       runExtensions(testCase, context, extensions) { result ->
-
-        // it's possible the listenerExecutor has been shut down here.
-        // If it has, we can only run them on another thread, better than a slap in the face
-        // but will run foul of https://github.com/kotlintest/kotlintest/issues/447
-        if (listenerExecutor.isShutdown) {
-          context.launch {
-            after(testCase, result)
-          }.join()
-        } else {
-          context.launch(listenerExecutor.asCoroutineDispatcher()) {
-            after(testCase, result)
-          }.join()
-        }
-
+        // invoke the "after" callbacks here on the main executor
+        context.launch(executor.asCoroutineDispatcher()) {
+          after(testCase, result)
+        }.join()
         onResult(result)
       }
 
@@ -103,12 +94,12 @@ class TestCaseExecutor(private val listener: TestEngineListener,
 
       listener.beforeTestCaseExecution(testCase)
 
-      // if we have more than one requested thread, we run the tests inside an executor,
-      // otherwise we run on the same thread as the listeners to avoid issues where before/after listeners
-      // require the same thread as the test case.
+      // if we have more than one requested thread, we run the tests inside a clean executor;
+      // otherwise we run on the same thread as the listeners to avoid issues where the before and
+      // after listeners  require the same thread as the test case.
       // @see https://github.com/kotlintest/kotlintest/issues/447
       val executor = when (testCase.config.threads) {
-        1 -> listenerExecutor
+        1 -> executor
         else -> Executors.newFixedThreadPool(testCase.config.threads)!!
       }
 
@@ -137,13 +128,12 @@ class TestCaseExecutor(private val listener: TestEngineListener,
         }
       }
 
-      // we need to interrupt the threads in the executor in order to effect the timeout
-      val scheduledInterrupt = scheduler.schedule({
-        error.compareAndSet(null, TimeoutException("Execution of test took longer than ${testCase.config.timeout}"))
-        // this will ruin the listener executor so after() won't run if the test times out but I don't
-        // know how else to interupt the coroutine context effectively. job.cancel() won't cut the mustard here.
-        executor.shutdownNow()
-      }, testCase.config.timeout.toMillis(), TimeUnit.MILLISECONDS)
+      // we schedule a timeout, (if timeout has been configured) which will fail the test with a timed-out status
+      if (testCase.config.timeout.nano > 0) {
+        scheduler.schedule({
+          error.compareAndSet(null, TimeoutException("Execution of test took longer than ${testCase.config.timeout}"))
+        }, testCase.config.timeout.toMillis(), TimeUnit.MILLISECONDS)
+      }
 
       supervisorJob.invokeOnCompletion { e ->
         error.compareAndSet(null, e)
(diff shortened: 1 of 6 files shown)


LeoColman:
@sksamuel @sschuberth 
Here's what I did here:
I created a scenario that I believe is a reproduction of the bug, however I'm not sure if it's actually a reproduction or another bug that was found while trying it.

This scenario exercises my main hypothesis of "I believe the tests are sharing a timeout". It works by creating a big timeout that will be "accumulated" between tests, and one of the tests will have a "Thread.sleep interrupted" message. This, I believe, is the executor being shutdown from another test.

When I added a `job.cancel()` inside TestCaseExecutor, this behavior stopped happening, and all tests complete flawlessly, because I won't run the interruption if the test succeeds.

LeoColman:
If you remove the `cancel`, a test from the added class will fail for apparently no reason.

kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
```diff
@@ -150,6 +150,10 @@ class TestCaseExecutor(private val listener: TestEngineListener,
       }
 
       supervisorJob.join()
+
+      // This cancel is necessary, or else the scheduler is reutilized by another test, and the shutdown will happen
+      // In the other test. This leads to tests "sharing" a timeout, and another test being timed out due to this.
```
sschuberth: Nit: s/In/in/

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/TestCaseExecutorTest.kt
```diff
@@ -288,4 +289,45 @@ class TestCaseExecutorTest : FunSpec() {
...
+
+  // The test executor was failing because as it reutilizes some threads from a thread pool.
+  // When using that thread pool, a task to cancel the thread is created, so that the engine can interrupt
+  // a test that is going forever.
+  // However, if the task
+
+  init {
+    // 100 millis sleep will "accumulate" between tests. If the context is still shared, one of them will fail
+    // due to timeout. This only happens on multi-thread, as on single-thread the interruption is syncronous, therefore
+    // it can't interrupt another thread in this spec
```
sschuberth: Bit: Missing dot at end of sentence.

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/TestCaseExecutorTest.kt
```diff
@@ -288,4 +289,45 @@ class TestCaseExecutorTest : FunSpec() {
...
     }
   }
+}
+
+class MultipleTestTimeoutTest : StringSpec() {
+
+  // The test executor was failing because as it reutilizes some threads from a thread pool.
+  // When using that thread pool, a task to cancel the thread is created, so that the engine can interrupt
+  // a test that is going forever.
+  // However, if the task
```
sschuberth: This sentence looks uncompleted.

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/TestCaseExecutorTest.kt
```diff
@@ -288,4 +289,45 @@ class TestCaseExecutorTest : FunSpec() {
...
+
+  init {
+    // 100 millis sleep will "accumulate" between tests. If the context is still shared, one of them will fail
+    // due to timeout. This only happens on multi-thread, as on single-thread the interruption is syncronous, therefore
+    // it can't interrupt another thread in this spec
+    "Test 1".config(timeout = 200.milliseconds, threads = 2) {
```
sschuberth: Probably a stupid question, but does `threads = 2` have any effect for a single test if `invocations` is still the default of `1`? I mean, if the test is only run once, what effect does `threads = 2` have here?

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/TestCaseExecutorTest.kt
LeoColman: true

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/TestCaseExecutorTest.kt
LeoColman: It doesn't. I added an extra thread because I know that internally, a thread pool will be used with 2+ threads, and the bug is on the thread pool. This is just me forcing it to use a thread pool

sschuberth: Ok, thanks. As I believe that piece of information to be critical to understand the code, could you please try to make that more clear from your source code comment?

LeoColman: Sure, makes sense

kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
```diff
@@ -150,6 +150,10 @@ class TestCaseExecutor(private val listener: TestEngineListener,
       }
 
       supervisorJob.join()
+
+      // This cancel is necessary, or else the scheduler is reutilized by another test, and the shutdown will happen
+      // in the other test. This leads to tests "sharing" a timeout, and another test being timed out due to this task.
+      scheduledInterrupt.cancel(false)
```
sksamuel: I'm confused by this. If the scheduled runnable still runs, all it does is shutdown the executor. The executor that will have already been shutdown, so what's the issue? There's no problem if it still runs right.

LeoColman: I'm also pretty confused by this, as the Executor should be a different one every time. However, the issue seems to be on `TestEngineListener`, as it's shared between tests. `executor` will be `listenerExecutor` for single-threaded tests, and these will be interrupted as the executor shuts down

LeoColman: Although I'm not completely sure if this is a `Threads` issue or a `Coroutines` issue of some sort, preventing the executor from shutting down if the test completes will fix the scenario that I created in `TestCaseExecutorTest`

LeoColman:
Running more tests now it made me realize this error has nothing to do with a threadpool. I'll remove it from the test scenario

LeoColman:
The error is specifically with `TestEngineListener`

sksamuel:
The error is what you said earlier I think.
The executor that we interrupt is not always a unique executor.
Sometimes it's the shared listener executor, which we _cannot_ shut down.

LeoColman:
Ok, so there are two options.
We either cancel the task if the test succeeds (current option)
Or only cancel the task if it's not `ListenerExecutor`.

I think it makes more sense to cancel the task

sksamuel:
Cancel won't do anything unless the user checks for it in their code. Which they won't 99% of the time.

The solution is to not share the listener executor. Always create a new one every time.
Then we can interrupt it and all will be well but... we can't call shutdown on the executor as that will prelude submitting the "after test" listener callbacks in the same thread.

Hmm...

LeoColman:
> Cancel won't do anything unless the user checks for it in their code.

What do you mean with this?
Cancel is used to stop the interruption job, not to cancel the thread/coroutine

This have nothing to do with user code, unless the code doesn't respons to a `Interrupt` call from thread, but if it's the case, a non-shared executor won't help anyways

Kerooker:
With this commit, I'm saying "Only interrupt the job if the test is really timing out"

sksamuel:
> > Cancel won't do anything unless the user checks for it in their code.
> 
> What do you mean with this?
> Cancel is used to stop the interruption job, not to cancel the thread/coroutine
> 
> This have nothing to do with user code, unless the code doesn't respons to a `Interrupt` call from thread, but if it's the case, a non-shared executor won't help anyways

Yes I meant the coroutine. We can cancel the scheduler but it won't help.

sksamuel:
I think my original approach is wrong, and kotlintest 3.2.0 is fundamentally broken on test suites that run for over 10 minutes that use threads=1 because they all share the listener executor.

Calling cancel on the scheduler - there's no point even adding it in the first place as we cannot interrupt this executor full stop.

This is my proposal. For each spec we create a new single threaded listener executor (delete the old shared executor).
We launch the `before` callbacks in this executor.
We then launch the tests using this executor or if threads>1 we create a new executor.
We add a scheduled event into the shared scheduler, which upon timeout _fails_ that test.
We launch the `after` callbacks in the listener executor.

Only _then_ do we call shutdown on the listener executor (and the test executor if threads>1).
Not shutdownNow but simply shutdown.
If the user has infinite loops then they're on their own.

I don't like it as it's just not correct to leave the threads running in the background but I'm not sure what else we can do.

LeoColman:
I don't think leaving threads running isn't a biiiig problem. As you said, users are on their own if their threads are not interruptable and run indefinetly. This is such a rare scenario that I don't think will even happen.

I'll try to implement your proposal and add it to this commit if you won't focus on this now.

LeoColman:
But if you want to take it and use this branch, let me know

LeoColman:
We should fix it ASAP, as this will break a lot of test suites in 3.2.0

LeoColman:
If you want to add the cancel on the scheduler as a hotfix now, while we implement the correct fix, I'm for it aswell.

LeoColman:
I'm trying to ignore the `ExecutorListener`, but even when doing this, tests in a same spec will still fail if the `cancel` doesn't happen. Unless we give an executor to each `Test`, and ignore the threads. I think `cancel` on the `Interrupt` is kinda OK for now.

Let me know if you think differently

sksamuel:
I can fix and release tonight if you want me to do it.
No point hot fixing, I'll do it properly for 3.2.1

LeoColman:
Sure thing, I think it will be better if you do it.
I don't have a lot of experience on these runners.

Use this branch if you want, as the test can be used from it.

sksamuel:
Ok

sksamuel:
All tests now pass using an implementation as I described earlier.

PR #593, opened 2019-01-18
Add extra scopes for BehaviorSpec

### LINKED ISSUE #562: Request for BehaviorSpec sample
Hi,

Version: "io.kotlintest:kotlintest-runner-junit5:3.1.11"
Kotlin Version: 1.3.11

I am curious how to use the `And` syntax with a BehaviorSpec.
```kotlin
given("a stateful object") {
  `when`("the stateful object is configured with A") {
      victim.setA("foo");
      and("the statefule object is configured with B") {
        victim.setB("bar");
         then("it should print hello") {
          victim.print() shouldBe "hello"
        }
      }
    }
}
```

When trying to do this the receiver context can't be determined:
`Unit can't be called in this context by implicit receiver. Use explicit context if necessary`.

Adding the context `this@given.and` does not resolve the issue and the test context is never executed.

I couldn't find a concrete example.

Thanks!

LeoColman:
To improve BehaviorSpec customizability, this commit adds extra scopes to BehaviorSpec. As discussed, users can concatenate `And` indefinitely, but it's ok. As `And` is optional to complete a test. This behavior is also observed in `FreeSpec`, as it can concatenate infinite test scopes too.

To do that, Given and And/When contexts received a new possibility of scoping, `And`. There are some variations of it, to build tests correctly: GivenAndContext and WhenAndContext. All of these are optional to use.

Fixes #562

### DIFF
--- a/doc/styles.md
+++ b/doc/styles.md
@@ -114,6 +114,26 @@ class MyTests : BehaviorSpec({
 Because `when` is a keyword in Kotlin, we must enclose with backticks. Alternatively, there are title case versions
 available if you don't like the use of backticks, eg, `Given`, `When`, `Then`.
 
+You can also use the `And` keyword in `Given` and `When` to add an extra depth to it:
+
+```kotlin
+class MyTests : BehaviorSpec({
+    given("a broomstick") {
+        and("a witch") {
+            `when`("The witch sits on it") {
+                and("she laughs hysterically") {
+                     then("She should be able to fly") {
+                     }
+                 }
+            }
+        }
+    }
+})
+
+```
+
+Note: `Then` scope doesn't have an `and` scope due to a gradle bug. For more information, see #594
+
 ### Free Spec
 
 `FreeSpec` allows you to nest arbitary levels of depth using the keyword `-` (minus), as such:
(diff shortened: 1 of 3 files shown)


LeoColman:
Added docs

sksamuel:
`TestType.Test` cannot (or should not) be nested inside other's.
They should be used for leaves only.
So I think allow 'and' everywhere but inside the 'Then' ?

LeoColman:
I was thinking about it. I think the correct TestType would be `TEST_AND_CONTAINER`, but I remember that you couldn't use it because it is bugged in Gradle.

What I had to do there is that `Then` can be either a final test or a scope for `And`. The tests are working, and I only did this instead of `Container` so the IDE would mark it as a test, not a container.

Perhaps there's another way?

LeoColman:
And what are the rules for TestType.TEST to not be nested inside another one? I know it's counterintuitive when you read it, but are there cases that it wouldn't work?

LeoColman:
As this is an intended feature from Gradle, could we add this anyways and mark it for fixing when gradle fixes its issue?

sksamuel:
In theory you can nest TEST inside TEST_AND_CONTAINER, but it's all broken in gradle and they haven't fixed it in over a year. I guess other features take priority, although I would have thought working test support would be high priority in build tool. At this rate they might not fix it until 2020 so we can't rely on it working any time soon.

Anyway, just don't use TEST inside TEST, so remove 'and' from within 'then'.

LeoColman:
That's a bummer.
I will remove it, but will let it commented for whenever gradle does it's thing.

LeoColman:
Nvm, comments are very ugly.

LeoColman:
Should be good to go.

sksamuel (approved):

PR #602, opened 2019-01-21
Add nullability matchers with Kotlin Contracts

### LINKED ISSUE #601: Assert not null or specific object type for further assertions on an object
Say:
```kotlin
val obj: Person? = ....
obj shouldNotBeNullAnd {
    it.name shouldBe "b"
    it.age shouldNotBe 43
}
```
This should and assert that Person is not null, and provide the ability to check the non-null result. This could also probably be done with Kotlin Contracts, just by doing a null assertion before other assertions. The difference between this and `result!! should { }` is that the previous is an actual assertion and the latter is a runtime failure.

Something similar could possibly done for `shouldBeInstanceOf` to make assertions on the already casted instance.

LeoColman:
With Kotlin Contracts it's possible to establish that "If this function passes, X is true". This commit implements nullability type matchers for the `Any?` type, using Contracts to allow non-null inference to users.

This commit didn't implement the same logic for matchers such as `x.shouldBeInstanceOf<Foo>()`, because currently it's impossible to implement this with Kotlin Contracts. https://youtrack.jetbrains.com/issue/KT-28298

Fixes #601

### DIFF
--- a/doc/matchers.md
+++ b/doc/matchers.md
@@ -27,6 +27,7 @@ For the extension function style, each function has an equivalent negated versio
 | `obj.shouldBeTypeOf<T>()` | Asserts that the given reference is exactly of type T. Subclass will fail. Ie, `1 should beOfType<Number>` would fail because although 1 _is_ a Number, the runtime type is not Number. |
 | `obj.shouldBeInstanceOf<T>` | Asserts that the given reference is of type T or a subclass of T. |
 | `obj.shouldHaveAnnotation(annotationClass)` | Asserts that the object has an annotation of the given type. |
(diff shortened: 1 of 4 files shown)


LeoColman:
@dave08
Take a look.
Special attention to the part that states `shouldBeInstanceOf` doesn't allow for Contracts to be used

dave08:
You might want to add another assertion after the `shouldNotBeNull()` that uses the non-null value so that if there are breaking changes to contracts, the tests will fail?

LeoColman:
Done

sksamuel (approved):

sksamuel:
3.3 only.

PR #603, opened 2019-01-22
Implement shouldNotThrow matchers

### LINKED ISSUE #205: shouldNotThrow and shouldNotThrowAny matchers
Having these matchers would be great 
-`shouldNotThrow` 
-`shouldNotThrowAny`

Currently there is not way of expressing that easily with kotlintest matchers

LeoColman:
This commit implements `shouldNotThrow`, `shouldNotThrowExactly` and `shouldNotThrowAny` matchers, with their `Unit` variations.

Effectively, for the user, it doesn't make a difference on calling this and calling the code directly, but this may be better for communicating intent explicitly.

These matchers won't let an uncaught failure to be silently caught. Any uncaught exception will still be propagated upwards, however, if it matches the `shouldThrow<T>` T parameter, it will be wrapped around an AssertionError (makes tests Yellow, not Red.).

It's never necessary to use this construct. It would be the same as every java method being `void foo() throws no Exception`, and it usually doesn't make sense to.

Fixes #205

### DIFF
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/AnyThrowableHandling.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/AnyThrowableHandling.kt
@@ -16,6 +16,23 @@ package io.kotlintest
 inline fun shouldThrowAnyUnit(block: () -> Unit) = shouldThrowAny(block)
 
 /**
+ * Verifies that a block of code does NOT throw any [Throwable]
+ *
+ * Use function to wrap a block of code that you want to make sure doesn't throw any [Throwable],
+ * where using [shouldNotThrowAny] can't be used for any reason, such as an assignment of a variable (assignments are
+ * statements, therefore has no return value).
+ *
+ * Note that executing this code is no different to executing [block] itself, as any uncaught exception will
+ * be propagated up anyways.
+ *
+ * @see [shouldNotThrowAny]
+ * @see [shouldNotThrowExactlyUnit]
+ * @see [shouldNotThrowUnit]
+ *
+ */
+inline fun shouldNotThrowAnyUnit(block: () -> Unit) = shouldNotThrowAny(block)
+
+/**
  * Verifies that a block of code throws any [Throwable]
  *
  * Use function to wrap a block of code that you want to verify that throws any kind of [Throwable].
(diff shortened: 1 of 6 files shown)


LeoColman:
Note that documentation is purposefully not updated, as this isn't a real feature. Following our conventions that Every *should* have a *shouldNot*, this is just a small fix.

LeoColman:
@sksamuel 
If you could give the javadocs a very quick read... I copy-pasted most of it, and maybe there's a small slip

kotlintest-assertions/src/main/kotlin/io/kotlintest/AnyThrowableHandling.kt
```diff
@@ -15,6 +15,23 @@ package io.kotlintest
...
 inline fun shouldThrowAnyUnit(block: () -> Unit) = shouldThrowAny(block)
 
+/**
+ * Verifies that a block of code does NOT throw any [Throwable]
+ *
+ * Use function to wrap a block of code that you want to make sure doesn't throw any [Throwable],
+ * where using [shouldNotThrowAny] can't be used for any reason, such as an assignment of a variable (assignments are
+ * statements, therefore has no return value).
+ *
+ * Note that executing this code is not very different than executing [block] by itself, as any uncaught exception will
```
sksamuel: * Note that executing this code is no different to executing the [block] by itself, as any uncaught exception will...

sksamuel (approved):

kotlintest-assertions/src/main/kotlin/io/kotlintest/AnyThrowableHandling.kt
LeoColman: It's somewhat different, as it will wrap exceptions in an AssertionError, but I'll change that

LeoColman:
Merging when this gets green.

kotlintest-assertions/src/main/kotlin/io/kotlintest/AnyThrowableHandling.kt
```diff
@@ -45,4 +62,33 @@ inline fun shouldThrowAny(block: () -> Any?): Throwable {
...
+ *
+ * Use function to wrap a block of code that you want to make sure doesn't throw any [Throwable].
+ *
+ * Note that executing this code is no different to executing [block] itself, as any uncaught exception will
+ * be propagated up anyways.
+ *
+ * **Attention to assignment operations:**
+ *
+ * When doing an assignment to a variable, the code won't compile, because an assignment is not of type [Any], as required
+ * by [block]. If you need to test that an assignment doesn't throw a [Throwable], use [shouldNotThrowAnyUnit] or it's
```
TAGC, line 79: *its

PR #606, opened 2019-01-24
Enables test bang on all specs

### LINKED ISSUE #605: Question: Bang and Focus outside of StringSpec
Hi,

Are bang and focus only supported in StringSpec? I like bang over `.config(enabled = false)` but it doesn't appear to work with the BehaviorSpec at the `given` level. I have not tried other specs, however. 

This doesn't appear to work:
```
given("!An ignored test") {
  `when`("executed") {
    then("it does not run") {
    }
  }
}
```

Thanks!

LeoColman:
Before this, some specs couldn't be marked as banged with a `!`. This happened due to the fact that we check for a exclamation mark at the beginning of a test name, such as "Foo" -> "!Foo" would be banged in FreeSpec.

However, some specs such as `BehaviorSpec` wouldn't allow this to work, as `Given("!Foo")` would be turned into `Given: !Foo`. When we checked for the start of the test name, it wouldn't contain `!`, therefore it wouldn't be ignored.

This commit changes all specs that would fall in that scenario, to move an exclamation mark from the beginning of the test chosen name to the beginning of the test final name (Before Given, for example. `Given("!Foo") -> "!Given: Foo"`), fixing this issue.

I chose to do it this way to keep consistent. "Why did this test get ignored?" would be better visualized when `Given` is prefixed with `!`, and not if we edited the bang mechanism to consider only the test name, and not the prefix.

Fixes #605

### DIFF
--- a/kotlintest-core/src/main/kotlin/io/kotlintest/specs/TestPrefixes.kt
+++ b/kotlintest-core/src/main/kotlin/io/kotlintest/specs/TestPrefixes.kt
@@ -1,0 +1,5 @@
+package io.kotlintest.specs
+
+internal fun createTestName(prefix: String, name: String): String {
+  return if(name.startsWith("!")) "!$prefix${name.drop(1)}" else "$prefix$name"
+}
--- a/kotlintest-core/src/main/kotlin/io/kotlintest/specs/AbstractBehaviorSpec.kt
+++ b/kotlintest-core/src/main/kotlin/io/kotlintest/specs/AbstractBehaviorSpec.kt
@@ -19,7 +19,7 @@ abstract class AbstractBehaviorSpec(body: AbstractBehaviorSpec.() -> Unit = {}) 
   fun given(name: String, test: suspend GivenContext.() -> Unit) = addGivenContext(name, test)
 
   private fun addGivenContext(name: String, test: suspend GivenContext.() -> Unit) {
-    addTestCase("Given: $name", { thisSpec.GivenContext(this).test() }, defaultTestCaseConfig, TestType.Container)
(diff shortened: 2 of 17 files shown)


kotlintest-core/src/main/kotlin/io/kotlintest/specs/TestPrefixes.kt
```diff
@@ -0,0 +1,5 @@
+package io.kotlintest.specs
+
+internal fun createTestName(prefix: String, name: String): String {
+  return if(name.startsWith("!")) "!$prefix${name.replaceFirst("!", "")}" else "$prefix$name"
```
sksamuel: `return if (name.startsWith("!")) "!$prefix${name.drop(1)}" else "$prefix$name"` maybe a bit more terse

sksamuel (approved):

kotlintest-core/src/main/kotlin/io/kotlintest/specs/TestPrefixes.kt
LeoColman: done

sksamuel:
I think we could put this on 3.2 as well and release 3.2.2
We'd need a release/3.2 branch cut from the last 3.2.1 release.

LeoColman:
We could also remove from master everything that should go to 3.3 only and move it to `release/3.3`

LeoColman:
I don't think there's much difference

sksamuel:
I want current dev to be master.
Master is special and should be treated specially.
If we don't use master, then what do people PR against. They won't know.

Issue #524, opened 2019-01-07
SystemProperty Test Helpers


sksamuel:
Might be useful to have a trait that offers some functionality to perform a test with a system property enabled, and then restores the system property to whatever it was before.

```
class MyTest : FunSpec({
  test("foo") {
     withSysProperty("key", "value") {
       // test here
     }
  }
```

sksamuel:
Your thoughts @Kerooker

Kerooker:
Find it great.
Would be good to have a plural form too `withSysProperties`

LeoColman:
Closed as per #608

PR #608, opened 2019-01-24
SystemProperty Test Helpers #524

sksamuel:


kotlintest-assertions/src/main/kotlin/io/kotlintest/assertions/sysprops.kt
```diff
@@ -0,0 +1,34 @@
+package io.kotlintest.assertions
+
+import java.util.*
+
+fun <T> withSystemProperty(key: String, value: String, thunk: () -> T): T {
```
LeoColman: Perhaps allow value to be nullable? One might want to override a specific System property to null

kotlintest-assertions/src/main/kotlin/io/kotlintest/assertions/sysprops.kt
```diff
@@ -0,0 +1,34 @@
...
+    return thunk()
+  } finally {
+    if (previous == null)
+      System.clearProperty(key)
+    else
+      System.setProperty(key, previous)
+  }
+}
+
+fun <T> withSystemProperties(props: List<Pair<String, String?>>, thunk: () -> T): T {
```
LeoColman, line 17: Also add an overload to `Map<String, String?>`?

LeoColman:
Do you think it's worth to add a `SystemPropertyExtension` of some sort?

LeoColman:
Missing documentation on how to use this

kotlintest-assertions/src/main/kotlin/io/kotlintest/assertions/sysprops.kt
sksamuel: Done.

kotlintest-assertions/src/main/kotlin/io/kotlintest/assertions/sysprops.kt
sksamuel: Done.

sksamuel:
> Do you think it's worth to add a `SystemPropertyExtension` of some sort?

Added system extension.

Documentation will come as a separate PR.

kotlintest-core/build.gradle
```diff
@@ -1,4 +1,5 @@
 dependencies {
+    compile project(':kotlintest-assertions')
```
LeoColman, line 2: We had a discussion about this a while back. Do you think it's ok for core to depend  on assertions now?

sksamuel:
Yes. The reason they're split is for people who want to use our assertions
but with junit (or spek even!).
But if you're using our test framework, then you're not going to mind one
more small dep.

LeoColman (approved):

Issue #587, opened 2019-01-16
Timezone Extension


sksamuel:
What do you think about adding something like this: https://junit-pioneer.org/docs/default-locale-timezone/

LeoColman:
I think it's great!

LeoColman:
The only point would be where to add it. Probably just a function to execute before a test, and we would have to clear it afterwards

sksamuel:
A TestCaseExtension that does it before every test and resets it after every test ?

LeoColman:
Might be good.
It's a very specific scenario, but a pain in the ass to test. I like it

LeoColman:
These were merged as per #609

PR #609, opened 2019-01-24
Timezone / Locale Extension #587

sksamuel:


LeoColman (approved):

LeoColman:
Great one!

Setting the Locale for what it was by default is good.

LeoColman:
Just one thing: this should be documented

Issue #611, opened 2019-01-27
Arrow Extension - null support


evanbennett:
In https://github.com/kotlintest/kotlintest/blob/master/kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/either/matchers.kt, should all uses of `Any` be `Any?` to support nullable types?

sksamuel:
Makes sense. Would you like to make a PR ?

evanbennett:
Sure.

PR #613, opened 2019-01-27
Allow Arrow Either extensions to support nullable types

evanbennett:
As requested for #611

### DIFF
--- a/kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/either/matchers.kt
+++ b/kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/either/matchers.kt
@@ -7,18 +7,18 @@ import io.kotlintest.matchers.beInstanceOf2
 import io.kotlintest.should
 import io.kotlintest.shouldNot
 
-fun <T> Either<Any, T>.shouldBeRight() = this should beRight()
-fun <T> Either<Any, T>.shouldNotBeRight() = this shouldNot beRight()
-fun <T> beRight() = beInstanceOf2<Either<Any, T>, Either.Right<T>>()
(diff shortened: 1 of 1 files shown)


evanbennett:
I am sorry, but I have no idea how I broke Travis CI.

kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/either/matchers.kt
```diff
@@ -60,10 +60,11 @@ inline fun <reified A> beLeftOfType() = object : Matcher<Either<Any?, Any?>> {
         Result(false, "Either should be Left<${A::class.qualifiedName}> but was Right(${value.b})", "")
       }
       is Either.Left -> {
-        if (value.a is A)
+        val valueA = value.a
+        if (valueA is A)
           Result(true, "Either should be Left<${A::class.qualifiedName}>", "Either should not be Left")
         else
-          Result(false, "Either should be Left<${A::class.qualifiedName}> but was Left<${value.a::class.qualifiedName}>", "Either should not be Left")
+          Result(false, "Either should be Left<${A::class.qualifiedName}> but was Left<${if (valueA == null) "Null" else valueA::class.qualifiedName}>", "Either should not be Left")
```
sksamuel, line 67: Can we replace the output with the usage of `convertValueToString` in kotlintest-assertions as this handles outputting null, empty string in a nicer way.

evanbennett: I have looked into this, but I cannot work out how this could be used here. The potential nullable is used to get it's qualified class name. If I call `convertValueToString(value.a)` there is no way to get the qualified class name as it would return a `String`.

sksamuel:
Merged, thank you.

PR #629, opened 2019-01-31
Move extensions to Kotlintest-Extensions module

### LINKED ISSUE #625: Rework extensions
We were discussing in #623 about reworking how extensions work, either creating a separate module or shoving all extensions in core.

Let's discuss a little bit on what we should do in this matter

LeoColman:
This commit moves all extensions to Kotlintest-Extensions module. This is necessary because as we add more extensions to the code, the Core module would get fuller by this. By keeping all extensions in the same place, it's easier to search and to refactor.

The moved extensions are safe to move, because neither `SystemProperties` nor `LocaleExtensions`, nor `TimezoneExtensions` are currently in production, therefore no users are using them yet.

All related tests were moved accordingly.

An important point to note is that the JVM-Runners module is now depending on the extensions modules. We did this to avoid users having to use more than one import in their dependencies. External/Integration extensions are still kept in their own modules, such as Spring and Allure.

Closes #625


LeoColman:
@sksamuel 
My only concern is that I might have screwed up the maven modules and jars. I'm not sure how they work completely

kotlintest-core/build.gradle
```diff
@@ -1,5 +1,4 @@
 dependencies {
-    compile project(':kotlintest-assertions')
```
sksamuel, line 2: Do we not want this now ?

kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/locale/TimezoneExtension.kt
```diff
@@ -1,22 +1,10 @@
-package io.kotlintest.extensions
+package io.kotlintest.extensions.locale
```
sksamuel, line 1: I'm not sure we need a package for a single file

kotlintest-extensions/src/test/kotlin/io/kotlintest/extensions/junitxml/JUnitXmlTest.kt
```diff
@@ -1,6 +1,5 @@
-package io.kotlintest.extensions.allure
+package io.kotlintest.extensions.junitxml
```
sksamuel: Same

sksamuel (approved):

kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/locale/TimezoneExtension.kt
LeoColman: There are two classes, so I created a package to hold 2 files

kotlintest-extensions/src/test/kotlin/io/kotlintest/extensions/junitxml/JUnitXmlTest.kt
LeoColman: Is junitxml related to allure in any way? I moved it to the a different package because it was a different thing

kotlintest-core/build.gradle
LeoColman: IDK if we want it, but we don't need it anymore.
We can add it, no difference to me, but it's not necessary

kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/locale/TimezoneExtension.kt
LeoColman: I'm afraid that we'll clog the extensions root package with too many things and then we'll start moving things to their own package later.

The same thing happened in KT-assertions, and I'm trying to avoid this happening here too

kotlintest-core/build.gradle
LeoColman: I can add it back, no problem.
I just thought that maybe it's not a dependency to core anymore

kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/locale/TimezoneExtension.kt
sksamuel: Yes fair enough.

kotlintest-extensions/src/test/kotlin/io/kotlintest/extensions/junitxml/JUnitXmlTest.kt
sksamuel: No, the junit xml extension can in theory be deleted. Gradle and maven do the same thing.

kotlintest-core/build.gradle
sksamuel: Why was it, I can't remember.

LeoColman: Due to SysProps using something from core, if I remember correctly

kotlintest-extensions/src/test/kotlin/io/kotlintest/extensions/junitxml/JUnitXmlTest.kt
LeoColman: What was the original reason for it to be created?
Is there any chance that anybody is using it? I'm for deleting it

sksamuel: It was created because I didn't realize at the time that gradle and maven did the same thing for us :)

kotlintest-core/build.gradle
sksamuel: Doesn't look like it. I may have done it just because I'm happy to have core depend on it. No reason :)

LeoColman: Do you think it's better to have it? I'm happy to add it too!

kotlintest-extensions/src/test/kotlin/io/kotlintest/extensions/junitxml/JUnitXmlTest.kt
LeoColman: Ok, I removed it.

Issue #623, opened 2019-01-29
System Environment extensions


LeoColman:
In the line of #524, we should allow configuration for System Environment Variables (which are differente from System Properties).

It's not trivial to do, because `System.getenv()` returns an immutable map, which can't be appended to.

The library [System Rules](https://stefanbirkner.github.io/system-rules/) does it with a JUnit rule, specifically in [this class](https://github.com/stefanbirkner/system-rules/blob/master/src/main/java/org/junit/contrib/java/lang/system/EnvironmentVariables.java).

I believe we could do something similar, and in the same pattern as `System Properties` (#524), with a Extension and helper methods.

sksamuel:
And a security manager extension too.
We should make a module with extensions, or shove them in core.

LeoColman:
I think shoving them in core is better.
Unless we make the Junit5 runner module depend on a new module, which I'm fine with too..

I've created #624 and #625 for f urther discussion on these

sksamuel:
Yeah I wouldn't make it a module the user has to add in manually, but maybe for separation of code we can have a kotlintest-extensions module.

LeoColman:
I agree

sksamuel:
We have a lot of features for 3.3 now.
Did you test intellij plugin?

LeoColman:
@sksamuel 
One thing that is scaring me a little bit about these extensions:
What if I develop for parallel testing?

For instance, if I need environment to be X in this test, Y in another test and they're both ran in parallel?
If this happen, the variable will be in an inconsistent state. The same goes for System Property extension.

I think user should know this, but should we warn in any way about this?

sksamuel:
We could possibly have the extension throw an exception if parallelism is enabled, because yes its a race condition. They would be able to use project level only.

LeoColman:
What do you mean using project level only?
Only `beforeProject` and `afterProject`?

Perhaps I can use it `beforeTest`, but I have to guarantee that no other test is using it. How to verify if parallelism is enabled?

sksamuel:
The extension can check Project.parallelism

LeoColman:
And if project is parallel we deny it's usage in tests, and allow usage in project only?

sksamuel:
Exactly.

LeoColman:
[...]

PR #633, opened 2019-02-01
Implement System Environment Extension

LeoColman:
This commit implements the System Environment Extensions. With some extension functions and a listener, this commit enables System Environment Override by users, something only possible via reflection.

An special attention is needed to the system environments, as they're susceptible to race conditions. This was specified in the documentations.

This commit also fixes a small issue with SystemProperties extension, in which they wouldn't accept a suspended function. A test for that behavior was added.

This commit also renames the tests package, to stop stacktrace filtering, so that the error is clear whenever one of these tests fail.

Implements #623

### DIFF
--- a/kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/SystemEnvironmentExtensions.kt
+++ b/kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/SystemEnvironmentExtensions.kt
@@ -1,0 +1,265 @@
+package io.kotlintest.extensions.system
+
+import io.kotlintest.TestCase
+import io.kotlintest.TestResult
+import io.kotlintest.extensions.TestListener
+import java.lang.reflect.Field
+
+/**
+ * Overrides System Environment with chosen key and value
+ *
+ * This is a helper function for code that uses Environment Variables. It overrides the specific [key] from [System.getenv]
+ * with the specified [value], only during the execution of [block].
+ *
+ * To do this, this function uses a trick that makes the System Environment editable, and overrides [key]. Any previous
+ * environment (anything not overridden) will also be in the environment. If the chosen key is in the environment,
+ * it will be overridden. If the chosen key is not in the environment, it will be included.
+ *
+ * After the execution of [block], the environment is set to what it was before.
+ *
+ * **ATTENTION**: This code is susceptible to race conditions. If you attempt to change the environment while it was
+ * already changed, the result is inconsistent, as the System Environment Map is a single map.
+ */
+inline fun <T> withEnvironment(key: String, value: String?, block: () -> T): T {
+  return withEnvironment(key to value, block)
+}
+
+/**
+ * Overrides System Environment with chosen key and value
+ *
+ * This is a helper function for code that uses Environment Variables. It overrides the specific key from [System.getenv]
+ * with the specified value, only during the execution of [block].
+ *
+ * To do this, this function uses a trick that makes the System Environment editable, and overrides the key. Any previous
+ * environment (anything not overridden) will also be in the environment. If the chosen key is in the environment,
+ * it will be overridden. If the chosen key is not in the environment, it will be included.
+ *
+ * After the execution of [block], the environment is set to what it was before.
+ *
+ * **ATTENTION**: This code is susceptible to race conditions. If you attempt to change the environment while it was
+ * already changed, the result is inconsistent, as the System Environment Map is a single map.
+ */
+inline fun <T> withEnvironment(environment: Pair<String, String?>, block: () -> T): T {
+  return withEnvironment(mapOf(environment), block)
+}
+
+/**
+ * Overrides System Environment with chosen keys and values
+ *
+ * This is a helper function for code that uses Environment Variables. It overrides the specific keys from [System.getenv]
+ * with the specified values, only during the execution of [block].
+ *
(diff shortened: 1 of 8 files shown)


LeoColman:
@sksamuel 
Initial implementation. Tell me what you think about it so far

sksamuel:
We can't use suspend in the extensions.
If you do this:
```kotlin
suspend fun <T> withSystemProperties(props: List<Pair<String, String?>>, thunk: suspend () -> T): T {
    return withSystemProperties(props.toMap(), thunk)
}
```
Then you can't do this use this in another lambda.

LeoColman:
If we don't let them suspend, we won't be able to use them in tests. I caught this scenario when trying to write `ShouldSpec`. Could you give me an example of what what would be a bad case to use it this way?

In ShouldSpec I tried something like:

```
"Foo" {
    withEnvironment("Boo", "Bar") {
        should("FAR") {

        }
    }
}
```

sksamuel:
I use them inside the tests, not outside.

sksamuel:
Try doing, `collections.forEach { withSysProp { ... } }`

LeoColman:
What if we make them inline?
Similar to how  it is for `shouldThrowAny` and similars?

sksamuel:
If you have a suspend function, and you pass it into a non suspend lambda, the compiler can't reason about when it will be executed. Even if the lambda is _currently_ inside a coroutine scope, there's no saying when that lambda will be invoked.

LeoColman:
I think that shouldThrow works nice in this sense

sksamuel:
I don't understand ?

LeoColman:
Look at the implementation of `shouldThrow`. It uses inline and enables
suspended functions, because they are inlined inside another suspend
function.



Maybe we can do the same with extensions? I'll test it

Em sex, 1 de fev de 2019 22:19, Stephen Samuel <notifications@github.com
escreveu:

sksamuel:
Oh I see. Clever.

LeoColman:
@sksamuel I think this is the version I'm comfortable with.

LeoColman:
I'm trying to verify why app veyor is failing. I think it's resetting my environment

LeoColman:
@sksamuel 
One thing I"m in doubt:
Do we need a version for each scope? I was thinking that maybe TestLevel would suffice, as it can also be used in ProjectLevel.

Ok, the function will be called more times than it should, but this will avoid confusion, no?

LeoColman:
(This is the reason I didn't add the others)

kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/SystemEnvironmentExtensions.kt
```diff
@@ -0,0 +1,191 @@
...
+ * To do this, this listener uses a trick that makes the System Environment editable, and overrides the keys. Any previous
+ * environment (anything not overridden) will also be in the environment. If the chosen key is in the environment,
+ * it will be overridden. If the chosen key is not in the environment, it will be included.
+ *
+ * After the execution of the test, the environment is set to what it was before.
+ *
+ * **ATTENTION**: This code is susceptible to race conditions. If you attempt to change the environment while it was
+ * already changed, the result is inconsistent, as the System Environment Map is a single map.
+ */
+class SystemEnvironmentListener(private val environment: Map<String, String>) : TestListener {
```
sksamuel: This is all great except I would make two versions.

SystemEnvironmentTestListener

SystemEnvironmentProjectListener.

The second one only sets the env at the start of the project, and resets at the end. This means you can use parallelism with it.
The first one is when you only want to override for one or two tests or something.

sksamuel (changes requested):

kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/SystemEnvironmentExtensions.kt
LeoColman: I didn't want to provide two versions as to not confuse users. But this makes sense. I"ll change it

LeoColman: How to test for project configuration tho?

sksamuel: You don't need to test for anything, just need to have two versions which use different callback methods. The user can decide which to use.

LeoColman:
@sksamuel SystemEnvironmentProjectListener added. However, I have no idea how to test it. It would dirty any other tests, and make them too coupled

sksamuel:
In the past I've done this by making a new sub module. Like I did for parallelisation.

LeoColman:
I don't know if there's a problem to dirty context with a very specific variable tho... For instance, just a `foo`  in the system environment.

Do you think it's too much of an issue?

LeoColman:
I added the test while dirtying the context.
Tell me if you're too much against it, and I'll create a module if necessary.

LeoColman:
Oh, I just saw that you said this test is not completely necessary. I'll remove it, as I believe that this kind of behavior was already tested in the other functions, and project listener execution is also already tested.

LeoColman:
@sksamuel 
Are you ok with these changes now?

sksamuel:
Yes looks good.

LeoColman:
Great. Merging this and doing the same to `SecurityManager` extension

Issue #634, opened 2019-02-04
new behavior spec without parentheses


leoxs22:
Scalatest has a syntax like this:

 ```
 "Chuck Norris" when {
     "hit by a truck" should {
          "be ok" in {
                //test code
           }

          "destroy the truck" in {
               //test code
          }
      }
}
```

I'll try to code it if someone can confirm that this could be a good adittion. I think that this syntax is cleaner than the BehaviorSpec, providing the same functionallity.

LeoColman:
I'd be glad to take a look in your code!
However, some things we should make sure of: BehaviorSpec will keep existing. How can we find a name for this kind of spec?

sksamuel:
@leoxs22 what is this spec called in scalatest ?

leoxs22:
@sksamuel in Scala it seems to be called WordSpec 
http://www.scalatest.org/at_a_glance/WordSpec

Maybe we could add it as an extension of the WordSpec, since it's similar but with more nested operations. If that nesting it's just optional, it could be a part of the same spec.

So, you can just use the "should" keyword, but if the test requires it, you can use the keywords "when" and "in".

I have to check if it's possible to do that while maintaining backward compatibility. If not, we can call it ExtendendWordSpec, or something like that.

So, "should" would be mandatory, but "when" and "in" would be optional.

leoxs22:
I've run some tests and it seems to work as expected when using the 3 functions.

![image](https://user-images.githubusercontent.com/9098977/52249372-879d0000-28d1-11e9-99b8-4abdd6b02a43.png)

![image](https://user-images.githubusercontent.com/9098977/52249377-8f5ca480-28d1-11e9-842e-48acc82000a5.png)

I still have to work in allowing only to use "should", and keeping the functionality of the "config" function in all the contexts, but at least I now know that I can code it :smile:

sksamuel:
So we already have a WordSpec.
It doesn't have the "when" syntax - I've been using ScalaTest since 2011 and I never realised word spec supported that :joy: 
So we could just add that to the existing word spec.

As for "in" - when I first wrote KotlinTest I made the conscious decision to avoid the use of In. I don't think it should be added now.

leoxs22:
Nice, so I'll delete the support for "In", and write some tests

sksamuel:
yes please, and if we don't have lowercase when, can we add that in backticks?

PR #635, opened 2019-02-05
WIP: Feature/634 Extended word spec

leoxs22:
Implementing https://github.com/kotlintest/kotlintest/issues/634

There are still some things to do and test, but I open the pull request to get some feedback before advancing more

In the WordSpecTest it can be seen how it would be used.

Intellij shows it like this:

image

I'm not sure if there is any way to avoid repeating so much code in the config function if we want to keep allowing to use "In" or nothing. I'm just learning kotlin, so I don't know if there is some magic feature to do that :laughing:

### DIFF
--- a/doc/styles.md
+++ b/doc/styles.md
@@ -72,6 +72,26 @@ class MyTests : WordSpec({
 })
 ```
 
+It also supports the keyword `When` allowing to add another level of nesting.
+
+```
+class MyTests : WordSpec({
+    "Hello" When {
+        "asked for length" should {
+            "return 5" {
+                "Hello".length shouldBe 5
+            }
+        }
+        "appended to Bob" should {
+            "return Hello Bob" {
+                "Hello " + "Bob" shouldBe "Hello Bob"
+            }
+        }
+    }
+    
+})
+```
+
 ### Feature Spec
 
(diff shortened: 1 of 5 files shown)


sksamuel:
Thank you for the PR. I've added comments in #634

leoxs22:
I'm not sure about how it's being shown in the IDE.

Using "should" it's shown as always:
```
A should
-- B
-- C
```

Using "when" it's shown as:

```
A
-- When B
-----Should C
-----Should D
--When E
-----Should F
```

I honestly think that the second option it's easier to read, but it should also be consistent inside the same spec style.

So, the options are:
-Keeping it with the difference
-Changing it to be
```
A when
-- B should
-----C
-----D
-- E should
-----F
```
-Changing the old style to

```
A 
-- Should B
-- Should C
```

What do you think?

sksamuel:
Let's not change what's there already as people may rely on it by parsing it or whatever.
I'd do

```
A when
  - B should
      C
      D
```

As that follows the way its written in code.

leoxs22:
I think it's ready for review. It ended up being a lot easier without the "In" and the other syntax.

Tell me if there is anything more that I should do

sksamuel:
lgtm

LeoColman (approved):

sksamuel:
Feel free to merge @Kerooker

LeoColman:
Alrighty!

sksamuel:
We should probably have a PR to update the styles page too, ready for when
we release 3.3

PR #638, opened 2019-02-06
CollectionMatchers: Lazily evaluate joined string representations

sschuberth:
For large data classes calling toString() might actually be quite
expensive, esp. if there are multiple of those in a collection. So even
if limited to 10 elements here, only lazily evaluate the failure message
strings as they are not used at all on success.

### DIFF
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/CollectionMatchers.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/CollectionMatchers.kt
@@ -32,11 +32,12 @@ fun <T> containsAll(ts: List<T>): Matcher<Collection<T>> = containAll(ts)
 
 fun <T> containAll(vararg ts: T) = containAll(ts.asList())
 fun <T> containAll(ts: Collection<T>): Matcher<Collection<T>> = object : Matcher<Collection<T>> {
-  override fun test(value: Collection<T>) = Result(
-      ts.all { value.contains(it) },
-      "Collection should contain all of ${ts.joinToString(", ", limit = 10) { stringRepr(it) }}",
-      "Collection should not contain all of ${ts.joinToString(", ", limit = 10) { stringRepr(it) }}"
-  )
+  override fun test(value: Collection<T>): Result {
+    val joinedString by lazy { ts.joinToString(", ", limit = 10) { stringRepr(it) } }
+    val failureMessage by lazy { "Collection should contain all of $joinedString" }
+    val negatedFailureMessage by lazy { "Collection should not contain all of $joinedString" }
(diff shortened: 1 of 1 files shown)


LeoColman:
Amazing contribution!
Thanks, @sschuberth

LeoColman:
Do you think there are other places where this would be usefull?

LeoColman:
In the collection matchers, I mean

LeoColman (approved):

sschuberth:
> Do you think there are other places where this would be usefull?
> In the collection matchers, I mean

For collection matchers, maybe doing the same for `sorted()` would still make sense. I'll amend the commit.

But ideally, the `Result` class should take closures instead of strings, and the closures should only be evaluated on error (once). But as that's a larger refactoring I didn't want to do that.

LeoColman:
I think you're correct.
Both that it could take a function instead of a value and that it's too much refactoring now :P

sksamuel:
I have to say I disagree. We're not writing Haskell, so we don't need to make everything lazy.
What happens if between the assertion and the error print out the collection changes. With lazy evaluation you bring in the possibility for mutations. This could very well happen if you were using assertSoftly and doing manipulations between assertions.

I understand the point the change - to avoid generating two sets of messages. So a better solution is to refactor the way we generate Result objects to only require a single message.

LeoColman:
I think that the main point of change is that sometimes an eager evaluation may take too long to run. However, @sschuberth Did you happen to be in a specific scenario in which toString was taking too much?

sschuberth:
@sksamuel, my point is actually more in the direction of what @Kerooker said. Let me give you some more background:

In the code base of [ORT](https://github.com/heremaps/oss-review-toolkit) we did override `toString()` in several data classes. That is a bad idea for several reasons, one being that [you cannot call](https://youtrack.jetbrains.com/issue/KT-12991) the original (compiler-generated) `toString()` implementation anymore, the other one being that it makes refactoring code much harder because it's very difficult to find all (implicit) calls to `toString()`, e.g. in interpolated strings. So what I did was to make our overridden `toString()` throw an exception to ensure it's not being called anymore, only to find out that kotlintest in many places (seemingly unnecessarily) calls `toString()` when creating error messages that the user might never ever see in the end. So that's where my original motivation comes from to not call `toString()` if not needed, and as a benefit, there's a change that not doing so improves performance (and reduces memory usage).

However, I have to admit that @sksamuel has a good point here saying that collections might have changed in between.

sksamuel:
So I agree with the motivation - don't generate messages if they're not required.
I'd be happy to think about a refactoring of the matchers/results to achieve this. I'll do the work if necessary, but between us we can outline a design.

sschuberth:
Sounds good. So for now I'll close this PR as it's not supposed to get merged as-is.

Issue #624, opened 2019-01-29
Security Manager extension


LeoColman:
In the line of #524 and #623, we should have a Security Manager extension.

The library [System Rules](https://stefanbirkner.github.io/system-rules/) does it with a JUnit rule, specifically in [this](https://github.com/stefanbirkner/system-rules/blob/master/src/main/java/org/junit/contrib/java/lang/system/ProvideSecurityManager.java) class.

We can do something similar, and in the same pattern of `System Properties` (#524) and `System Environment Variables` (#623), with an extension and helper methods.

sksamuel:
One thing I'd like to think about is project level vs spec level extensions.

Maybe you want to override security manager for a single spec. Or maybe for the entire test suite. Currently we don't offer that.

PR #640, opened 2019-02-08
Implement System Security Manager Extensions

LeoColman:
This commit implements the System Security Manager Extensions.

A special attention is needed to the System Security Manager, as it's only one for the JVM, it's susceptible to race conditions. This was specified in the documentation.

Implements #624

### DIFF
--- a/kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/SecurityManagerExtensions.kt
+++ b/kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/SecurityManagerExtensions.kt
@@ -1,0 +1,93 @@
+package io.kotlintest.extensions.system
+
+import io.kotlintest.TestCase
+import io.kotlintest.TestResult
+import io.kotlintest.extensions.TestListener
+
+/**
+ * Replaces System Security Manager with [securityManager]
+ *
+ * This function replaces the System Security Manager with the specified [securityManager], executes [block] and then
+ * returns the original manager to its place. This will also happen if the original manager is null.
+ *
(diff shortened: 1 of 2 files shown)


kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/SecurityManagerExtensions.kt
```diff
@@ -0,0 +1,54 @@
...
+ * This is a Listener for code that uses the System Security Manager. It replaces the System Security Manager with the
+ * desired [securityManager].
+ *
+ * After the execution of the test, the System Security Manager is set to what it was before. This will also happen if
+ * the original manager is null.
+ *
+ * **Attention**: This code is subject to race conditions. As Java's System Manager is only one per JVM, this code
+ * will replace it. However, all other tests are going to use it if tests are running in parallel.
+ */
+class SecurityManagerListener(
```
sksamuel: I would do the same as for env vars and have two listeners, one which changes the entire project (or per spec) and one that does it per test.

sksamuel (changes requested):

kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/SecurityManagerExtensions.kt
LeoColman: I agree. If the implementation is ok for Environment, I'll do the same thing here

LeoColman:
@sksamuel 
Implemented the changes. If you're ok with them, I'll get this merged

PR #653, opened 2019-02-13
Remove all deprecated matchers/assertions

### LINKED ISSUE #651: Remove old shouldBe methods
For 3.3 are we removing the deprecated shouldBe methods that have been deprecated since 3.0 ?
@Kerooker

LeoColman:
This commit removes all `shouldBe` kind of matchers/assertions that were deprecated. Some were already moved to another package, some are not the correct usage anymore.

They're deprecated quite some time, and are now being removed.

Closes #651


LeoColman:
I don't think there is any other matcher to remove

LeoColman:
But we can discuss if we want to remove `isInstancePerTest` and the deprecated TestListener functions

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/StringMatchers.kt
```diff
@@ -23,8 +23,6 @@ fun startWith(prefix: String) = neverNullMatcher<String> { value ->
 }
 
 fun haveSubstring(substr: String) = include(substr)
-@Deprecated("use should include(substring)", ReplaceWith("include(substr)", "io.kotlintest.matchers.string.include"))
```
sksamuel, line 26: How long has this one been deprecated for ?

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/collections/matchers.kt
```diff
@@ -62,11 +51,6 @@ fun <T> List<T>.shouldHaveElementAt(index: Int, element: T) = this should haveEl
 
 fun <T> List<T>.shouldNotHaveElementAt(index: Int, element: T) = this shouldNot haveElementAt(index, element)
 
-@Deprecated("Use shouldHaveElementAt", ReplaceWith("this.shouldHaveElementAt(index, element)", "io.kotlintest.should"))
-fun <T, L : List<T>> L.shouldContainElementAt(index: Int, element: T) = this should haveElementAt(index, element)
-
-@Deprecated("Use shouldNotHaveElementAt", ReplaceWith("this.shouldNotHaveElementAt(index, element)", "io.kotlintest.should"))
-fun <T, L : List<T>> L.shouldNotContainElementAt(index: Int, element: T) = this shouldNot haveElementAt(index, element)
```
sksamuel, line 69: Are these relatively recent deprecations ?

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/collections/matchers.kt
```diff
@@ -62,11 +51,6 @@ fun <T> List<T>.shouldHaveElementAt(index: Int, element: T) = this should haveEl
 
 fun <T> List<T>.shouldNotHaveElementAt(index: Int, element: T) = this shouldNot haveElementAt(index, element)
 
-@Deprecated("Use shouldHaveElementAt", ReplaceWith("this.shouldHaveElementAt(index, element)", "io.kotlintest.should"))
-fun <T, L : List<T>> L.shouldContainElementAt(index: Int, element: T) = this should haveElementAt(index, element)
```
sksamuel, line 66: Are these relatively recent deprecations ?

sksamuel (changes requested):

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/collections/matchers.kt
LeoColman: 09/05/2018

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/collections/matchers.kt
LeoColman: 05/2018

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/StringMatchers.kt
LeoColman: 04/2018

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/collections/matchers.kt
sksamuel: So all these deprecations went into 3.1.0 so I think we can remove
I just don't want to remove anything we only deprecated in 3.2.0 as we only released that a month ago.
And we released 3.1.0 on 11th May - so anything deprecated after that might not have been seen by a user until Jan this year.

sksamuel:
> But we can discuss if we want to remove `isInstancePerTest` and the deprecated TestListener functions

I think no, as they only got deprecated a month ago. 3.4 for those.

LeoColman:
No prob. Let's merge this?

PR #655, opened 2019-02-13
Add project config option for `assertSoftly` #512

### LINKED ISSUE #512: Add project config option for `assertSoftly`
It would be nice to be able to enable "soft asserts" for each test case once in project config, instead of spreading  `assertSoftly{}` around in every single test case.

sksamuel:
Resolves #512


kotlintest-assertions/src/main/kotlin/io/kotlintest/dsl.kt
```diff
@@ -162,7 +162,7 @@ fun diffLargeString(expected: String, actual: String, minSizeForDiff: Int = 50):
 
   val useDiff = expected.lines().size >= minSizeForDiff
       && actual.lines().size >= minSizeForDiff &&
-      System.getProperty("kotlintest.assertions.multilinediff") != "simple"
+      System.getProperty("kotlintest.assertions.multi-line-diff") != "simple"
```
LeoColman, line 165: Any reason for this to be brought together? Intentional or accident?

LeoColman (approved):

kotlintest-assertions/src/main/kotlin/io/kotlintest/dsl.kt
sksamuel: If you mean why did I change it on this commit, it is because I only added it last night (so not in any release yet), and when I added the flag for _this_ commit, I thought putting in the hyphens made it look better, so I did the same change here.

LeoColman:
Approved all.

PR #660, opened 2019-02-16
fix containOnlyOnce which return true when no occurrence

mrduguo:
* added test to ensure it fail with expected message


sksamuel:
Good spot :+1:

mrduguo:
Thanks for the quick actions and great works!

PR #684, opened 2019-02-18
Fix Spring mocking injection @MockBean @MockkBean

tonilopezmr:
### 💣 Issue

* Using the current version only supports @Autowired injections but not mocked injections because the implementation only takes care about Autowired objects.

### :tophat: What is the goal?

* Inject mocks automatically with [@MockBean](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/test/mock/mockito/MockBean.html) / [@MockkBean](https://github.com/Ninja-Squad/springmockk/blob/master/src/main/java/com/ninjasquad/springmockk/MockkBean.java#L77) (thanks to the [latest release](https://spring.io/guides/tutorials/spring-boot-kotlin/))

### :memo: How is it being implemented?

* Use the non-deprecated method [beforeSpec(spec: Spec)](https://github.com/kotlintest/kotlintest/blob/413910fecc5cf1f2379e4ea06f2622d61c3d35e9/kotlintest-core/src/main/kotlin/io/kotlintest/extensions/TestListener.kt#L57)
* Use [prepareTestInstance](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/context/TestContextManager.html#prepareTestInstance-java.lang.Object-) method which prepare the entire execution stack before each any individual test, inject dependencies, mocks, etc...

I hope we can improve Kotlintest integration with Spring ❤️, thanks!

### DIFF
--- a/kotlintest-extensions/kotlintest-extensions-spring/src/main/kotlin/io/kotlintest/spring/SpringListener.kt
+++ b/kotlintest-extensions/kotlintest-extensions-spring/src/main/kotlin/io/kotlintest/spring/SpringListener.kt
@@ -11,11 +11,9 @@ import kotlin.reflect.full.primaryConstructor
 
 object SpringListener : TestListener {
 
-  override fun beforeSpec(description: Description, spec: Spec) {
+  override fun beforeSpec(spec: Spec) {
     try {
-      val manager = TestContextManager(spec.javaClass)
-      val ac = manager.testContext.applicationContext
-      ac.autowireCapableBeanFactory.autowireBean(spec)
+      TestContextManager(spec.javaClass).prepareTestInstance(spec)
     } catch (t: Throwable) {
       t.printStackTrace()
     }


LeoColman (approved):

LeoColman:
Great one!

@sksamuel Let's sneak this into 3.3.0?

sksamuel:
This is most beautifully formatted pull request ever.

LeoColman:
And commit + commit message

sksamuel:
We should do the same for the `ConstructorExtension`

tonilopezmr:
[We use it at Karumi a lot](https://github.com/Karumi/project-quality-assurance)

mindhaq:
Is this already part of 3.3.0-RC3?

If so, it's not working for me. I still get this when the test makes the first mock-wiring call:

    kotlin.UninitializedPropertyAccessException: lateinit property service has not been initialized

sksamuel:
Yes it's included in 3.3.0-RC3

tonilopezmr:
@mindhaq Did you added the `SpringListener` ? 

If you could share the code, I can take a look. I'm using 3.3.0-RC3, and it is working.

mindhaq:
@tonilopezmr ah! my bad, I though adding the `SpringAutowireConstructorExtension` would be enough. Should have read the docs more thoroughly.

Thanks for the hint, now it works like a charme.

sksamuel:
The `SpringListener` does field injection and the `SpringAutowireConstructorExtension` does constructor based injection.

PR #686, opened 2019-02-18
Feature/pitest

sksamuel:


LeoColman (approved):
I don't really know how Pitest works, but the code seems fine to me.

sksamuel:
Ok I'll merge once green - the code doesn't touch anything outside it's own module, so the impact is minimal.

PR #693, opened 2019-02-22
Feature/689 constant now listeners

LeoColman:


doc/extensions.md
```diff
@@ -260,4 +260,36 @@ And with the listeners
     TimeZoneTestListener(TimeZone.getTimeZone(ZoneId.of("America/Sao_Paulo")))
   )
 
+```
+
+### Current instant listeners
+
+Sometimes you may want to user `java.time` classes' `now` static function for multiple reasons, such as setting the creation date of an entity
```
sksamuel: Sometimes you may want to use the `now` static functions located in `java.time` ...

doc/extensions.md
```diff
@@ -260,4 +260,36 @@ And with the listeners
...
+
+`data class MyEntity(creationDate: LocalDateTime = LocalDateTime.now())`.
+ 
+But what to do when you want to test that value? `now` will be different
+each time you call it!
+
+For that, KotlinTest provides `ConstantNowListener` and `withConstantNow` functions.
+
+While executing your code, your `now` will always be the value that you want to test against!!
```
sksamuel: No need for !! i don't think.

kotlintest-extensions/src/test/kotlin/com/sksamuel/kt/extensions/time/ConstantNowExtensionsTests.kt
```diff
@@ -0,0 +1,267 @@
...
+  override fun afterSpecClass(spec: Spec, results: Map<TestCase, TestResult>) {
+    HijrahDate.now() shouldNotBeSameInstanceAs myNow
+    LocalDateTime.now() shouldNotBeSameInstanceAs myNow2
+  }
+  
+  override fun listeners() = listOf(
+          ConstantNowTestListener(myNow),
+          ConstantNowTestListener(myNow2)
+  )
+}
```
sksamuel, line 267: Really good idea and clever implementation. We can put this in 3.3 if you like.

sksamuel (changes requested):

kotlintest-extensions/src/test/kotlin/com/sksamuel/kt/extensions/time/ConstantNowExtensionsTests.kt
LeoColman: Sure!

My only issues with the implementation is that technically speaking, `Temporal` doesn't force any implementation to contain a static `now`. All `Temporal`s from Java do, but maybe a random temporal appears, and it won't work.

I said that in the docs, but idk if that's a concern.

I didn't want to create 17 implementations of 17 types with 17 ultra-similar executions

doc/extensions.md
LeoColman: much better. Brainfarted there

LeoColman:
Alright! Merging this guy.

Issue #665, opened 2019-02-16
between matcher for dates


sksamuel:
`mydate.isBetween(date1, date2)`

LeoColman:
With this you mean Java's Date?

sksamuel:
Yes, LocalDate, Instant, etc

LeoColman:
Oh, ok. So not including the deprecated `java.util.Date`

sksamuel:
No

LeoColman:
We should have a matcher

`beBetween` for all the types, and a `shouldBeBetween`/`shouldNotBeBetween` for all types aswell.

LeoColman:
Should we add a tolerance value?

sksamuel:
don't need a tolerance on dates I don't think.

LeoColman:
If we don't add tolerance, the user will have some hassle when using things like `LocalDateTime.now()`, as this exact now may be a little bit different from the `now` in some time from here.

Are we ok with that?

sksamuel:
Well you can't really use now() for testing can you.
And if you have tolerance, how much tolerance? If it's a day your tests
will fail after a day. If it's a month they'll start failing after a month.
So it doesn't really work.

LeoColman:
Sometimes you're testing thing likes

"And the creation date should be now", because inside that testing code you use `now()` and such. But I guess it doesn't make sense for `between`

sksamuel:
Yes I see. Maybe then it would be useful I agree, so we could add something
like that.

`date.isApprox(now())`

LeoColman:
Closed as per #694

PR #694, opened 2019-02-24
Added LocalDate, LocalDateTime, ZonedDateTime and OffsetDateTime between matchers and the respective tests

luiz-neves:
Solves #665


LeoColman:
:+1: :heart:

PR #695, opened 2019-02-25
Allows instanceOf and typeOf matchers to use the casted value

### LINKED ISSUE #666: isInstanceOf matcher should allow lamba for further tests

LeoColman:
This commit also adds some documentation to all related methods.

Fix #666

### DIFF
--- a/kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/types/TypeMatchersTest.kt
+++ b/kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/types/TypeMatchersTest.kt
@@ -60,7 +60,7 @@ class TypeMatchersTest : WordSpec() {
       "Allow execution with a lambda" {
         val list = arrayListOf(1, 2, 3)
         
-        list.shouldBeInstanceOf<ArrayList<Int>> {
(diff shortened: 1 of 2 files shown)


LeoColman:
@dave08

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/types/TypeMatchersTest.kt
```diff
@@ -66,6 +56,14 @@ class TypeMatchersTest : WordSpec() {
           arrayList.shouldNotBeInstanceOf<ArrayList<*>>()
         }
       }
+      
+      "Allow execution with a lambda" {
+        val list = arrayListOf(1, 2, 3)
+        
+        list.shouldBeInstanceOf<ArrayList<Int>> { it: ArrayList<Int> ->
```
LeoColman, line 63: type was made explicit to enforce that this type won't change in the future, even though it's redundant for users

sksamuel:
This allows us to do `t.shouldBeInstanceOf<T> { t -> ... }` ?

LeoColman:
Yes. An example of this is on the test. I made the type explicit to ensure it won't change in the future

sksamuel (approved):
Good work as always.

PR #700, opened 2019-02-27
Allows focusing tests with prefix

### LINKED ISSUE #699: Feature spec not enabling focus
When trying to use FeatureSpec, it's impossible to focus a test.

This is likely due to the way we setup the test code:

```addTestCase(createTestName("Feature: ", name)```

we don't allow for F: to be placed before Feature

### PR DESCRIPTION
**Pull request by LeoColman**
Fixes #699

### DIFF
--- a/kotlintest-core/src/main/kotlin/io/kotlintest/specs/TestPrefixes.kt
+++ b/kotlintest-core/src/main/kotlin/io/kotlintest/specs/TestPrefixes.kt
@@ -1,5 +1,9 @@
 package io.kotlintest.specs
 
-internal fun createTestName(prefix: String, name: String): String {
-  return if(name.startsWith("!")) "!$prefix${name.drop(1)}" else "$prefix$name"
+fun createTestName(prefix: String, name: String): String {
+  return when {
+    name.startsWith("!") -> "!$prefix${name.drop(1)}"
+    name.startsWith("f:") -> "f:$prefix${name.drop(2)}"
+    else -> "$prefix$name"
+  }
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by LeoColman** [comment]
All specs with prefix already delegate to `createTestNAme` as per https://github.com/Kotlintest/Kotlintest/commit/31d524513970947442385e849e8fca304dea1681

This brings focus into it.

File: kotlintest-core/src/main/kotlin/io/kotlintest/specs/TestPrefixes.kt [hunk 1]
```diff
@@ -1,5 +1,9 @@
 package io.kotlintest.specs
 
-internal fun createTestName(prefix: String, name: String): String {
-  return if(name.startsWith("!")) "!$prefix${name.drop(1)}" else "$prefix$name"
+fun createTestName(prefix: String, name: String): String {
+  return when {
+    name.startsWith("!") -> "!$prefix${name.drop(1)}"
```
sksamuel, line 5: Should there be a space here between the prefix ?
Would be good to have a test that works against FeatureSpec or something that actually uses this.

sksamuel (changes requested):

kotlintest-core/src/main/kotlin/io/kotlintest/specs/TestPrefixes.kt
LeoColman: There's no need for a space.

```
fun isFocused() = name.startsWith("f:")
fun isBang(): Boolean = name.startsWith("!")
```

sksamuel: What I mean is, will it end up being "Feature:TestHere" rather than "Feature: Test here"

LeoColman: Yes, but all the prefixes are like `"Feature: "`, and not `"Feature:"`. This was the original behavior before I chagned the code, so I kept it that way

sksamuel: Great I was just checking.

Might still be worth a test though ;)

sksamuel:
Merge when you want.

Issue #713, opened 2019-03-06
WordSpec doesn't fully support coroutines


thibseisel:
Some test scopes in `WordSpec` does not fully support calling suspending functions:

```kotlin
class WordSpecExampleTest : WordSpec() {

  init {
    "given a potato" `when` {
      delay(100) // OK
      "planted it" should {
        delay(100) // OK
        "grow" {
          delay(100) // Error: Suspension functions can be called only within coroutine body
        }
        "not be eaten".config(enabled = false) {
          delay(100) // Error: Suspension functions can be called only within coroutine body
        }
      }
    }
  }
}
```

One workaround is to wrap test code in a `launch { ... }` block (that is properly linked to the coroutine scope of `AbstractWordSpec.FinalTestContext`) but when the test fails, the assertion message is hidden behind a `JobCancelationException` which is not very readable.

LeoColman:
Nice catch. Do you want to fix this, or leave it to us?

thibseisel:
I'd like to, but I am unable to push my topic branch due to an HTTP 403 (Permission denied)

LeoColman:
You need to push it to your fork, and submit a pull request to this repository from it

sksamuel:
Fixed. Will be included in 3.3.2

PR #714, opened 2019-03-06
Makes test blocks suspend in WordSpec

thibseisel:
Fixes issue #713 .


LeoColman:
It would also be interesting to add a test to guarantee that this behavior will be preserved in the future. We must ensure that the suspended behavior is indeed there

sksamuel:
I will add the test.

sksamuel:
:+1:

PR #719, opened 2019-03-09
`Gen` kind and type class support + Extensions + matchers for many ΛRROW datatypes

raulraja:
The following PR introduces

- [x] `Gen` as a higher kinded type that supports ad-hoc polymorphism.
- [x] `@extension` for all the type classes that `Gen` is able to support in the `Functor` hierarchy.
- [x] `Matcher` instances and Kotlin Test specific utils for the most common Λrrow data types.
  - [x] Nel
  - [x] Either
  - [x] Option
  - [x] Try
  - [x] Validated


raulraja:
@sksamuel if this is accepted you may want to squash and merge as the commit history is not that relevant.

raulraja:
looks like some code ended in a relocated module, not ready to merge yet

kotlintest-assertions/src/main/kotlin/io/kotlintest/properties/Gen.kt
```diff
@@ -46,7 +47,8 @@ class BigIntegerGen(maxNumBits: Int) : Gen<BigInteger> {
  * The [Int] generator example should return a random int
  * from across the entire integer range.
  */
-interface Gen<T> {
+@higherkind
```
pakoito, line 50: We can HK interfaces now?

pakoito: In arrow we construct Monoid manually: https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Monoid.kt#L3-L5

raulraja: @pakoito AFAIK we were able to before but not sure about that particular case for Monoid

raulraja: You manually added the encoding there probably because you declare extensions in the same module as `@higherkind`, that was the actual limitation, we have not figured out annotation order with kapt.

raulraja:
@sksamuel This is now ready for review, thanks for your consideration and let us know if we can clarify anything. 🍻 !

sksamuel:
The 200 commits might make it harder to review lol.

sksamuel:
By the way, if you want a beer, what about early April. You said a while ago you're based in London. My colleagues would no doubt come out @stheppi
Last chance before I move abroad :)

pakoito:
I have one full Arrow explainer I promised on Twitter, and we are planning to do it on a pub on April 2nd if you want to join in. If the date isn't good I can do another day the week after!

sksamuel:
Ok so April 2nd is the date the removers come to my house to collect everything!
April 3rd I run a charity quiznight.
I'm on a cruise 5th April to the 11th. 
Other than that I'm free :)
I'm doing drinks near old street on 15th April with the Lenses team (my employeers) but can be another date.

pakoito:
Scheduling plans in London, am I right? hahahah Is April 12th good? Else we can do that weekend too.

kotlintest-assertions-arrow/build.gradle
```diff
@@ -2,4 +2,8 @@ dependencies {
     compile project(":kotlintest-assertions")
     compile 'io.arrow-kt:arrow-typeclasses:0.8.1'
     compile 'io.arrow-kt:arrow-data:0.8.1'
-}
\ No newline at end of file
+    testCompile 'io.arrow-kt:arrow-test:0.8.1'
+    kapt    'io.arrow-kt:arrow-annotations-processor:0.8.1'
```
sksamuel: Will this leak into downstream projects, or is it only used when building KotlinTest itself?

kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/gen/gen.kt
```diff
@@ -1,15 +1,76 @@
...
+import arrow.Kind
+import arrow.core.Either
+import arrow.extension
+import arrow.typeclasses.*
+import io.kotlintest.properties.ForGen
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.GenOf
+import io.kotlintest.properties.fix
+
+@extension
```
sksamuel: What does @extension do

kotlintest-assertions/src/main/kotlin/io/kotlintest/properties/Gen.kt
```diff
@@ -53,7 +54,7 @@ class BigIntegerGen(maxNumBits: Int) : Gen<BigInteger> {
  * The [Int] generator example should return a random int
  * from across the entire integer range.
  */
-interface Gen<T> {
+@higherkind interface Gen<T> : GenOf<T> {
```
sksamuel: Where is GenOf defined?

kotlintest-assertions-arrow/build.gradle
pakoito: Only when building itself, once the artifact is published it's not needed anymore.

kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/gen/gen.kt
pakoito: It generates top level extension functions for all the helper functions defined in the typeclass for the specified datatype.

You give one function, like `map` for `Functor`, and you get the others for free: https://arrow-kt.io/docs/apidocs/arrow-typeclasses/arrow.typeclasses/-functor/index.html#functions

So now you can do `myGen.tupleLeft(1)` to generate a `Tuple2` where the left side is a constant `1` for every value generated.

kotlintest-assertions/src/main/kotlin/io/kotlintest/properties/Gen.kt
pakoito: It's created by processing `@higherkind`:

```
@higherkind data class ListK<A>(val list: List<A>): ListKOf<A>

// Generates the following code:
//
// class ForListK private constructor() { companion object {} }
// typealias ListKOf<A> = Kind<ForListK, A>
// fun ListKOf<A>.fix() = this as ListK<A>
```

sksamuel: So by adding @higherkind to `Foo`, `FooOf<A>` is generated, and is a type alias to Kind<ForFoo, A>, where `ForFoo` is another automatically generated class. Why does the type alias require ForFoo? Is this to extract the type parameter basically?

kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/gen/gen.kt
sksamuel: Because the other functions can be implemented using map, you only need to provide map and the rest can be derived? Why do you need @extension and not just a regular abstract method with other definitions in the interface/typeclass ?

pakoito: We did that for a while. For those functions to be available you needed to do:

```
Option.functor().run { myGen.tupleLeft(1) }
```

So at some point we put all typeclasses together and made it a bit nicer:

```
ForOption {
  myGen.tupleLeft(1)
}
```

until we finally settled with creating top-level extension functions.

kotlintest-assertions/src/main/kotlin/io/kotlintest/properties/Gen.kt
pakoito: This is the full explanation:
https://arrow-kt.io/docs/patterns/glossary/#higher-kinds

The type alias is a nicety to make it easier to read and write extension functions for one datatype. 

```
fun <A, B, C> ListKOf<A>.doThing(another: ListKOf<B>): ListK<C>
```
Parameters are contravariant so you always take a Kind, and returns are covariant so you can return the specific class. This makes the extfuns work pretty much for everywhere, and their returns don't require `fix()` anymore. It hides some of the machinery.

```
val a: Kind<ForOption, ListK<User>> = userOptions.sequence(Option.applicative())
val b: Option<Tuple2<Int, ListK<User>> = a.tupleLeft(2)
```

pakoito: The key here is that Kotlin allows extfuns to be specialized for a generic, so `Kind<ForOption, A>` can have a different implementation that `Kind<ForListK, A>` even if they're defined on the same file.

sksamuel: It's coming together now in my tiny brain.
You're doing all this machinery basically because you can't do this:

```kotlin
interface Functor<F<A>> {
  fun <B> map(f: (A) -> B): F<B>
}
```

sksamuel:
I'm happy for this to be merged now but let's wait until we release 3.3.2 and then master will become the 3.4.0 branch. This PR shouldn't be included in 3.3.x as it bumps arrow to 0.9.x

kotlintest-assertions/src/main/kotlin/io/kotlintest/properties/Gen.kt
pakoito:
```
interface Functor<F<_>> {
  fun <A, B> F<A>.map(f: (A) -> B): F<B>
}
```

sksamuel: Ok yes, because you don't care about A when you're implementing ListFunctor or whatever.

PR #723, opened 2019-03-19
Fix kotlintest gradle plugin version

HormCodes:
I wanted to use KotlinTest Gradle example but in build.gradle was bad kotlintest plugin version... So I fixed it...

https://plugins.gradle.org/plugin/io.kotlintest


Kerooker:
Oh, th at's a ridiculously old one hahaha

sksamuel:
Thanks, this plugin is still is still a bit beta.

PR #730, opened 2019-03-26
Move Float matchers to common

LeoColman:


LeoColman:
This also fixes a likely unwanted behavior

PR #732, opened 2019-03-28
Move String matchers to common

LeoColman:
This commit moves String and associated code to the common modules. An important point to notice is that this commit will introduce some breaking changes, as some files were moved together to the common modules.

It was assumed that this is OK, as perhaps KotlinTest these changes are for KotlinTest 4.0, in which some breaking changes are expected.


sksamuel:
Are the breaking changes limited to changing the package names ?

LeoColman:
Yes

LeoColman:
No change to functionality was made (yet, at least)

PR #736, opened 2019-03-28
Rename matchers packages to singular case

LeoColman:


sksamuel:
I think this should be reverted.
I don't think we should change things just because its 4.0 and we can. There's going to be tons of breaking changes anyway. This adds no functionality.

PR #740, opened 2019-04-01
"start with" example does not compile with kotlintest 3.3.0

DieBauer:
Example does not compile with kotlintest 3.3.0.


LeoColman:
Thanks for the contribution!

PR #749, opened 2019-04-10
fix typo for collection shouldBeLargerThan

mrduguo:
Source code is spell correct, only documentation is wrong.

https://github.com/kotlintest/kotlintest/blob/master/kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/matchers/collections/matchers.kt#L225


LeoColman:
Thanks for the contribution!

We brainfarted there. Good catch!

PR #754, opened 2019-04-19
add the ability for eventually to defaultly handle AssertionErrors

### LINKED ISSUE #753: eventually should handle AssertionError by default
Is there a good reason that kotlintest `eventually` does not handle AssertionError by default. I was reading another issue #145 and it seemed like it was talked about, but there was no clear reason to why this was not adopted...

To re-state what was said there, it would be nice to write:
```kotlin
    "find the 1000 number" {
        val iterator = (1..1000).iterator()
        eventually(2.seconds) {
            iterator.next() shouldBe 1000
        }
    }
```

instead of:
```kotlin
    "find the 1000 number" {
        val iterator = (1..1000).iterator()
        eventually(2.seconds, AssertionError::class.java) {
            iterator.next() shouldBe 1000
        }
    }
```

To me I can't think of too many use cases where I would not want to do an assertion in that block, thoughts?

codyoss:
A strong use case for using `eventually` is so that something you are
asserting eventually becomes true. Before this change you would have to
explicitly pass `AssertionError` as an exception to handle. Now it will
do this automatically. This enables you to pass a different exception if
you should so choose.

Fixes #753

### DIFF
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/Eventually.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/Eventually.kt
@@ -13,7 +13,7 @@ fun <T, E : Throwable> eventually(duration: Duration, exceptionClass: Class<E>, 
     try {
       return f()
     } catch (e: Throwable) {
-      if (!exceptionClass.isAssignableFrom(e.javaClass)) {
+      if (!exceptionClass.isAssignableFrom(e.javaClass) && !AssertionError::class.java.isAssignableFrom(e.javaClass)) {
         // Not the kind of exception we were prepared to tolerate
         throw e
       }
--- a/kotlintest-core/src/main/kotlin/io/kotlintest/AbstractProjectConfig.kt
+++ b/kotlintest-core/src/main/kotlin/io/kotlintest/AbstractProjectConfig.kt
@@ -59,10 +59,11 @@ abstract class AbstractProjectConfig {
   /**
    * When set to true, failed specs are written to a file called spec_failures.
    * This file is used on subsequent test runs to run the failed specs first.
-   * To disable this feature, set this to false, or set the system property
-   * 'kotlintest.write.specfailures=false'
+   *
+   * To enable this feature, set this to true, or set the system property
+   * 'kotlintest.write.specfailures=true'
    */
-  open fun writeSpecFailureFile(): Boolean = true
+  open fun writeSpecFailureFile(): Boolean = false
 
   /**
    * Sets the order of top level tests in a spec.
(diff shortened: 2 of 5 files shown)


codyoss:
Two things I thought I would point out:

1. I don't think `./gradlew build` actually runs these tests I made edits to. I made some changes locally that I did not commit to make sure they did work. I am not sure if this is on purpose or not, but I just thought I would point it out.

2. I would consider deprecating `eventually(duration: Duration, predicate: (T) -> Boolean, f: () -> T): T` . To me there is not a super clear idea why I would do this over just a "normal eventually" with some type of assertion in it. Also, this method is not handling any exceptions, which makes its behavior different than the other two overloaded functions that do. If it is desirable I could include this in the PR, lmk.

LeoColman:
Ok, so the tests aren't running because we are currently changing KT to MultiPlatform, and we're still adjusting test execution on the master branch.

LeoColman:
What you can do right now is instead of basing off `master`, base it off `release/3.3` if you want to test things quickly, and we can merge into master later on

codyoss:
Just did that locally and it is all good :+1:  Or are you saying I should close this PR and open it against `release/3.3`?

LeoColman:
We can point it to 3.3 here, so that we take a look at that integration right away, and merge it as soon as possible. What do you think?

codyoss:
Hmmm does not seem clean...

LeoColman:
I think you need to base your branch off of the release branch, and not master. This is making Github try to merge everything from master into 3.3

codyoss:
:fireworks:

codyoss:
Travis seems to be failing, but not due to tests. Looks related to: https://travis-ci.community/t/install-of-openjdk11-is-failing/1914/6

LeoColman:
Travis issue solved itself. I set it to build again.

LeoColman:
Can we add a test with a custom exception, just to ensure it's still working? We were treating `AssertionError` as the custom exception, but I'd like to have a test which will accept another exception to be passed.

codyoss:
I could add one, but I think https://github.com/kotlintest/kotlintest/pull/754/files#diff-11b9d0bed259e1191ce787f4f94cb147R55 covers that case, right?

Correction, nevermind I will add one in a few.

LeoColman:
You're correct.

I think this is good to go!

codyoss:
This got a little ugly since I started from basing off master and now pushing to the release branch, but I think the requested changes are done.

LeoColman:
It's all good. I'm just thinking of how I'm gonna merge this, as we're moving assertions to MPP. This is all fine and you don't need to do anything else!

LeoColman:
Ok, so here's the thing: We changed the master branch to normal. Due to that, I changed your PR to be to master instead of to release. I'll just get CI to build again, and we'll see if this can successfully be merged.

LeoColman:
Merged manually.

PR #757, opened 2019-05-05
added support to change sizes of generated lists, sets, maps

lenalebt:
I missed the possibility to restrict the size of a list when using property-based tests, so I added that feature, and did so for sets and maps as well. I also added a map generator that generated from a `Gen<Pair>K,V>>` to allow to generate maps with correlated keys and values.

I am unsure how much you care about binary compatibility (since in the end kotlintest is most probably not used as a library?) and was unsure if default parameters are binary-compatible, so I added `@JvmOverloads` to make sure it's binary compatible. Can remove that if you want to.

Hope it helps.


LeoColman:
I don't think we mind that much about binary compatibility, but let's try our best to not break any user. As I see, there is no reason for this to break users, right? As it has a default parameter.

I don't think we need JVM overloads here, but I don't think it hurts either.

LeoColman:
We have some tests breaking.
Execute tests locally with `./gradlew test`, but the last output I saw from travis was

> com.sksamuel.kotlintest.properties.GenTest > Generate sets up to the given length FAILED
    java.lang.IllegalArgumentException: bound must be positive
        at java.base/java.util.Random.nextInt(Random.java:388)
        at kotlin.random.AbstractPlatformRandom.nextInt(PlatformRandom.kt:45)
        at kotlin.random.Random$Default.nextInt(Random.kt:246)
        at io.kotlintest.properties.Gen$Companion$set$1$random$1.invoke(Gen.kt:552)
        at io.kotlintest.properties.Gen$Companion$set$1$random$1.invoke(Gen.kt:549)
        at kotlin.sequences.GeneratorSequence$iterator$1.calcNext(Sequences.kt:544)
        at kotlin.sequences.GeneratorSequence$iterator$1.hasNext(Sequences.kt:562)
        at kotlin.sequences.FlatteningSequence$iterator$1.ensureItemIterator(Sequences.kt:278)
        at kotlin.sequences.FlatteningSequence$iterator$1.hasNext(Sequences.kt:265)
        at kotlin.sequences.TakeSequence$iterator$1.hasNext(Sequences.kt:374)
        at io.kotlintest.properties.PropertyTestingAssertAllKt._assertAll(PropertyTestingAssertAll.kt:346)
        at io.kotlintest.properties.PropertyTestingAssertAllKt.assertAll(PropertyTestingAssertAll.kt:41)
        at com.sksamuel.kotlintest.properties.GenTest$22$2$1.invoke(GenTest.kt:502)
        at com.sksamuel.kotlintest.properties.GenTest$22$2$1.invoke(GenTest.kt:39)
        at TestsKt.testAndShrink(tests.kt:12)
        at io.kotlintest.properties.PropertyTestingAssertAllKt._assertAll(PropertyTestingAssertAll.kt:52)
        at io.kotlintest.properties.PropertyTestingAssertAllKt.assertAll(PropertyTestingAssertAll.kt:41)
        at com.sksamuel.kotlintest.properties.GenTest$22$2.invokeSuspend(GenTest.kt:501)
        at com.sksamuel.kotlintest.properties.GenTest$22$2.invoke(GenTest.kt)

lenalebt:
So, I upped the test runs for the size parameter - did not fail locally for me, although I ran them multiple times. I forgot about the special case of a "random" set/list/map with zero elements. I now explicitly forbid that case, since it does not make too much sense to explicitly generate (only) random empty lists/sets/maps. They still are being generated.

Regarding the binary compatibility of default parameters: I don't know how kotlin behaves there (I am not too deep into JVM internals), but my expectation is that without `@JvmOverloads` the previously-existing method without any parameter might not be found anymore. But whatever - with that annotation, it will work anyways.

LeoColman:
Perhaps it didn't fail locally because you were running tests with old code. IntelliJ sometimes does that, mainly with large codebases. Check if you need to delete the `out` folder. It may keep old compiled classess instead of compiling new ones

lenalebt:
Maybe, thanks for the suggestion :+1:

sksamuel:
Binary compatibility would only be an issue if someone had compiled tests against one version of KT and ran them against a newer version of KT. No one does that so we're fine.

LeoColman (approved):

LeoColman:
I'm ok with this, @sksamuel 
What do you think?

LeoColman (approved):

sksamuel (approved):

lenalebt:
nice, thanks :)

PR #758, opened 2019-05-06
Implement Gen.take(n) function

### LINKED ISSUE #739: Introduce `Gen.take(int)` function
Sometimes people won't want to use `assertAll` or similars when using a Gen, but does want the generated values anyway (such as collection them in a list).

What about introducing a member function (can be an extension function as well) to allow collection of values in a Gen?

Perhaps

```
// Contains all IntGenerator.constants() + X random values from IntGenerator.random(), in which X = 1000 - IntGenerator.constants().size()
val generatedIntList: List<Int> = IntGenerator().take(1000)
```

LeoColman:
Fixes #739


sksamuel:
This is nice.

sksamuel (approved):

PR #773, opened 2019-05-16
Add more primative gens

### LINKED ISSUE #772: Generators for Byte, Short, and experimental unsigned types
It seems straightforward to add them. I can submit a pull request if desired.

magneticflux-:
Closes #772


sksamuel:
lgtm merge when you're happy @Kerooker

sksamuel:
Thanks for the pr @magneticflux-

Kerooker:
Could we add some tests for these new generators?

LeoColman (changes requested):
We should add some tests for the new generators, to ensure that we won't break them in the future. If you can't create the tests right now or whatever reason, please let us now, and we can do it for you :)

magneticflux-:
I added to some existing tests, but I didn't see any other tests for the original Int and Long gens to parallel for Short and Byte. If you want more tests, just link where I should put them and I'll add them!

LeoColman:
I think we're ok for now.
I'll add some tests for all of these later.

Thanks for the PR, @magneticflux- !

PR #784, opened 2019-05-26
added `asClue` helper and rewrote `withClue` to allow nesting

### LINKED ISSUE #783: withClue: Clues should be nestable
Consider this case (taken from the tests):

```
      withClue("clue outer:") {
        shouldThrow<AssertionError> {"1" shouldBe "2" }.message shouldBe "clue outer: expected: \"2\" but was: \"1\""
        withClue("clue inner:") {
          shouldThrow<AssertionError> {"3" shouldBe "4" }.message shouldBe "clue inner: expected: \"4\" but was: \"3\""
        }
        shouldThrow<AssertionError> {"5" shouldBe "6" }.message shouldBe "clue outer: expected: \"6\" but was: \"5\""
      }
```
The error message from the inner clue does not contain the outer clue, as in it is: `"clue inner: expected: \"4\" but was: \"3\""` and imho should be along the lines of `"clue outer: clue inner: expected: \"4\" but was: \"3\""`.

Sometimes it is very helpful when you get more information about what exactly went wrong in your test case. We're using a similar approach in our tests that allows nesting clues, which really helps seeing what exactly failed.
This is most useful in more complicated integration tests. The clue in these cases would not be a simple string, but an object (e.g. the complete http response of a service where we checked the status code):

```
val httpResponse = doSomething()
httpResponse.asClue {
  it.status.shouldBe(307)
  it.header.location.asClue { header ->
    header.isPresent.shouldBe(true)
    header.get().shouldContain("http://")
  }
}
```
Would you be open for a PR in that direction?

lenalebt:
fixes #783

I changed to formatting of the former `withClue` to add newlines after each clue, since I think this is more readable, especially when using data classes as clue. Additionally, I allowed returning something from the body (in both cases). It is not strictly needed for these cases, but sometimes comes in handy to write code in an immutable way and does not hurt either.

### DIFF
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/matchers.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/matchers.kt
@@ -4,13 +4,16 @@ package io.kotlintest.matchers
 
 import io.kotlintest.*
 
-fun withClue(clue: String, thunk: () -> Any) {
-  val currentClue = ErrorCollector.clueContext.get()
+fun <ReturnType> withClue(clue: Any, thunk: () -> ReturnType): ReturnType {
+    return clue.asClue { thunk() }
+}
+
+fun <ClueType, ReturnType> ClueType.asClue(body: (ClueType) -> ReturnType): ReturnType {
   try {
-    ErrorCollector.clueContext.set("$clue ")
-    thunk()
+    ErrorCollector.clueContext.get().push(this)
+    return body(this)
   } finally {
-    ErrorCollector.clueContext.set(currentClue)
+    ErrorCollector.clueContext.get().pop()
   }
 }
 
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/matchers.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/matchers.kt
@@ -4,14 +4,28 @@ package io.kotlintest.matchers
 
 import io.kotlintest.*
 
+/**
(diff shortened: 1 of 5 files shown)


kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/matchers.kt
```diff
@@ -4,13 +4,16 @@ package io.kotlintest.matchers
 
 import io.kotlintest.*
 
-fun withClue(clue: String, thunk: () -> Any) {
-  val currentClue = ErrorCollector.clueContext.get()
+fun <ReturnType> withClue(clue: Any, thunk: () -> ReturnType): ReturnType {
+    return clue.asClue { thunk() }
+}
+
+fun <ClueType, ReturnType> ClueType.asClue(body: (ClueType) -> ReturnType): ReturnType {
```
LeoColman: Could you please add some javadocs to this method? It may improve usability when users try to verify what this does inside the IDE (CTRL + Q).

LeoColman (changes requested):
Great implementation and tests!

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/matchers.kt
lenalebt: Will do!

lenalebt:
Thanks :). Trying to give back some of the joy I'm having when using the framework ;).

LeoColman:
Good for me.

Whenever you want it, @sksamuel

LeoColman (approved):

sksamuel:
lgtm

PR #792, opened 2019-05-29
Add infix map matchers using Pair

### LINKED ISSUE #791: Map matcher should{Not}Contain support for Pairs
At the moment, the most terse map matcher is of the form:

```kotlin
map.shouldContain("key", "value")
```

It would be nice to add support to be more fluid by using `Pair` to say:

```kotlin
map shouldContain ("key" to "value")
```

This is a very small fix which I can create a PR for, but just checking if this is something that will be accepted.

rses:
Fixes  #791


LeoColman:
Thanks a lot, @rses !

As soon as this gets green I'll be merging!

LeoColman (approved):

sksamuel:
lgtm

Issue #790, opened 2019-05-29
withConstantNow only mock the NoParameterNowFunction


kelly-yinn:
I used the `Current instant listeners` to test the `now value` in my test code.

And I declared as following:
```
override fun listeners() = listOf(
        ConstantNowTestListener(now)
    )
```

But I found that if the call is `ZonedDateTime.now(zoneId)` not  `ZonedDateTime.now()`, what I obtained is not the `fixed now value`.

By analyzing the source code, I found it seems that we just mock the `no parameter now function`.  
So in order to be consistent with my test code. I first call `ZonedDateTime.now()` and then convert it according to the given `zoneId`. 

So I think maybe it would be better to mock more `now function`.

LeoColman:
This makes sense.

Would you like to make a pull request?

kelly-yinn:
Thanks for your reply. 

I'd like to have a try and make a pull request.

LeoColman:
Great!
Let me know if you have any troubles with it!

kelly-yinn:
Hi, so sorry for my late reply. 
I have created a pull request: https://github.com/kotlintest/kotlintest/pull/798. 

And I mocked `now(zoneId)` for 3 classes: `ZonedDateTime`, `OffsetDateTime` and `LocalDateTime`, also I put the reasons why I did not mocked the `now(zoneId)` for other classes in the PR description.

There may some errors，typos  and ugly code exist for my lack of experience, so sorry for that and wait for your feedbacks.

sksamuel:
@Kerooker any idea on this one ?

LeoColman:
I think this was depending on the PR, but I think it was not completely done and now is distant from the master branch

sksamuel:
I think it was that you can't use now(zoneId) with the listener.

LeoColman:
Yes, that was the feature request

kelly-yinn:
Yes, it was that we can't use `now(zoneId)` and `now(clock)` with the listener. Let me have a try on Sunday :)

PR #798, opened 2019-06-02
For issue #790, mock now(zoneId) for ZonedDateTime, OffsetDateTime and LocalDateTime

kelly-yinn:
Mocked `XXXDateTime.now( zoneId )`

And for other classes that tested in test class `ConstantNowExtensionsTests.kt`, such as `LocalDate`, `LocalTime`, `Year`,  It seems that converting them is impossible.  And only both the `date` and `time` are provided, the converting can be achived. For this point, I need your verification.

For these `cannot convert` classes, I returned the parameter `value` directly, it is not appropriate, I think maybe throw an exception is acceptable, but I am not sure, so need your suggestions.


LeoColman:
I suggest a different approach:

Perhaps instead of depending on context for this extension (the specific context when one want a differente Zone), what if we create two different extensions?

One for constantNow() and one for constantNow which contains the TimeZone/ZoneID/...

What do you think?

LeoColman:
This approach seems somewhat magic IMO. The caller won't know exactly what the method will do inside

kelly-yinn:
Yes, that makes sense. Maybe caller would be confused and hard to locate issues when unpredictable  condition happens for my approach. I will update this. Thank you!

LeoColman:
Let's just think what we want to call that extension, as to make it clear.

And perhaps we can make it work only for the types that contain a configurable TimeZone, as you explained in your PR.

LeoColman:
Also, you're doing just fine for your first PR.

Thanks for volunteering to help us with this issue!

kelly-yinn:
You mean that for some particular types we use the configurable TimeZone, but for other types, the callers need to define the value in advance?

LeoColman:
No, I mean to not provide an extension if the value os not convertible, such as `Year`. We'll have to specify an extension for each possibility of TimeZone convertable types

kelly-yinn:
So...we just convert the convertable class, leave the other types alone? I don't know if I misunderstand you.

LeoColman:
I mean:

- Create a new extension for LocalDateTime (which is convertable)
- Create this new extension with an overload for ZonedDateTime (which is also convertible)
- Don't create this new extension for a type like `Year` (which is not convertible in this fashion)

The current extension accepts any `Temporal` because all `Temporal` have an static method called `now`. Not all of them have an static method called `now(parameter)`, so we can't use this generic approach on the ones that don't have the parameter for the now method.

Is this clearer?

kelly-yinn:
I have another idea, the Caller just provide a now, a `detailed now`, maybe, ZonedDateTime type. And then, all the other calls, like  `Year.now()`  returns the year based on the `detailed now`, `Year.now(zoneid)` returns the year in the responding zone according to the `Year.now()`....

ylt1534:
I am not sure if my idea is reasonable...And I will check your suggestions tomorrow. 
It's 0:26 am in China, I need to sleep now. Good night.

kelly-yinn:
To rephrase a little bit,  how about having `withConstantNow` accepting only `Instant` or `ZonedDateTime`? In which case, no matter on which type `now`, `now(ZoneId)` or `now(Clock)` is called, we can always get back a meaningful return value. Even when a user has mixed using `Year.now()` and `LocalDateTime.now(ZoneId)` in the test

LeoColman:
I didn't quite understand. Could you give us an example, please?

kelly-yinn:
Ok. 

First we state the current definition. The current definition of `withConstantNow` is : `withConstantNow(now: Time, block: () -> T)` , the parameter `now` is `Time` type.  

But as we discussed before, there are just 3 time types can achieve the converting: 
`ZonedDateTime`, `OffsetDateTime`  and `LocalDateTime`.  
That is if the client provides us a `now` with `ZonedDateTime`  type, and there exists `ZonedDateTime.now(zoneId)` in their source code, we can return the real `now` in corresponding time zone as my PR does. 

But for other time types, like `Year `, if the client provides us a `now` with `Year` type, and called `Year.now(zoneId)` in their source code, we cannot provide the real `now` value in the given time zone.  For example, if you just tell me now is 2008 year in the United States，I cannot tell you which year is in China because you just give me a year, not the detailed time. But if you tell me now is `2018.10.01 16:35:40` in the United States, I can tell you the current year in China. 

So, can we change the definition to `withConstantNow(now: ZonedDateTime, block: () -> T)`  to let the client provide us a time with `ZonedDateTime` so that we can convert it to any format they want ( the year/month/date/time of any time zone, or with any clock). 

I don't know if I express it clearly and what do you think?

sksamuel:
It makes sense, and you're right. Temporal is not sufficient to set now, it should be a DateTime variant.

kelly-yinn:
Ok，and I will update my PR.

LeoColman:
I think you made it clear on your point of view.

However, we don't want to lose our current `withConstantNow` when there's no datetime. We must have  both possibilities

`withConstantNow(Year.of(xyz))` should still be possible, as `Year` have the `now` function. The same is true for `LocalDate`, for example. We don't want to lose the ability to mock `LocalDate.now` just to fit the others.

Modify the method to mock the `X.now(zone)` only if it's possible to have a zone. Don't remove the current possibilities!

kelly-yinn:
Yes I know，the client code would be broken if I remove the published api so I won’t remove it.

LeoColman:
Hey, @ylt1534 !

I know it's a bit late for your PR, but it kind of got buried :sweat_smile: 

Could you please rebase it so we can merge?

PR #804, opened 2019-06-04
Bugfix/771 test error beforetest

sksamuel:


sksamuel:
I created a PR so I could review it on here.

kotlintest-runner/kotlintest-runner-junit5/src/main/kotlin/io/kotlintest/runner/junit5/JUnitTestRunnerListener.kt
```diff
@@ -193,9 +187,11 @@ class JUnitTestRunnerListener(private val listener: EngineExecutionListener,
       logger.error("Spec descriptor cannot be null $description")
       throw RuntimeException("Spec descriptor cannot be null")
     } else {
-      val result = when (t) {
-        null -> TestExecutionResult.successful()
-        else -> TestExecutionResult.failed(t)
+      val firstTestError = results.find { it.result.error != null }?.result?.error
```
sksamuel: I am pretty sure that the JUnit specs say that a parent must pass even if it's children fail.
This is why I do not search the children for a failed test.

kotlintest-runner/kotlintest-runner-console/src/main/kotlin/io/kotlintest/runner/console/TeamCityConsoleWriter.kt
```diff
@@ -1,17 +1,14 @@
...
+import io.kotlintest.*
 import java.io.PrintWriter
 import java.io.StringWriter
 import kotlin.reflect.KClass
 
 class TeamCityConsoleWriter : ConsoleWriter {
 
+  private val results = HashSet<TestResult>()
+
   private var errors = false
```
sksamuel: Is this read anywhere?

kotlintest-runner/kotlintest-runner-junit5/src/main/kotlin/io/kotlintest/runner/junit5/JUnitTestRunnerListener.kt
LeoColman: Perhaps. Let me see what I can do.

LeoColman: The thing is, if we don't fail the parent, the test case won't ever be created

LeoColman: and it won't show in IntelliJ, that's why this fix was necessary

LeoColman: perhaps we can add the dummy failure like we do with our runner

Issue #522, opened 2019-01-06
SkipTestException


sksamuel:
Should we support skipling a test by thowing this exception like junit.

Kerooker:
Something like `@Expected`?

sksamuel:
I don't know what @expected is.
I just mean inside a test you can throw `SkipTestException` and instead of failing, it acts like the test is ignored.

Kerooker:
Ah, I see...
I've never seen a test with this, but I don't think it will be hard to implement

LeoColman:
Reviving this issue:

Do we actually need this? Would this be useful to any of our users? I don't see this as being really necessary, as it's not very different from a return inside a test, is it?

Tell me if we need this, or if we can close this issue for now.

sksamuel:
I suppose you can just return yeah so close away. Although other test
frameworks have it. Maybe they found a reason for it.

sksamuel:
You can include a reason if you use an exception.

LeoColman:
I mean, it's very easy to implement, I just want to see if we want it. Perhaps it's so easy to implement that we can just do it xD

sksamuel:
Yeah can do, as in the exception you can include a message which is then
displayed as the skip reason in the output.

PR #806, opened 2019-06-04
[Next Release Only] Documents SkipTestException

LeoColman:
Part of #522

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -787,6 +787,22 @@ class BangExample : StringSpec({
 })
 ```
 
+### SkipTestException
+Sometimes you want to interrupt a test in runtime, as perhaps you don't know at compile-time if the test should be executed. For this, KotlinTest provides a way to interrupt it by throwing an exception: The `SkipTestException`.
+
+```kotlin
+class SkipTestExceptionExample : StringSpec({
+   
+  "Test should be skipped" {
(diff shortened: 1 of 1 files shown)


doc/reference.md
```diff
@@ -787,6 +787,22 @@ class BangExample : StringSpec({
...
 
+### SkipTestException
+Sometimes you want to interrupt a test in runtime, as perhaps you don't know at compile-time if the test should be executed. For this, KotlinTest provides a way to interrupt it by throwing an exception: The `SkipTestException`.
+
+```kotlin
+class SkipTestExceptionExample : StringSpec({
+   
+  "Test should be skipped" {
+    if(noBananasEaten()) {
+      throw SkipTestException("This test was skipped due to the amount of bananas I ate today.")
```
sksamuel: It's a bit of an odd example.
I'd say something like if (isLocal()) throw SkipTestException("Cannot run this test when in local mode")

sksamuel (changes requested):

doc/reference.md
LeoColman: I randomly came up with a random example.

Your is better. I'll make the change

LeoColman: done.

sksamuel:
Cool

Issue #771, opened 2019-05-14
Error handling in tests


leoxs22:
Hello!

I found a really serious bug on the test runner, when there is an exception in some test.

The effect is really hard to detect.

I have this stacktrace:

```
17:18:32.856 [main] ERROR io.kotlintest.runner.jvm.TestEngine - Error during test engine run
java.lang.reflect.InvocationTargetException: null
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[na:1.8.0_144]
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) ~[na:1.8.0_144]
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[na:1.8.0_144]
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423) ~[na:1.8.0_144]
	at kotlin.reflect.jvm.internal.FunctionCaller$Constructor.call(FunctionCaller.kt:66) ~[kotlin-reflect-1.2.51.jar:1.2.51-release-125 (1.2.51)]
	at kotlin.reflect.jvm.internal.KCallableImpl.call(KCallableImpl.kt:107) ~[kotlin-reflect-1.2.51.jar:1.2.51-release-125 (1.2.51)]
	at kotlin.reflect.jvm.internal.KCallableImpl.callDefaultMethod(KCallableImpl.kt:149) ~[kotlin-reflect-1.2.51.jar:1.2.51-release-125 (1.2.51)]
	at kotlin.reflect.jvm.internal.KCallableImpl.callBy(KCallableImpl.kt:111) ~[kotlin-reflect-1.2.51.jar:1.2.51-release-125 (1.2.51)]
	at kotlin.reflect.full.KClasses.createInstance(KClasses.kt:283) ~[kotlin-reflect-1.2.51.jar:1.2.51-release-125 (1.2.51)]
	at io.kotlintest.runner.jvm.JvmKt.instantiateSpec(jvm.kt:15) ~[kotlintest-runner-jvm-3.3.2.jar:na]
	at io.kotlintest.runner.jvm.TestEngine.createSpec(TestEngine.kt:112) ~[kotlintest-runner-jvm-3.3.2.jar:na]
[...]

PR #807, opened 2019-06-04
Fixes tests passing when before test fails

LeoColman:
The problem with this issue was that any exception that happened in either `BeforeTest` or `AfterTest` was being silently ignored due to Coroutine Launch instead of `async`.

This commit fixes that error, and also modifies listener execution to notify `JUnit` and our executor to be able to identify these errors and show that in the interface.

Part of #771

### DIFF
--- a/kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
+++ b/kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
@@ -8,6 +8,7 @@ import io.kotlintest.internal.unwrapIfReflectionCall
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.async
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import org.slf4j.LoggerFactory
 import java.util.concurrent.*
 import java.util.concurrent.atomic.AtomicReference
@@ -41,9 +42,16 @@ class TestCaseExecutor(private val listener: TestEngineListener,
     try {
 
       // invoke the "before" callbacks here on the main executor
-      context.launch(executor.asCoroutineDispatcher()) {
-        before(testCase)
-      }.join()
+      try {
+        withContext(context.coroutineContext + executor.asCoroutineDispatcher()) {
+          before(testCase)
+        }
+      }catch (t: Throwable) {
+        // Necessary to tell listeners that the test case would be executed if there wasn't an error on
+        // the onBefore
+        listener.invokingTestCase(testCase, 1)
+        throw t
+      }
 
       val extensions = testCase.config.extensions +
           testCase.spec.extensions().filterIsInstance<TestCaseExtension>() +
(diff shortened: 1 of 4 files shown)


kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
```diff
@@ -41,9 +42,16 @@ class TestCaseExecutor(private val listener: TestEngineListener,
     try {
 
       // invoke the "before" callbacks here on the main executor
-      context.launch(executor.asCoroutineDispatcher()) {
-        before(testCase)
-      }.join()
+      try {
```
sksamuel, line 45: Does this actually make any difference? Wouldn't an error inside before() just get silently dropped like before because the exception doesn't propagate out of the thread it runs on ?

sksamuel (changes requested):

kotlintest-runner/kotlintest-runner-jvm/src/main/kotlin/io/kotlintest/runner/jvm/TestCaseExecutor.kt
LeoColman: It does. JUnit is not displaying the test because it was never started.  I'm capturing the `before` error to start the test, and then I'll let it propagate further up the stack.

I don't want to start it if it didn't errored in `before`, as the test might be ignored. But if `before` is an error, it's not possible for it to be ignored.

The exception will propagate, as we'll rethrow it. We're just capturing it here to ensure the test is started (invoking)

LeoColman:
There are some tests for these guys, but for interface testes I did them mainly manually. If you can validate these scenarios, it would be great too.

And I think that after we merge this, we can release the next patch, perhaps?

LeoColman:
@sksamuel Do you think this is a good enough fix?

sksamuel:
I don't think we should be failing the spec if there's _any_ test error, which is what we are now doing. We should only fail the spec if there is an error in the spec init method, or in beforeSpec or afterSpec. I believe this is the JUnit spec - a parent should not fail if it's children fail (odd but that's their rules).

Also, the changes to the withContext shift I don't believe will help. Because we execute the before and after callbacks on our own dispatcher `executor.asCoroutineDispatcher()` then any exception will be swallowed by the default coroutine exception handler.

LeoColman:
The `withContext` will work, because it's the same as `async { }` -> `await()`. I tested that (and created a test scenario for it).

sksamuel:
Ah ok. And the other point about failing the spec?

LeoColman:
Just an addendum to my previous comment
I made async -> await initially, but IntelliJ suggested to change to withContext as they were the same thing.

About the failing spec, I'm trying to only fail the test, as `beforeTest` is somewhat related to it IMO. What would be a solution to this? I'm not sure if I can find another way. Open to suggestions xD

sksamuel:
Ok so the withContext thing is good then.

As for the failing spec, see the docs here:
https://junit.org/junit5/docs/5.0.3/api/org/junit/platform/launcher/TestExecutionListener.html#executionFinished-org.junit.platform.launcher.TestIdentifier-org.junit.platform.engine.TestExecutionResult-

Might be worth linking to that from within the code actually.
I think what we can do is set a var that says "specException=e" which is only set if the exception is thrown in before / after / init, and then use that for setting the failure status of a spec.

LeoColman:
That may be a good solution. Let me fiddle with it for a bit

sksamuel:
Your PR was basically right, so I took it and adapted it in #862

PR #814, opened 2019-06-07
Documents eventually functions

### LINKED ISSUE #781: Add more docs on eventually
There seems to be an entire project which does exactly what our one single function `eventually` does. Let's add some more use cases to docs for our eventually function in case we've missed some cases.

https://github.com/awaitility/awaitility/wiki/Usage

matheusmessora:
Improves eventually documentation for Non-Deterministic tests.

Closes #781

### DIFF
--- a/doc/nondeterministic.md
+++ b/doc/nondeterministic.md
@@ -14,21 +14,50 @@ Plus it means that your test will sit around waiting on the timeout even if the 
 Another common approach is to use countdown latches and this works fine if you are able to inject the latches in the appropriate places.
 
 As an alternative, KotlinTest provides the `eventually` function which will repeatedly
-test the code until it either passes, or the timeout is reached. This is perfect for nondeterministic code. For example:
+test the code until it either passes, or the timeout is reached. This is perfect for nondeterministic code.
+
+
+### Examples
+
+#### Simple example
+
+Let's assume that we send a message to an asynchronous service. After the message is processed, a new row is inserted into user table.
+
+We can check this behaviour with our `eventually` function. 
 
 ```kotlin
 class MyTests : ShouldSpec() {
   init {
-    should("do something") {
+    should("check if user repository has one row") {
       eventually(5.seconds) {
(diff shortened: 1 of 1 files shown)


doc/nondeterministic.md
```diff
@@ -14,21 +14,50 @@ Plus it means that your test will sit around waiting on the timeout even if the
...
       }
     }
   }
 }
 ```
 
-Continually <a name="eventually"></a>
+#### Exceptions
+
+By default, `eventually` will ignore any exception that is thrown inside the function. If you want to be more specific, you can tell `eventually` to ignore specific exceptions until the required assertion passes.
```
LeoColman, line 42: This is not true. If an exception is thrown inside eventually, it will propagate the exception, and crash the test. 

You have to specify which exception you are expecting to happen.

By default we ignore `AssertionError` only.

LeoColman (changes requested):

doc/nondeterministic.md
matheusmessora: @Kerooker I double checked it. `Eventually` without the second argument, pass Exception as the default one. So it actually ignores both AssertionError and any Exception (if the second argument is not passed).

See `Eventually.kt`
```
fun <T> eventually(duration: Duration, f: () -> T): T = eventually(duration, Exception::class.java, f)
```
Also, there is a test that validate it.
```
      "pass tests that completed within the time allowed"  {
        val end = System.currentTimeMillis() + 2000
        eventually(Duration.ofSeconds(3)) {
          if (System.currentTimeMillis() < end)
            throw RuntimeException("foo")
        }
      }
```

The problem is, if you call eventually with a specific exception, like IOException, it overrides the default Exception generic one.

We could improve the function, forcing the user to explicit ignores exceptions. Awaitility follows this rule. See https://github.com/awaitility/awaitility/wiki/Usage#example-9---ignoring-exceptions

LeoColman (approved):

doc/nondeterministic.md
LeoColman: You're correct. I changed my review.

Issue #612, opened 2019-01-27
Arrow Extension - `should beRight` lambda expression support


evanbennett:
To test the returned value of an `Either` I am using:
```
val either = passwordColumn.loadType(com.google.gson.JsonPrimitive("73^pickles"))
either.shouldBeRight()
val password = (either as Either.Right<Password?>).b
password.shouldNotBe(null)
password!!.hashedPassword.shouldNotBe(null)
password.useExistingValue.shouldBe(false)
```

It would be nicer to be able to use lambda expression support, such as:
```
passwordColumn.loadType(com.google.gson.JsonPrimitive("73^pickles")).shouldBeRight {
	it.shouldNotBe(null)
	it!!.hashedPassword.shouldNotBe(null)
	it.useExistingValue.shouldBe(false)
}
```

Note: `passwordColumn.loadType(...)` returns `Either<ValidationError, Password?>`.

evanbennett:
I have looked into the code, and come up with:
```
fun <B> Either<Any?, B>.shouldBeRight(checkB: (B) -> Unit) = this should beRight(checkB)
// fun <B> Either<Any?, B>.shouldNotBeRight(checkB: (B) -> Unit) = this shouldNot beRight(checkB)
fun <B> beRight(checkB: (B) -> Unit) = object : Matcher<Either<Any?, B>> {
	override fun test(value: Either<Any?, B>): Result {
		return when (value) {
			is Either.Left -> Result(false, "Either should be Right(...) but was Left(${value.a})", "")
			is Either.Right -> {
				checkB(value.b)
				Result(true, "", "")
			}
		}
	}
}
```
My feeling is that this should not be negatable, because I do not see how to negate `checkB`.

sksamuel:
How about we do it even better by using contracts.

@Kerooker you are the contract man. What do you think about doing something like

```
a.shouldBeRight()
// contract to say a is now anInstanceOf[Right]
a.someRightOnlyOp()
```
?

LeoColman:
Unfortunately contracts cannot be used to infer type with generics yet
https://youtrack.jetbrains.com/issue/KT-28298

We had this exact same issue at #602.

We can wait on this feature (I do believe it's going to be supported, as per https://github.com/Kotlin/KEEP/issues/139#issuecomment-436897878 and it's answer) or add the lambda as a current workaraound

sksamuel:
I think we can add the lambda, then also add the contract version later.

sksamuel:
[...]

PR #834, opened 2019-06-15
Feature/arrowcontracts

sksamuel:
I've added contracts for Validated, Either and Option.
These contracts allow us to extract the field from these datatypes, but because of the issue with contracts not supporting generics yet, the extracted field will be `Any?`.
This is still a big improvement I think, and also means when contracts support generics (1.4?) we can easily update these matchers.

Current example:

```kotlin
val either = "foo".right()
either.shouldBeRight()
(either as Either.Right<String>).b shouldBe "foo"
(either as Either.Right<String>).b.length shouldBe 3
```

Example with this PR:

```kotlin
val either = "foo".right()
either.shouldBeRight()
either.b shouldBe "foo"
(either.b as String).length shouldBe 3
```

Example in the future:

```kotlin
val either = "foo".right()
either.shouldBeRight()
either.b shouldBe "foo"
either.b.length shouldBe 3
```

This partially addresses the feature request of #612

### DIFF
--- a/kotlintest-assertions/kotlintest-assertions-arrow/build.gradle
+++ b/kotlintest-assertions/kotlintest-assertions-arrow/build.gradle
@@ -23,4 +23,11 @@ dependencies {
     testCompile project(":kotlintest-runner:kotlintest-runner-junit5")
 }
 
+compileKotlin {
+    kotlinOptions {
+        freeCompilerArgs += '-Xuse-experimental=kotlin.Experimental'
+    }
+}
+
+
 apply from: '../../publish.gradle'
--- a/kotlintest-assertions/kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/either/matchers.kt
+++ b/kotlintest-assertions/kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/either/matchers.kt
@@ -6,14 +6,20 @@ import io.kotlintest.Result
 import io.kotlintest.matchers.beInstanceOf2
 import io.kotlintest.should
 import io.kotlintest.shouldNot
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
 
-fun <T> Either<Any?, T>.shouldBeRight() = this should beRight()
+@UseExperimental(ExperimentalContracts::class)
+fun <T> Either<*, T>.shouldBeRight() {
+  contract {
+    returns() implies (this@shouldBeRight is Either.Right<*>)
+  }
+  this should beRight()
+}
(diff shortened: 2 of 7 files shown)


kotlintest-assertions/kotlintest-assertions-arrow/build.gradle
```diff
@@ -23,4 +23,11 @@ dependencies {
     testCompile project(":kotlintest-runner:kotlintest-runner-junit5")
 }
 
+compileKotlin {
+    kotlinOptions {
+        freeCompilerArgs += '-Xuse-experimental=kotlin.Experimental'
```
LeoColman, line 28: Is this actually necessary?

I remember adding a experimental contract somewhere else in our code, and I don't think I needed to add this.

kotlintest-assertions/kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/either/matchers.kt
```diff
@@ -6,14 +6,20 @@ import io.kotlintest.Result
 import io.kotlintest.matchers.beInstanceOf2
 import io.kotlintest.should
 import io.kotlintest.shouldNot
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
+
+@UseExperimental(ExperimentalContracts::class)
+fun <T> Either<*, T>.shouldBeRight() {
+  contract {
+    returns() implies (this@shouldBeRight is Either.Right<*>)
```
LeoColman, line 15: We wanted this to be the generic type, but it's not possible due to contracts not supporting it yet, right?

LeoColman (approved):

kotlintest-assertions/kotlintest-assertions-arrow/build.gradle
sksamuel: Yeah you need it where the contract is defined, but not where it is used.

kotlintest-assertions/kotlintest-assertions-arrow/src/main/kotlin/io/kotlintest/assertions/arrow/either/matchers.kt
sksamuel: Yes, so once contracts support generics we can just change `Either.Right<*>` to `Either.Right<T>` and it will be generic.

LeoColman:
Great.

Let's merge this!

sksamuel:
Please do if you're happy. We can apply this strategy to other parts too.

PR #839, opened 2019-06-18
Ensure specs that fail in initialisation fail a Maven build (#832)

Mahoney:
KotlinTest fails to notify the Maven Surefire plugin properly via the
JUnit platform if a spec throws an exception during initialisation.

This is because whilst `TestEngine` does try to register a spec that
fails during initialisation, it accidentally skips the beforeSpecClass
and afterSpecClass methods on the `TestEngineListener` because it is
wrapped in an `IsolationTestEngineListener` which has guards around
those methods, and the guards always return false because the spec was
never created.

See https://github.com/Mahoney/maven-kotlintest for an example of the
problem.

(cherry picked from commit b99f7f7595d3f15f7b453f014c7ce9647ccc0c23)


Mahoney:
Wondering if this could be in 3.3.3?

sksamuel:
Yes it will be.

LeoColman:
I guess it wasn't?
Can we fix this guy for 3.4?

sksamuel:
I incorporated the changes into the coroutine fixes I think.

PR #843, opened 2019-06-19
Add possibility of configuration to Environment Listeners

### LINKED ISSUE #829: Add mode to SystemEnvironmentProjectListener etc
The mode should allow us to control when the env is added - always or only if it doesn't already exist.

`SystemEnvironmentProjectListener("ENV_NAME", "test", Mode.OVERWRITE)`

vs

`SystemEnvironmentProjectListener("ENV_NAME", "test", Mode.CREATE)`

LeoColman:
Closes #829

### DIFF
--- a/kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/MapOverrides.kt
+++ b/kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/MapOverrides.kt
@@ -1,12 +1,60 @@
 package io.kotlintest.extensions.system
 
-@PublishedApi
-internal infix fun Map<String,String>.overridenWith(map: Map<String, String?>): MutableMap<String, String> {
-  return toMutableMap().apply { putReplacingNulls(map) }
+sealed class OverrideMode {
+
+  @PublishedApi
+  internal abstract fun override(originalValues: Map<String, String>, newValues: Map<String, String?>): MutableMap<String, String>
+
+  /**
+   * Sets specific values and overrides pre-existent ones, if any
+   *
+   * Any values that are not present in the overrides will be left untouched.
+   */
+  object SetOrOverride : OverrideMode() {
+    override fun override(originalValues: Map<String, String>, newValues: Map<String, String?>) =
+            originalValues.toMutableMap().apply { putReplacingNulls(newValues) }
+  }
+
+  /**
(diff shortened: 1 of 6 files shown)


LeoColman:
I think that I might be missing some tests here, but I'm not sure

kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/MapOverrides.kt
```diff
@@ -1,12 +1,34 @@
 package io.kotlintest.extensions.system
 
+
+enum class SystemOverrideMode { ALLOW_OVERRIDE, DENY_OVERRIDE }
```
sksamuel: The two possibilities, actually lets say three.

If already set ignore
If already set override
Error if already set

So maybe the nomenclature can be:

- SET_OR_OVERRIDE
- SET_OR_ERROR (default)
- SET_OR_IGNORE

?

sksamuel (changes requested):
As usual from your PRs it's great, but I think we should allow all three options now.

kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/MapOverrides.kt
LeoColman: My only point to this is that `SET_OR_OVERRIDE` should be set as default. It's the default now, and I don't think we need to change that behavior and potentially break clients

sksamuel:
Agreed.

kotlintest-extensions/src/main/kotlin/io/kotlintest/extensions/system/MapOverrides.kt
sksamuel: Yes ok agreed.

LeoColman:
@sksamuel I think it's done.

Please review again!

LeoColman:
I decided that setting the default to `SetOrError` is better. It should be the expected, and if clients' code break, it is a good thing

sksamuel:
Do you want me to finish this off?

sksamuel (approved):

LeoColman:
Did I miss anything?

sksamuel:
Don't think so

PR #847, opened 2019-06-20
Changed autoClose to accept an AutoCloseable

xenomachina:
Previously autoClose's parameter was a Closeable. Unfortunately, some
classes implement AutoCloseable but not Closeable (JDBC and
testcontainers are a couple of examples).

Closeable extends AutoCloseable, so autoClose will continue working with
Closeable objects.


sksamuel:
lgtm @Kerooker

PR #848, opened 2019-06-20
Added shouldBeZero and shouldNotBeZero matcher for number types (#819)

### LINKED ISSUE #819: ShouldBeZero matcher
`Int`, `Double`, `Long`, (...) should have a special matcher to verify if it's zero.

This matcher should also have an opposite, to verify it's not zero.

```
myInt.shouldBeZero()
myDouble.shouldNotBeZero()
```

Special cases to pay attention to: `Double.NaN` and `Double.POSITIVE_INFINITY`/`Double.NEGATIVE_INFINITY`.

These can be added to the `kotlintest-assertions` module. There can be a matcher for each numeric type, or a `Number` matcher that match them all.

An example on how to start the `isZero` method:

```
fun Number.isZero(): Boolean {
  return when(this) {
    is Double -> this == 0.0    // Verification for NaN
    is Float -> this == 0.0     // Verification for NaN
    else -> this.toLong() == 0L
  }
}
```

faogustavo:
Closes #819

I've also added the same matcher for floats (idn if it was necessary, because if it was not described on the task).
Also, there was a matcher file with just a couple of rules on the `numerics` package, so I joined with the existing one with more rules. Did the same with floats with to keep the same pattern from other numerics.


kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/FloatMatchers.kt
```diff
@@ -29,3 +51,6 @@ class FloatToleranceMatcher(val expected: Float, val tolerance: Float) : Matcher
 
   infix fun plusOrMinus(tolerance: Float): FloatToleranceMatcher = FloatToleranceMatcher(expected, tolerance)
 }
+
+infix fun Float.shouldBeExactly(x: Float) = this shouldBe exactly(x)
+infix fun Float.shouldNotBeExactly(x: Float) = this shouldNotBe exactly(x)
```
hjJunior: Pls, add new empty line at the end of file

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/IntMatchers.kt
```diff
@@ -26,4 +28,23 @@ fun beGreaterThan(x: Int) = object : Matcher<Int> {
...
+      )
+}
+
+fun exactly(x: Int) = object : Matcher<Int> {
+  override fun test(value: Int) = Result(
+      value == x,
+      "$value should be equal to $x",
+      "$value should not be equal to $x"
+  )
 }
```
hjJunior: Same again :)

hjJunior (approved):
It looks great, just give another look about new line at end of file

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/FloatMatchers.kt
```diff
@@ -2,13 +2,35 @@ package io.kotlintest.matchers
...
 fun exactly(d: Float): Matcher<Float> = object : Matcher<Float> {
   override fun test(value: Float) = Result(value == d, "$value is not equal to expected value $d", "$value should not be equal to $d")
 }
 
+fun lt(x: Float) = beLessThan(x)
+fun beLessThan(x: Float) = object : Matcher<Float> {
+  override fun test(value: Float) = Result(value < x, "$value should be < $x", "$value should not be < $x")
+}
+
+fun lte(x: Float) = beLessThanOrEqualTo(x)
```
sksamuel: These already exist in floats/matchers.kt.

The FloatMatchers.kt file is basically deprecated because it doesn't allow for multiple matchers of the same type but different type to be imported.

sksamuel (changes requested):
The zero stuff is awesome and can be merged.
The float stuff needs change as per request.
Thanks!

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/FloatMatchers.kt
faogustavo: Só, it's preferable to keep these rules in the floats package? Shouldn't we do the same with other types? Like for ints, doubles, etc?

sksamuel:
Yes definitely.

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/FloatMatchers.kt
gustavo-valvassori-codeminer42: Done!

gustavo-valvassori-codeminer42:
@sksamuel can you see if it's ok now?

faogustavo:
@sksamuel can you see if it's ok now?

sksamuel:
:+1:

PR #857, opened 2019-06-22
Added shouldContainFiles matcher for path (#854)

### LINKED ISSUE #854: New file matchers
dir.shouldContainFilenames("a.txt", "b.gif")

faogustavo:
Closes #854

### DIFF
--- a/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/file/matchers.kt
+++ b/kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/file/matchers.kt
@@ -233,3 +233,24 @@ fun containFileDeep(name: String): Matcher<Path> = object : Matcher<Path> {
       fileExists(value.toFile()), "File $name should exist in $value", "File $name should not exist in $value"
   )
 }
+
+fun Path.shouldContainFiles(vararg files: String) = this should containsAllFiles(files.asList())
+fun Path.shouldNotContainFiles(vararg files: String) = this shouldNot containsAllFiles(files.asList())
+fun containsAllFiles(files: List<String>) = object : Matcher<Path> {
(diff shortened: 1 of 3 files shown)


kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/file/matchers.kt
```diff
@@ -233,3 +233,24 @@ fun containFileDeep(name: String): Matcher<Path> = object : Matcher<Path> {
       fileExists(value.toFile()), "File $name should exist in $value", "File $name should not exist in $value"
   )
 }
+
+fun Path.shouldContainFiles(vararg files: String) = this should containsAllFiles(files.asList())
+fun Path.shouldNotContainFiles(vararg files: String) = this shouldNot containsAllFiles(files.asList())
+fun containsAllFiles(files: List<String>) = object : Matcher<Path> {
```
sksamuel: Can we call this containFiles so then people can do `dir should containFiles(a, b)` which reads a bit nicer.

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/file/matchers.kt
```diff
@@ -233,3 +233,24 @@ fun containFileDeep(name: String): Matcher<Path> = object : Matcher<Path> {
...
+  override fun test(value: Path): Result {
+    val existingFiles = files.intersect(value.toFile().list().asIterable())
```
sksamuel: I didn't check but can list() return null ?

faogustavo: Yes, it can. Just saw that. I'll push a fix in a few minutes.

PR #864, opened 2019-06-26
Add throwable matchers (#811)

### LINKED ISSUE #811: ShouldThrow matchers should allow testing for message and cause
I said something about this at #810, but it can be a different issue.

I usually see code used with `shouldThrow` being used in this fashion:

```
val throwable = shouldThrowAny { foo() }

throwable.message shouldBe "fooBar"
```

We should come up with a syntax to allow the `shouldThrow` matchers (all of them, of course) to explicitly test for the message.

An idea would be to optionally receive a message parameter:

```
shouldThrowAny(message = "fooBar") { foo() }
```

But I don't like this approach.

What I usually see is also
```
shouldThrowAny {
foo()
}.message shouldBe "fooBar"
```

But I don't know if it's descriptive enough.

Thoughts? @sksamuel

faogustavo:
Closes #811


LeoColman:
That was fast lol

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
```diff
@@ -0,0 +1,83 @@
+package io.kotlintest.matchers.throwable
+
+import io.kotlintest.Matcher
+import io.kotlintest.Result
+import io.kotlintest.should
+import io.kotlintest.shouldNot
+
+inline fun <reified T : Throwable> Throwable.shouldBeOfType() = this should beOfType<T>()
```
LeoColman: I think we already have a `shouldBeOfType`.

`obj.shouldBeInstanceOf<T>` at the beggining of `matchers.md` is actually this.

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
```diff
@@ -0,0 +1,83 @@
...
+inline fun <reified T : Throwable> Throwable.shouldNotBeOfType() = this shouldNot beOfType<T>()
+inline fun <reified T : Throwable> beOfType() = object : Matcher<Throwable> {
+  override fun test(value: Throwable) = Result(
+      value is T,
+      "Value should be of type ${T::class}, but instead got ${value::class}",
+      "Value should not be of type ${T::class}"
+  )
+}
+
+inline fun <reified T : Throwable> Throwable.shouldBeExactlyOfType() = this should beOfTypeExactly<T>()
```
LeoColman: `obj.shouldBeTypeOf<T>` covers this already

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
```diff
@@ -0,0 +1,83 @@
...
+fun Throwable.shouldNotHaveCause() = this shouldNot haveCause()
+fun haveCause() = object : Matcher<Throwable> {
+  override fun test(value: Throwable) = Result(
+      value.cause != null,
+      "Throwable should have a cause",
+      "Throwable should not have a cause"
+  )
+}
+
+inline fun <reified T : Throwable> Throwable.shouldHaveCauseOfType() = this should haveCauseOfType<T>()
```
LeoColman: To keep the pattern above, this should be `shouldHaveCauseInstanceOf`

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
```diff
@@ -0,0 +1,83 @@
...
+    )
+    else -> Result(
+        value.cause is T,
+        "Throwable cause should be of type ${T::class}, but instead got ${value::class}",
+        "Throwable cause should be of type ${T::class}"
+    )
+  }
+}
+
+inline fun <reified T : Throwable> Throwable.shouldHaveCauseOfExacltyType() = this should haveCauseOfTExacltyType<T>()
```
LeoColman: Should be renamed to `shouldHaveCauseOfType` to keep consistency (as mentioned above)

LeoColman (changes requested):

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
faogustavo: Nice!

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
faogustavo: This is to check if the classes are exaclty the same. If you use only `shouldBeOfType` and throw a `FileNotFoundException`, it will accept `IOException`. This is to verify the exactly same class.

LeoColman: I don't think it will.
Unless our docs are lying xP

| Types ||
| ------- | ---- |
| `obj.shouldBeSameInstanceAs(other)` | Compares objects by identity, that is, they are the same exact reference. |
| `obj.shouldBeTypeOf<T>()` | Asserts that the given reference is exactly of type T. Subclass will fail. Ie, `1 should beOfType<Number>` would fail because although 1 _is_ a Number, the runtime type is not Number. |
| `obj.shouldBeInstanceOf<T>` | Asserts that the given reference is of type T or a subclass of T. |
| `obj.shouldHaveAnnotation(annotationClass)` | Asserts that the object has an annotation of the given type. |
| `obj.shouldBeNull()` | Asserts that a given reference is null.|

LeoColman: `ShouldBeTypeOf` shouldn't allow for subclasses, while `shouldBeInstanceOf` should

faogustavo: Oh, right, changed both of them :man_facepalming:

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
faogustavo: done!

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
faogustavo: done!

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
faogustavo: done!

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
faogustavo: done!

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
```diff
@@ -0,0 +1,63 @@
...
+import io.kotlintest.Result
+import io.kotlintest.should
+import io.kotlintest.shouldNot
+
+infix fun Throwable.shouldHaveMessage(message: String) = this should haveMessage(message)
+infix fun Throwable.shouldNotHaveMessage(message: String) = this shouldNot haveMessage(message)
+fun haveMessage(message: String) = object : Matcher<Throwable> {
+  override fun test(value: Throwable) = Result(
+      value.message == message,
+      "Throwable should have message \"$message\", but instead got ${value.message}",
```
LeoColman: I believe this should be treated in a special way. I have to look where it is, but we have a `transformString` of some type somewhere in our code.

I saw that in a recent commit from @sksamuel .

This `${value.message}` might become confusing if it's either null or empty

"Throwable should have message $message but instead got ."

The function I cited already handles this kind of case, I just need to remember which is it

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
```diff
@@ -0,0 +1,63 @@
...
+      "Throwable should not have message \"$message\""
+  )
+}
+
+fun Throwable.shouldHaveCause(block: ((Throwable) -> Unit)? = null) {
```
LeoColman: Perhaps it's better to use
```
block: (Throwable) -> Unit = {}
```

What do you think?

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
```diff
@@ -0,0 +1,63 @@
...
+        "Throwable cause should be of type ${T::class}, but instead got ${value::class}",
+        "Throwable cause should be of type ${T::class}"
+    )
+  }
+}
+
+inline fun <reified T : Throwable> Throwable.shouldHaveCauseOfType() = this should haveCauseOfType<T>()
+inline fun <reified T : Throwable> Throwable.shouldNotHaveCauseOfType() = this shouldNot haveCauseOfType<T>()
+inline fun <reified T : Throwable> haveCauseOfType() = object : Matcher<Throwable> {
+  override fun test(value: Throwable) = when {
```
LeoColman, line 40: The first part of this when is duplicated in two points.
It can be extracted to a method.

What do you think?

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/throwable/ThrowableMatchersTest.kt
```diff
@@ -0,0 +1,178 @@
...
+import io.kotlintest.shouldThrowAny
+import io.kotlintest.shouldThrowExactly
+import io.kotlintest.specs.FreeSpec
+import java.io.FileNotFoundException
+import java.io.IOException
+
+class ThrowableMatchersTest : FreeSpec() {
+  init {
+    "shouldThrowAny" - {
+      "shouldBeExactlyOfType" - {
```
LeoColman: If I remember correctly, the final leaf of a FreeSpec must not contain a `-`.

Would you mind changing this? If you rather use another style, there's no problem

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
faogustavo: done!

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
faogustavo: done!

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
faogustavo: done!

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/throwable/ThrowableMatchersTest.kt
faogustavo: done!

doc/matchers.md
```diff
@@ -189,3 +189,12 @@ For the extension function style, each function has an equivalent negated versio
...
 | `thread.shouldBeTerminated()` | Asserts that the thread has been terminated. |
+
+
+| Throwables / Exceptions | |
+| ---- | --- |
+| `throwable.shouldHaveMessage(message)` | Asserts that the throwable message is the same of the given one. |
+| `throwable.shouldHaveCause()` | Asserts that the throwable have a cause. |
+| `throwable.shouldHaveCause { block }` | Asserts that the throwable have a cause, and pass it as parameter to the block |
+| `throwable.shouldHaveCauseInstanceOf<T>()` | Asserts that the throwable have a cause and it is of type T. |
+| `throwable.shouldHaveCauseOfType<T>()` | Asserts that the throwable have a cause and it is **exaclty** of type T. |
```
sksamuel: should be spelled exactly

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/matchers/throwable/ThrowableMatchersTest.kt
```diff
@@ -0,0 +1,158 @@
...
+        Result.failure<Any>(CompleteTestException()).exceptionOrNull()!!.shouldHaveCause {
+          it shouldHaveMessage "file.txt not found"
+        }
+      }
+      "shouldNotHaveCause" {
+        Result.failure<Any>(TestException()).exceptionOrNull()!!.shouldNotHaveCause()
+        Result.failure<Any>(FileNotFoundException("this_file.txt not found")).exceptionOrNull()!!.shouldNotHaveCause()
+      }
+      "shouldHaveCauseInstanceOf" {
+        Result.failure<Any>(CompleteTestException()).exceptionOrNull()!!.shouldHaveCauseInstanceOf<FileNotFoundException>()
```
sksamuel, line 140: What's the difference between instanceOf and ofType?

faogustavo: One check if is an instance of the provided class or it's parent (instanceOf), the other (typeOf) check if is exactly the same class. You can take the base of `shouldBeInstanceOf` and `shouldBeTypeOf` for any type.

doc/matchers.md
faogustavo: done!

sksamuel:
Would is subtype and is type.

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
```diff
@@ -0,0 +1,54 @@
...
+    value.cause == null -> resultForThrowable(value.cause)
+    else -> Result(
+        value.cause!!::class == T::class,
+        "Throwable cause should be of type ${T::class}, but instead got ${value::class}",
+        "Throwable cause should be of type ${T::class}"
+    )
+  }
+}
+
+fun resultForThrowable(value: Throwable?) = Result(
```
LeoColman: I think this may be a private function

faogustavo: One of the matchers that use this function is inline and if I put private on this, I receive a compilation error.

Kerooker:
You can make it internal with @PublishedApi

Em qui, 27 de jun de 2019 08:30, Gustavo Fão Valvassori <
notifications@github.com> escreveu:

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/throwable/matchers.kt
LeoColman: You can make it internal with @PublishedApi

LeoColman: That way it's at least not public. We did that for many situations like this

faogustavo: done!

sksamuel:
Are we happy to merge this @Kerooker

LeoColman:
lgtm

PR #878, opened 2019-07-05
Ignore to print empty configs (#877)

### LINKED ISSUE #877: Improvements to ProjectConfig output
I'm loving the new project config output @faogustavo 

```
~~~ Discovered this project configurations ~~~
-> Parallelism: 1 thread
-> Test order: LexicographicSpecExecutionOrder
-> Soft assertations: False
-> Write spec failure file: True
-> Fail on ignored tests: False
-> Extensions
  - io.kotlintest.extensions.SystemPropertyTagExtension
  - io.kotlintest.extensions.RuntimeTagExtension
-> Listeners
-> Filters
```

I think it would be good if we skip output on things that are empty - in this case Listeners and Filters.

faogustavo:
Closes #877


sksamuel:
Great

PR #881, opened 2019-07-08
Remove the deprecated JUnit surefire provider from reference

notquiteamonad:
It's recommended to use the built-in provider from version 2.22.0 of the maven-surefire-plugin onwards.


sksamuel:
👍

Issue #886, opened 2019-07-12
Add support to Spring Boot's TestExecutionListener


LeoColman:
Spring Boot defines an API for test execution listener. It seems to be similar to ours, but KT is not working with them at the current state.

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/context/TestExecutionListener.html

An example of this feature can be found at @Sql, at this guide:
https://docs.spring.io/spring/docs/current/spring-framework-reference/testing.html#testcontext-executing-sql-programmatically

sksamuel:
I don't understand what they do. Are we supposed to add a KotlinTestExecutionListener, or is KT meant to invoke the current TestExecutionListeners?

LeoColman:
I think the latter.

It seems that one can register some listeners to Spring, and it will do something to execute them. As stated in that link, 

>Note that not all testing frameworks support all lifecycle callbacks defined in this API.

We should be one of the frameworks that does.

I'm figuring out how they're executed now.

We would just need to add their behavior to our SpringListener

tibtof:
Do you need help with this? I also need this feature and I would happily contribute.

sksamuel:
@tibtof I think this is just about done, but there's plenty of other issues if you'd like to contribute again :)

LeoColman:
Once we merge this, it would be awesome if  you could validate it in your usage. As I said, I couldnt validate many scenarios

LeoColman:
If you can test with master snapshot it would be great, @tibtof

tibtof:
@Kerooker I'll do my best to test this in the following week. I'll let you know of my findings.

PR #887, opened 2019-07-13
Modifies SpringListener to include Spring's Test Listeners

LeoColman:
Originally, our SpringLister would only execute one part of the Spring Test Context's API. This was insufficient for some of Spring's implementations, such as the TestExecutionListener api.

This commit changes the way the listener works to enable that kind of listener. It was very similar to our listener execution, so we just had to hook the executions.

Users that didn't use these features before won't be impacted in any way.

Implements #886

### DIFF
--- a/kotlintest-extensions/kotlintest-extensions-spring/src/main/kotlin/io/kotlintest/spring/SpringListener.kt
+++ b/kotlintest-extensions/kotlintest-extensions-spring/src/main/kotlin/io/kotlintest/spring/SpringListener.kt
@@ -1,8 +1,11 @@
 package io.kotlintest.spring
 
 import io.kotlintest.Spec
+import io.kotlintest.TestCase
+import io.kotlintest.TestResult
 import io.kotlintest.extensions.ConstructorExtension
 import io.kotlintest.extensions.TestListener
+import io.kotlintest.extensions.TopLevelTest
 import org.springframework.beans.factory.config.AutowireCapableBeanFactory.AUTOWIRE_CONSTRUCTOR
 import org.springframework.test.context.TestContextManager
 import kotlin.reflect.KClass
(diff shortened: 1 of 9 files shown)


LeoColman:
I'm not completely sure this will solve all the issues, but I know that now the listeners are being executed.

I'll ask my peer (he reported that this wasn't working) to take a look and see if this is likely to solve his bug

kotlintest-extensions/kotlintest-extensions-spring/src/main/kotlin/io/kotlintest/spring/SpringListener.kt
```diff
@@ -1,22 +1,50 @@
...
 import io.kotlintest.extensions.TestListener
+import io.kotlintest.extensions.TopLevelTest
 import org.springframework.beans.factory.config.AutowireCapableBeanFactory.AUTOWIRE_CONSTRUCTOR
 import org.springframework.test.context.TestContextManager
 import kotlin.reflect.KClass
 import kotlin.reflect.full.primaryConstructor
 
 object SpringListener : TestListener {
 
+  // Necessary to keep specs and contexts in a map because this Listener
```
sksamuel: The comment doesn't make sense

kotlintest-extensions/kotlintest-extensions-spring/src/test/kotlin/io/kotlintest/spring/SpringTestExecutionListenerTest.kt
```diff
@@ -0,0 +1,69 @@
...
+
+class DummyTestExecutionListener : SpringTestExecutionListener {
+
+  override fun beforeTestExecution(testContext: TestContext) {
+    count++
+  }
+
+  override fun afterTestClass(testContext: TestContext) {
+    count++
+  }
```
sksamuel: Each of these counts should be separate otherwise we don't know if the correct ones were invoked just that the correct quantity was.

kotlintest-extensions/kotlintest-extensions-spring/src/main/kotlin/io/kotlintest/spring/SpringListener.kt
```diff
@@ -1,22 +1,50 @@
...
+
+  override fun afterSpec(spec: Spec) {
+    spec.testContext.afterTestClass()
+  }
+
+  private val Spec.testContext: TestContextManager
+    get() = testContexts.getValue(this)
+
+  // We don't run methods, but we need to pass one to TestContextManager, so we'll pass any.
+  private val method = SpringListener::class.java.methods[0]
```
sksamuel: Should we pick a method we always know will be accessible.  [0] might pick a method that's private or something. I'd suggest using hashcode

sksamuel (changes requested):

kotlintest-extensions/kotlintest-extensions-spring/src/main/kotlin/io/kotlintest/spring/SpringListener.kt
LeoColman: I think I erased the next line. oops.
This map is needed because each spec needs its own context. But as this listener is a singleton, we can't associate each listener to each spec.

To avoid breaking code, I added this map to keep those separate contexts

kotlintest-extensions/kotlintest-extensions-spring/src/main/kotlin/io/kotlintest/spring/SpringListener.kt
LeoColman: Good one. I'll make the change

kotlintest-extensions/kotlintest-extensions-spring/src/test/kotlin/io/kotlintest/spring/SpringTestExecutionListenerTest.kt
LeoColman: Good one. I made the change and found a bug because of it.

LeoColman:
I think this is ready, @sksamuel

sksamuel:
Lgtm

Issue #793, opened 2019-05-29
Eventually should allow configuration for space between requests


LeoColman:

matheusmessora:
Based on Awailitiy, this is called Pooling.
Also, should be great if we create a default configuration from Duration and Pooling. 
At the moment, `eventually` calls a while(true) method.

sksamuel:
Yeah we must improve eventually. I looked at awatility and it seems to be basically the same as our eventually.

I think what we need to support are different durations, different polling methods, and what awaitility calls conditional evaluation listeners. https://github.com/awaitility/awaitility/wiki/Usage#condition-evaluation-listener

Happy for anyone to pick up this or I can tackle it.

matheusmessora:
I'll try to work on that

LeoColman:
We're here to help if needed.

matheusmessora:
The actual eventually receives all the data inside the method parameters. For a non-breaking change, we can add more parameters with default values.

My suggestion is to create a "new" eventually. Maybe _`eventual`_ returning a factory with customizable methods. Similar to awaitility.
This will improve the extensibility of the _eventual_ function.
The polling interval by the default is already configured. 

The result could be something like this:
`eventual() { 1 == 1 }` 

`eventual().during(ofSeconds(2)) { 1 == 1}`

`eventual().during(2).ignoreExceptions() { 1 == 1}`

`eventual().ignoreExceptions(IOException.class) { 1 == 1}`

What do you guys think?

LeoColman:
Perhaps we can use the `eventually` construct, but deprecating the current `eventually` in favor of the builder alternative, and remove it in the future.

But I believe that using a `builder` is too much Javalistic, and we should come up with a better DSL for it (I don't have any suggestions yet :P)

matheusmessora:
Do you have a example of a good one?
_PS: I came from Java. Don't judge me_

sksamuel:
eventually(polling = exponential(8), delay = 4.seconds, ..... and so on) {

}

Then you can do

eventually {

}

If you're happy with the defaults.

And it's 100% backwards compatible.

LeoColman:
I liked your example, @sksamuel

sksamuel:
I would like to see this in 3.4 so I can pick this up if it's stalled.

PR #888, opened 2019-07-15
Introduced await as a more feature rich version of eventually for whe…

sksamuel:
…n you want to use custom polling periods #793


LeoColman:
Did you bring these guys straight from Awaitility? Some of them?

kotlintest-assertions/src/main/kotlin/io/kotlintest/eventually/Eventually.kt
```diff
@@ -1,5 +1,8 @@
-package io.kotlintest
+package io.kotlintest.eventually
```
LeoColman: Moving this guy to another package breaks client's code. 

If it breaks already, we don't even need to deprecate, I suppose

kotlintest-assertions/src/main/kotlin/io/kotlintest/eventually/FibonacciPollInterval.kt
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Create an instance of the [FibonacciPollInterval] with a supplied time unit.
+ *
+ * @param offset   The fibonacci offset. Eg if offset is 5 and poll count is 1 then the duration will be base * fib(6).
+ * @param base The duration that is multiplied by the fibonacci value
+ */
+class FibonacciPollInterval(private val base: Duration, private val offset: Int) : PollInterval {
+
+  init {
+    if (offset <= -1) {
```
LeoColman: require(offset >0) { "message" }

kotlintest-assertions/src/main/kotlin/io/kotlintest/eventually/PollInterval.kt
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * A [PollInterval] represents how often Awaitility will pause before reevaluating the supplied condition.
+ * <p/>
+ * Note that the name "poll interval" is a bit misleading. It's actually a delay between two successive condition evaluations.
+ * I.e if the condition evaluation takes 5 ms and a fixed poll interval of 100 ms is used then the next condition evaluation will happen at
+ * (approximately) 105 ms. It's called <tt>PollInterval</tt> for historic reasons.
+ *
+ * @since 1.7.0
+ */
+interface PollInterval {
```
LeoColman: We need to improve docs on this class

LeoColman:
I think that await might not be the best name here.

It is very used in CoRoutines

kotlintest-assertions/src/main/kotlin/io/kotlintest/eventually/Eventually.kt
sksamuel: I'll add a deprecated alias.

kotlintest-assertions/src/main/kotlin/io/kotlintest/eventually/FibonacciPollInterval.kt
sksamuel: :+1:

sksamuel:
await is only used in coroutines as part of deferred, not as a standalone function, so it won't clash. I could call it `until`.

sksamuel:
I've renamed it `until`, added better (not copied) docs, and moved eventually back to the original package.

Please merge if you're happy.

PR #890, opened 2019-07-15
Expressive Arrow Try Asserts

tibtof:
I tried the assertions today and I felt that the assertion can be a bit more expressive.
I added some more asserts, usage examples can be found in tests.
I had a bit of a dilemma when testing success/failure values, between:
`Try.Success("foo") shouldBeSuccess { it shouldBe "foo" }`
and 
`Try.Success("foo") shouldBeSuccess { this shouldBe "foo" }`

I also noticed that diffs are not properly highlighted in IntelliJ for `shouldBeSuccess [value]`. That's why I delegated testing failure equality to `equalityMatcher`. The result can be compared easier with IntelliJ, at the cost of losing the `Failure()` wrapper.

Other changes:
- use contracts to expose Failure was failing, I added a data class Exception to fix it
- added `.editorconfig` with some basic rules

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+
+indent_style = space
+indent_size = 2
+
+trim_trailing_whitespace = true
+insert_final_newline = true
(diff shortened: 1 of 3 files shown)


sksamuel:
Thanks for the editor config. Very useful as this project is 2 spaces :)

I think it is better than this, so that's good. Great PR.

tibtof:
Hi @sksamuel, thanks for merging my PR. I could try to add this kind of asserts to other Arrow types in the next few days if you think it's useful. I only use it for `Try` right now.

sksamuel:
I think the others already have them. At least Either definitely does as I added it myself.

Issue #853, opened 2019-06-21
New string matchers


sksamuel:
string.shouldBePalindrome()
string.shouldBeOneLine()

LeoColman:
should be palindrome lol

LeoColman:
Have you actually found a use case to this?

sksamuel:
No :)

sksamuel:
One line yes.

PR #892, opened 2019-07-15
Added string matchers for single line and size between #853

sksamuel:


kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/string/matchers.kt
```diff
@@ -154,10 +159,26 @@ infix fun String?.shouldNotHaveMinLength(length: Int) = this shouldNot haveMinLe
...
   Result(
       value.length >= length,
-      "${convertValueToString(value)} should have minimum length of $length",
-      "${convertValueToString(value)} should have maximum length of ${length - 1}")
+      "${value.show()} should have minimum length of $length",
+      "${value.show()} should have maximum length of ${length - 1}")
 }
 
+
+fun String?.shouldHaveLengthBetween(min: Int, max: Int) = this should haveLengthBetween(min, max)
```
LeoColman, line 167: Perhaps we can have an overload that uses range? it would allow a infix notation

```
"foo" shouldHaveLengthIn 1..5
```

LeoColman (approved):

kotlintest-assertions/src/main/kotlin/io/kotlintest/matchers/string/matchers.kt
sksamuel: Good idea. I've added that.

Issue #889, opened 2019-07-15
Lack of LocalTime assertions


alladywek:
I faced with problem that kotlintest doesn't have assertions for LocalTime class

LeoColman:
What kind of assertions do you suggest for LocalTime?

Em seg, 15 de jul de 2019 08:14, Abbas Abbasov <notifications@github.com>
escreveu:

alladywek:
```
LocalTime.shouldBeAfter(LocalTime)
LocalTime.shouldBeAfter(LocalTime)
LocalTime.shouldBeAfter(LocalTime)
LocalTime.shouldBeAfter(LocalTime)
```

alladywek:
```
LocalTime.shouldBeAfter(LocalTime)
LocalTime.shouldBeBefore(LocalTime)
LocalTime.shouldBeBetween(LocalTime)
LocalTime.shouldHaveSameSecondsAs(LocalTime)
LocalTime.shouldHaveSameHoursAs(LocalTime)
LocalTime.shouldHaveSameMinutesAs(LocalTime)
LocalTime.shouldHaveSameNanosAs(LocalTime)
```
and their versions with `Not`, they are the similar as from `io.kotlintest.matchers.date`

LeoColman:
Great matchers!

Do you want to do it? We are open to prs!

If you are not, no worries, we can take it from here

Em seg, 15 de jul de 2019 10:10, Abbas Abbasov <notifications@github.com>
escreveu:

alladywek:
I can do it

sksamuel:
We could add this to 3.4 if it's ready in the next couple of days !

alladywek:
> We could add this to 3.4 if it's ready in the next couple of days !

I think I will do it today/tomorrow

sksamuel:
When you do this @alladywek can you add them in a seperate file `localtime.kt` inside the date folder. Thanks.

alladywek:
> When you do this @alladywek can you add them in a seperate file `localtime.kt` inside the date folder. Thanks.

yeah, I saw your message after PR =) I will change it.

PR #893, opened 2019-07-15
(#889) add LocalTime asserts

alladywek:


sksamuel (commented):
Should we add haveSeconds haveMinutes etc. I can't remember if I had added them in the past.

LeoColman:
I think we should, as LocalTimes are made for this kind of glanularity

LeoColman:
But `time.minutes shouldBe 12` isn't that hard to type either. It wouldn't hurt to add though

sksamuel:
True

sksamuel:
Yeah time.minutes shouldBe 12 is fine.
Merging this.

PR #895, opened 2019-07-16
Minor correction of indentations and typos errors.

dagonco:
Following the documentation to test _kotlintest_ in a pet project I found some typos. Like a left over curly brace, a misspelled word and also the indentation was not the same for all examples in some cases.


sksamuel:
Thanks for taking the time to fix!

Issue #891, opened 2019-07-15
Add matcher for Dates to ignore timezone


sksamuel:
Created from #880

PR #896, opened 2019-07-16
Add matcher for Dates to ignore timezone #891

tibtof:


sksamuel:
We would need an extension function version too.

tibtof:
@sksamuel It is an extension function on ZonedDateTime. Am I missing something?

sksamuel:
Actually, never mind, that is perfectly fine :)

PR #899, opened 2019-07-17
Add matcher to part of the date

### LINKED ISSUE #828: New date matchers
We should make sure we have matchers for all the parts of a date.

Like 

`date.shouldHaveMinute(1)`
`date.shouldHaveDayOfWeek(3)` and so on

wevnasc:


wevnasc:
I made the new matches but seems that have someones created some that are alike.

Is there a matcher that still make sense?

sksamuel:
All looks good.

PR #904, opened 2019-07-19
Add Array overloads to collection matchers

### LINKED ISSUE #900: Our collection matchers don't work on arrays

LeoColman:
Closes #900


LeoColman:
What I did here was to exactly duplicate all the matchers, but change them to Array<T>. I also added some overloads in which the parameter type was Collection to suppor Array as well.

LeoColman:
One thing I didn't do was add overloads to all primitive array types, such as `ByteArray` or `IntArray`. There would be too many overloads if i did this.

It can be done if we want tho

sksamuel (approved):

LeoColman:
I don't think we need documentation for this, right? One can use the collections documentation

sksamuel:
Yeah. Can mention they work.

PR #917, opened 2019-07-23
Output response body on matcher failure log

### LINKED ISSUE #876: Improve ktor matcher
When you do this in ktor `response.shouldHaveStatus(200)` it would be useful if the error message contained the body as well, as that often contains a reason why it's not the expected code.

doyaaaaaken:
Fixes https://github.com/kotlintest/kotlintest/issues/876


sksamuel:
What does this address ?

doyaaaaaken:
@sksamuel 

In #876 , you said
> When you do this in ktor response.shouldHaveStatus(200) it would be useful if the error message contained the body as well

So, I change error message to contain response body `TestApplicationResponse.content` .

sksamuel:
Oh. Is the content only available on the test version ?

doyaaaaaken:
Ah! Sorry! 😮  I'll fix my code.

doyaaaaaken:
Yes, `ApplicationResponse` does not have content.

sksamuel:
Ok cool. Thanks, merging now.

Issue #850, opened 2019-06-21
Increasing/Decreasing list/seq matchers


sksamuel:
list.shouldBeMonotonicallyIncreasing // 1,2,2,2,3,3,5,6
list.shouldBeStrictlyIncreasing // 1,2,7,8

same for decreasing.

Could work on any type that we have a comparator for as well as the built in common types.

PR #922, opened 2019-07-25
Increase and decrease seq matcher (#850)

doyaaaaaken:
@sksamuel 
Hi, I implement feature of issue #850.
This is continuation from PR https://github.com/kotlintest/kotlintest/pull/920.


sksamuel (approved):
Awesome work.

We should probably have versions that don't require T to be Comparable but instead accept a comparator.

doyaaaaaken:
> We should probably have versions that don't require T to be Comparable but instead accept a comparator.

I implement this version's methods.

sksamuel (approved):

sksamuel:
:+1: fantastic

PR #924, opened 2019-07-25
Update doc matchers json

jesuswasrasta:
I verified the behavior running existing tests, then wrote the documentation.
In general, it would be nice having some examples for every assertion, too. What do you think?

### DIFF
--- a/kotlintest-core/src/main/kotlin/io/kotlintest/Project.kt
+++ b/kotlintest-core/src/main/kotlin/io/kotlintest/Project.kt
@@ -54,7 +54,7 @@ object Project {
   private val _projectlisteners = mutableListOf<ProjectListener>()
   private val _filters = mutableListOf<ProjectLevelFilter>()
   private var _specExecutionOrder: SpecExecutionOrder = LexicographicSpecExecutionOrder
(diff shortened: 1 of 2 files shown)


sksamuel:
Yes examples is a great idea. Maybe the matches.md file can link through to
more detailed examples

kotlintest-core/src/main/kotlin/io/kotlintest/Project.kt
```diff
@@ -54,7 +54,7 @@ object Project {
   private val _projectlisteners = mutableListOf<ProjectListener>()
   private val _filters = mutableListOf<ProjectLevelFilter>()
   private var _specExecutionOrder: SpecExecutionOrder = LexicographicSpecExecutionOrder
-  private var writeSpecFailureFile: Boolean = true
+  private var writeSpecFailureFile: Boolean = false
```
LeoColman, line 57: I'm not sure what this has to do with the feature

sksamuel: That's on master too.

sksamuel:
great PR thanks

PR #930, opened 2019-07-31
Improve forAll method error message (#929)

### LINKED ISSUE #929: forAll displays all errors at once, even when they are too many
Hello,

Today, using `forAll` can be a bit annoying, as it displays all failing cases at once.

Example

```kotlin
forAll(Gen.string().take(1000)) {
    it shouldBe "lovely"
}
```

You'll get the following output:

```
java.lang.AssertionError: 0 elements passed but expected 1000

The following elements passed:
--none--

The following elements failed:
<empty string> => expected: "lovely" but was: ""
(900 more lines)
```

When this happens, the output should only provide a few failing examples, and not all the 1000 examples (would be worse if more were requested).

It's sort of like "shrinking" the possible values for which it doesn't work, and it would make it a bit more user-friendly.

doyaaaaaken:
Fixes #929

# Log results (test result run in this PR)

### Pattern1

I run this test code.
```
forAll((1..30).toList()) {
      it shouldBeLessThanOrEqual 15
}
```

And I got below error log.
```
Caused by: java.lang.AssertionError: 15 elements passed but expected 30

The following elements passed:
1
2
3
4
5
6
7
8
9
10
... and 5 more passed elements

The following elements failed:
16 => 16 should be <= 15
17 => 17 should be <= 15
18 => 18 should be <= 15
19 => 19 should be <= 15
20 => 20 should be <= 15
21 => 21 should be <= 15
22 => 22 should be <= 15
23 => 23 should be <= 15
24 => 24 should be <= 15
25 => 25 should be <= 15
... and 5 more failed elements
	at com.sksamuel.kotlintest.properties.GenTest.<init>(GenTest.kt:44)
	... 19 more
```

### Pattern2

I run this test code.
```
forAll((1..30).toList()) {
      it shouldBeLessThanOrEqual 15
}
```

And I got below error log.
```
Caused by: java.lang.AssertionError: 1 elements passed but expected 2

The following elements passed:
1

The following elements failed:
2 => 2 should be <= 1
	at com.sksamuel.kotlintest.properties.GenTest.<init>(GenTest.kt:44)
	... 19 more
```


kotlintest-assertions/src/main/kotlin/io/kotlintest/inspectors/funcs.kt
```diff
@@ -16,22 +16,29 @@ fun <T> runTests(col: Collection<T>, f: (T) -> Unit): List<ElementResult<T>> {
 }
 
 fun <T> buildAssertionError(msg: String, results: List<ElementResult<T>>): String {
-
   val passed = results.filterIsInstance<ElementPass<T>>()
   val failed = results.filterIsInstance<ElementFail<T>>()
+  val maxOutputResult = 10
```
LeoColman: This definitely should be a configurable amount, perhaps via SystemProperties

doyaaaaaken: Thank you your review!
Does below commit match with your intention?
8e5d3a4

LeoColman:
I believe we're missing the testes here, right? That's why the build is breaking?

doyaaaaaken:
I'm sorry, I don't understand this situation. Please, could you tell me what to do?
 
> I believe we're missing the testes here, right?

Should I write test for the case of 'more than 10 successful and failed elements' and 'less than 10  successful and failed elements' like this test code?
https://github.com/kotlintest/kotlintest/blob/69f93a865a46a62c0ddeebca78cd74a2ea1ae70b/kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/properties/GenTest.kt#L170

> That's why the build is breaking?

Maybe the test failure is caused by other reason??
What should I do?
```
com.sksamuel.kotlintest.matchers.date.DateMatchersTest > LocalDateTime.shouldBeToday() should not match the past FAILED
1792    java.lang.AssertionError: Expected exception java.lang.AssertionError but a DateTimeException was thrown instead.
1793        at com.sksamuel.kotlintest.matchers.date.DateMatchersTest$42.invokeSuspend(DateMatchersTest.kt:309)
1794        at com.sksamuel.kotlintest.matchers.date.DateMatchersTest$42.invoke(DateMatchersTest.kt)
1795
1796        Caused by:
1797        java.time.DateTimeException: Invalid value for DayOfMonth (valid values 1 - 28/31): 0
1798            at java.base/java.time.temporal.ValueRange.checkValidValue(ValueRange.java:311)
1799            at java.base/java.time.temporal.ChronoField.checkValidValue(ChronoField.java:717)
1800            at java.base/java.time.LocalDate.of(LocalDate.java:270)
1801            at java.base/java.time.LocalDate.withDayOfMonth(LocalDate.java:1124)
1802            at java.base/java.time.LocalDateTime.withDayOfMonth(LocalDateTime.java:1023)
1803            at com.sksamuel.kotlintest.matchers.date.DateMatchersTest$42$2.invoke(DateMatchersTest.kt:310)
1804            at com.sksamuel.kotlintest.matchers.date.DateMatchersTest$42$2.invoke(DateMatchersTest.kt:9)
1805            at io.kotlintest.FailuresKt.shouldFail(Failures.kt:116)
1806            ... 2 more
```

kotlintest-assertions/src/main/kotlin/io/kotlintest/inspectors/funcs.kt
```diff
@@ -16,22 +16,29 @@ fun <T> runTests(col: Collection<T>, f: (T) -> Unit): List<ElementResult<T>> {
 }
 
 fun <T> buildAssertionError(msg: String, results: List<ElementResult<T>>): String {
-
   val passed = results.filterIsInstance<ElementPass<T>>()
   val failed = results.filterIsInstance<ElementFail<T>>()
+  val maxOutputResult = System.getProperty("kotlintest.assertions.maxOutputResult")?.toInt() ?: 10
```
Jiehong: Shouldn't this new variable be documented?

kotlintest-assertions/src/main/kotlin/io/kotlintest/inspectors/funcs.kt
```diff
@@ -16,22 +16,29 @@ fun <T> runTests(col: Collection<T>, f: (T) -> Unit): List<ElementResult<T>> {
...
   if (failed.isEmpty()) {
     builder.append("--none--")
   } else {
-    builder.append(failed.joinToString("\n") { convertValueToString(it.value) + " => " + exceptionToMessage(it.error) })
+    builder.append(failed.take(maxOutputResult).joinToString("\n") { convertValueToString(it.value) + " => " + exceptionToMessage(it.error) })
+    if (failed.size > maxOutputResult) {
+      builder.append("\n... and ${failed.size - maxOutputResult} more failed elements")
+    }
   }
   throw Failures.failure(builder.toString())
```
Jiehong: Add some tests showing that you now throw the correct error messages when too many failures occur.

Jiehong:
@doyaaaaaken : I don't think the tests fail, because of your code, so that's a false negative (I had the same issue in my PR)

kotlintest-assertions/src/main/kotlin/io/kotlintest/inspectors/funcs.kt
doyaaaaaken: I think I should. I write document as method's comment.
ec5caf3

kotlintest-assertions/src/main/kotlin/io/kotlintest/inspectors/funcs.kt
doyaaaaaken: Add some tests in 1a585bc

doyaaaaaken:
Thanks for review! I improve my code.

kotlintest-tests/kotlintest-tests-core/src/test/kotlin/com/sksamuel/kotlintest/inspectors/InspectorsTest.kt
```diff
@@ -70,6 +70,52 @@ class InspectorsTest : WordSpec() {
...
+          "4 => java.lang.NullPointerException\n" +
+          "5 => java.lang.NullPointerException\n" +
+          "6 => java.lang.NullPointerException\n" +
+          "7 => java.lang.NullPointerException\n" +
+          "8 => java.lang.NullPointerException\n" +
+          "9 => java.lang.NullPointerException\n" +
+          "10 => java.lang.NullPointerException\n" +
+          "... and 2 more failed elements"
+      }
+      "failed results are truncated when failed list length is over 10" {
```
Jiehong: This is the exact same test as in line 73. I guess you wanted to show that the passing case also truncate message, so maybe a too quick copy/paste :P

doyaaaaaken: This code is as I intended. I intended to write test for both `Array.forAll` and `Collection.forAll` in accordance with other tests.
But, I agree with you, this test is not useful.

So, I'll write test about passing case.

doyaaaaaken: f113425

doyaaaaaken:
Than you. I fix my code.

sksamuel:
@Kerooker can we fix the conflict and use this ?

LeoColman:
Probably, but I don't think I can modify this commit.

sksamuel:
I have taken this PR and merged it manually.

Issue #921, opened 2019-07-25
Add infix modifier to arrow matchers


Jiehong:
Hello,

It would be nice to be able to use arrow matchers just like the rest of matchers in kotlintests. For example:

```kotlin
val x = Either.left("value")
x shouldBeLeft "expected"
```

instead of:

```kotlin
val x = Either.left("value")
x.shouldBeLeft("expected")
```

But this isn't possible due to the missing `infix` modifier on Arrow matchers.

LeoColman:
Agreed!
Do you want to do it?

faogustavo:
@sksamuel I think we can close this one now.

PR #931, opened 2019-07-31
Add infix modifier to more Arrow matchers (#921)

Jiehong:
Offer a fix for https://github.com/kotlintest/kotlintest/issues/921.

I also tried to improve a few unit-tests to call some non-called matchers, to slightly improve the coverage.


Jiehong:
Tests don't seem to fail, because of my changes!

LeoColman:
@sksamuel

sksamuel:
lgtm

PR #936, opened 2019-08-06
Set .kotlintest folder to be generated inside build

### LINKED ISSUE #849: Configurable location of .kotlintest dir
From the source,  it looks like the [current location](https://github.com/kotlintest/kotlintest/blob/88afb914d145f1e1244c779358cd05ab1e040038/kotlintest-runner/kotlintest-runner-junit5/src/main/kotlin/io/kotlintest/runner/junit5/JUnitTestRunnerListener.kt#L112) is relative to CWD .

Ideally, `.kotlintest` dir should be created inside build directory so that it can get cleaned when we run the build tool's `clean` task. As this is junit runner which might not have context of Gradle/Maven build dirs, hence the path of `.kotlintest` dir should be made configurable.

LeoColman:
Closes #849

### DIFF
--- a/kotlintest-runner/kotlintest-runner-junit5/src/main/kotlin/io/kotlintest/runner/junit5/JUnitTestRunnerListener.kt
+++ b/kotlintest-runner/kotlintest-runner-junit5/src/main/kotlin/io/kotlintest/runner/junit5/JUnitTestRunnerListener.kt
@@ -1,13 +1,7 @@
 package io.kotlintest.runner.junit5
 
 import arrow.core.Try
-import io.kotlintest.Description
-import io.kotlintest.Project
-import io.kotlintest.Spec
-import io.kotlintest.TestCase
-import io.kotlintest.TestResult
-import io.kotlintest.TestStatus
-import io.kotlintest.TestType
+import io.kotlintest.*
 import io.kotlintest.runner.jvm.TestEngineListener
 import org.junit.platform.engine.EngineExecutionListener
 import org.junit.platform.engine.TestDescriptor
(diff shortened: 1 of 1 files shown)


kotlintest-runner/kotlintest-runner-junit5/src/main/kotlin/io/kotlintest/runner/junit5/JUnitTestRunnerListener.kt
```diff
@@ -109,9 +103,9 @@ class JUnitTestRunnerListener(private val listener: EngineExecutionListener,
   }
 
   private fun writeSpecFailures(failures: List<ResultState>): Try<Any> = Try {
-    val dir = Paths.get(".kotlintest")
-    dir.toFile().mkdirs()
-    val path = dir.resolve("spec_failures").toAbsolutePath()
+    val kotlintestBuildDirectory = Paths.get("build/.kotlintest")
```
sksamuel, line 106: build isn't used by everyhing
maven uses `target`
some other things use `out`

sksamuel (changes requested):

kotlintest-runner/kotlintest-runner-junit5/src/main/kotlin/io/kotlintest/runner/junit5/JUnitTestRunnerListener.kt
LeoColman: Hmm.. You're right. What can we do in this case then?

sksamuel: I think we should have the gradle plugin provide it as a parameter to the test runner. That's the best way to do it, but we don't control the junit platform plugin. It might already supply it.

LeoColman:
I'm closing this as it isn't as trivial as creating on the build directory. We can discuss this further in the future (maybe make it configurable in some way)

PR #953, opened 2019-08-23
Enhances SpringListener to work correctly with all Spring's Listeners

### LINKED ISSUE #950: Kotlintest does not with spring data jpa
`@DataJpaTest` does not appear to work with kotlintest. 
The tests fail with the error:

    javax.persistence.TransactionRequiredException: No EntityManager with actual transaction available for current thread - cannot reliably process 'persist' call
    
Note that the standard JUnit 5 test in the same project works.

See [sample project][sample project] for runnable project

[sample project]: https://github.com/tunaranch/spring-boot-kotlin-jpa-kotlintest

LeoColman:
More information at https://github.com/kotlintest/kotlintest/issues/950

Spring executes transactions (and this include anything inside @DataJPATest) through a Listener called TransactionalTestExecutionListener. This comes pre-registered with a SpringBootTest, and KotlinTest was executing this listener correctly as per #887.

However, only executing the Listener wasn't enough, as it couldn't autowire the EntityManager. Initially, this was happening because our SpringListener had a fake "test method" that was a pointer to java.lang.Object.hashcode method. We made this hack to allow us to execute code inside Spring Test Framework, and it seemed fine on our tests.

Due to hashcode being a java.lang.Object method, Spring couldn't find the transaction attribute, as it verified directly that the method wasn't from Java Object.

With this in mind, we must use an object from another class. My first attempt was to use a dummy method, declared in the SpringListener itself. But this also wouldn't work, as the listener would never find the TransactionAttribute, as the method doesn't (and shouldn't) declare a @Transaction, and SpringListener also didn't declare that annotation. The only way to have a method with that annotation would be inside user's Spec definition, where the user would choose it.

We don't have methods, as we define everything through a DSL. But that didn't seem an issue, as Spring's Listener would get the Transaction Attribute from the method's class if it doesn't declare it. @DataJPATest already declare transaction attributes, so if we managed to get a method inside it, it would work without having to declare @Transaction.

If we could guarantee that our users were overriding a method, we could get its reference, but we couldn't force that, and referecing a method from Spec wouldn't work (as Spec doesn't declare the attributes).

The only way to solve this issue that I found was to inject a dummy method into the class, and get it through reflection, so that Spring Listener would correctly find the Transaction Attributes, and that's what I've done.

Fixes #950

### DIFF
--- a/kotlintest-extensions/kotlintest-extensions-spring/build.gradle
+++ b/kotlintest-extensions/kotlintest-extensions-spring/build.gradle
@@ -2,6 +2,7 @@ plugins {
     id 'java'
     id 'org.jetbrains.kotlin.multiplatform'
     id 'java-library'
+    id "org.jetbrains.kotlin.plugin.spring" version "1.3.41"
 }
 
 repositories {
@@ -23,6 +24,7 @@ kotlin {
                 implementation kotlin('reflect')
                 implementation 'org.springframework:spring-test:5.1.7.RELEASE'
                 implementation 'org.springframework:spring-context:5.1.7.RELEASE'
+                implementation 'net.bytebuddy:byte-buddy:1.10.1'
             }
         }
 
--- a/kotlintest-extensions/kotlintest-extensions-spring/src/jvmMain/kotlin/io/kotlintest/spring/SpringListener.kt
+++ b/kotlintest-extensions/kotlintest-extensions-spring/src/jvmMain/kotlin/io/kotlintest/spring/SpringListener.kt
@@ -5,37 +5,42 @@ import io.kotlintest.TestCase
 import io.kotlintest.TestResult
 import io.kotlintest.extensions.ConstructorExtension
 import io.kotlintest.extensions.TestListener
-import io.kotlintest.extensions.TopLevelTest
+import net.bytebuddy.ByteBuddy
+import net.bytebuddy.description.modifier.Visibility
+import net.bytebuddy.dynamic.loading.ClassLoadingStrategy
+import net.bytebuddy.implementation.FixedValue
+import org.slf4j.LoggerFactory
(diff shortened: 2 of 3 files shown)


sksamuel:
So it tries to find a method in the class, can't find any (because we don't have any) and then returns null, so the entity manager is never autowired in? Shouldn't it / wouldn't it throw some error if it couldn't autowire the class ?

I'm also confused why it's looking for a method in the first place and why having an arbitrary method is enough for it to work. If you have 10 methods in the spec, which one would it pick, would it matter, and if not, why not.

LeoColman:
It throws an error in runtime when it can't resolve the entity manager.

It doesn't try to find a method in class, Spring expects us to pass a method to it (as it's used to JUnit, that uses methods). Initially we decided to use any method (hashcode), and most of the Listeners worked.

However, some listeners such as DataJpaTest tries to find specific information in the method description (`@Transactional`). If it doesn't find it, it tries to search for the class' annotations that provides the information.

It also won't accept any `java.lang.Object` methods.

The way we had previously was:

method = `java.lang.Object.hashcode` -> Not working

An alternative would be:

method = `Spec.anyMethod` -> No `@Transaction` in it -> Search for it in class -> find nothing -> fails

My fix does the following:

method = `Injected method in MySpec` -> No `@Transaction` in it -> Search for it in class -> finds `@Transaction` or `DataJpaTest` -> works

LeoColman:
So, we need a method with `@Transaction`, or a method in a class that has `@Transaction`. The only way to do it is trying to read that from the User class, but it can be done by class, only by method.

It's kinda weird

sksamuel:
Thanks thats a great explanation.
And amazing job by the way.

sksamuel (approved):

sksamuel:
Merging so this doesn't get out of sync.

PR #955, opened 2019-08-24
Feature/versions

sksamuel:
bumped kotlin to 1.3.50
coroutines 1.3.0 has just come out today so bumped that too
and bumped junit platform to latest


sksamuel:
Build is green, see no harm in merging.

glasser:
Looks like there hasn't been a release since this (it didn't get into 3.4.2?)
I'm looking into slowly migrating to kotlintest but it would be nice to not be getting a conflict in my app which is already on junit platform 1.5.2.

sksamuel:
@kerooker lets do a 3.4.3

PR #956, opened 2019-08-24
Using kotlin durations now they are included in 1.3.50

sksamuel:


sksamuel:
This is green and reverts a previous change so merging in.

Issue #912, opened 2019-07-22
New feature: add string generator from regexp


Jiehong:
Hello,

It would be very convenient if a generator was present that would provide strings from a regular expression, something like `Gen.string(regexp: String)` for example, or from `Gen.string(regexp: Regex)`.

Currently, we have to define our own generator based on something like https://www.baeldung.com/kotlin-random-alphanumeric-string.

LeoColman:
Good suggestion!

We should probably take in a regex instead of a string tho

sksamuel:
How would you implement this ?
Generate random strings until they match ?
Or parse the regex yourself to try and generate strings that match it? Seems like both approaches would be problematic.

LeoColman:
I think there are ways to do this
https://stackoverflow.com/questions/22115/using-regex-to-generate-strings-rather-than-match-them

sksamuel:
Ok, this looks good:
https://github.com/mifmif/Generex

BurstX:
BTW, Microsoft created a CLI tool to generate strings that match one or more regexes https://www.microsoft.com/en-us/download/details.aspx?id=52296

PR #957, opened 2019-08-24
Add generator for multiples and factors #942; Move `kotlintest-assert…

### LINKED ISSUE #942: Add generator which is a multiple of k

### LINKED ISSUE #949: Move `kotlintest-assertions` tests to its module
As the assertion module grows, having tests for it all over the place is making it difficult for new contributors to find where to place those tests.

As we don't depend on KotlinTest itself to run them (as we do with Core and Runners), we can easily move these tests closer to the source code they test

sksamuel:
…ions` tests to its module #949

Closes #942 
Closes #949


sksamuel:
Added regex generator.
Closes #912

sksamuel:
Merging so I can continue work. Build is green. Nothing controversial here.

PR #958, opened 2019-08-24
[TableDrivenTesting] - Add an example of generating individual tests like Spock's @Unroll

3ygun:
At work we currently utilize [Groovy's Spock Framework](). The documentation makes it very clean how to generate multiple individual tests from their [data driven testing framework with `@Unroll`](http://spockframework.org/spock/docs/1.3/data_driven_testing.html#_method_unrolling). I looked into the kotlintest documentation and wasn't able to find a direct example of how to get this behavior. Turns out it it supported with less fuss then the Groovy version (as you have access to Kotlin code at all points) but wanted to add the example for future readers looking to switch.

Please let me know if more detail should be added to the example. And/or it should be placed elsewhere. etc.

TL;DR: Using table driven testing is great. Producing less "total tests numbers" then your co-workers thanks to the nesting of table driven testing not showing up in CI test counts, etc. may not. Thus document how you can use table driven testing to produce one test per row of input data. Achieving the best of both worlds!

Q: Did I edit the 3.1 docs instead of the most recent docs?


sksamuel:
Awesome thanks.

PR #959, opened 2019-08-25
Added assertion counting mode #852

### LINKED ISSUE #852: Assertions count
It would be good if we could fail a test if no assertions were called in that test (perhaps optionally).
Sometimes you may write a test with some callback or async code and it's not obvious that the assertions were not invoked.

sksamuel:
Closes #852


LeoColman:
Is there an actual use case on warning instead of erroring for assertions?

LeoColman (approved):

sksamuel:
Possibly people don't want it to fail the build, but still want it flagged it.

PR #960, opened 2019-08-25
Allow seeding of the random values in gens #815

### LINKED ISSUE #815: Must allow seeding of the random values in gens

sksamuel:
Closes #815


LeoColman:
I think that just creating the parameter may not be good enough
What about
- Making it default to null
- Making it non-nullable and default to current millis (just like JVM Random)
- Making it a Random and default it to `kotlin.Random` instance

I think this last one may be better, as we'll depend on anything that generates a Random number, and that can be simulated or mocked, for example(although I have no idea on why the person would want to do this)

sksamuel:
So at the moment it's nullable and defaults to null. See line 45 in Gen interface.

I would be open to changing it to a Random instance if you think that's better, although why you would want to mock Random when you can just control random by the seed I'm not sure.

LeoColman:
Ah, I thought that if it defaulted to null in the interface, you still had to set it to null in all subclasses.

Just tested, and it's not necessary.

LeoColman:
As per the Random instance, I just suppose sometimes, maybe, a random use case will appear.

Maybe changing the Random to a ThreadLocalRandom, perhaps changing it to a random that counts how many times something was called (idk lol)...

I'm just scared to make it this way and not being able to change it in the future, but I'm OK to just merge this too

sksamuel:
Ok I guess it makes sense to accept a Random.

LeoColman:
With this, would it be possible to print or output the seed some way if the test fails? Or perhaps print the used seed every time a `Gen` is created.

I'm thinking that the user will probably want to repeat the test until it is fixed if a certain seed fails

If this sounds too complex, let's leave it for another issue

sksamuel:
I've updated to use Random's instead of seeds.

sksamuel:
You're right, the user will want to know the seed so they can reuse it.
However by using a Random, you can't get access to the seed. There is no method to return it.

So I guess we need to decide what is more important:
- being able to see the seed and reuse it
- being able to pass a ThreadLocalRandom in.

And in fact you can't use a TLR anyway, as we're using Kotlin's MPP randoms which are different.

LeoColman:
If the user passes a random in, they will know/control the seed on their
side

sksamuel:
But we can't get it to print it out.

sksamuel:
I've looked at other frameworks and they use a long.
That is because if we let the user specify the random, then we can't get access to the seed that was used in order to print it out. The idea is you use a random seed until it fails, then you take that random seed and use it to make the test repeatable.

PR #964, opened 2019-08-30
Implement should contain any matcher

### LINKED ISSUE #946: New matcher: shouldContainAny and shouldNotContainAny
A handy matcher to say that a collection should contain any of the provided elements - at least one is sufficient to pass.

LeoColman:
Closes #946


sksamuel:
Maybe shouldContainAnyOf ?

LeoColman:
You think it's better?

sksamuel:
Yes because shouldContain is one element and this is any from the list.
It's clearer to say anyOf

sksamuel:
If you rename this then you can merge.

LeoColman:
Yep. Merging when green

PR #966, opened 2019-09-01
Fix typos in styles.md

Inego:


LeoColman:
You sure did proofread the entire file, huh?

Thanks for the contribution, Inego!

Inego:
> You sure did proofread the entire file, huh?

That's just the way I read documentation :-)

PR #976, opened 2019-09-03
Added new spec DSL for overriding values and functions

sksamuel:


build.gradle
```diff
@@ -51,7 +51,7 @@ allprojects {
     group = "io.kotlintest"
 
     if (isTravis) {
-        version "4.0.0." + travisBuildNumber + "-SNAPSHOT"
+        version "4.0." + travisBuildNumber + "-SNAPSHOT"
```
LeoColman, line 54: Was this intentional?

kotlintest-core/src/commonMain/kotlin/io/kotlintest/Spec.kt
```diff
@@ -74,7 +74,7 @@ interface Spec : TestListener {
 
    fun description(): Description = Description.fromSpecClass(this::class)
 
-   val assertionMode: AssertionMode? get() = null
```
LeoColman, line 77: I believe you added this recently, so there would be no problems to refactor this, right? It didn't go into a release version

kotlintest-core/src/commonMain/kotlin/io/kotlintest/core/specs/AbstractSpecDsl.kt
```diff
@@ -0,0 +1,106 @@
...
+   override fun testCaseOrder(): TestCaseOrder? = testCaseOrder
+   override fun isolationMode(): IsolationMode? = isolationMode
+   override fun assertionMode(): AssertionMode? = assertionMode
+
+   override fun listeners(): List<TestListener> = listeners
+   override fun extensions(): List<SpecLevelExtension> = extensions
+
+   override fun beforeTest(testCase: TestCase) {
+      super.beforeTest(testCase)
+      beforeTestFn(testCase)
```
LeoColman, line 36: Perhaps this is kind of dangerous. It wouldn't be possible to mix old syntax with new syntax this way (maybe we don't want to?), as any override could not call `super`

kotlintest-core/src/commonMain/kotlin/io/kotlintest/core/specs/AbstractSpecDsl.kt
```diff
@@ -0,0 +1,106 @@
...
+   fun set(assertionMode: AssertionMode) {
+      this.assertionMode = assertionMode
+   }
+
+   fun set(testCaseOrder: TestCaseOrder) {
+      this.testCaseOrder = testCaseOrder
+   }
+
+   fun tags(vararg tags: Tag) {
+      this.tags = tags.toSet()
```
LeoColman, line 76: Do we think that perhaps we want Tags, Listeners and Extensions to be called more than once, and this could be an increment instead of an asignment?
`this.tags += tags.toSet()`. I don't have a concrete use case for this in my mind yet, but I'm thinking that perhaps when we we want to compose different pieces of code (we don't support that yet, I don't remember if it's proposed), we might want to compose this too.

It wouldn't make a difference for our main use case with this DSL

build.gradle
sksamuel: Yes, I shouldn't have set it to 4.0.0 in the first place.

kotlintest-core/src/commonMain/kotlin/io/kotlintest/Spec.kt
sksamuel: Correct. I only added it for 4.0 anyway.

kotlintest-core/src/commonMain/kotlin/io/kotlintest/core/specs/AbstractSpecDsl.kt
sksamuel: An override could call super, but no one is going to override before test in both places.

kotlintest-core/src/commonMain/kotlin/io/kotlintest/core/specs/AbstractSpecDsl.kt
sksamuel: We could have tags that replaces it, and addTag, withTag or just tag to add ?

Issue #973, opened 2019-09-02
Add samples Generator


sksamuel:
Gen.samples(vararg xs :X) will take 1 or more Xs and then loop around providing those (starting from the beginning again once the end is reached).

LeoColman:
So
`Gen.samples(1, 2, 3, 4, 5)` would lead to `[1, 2, 3, 4, 5, 1, 2, 3, 4, ...]`?

sksamuel:
Yes

LeoColman:
Closed as per #983

PR #983, opened 2019-09-10
#973 Adds a samples type Gen which repeatably gives given samples

ashishkujoy:
Related to https://github.com/kotlintest/kotlintest/issues/973

### DIFF
--- a/kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
+++ b/kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
@@ -587,7 +587,7 @@ fun Gen.Companion.factors(k: Int): Gen<Int> = object : Gen<Int> {
 }
 
 /**
- * Returns a [Gen] which returns the sample values in same order as they are passed in, once all sample values are used
+ * Returns a [Gen] which returns the sample values in the same order as they are passed in, once all sample values are used
  * it repeats elements from start.
  */
 fun <T> Gen.Companion.samples(vararg sampleValues: T) = object : Gen<T> {
--- a/kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
+++ b/kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
@@ -585,3 +585,22 @@ fun Gen.Companion.factors(k: Int): Gen<Int> = object : Gen<Int> {
          .filter { k % it == 0 }
    }
 }
+
+fun <T:Any> Gen.Companion.samples(vararg sampleValues: T) = object : Gen<T> {
+    private fun nextNumberGenerator(): () -> T  {
+        var currentIndex = 0;
+        return {
+            val nextIndex = currentIndex % sampleValues.size
+            val nextValue = sampleValues[nextIndex]
+            currentIndex += 1
+            nextValue
+        }
+    }
+
+    override fun random(seed: Long?): Sequence<T> {
+        val a = nextNumberGenerator()
+        return generateSequence(a)
+    }
+
+    override fun constants(): Iterable<T> = sampleValues.asIterable()
+}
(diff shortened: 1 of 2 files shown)


LeoColman:
I believe this is related to an issue, right?

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
```diff
@@ -585,3 +585,26 @@ fun Gen.Companion.factors(k: Int): Gen<Int> = object : Gen<Int> {
          .filter { k % it == 0 }
    }
 }
+
+/**
+ * Returns a [[Gen]] which returns the sample values in same order as they are passed in, once all sample values are used
```
LeoColman:
```suggestion
 * Returns a [Gen] which returns the sample values in the same order they are passed in, once all sample values are used
```

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
```diff
@@ -585,3 +585,26 @@ fun Gen.Companion.factors(k: Int): Gen<Int> = object : Gen<Int> {
...
+ * it repeats elements from starting.
```
LeoColman:
```suggestion
 * it repeats elements from start.
```

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
```diff
@@ -585,3 +585,26 @@ fun Gen.Companion.factors(k: Int): Gen<Int> = object : Gen<Int> {
...
+ */
+fun <T:Any> Gen.Companion.samples(vararg sampleValues: T) = object : Gen<T> {
```
LeoColman: This should probably be `<T: Any?>`, right? We might want nullable values in the sample

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
```diff
@@ -585,3 +585,26 @@ fun Gen.Companion.factors(k: Int): Gen<Int> = object : Gen<Int> {
...
+    private fun nextNumberGenerator(): () -> T  {
+        var currentIndex = 0;
+        return {
```
LeoColman, line 596: Any reason this return statement opens a `{` instead of returning `nextValue` at the end?

kotlintest-assertions/src/jvmTest/kotlin/com/sksamuel/kotlintest/properties/GenTest.kt
```diff
@@ -602,6 +606,12 @@ class GenTest : WordSpec() {
         }
       }
     }
+
+    "Gen.samples(sampleValues)" should {
```
LeoColman, line 610: Could you please add some more scenarios? Such as verifying that the order is correct, and verifying that the constants are exactly the same as passed in.

```
val gen = Gen.samples(1, 2, 3, 4, 5)
gen.random().take(10).toList() shouldBe (1, 2, 3, 4, 5, 1, 2, 3, 4, 5)
```

LeoColman (changes requested):

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
ashishkujoy: That's because generateInfiniteSequence needs a lambda to generate the next element. That's the reason I am returning a lambda instead of the next element. I am changing its name to reflect that.

LeoColman: Oh, right

LeoColman:
I think this is good to go.
@sksamuel Wanna take a look?

LeoColman (approved):

Issue #937, opened 2019-08-07
Matchers for elements in a collection


LeoColman:
We have matchers for collections that checks if an element is in them, such as 

```
collection shouldContain element
```

What about a matcher that does the same thing, but on the point of view of the element?

```
element shouldBeIn collection
```

We already have exact instance check:
```
element shouldBeOneOf collection
```

PR #984, opened 2019-09-10
#937 Adds a shouldBeIn matcher

ashishkujoy:
This matcher checks if a element is in given collection or not by doing value check instead of reference check

### DIFF
--- a/kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/matchers/collections/matchers.kt
+++ b/kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/matchers/collections/matchers.kt
@@ -400,6 +400,26 @@ infix fun <T> Collection<T>.shouldContainAll(ts: Collection<T>) = this should co
 infix fun <T> Array<T>.shouldNotContainAll(ts: Collection<T>) = asList().shouldNotContainAll(ts)
 infix fun <T> Collection<T>.shouldNotContainAll(ts: Collection<T>) = this shouldNot containAll(ts)
 
+fun <T> Array<T>.shouldContainAnyOf(vararg ts: T) = asList().shouldContainAnyOf(ts)
+fun <T> Collection<T>.shouldContainAnyOf(vararg ts: T) = this should containAnyOf(ts.asList())
+fun <T> Array<T>.shouldNotContainAnyOf(vararg ts: T) = asList().shouldNotContainAnyOf(ts)
+fun <T> Collection<T>.shouldNotContainAnyOf(vararg ts: T) = this shouldNot containAnyOf(ts.asList())
+infix fun <T> Array<T>.shouldContainAnyOf(ts: Collection<T>) = asList().shouldContainAnyOf(ts)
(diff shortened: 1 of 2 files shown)


kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/matchers/collections/matchers.kt
```diff
@@ -491,6 +475,83 @@ fun <T> beOneOf(collection: Collection<T>) = object : Matcher<T> {
...
+ * test will pass.
+ *
+ * An empty collection will always fail. If you need to check for empty collection, use [Collection.shouldBeEmpty]
+ *
+ * @see [shouldBeOneOf]
+ * @see [shouldNotBeOneOf]
+ */
+fun <T> beIn(collection: Collection<T>) = object : Matcher<T> {
+  override fun test(value: T): MatcherResult {
+    if (collection.isEmpty()) throwEmptyCollectionError()
```
LeoColman, line 597: I'm thinking that at some point we will need to change this to `collection.verifyIsNotEmpty()`. This if is used at so many places xD

If you want to do it, feel free, but it's not necessary for this PR.

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/matchers/collections/matchers.kt
```diff
@@ -491,6 +475,83 @@ fun <T> beOneOf(collection: Collection<T>) = object : Matcher<T> {
...
+
+    val match = collection.any { it == value }
```
LeoColman: Wouldn't a `it in collection` work? This will probably optimize collections based on `hashcode`

LeoColman (approved):

LeoColman:
Great PR! Awesome documentation as well.

LeoColman:
I approved the code, but somethings we should consider:

- There should be a PR to add this to the documentation as well. I like to create 2 different PRs so that the docs don't get not-released content to it.
- I'm wondering if we should create an array overload in addition to List and Vararg

ashishkujoy:
> I approved the code, but somethings we should consider:
> 
>     * There should be a PR to add this to the documentation as well. I like to create 2 different PRs so that the docs don't get not-released content to it.
> 
>     * I'm wondering if we should create an array overload in addition to List and Vararg

@Kerooker 
Array overload will result in jvm signature clash. Which can be avoided using different name(with @jvmName) for overloaded method. [see here](https://discuss.kotlinlang.org/t/vararg-vs-array-parameter/4068/16).
Not sure of how important is this overload.

LeoColman:
The jvm clashes, but kotlin will force us to use a `*arr` to use it.


We have some prerrogatives to add the array overload, as matchers for
collection all have the overload.

Given it's very easy to implement, I don't see why not. I can do it when I
merge, if you want :)

LeoColman:
Ok, so we are now pending the documentation changes.
I got you covered ;)

Thanks a lot for this contribution! This is a very high quality PR!

Issue #947, opened 2019-08-18
Gen.file(dir)


sksamuel:
A generator that provides random files from a directory

fun Gen.Companion.file(dir) : Gen<File> ...
fun Gen.Companion.file(dir, recursive: Boolean) : Gen<File> ...

ashishkujoy:
@sksamuel  As of now all gens are in commonMain, where we don't have access to java modules.
Which means this implementation will need to be in jvmMain. But that would look odd.

sksamuel:
Not all gens are in common. Quite a few JVM specific ones are in jvmMain

PR #985, opened 2019-09-11
#947 Adds Gen.file which randomly gives files from a given directory.

ashishkujoy:
Not sure of how to write test for this without being dependent on external file (test resource).

### DIFF
--- a/kotlintest-assertions/src/jvmMain/kotlin/io/kotlintest/properties/jvmgens.kt
+++ b/kotlintest-assertions/src/jvmMain/kotlin/io/kotlintest/properties/jvmgens.kt
@@ -152,11 +152,7 @@ fun Gen.Companion.regex(regex: Regex) = regex(regex.pattern)
 /**
  * Returns a stream of values where each value is a randomly
  * chosen File object from given directory. If the Directory does not exist, an empty sequence will be returned instead.
-<<<<<<< HEAD
  * If recursive is true(default value is false) it gives files from inner directories as well recursively.
-=======
- * If recursive is true it gives files from inner directories as well recursively.
->>>>>>> #947 Adds unit test for Gen.file
  */
 fun Gen.Companion.file(directoryName: String, recursive: Boolean = false): Gen<File> = object : Gen<File> {
    override fun constants(): Iterable<File> = emptyList()
--- a/kotlintest-assertions/src/jvmMain/kotlin/io/kotlintest/properties/jvmgens.kt
+++ b/kotlintest-assertions/src/jvmMain/kotlin/io/kotlintest/properties/jvmgens.kt
@@ -2,16 +2,11 @@ package io.kotlintest.properties
 
 import java.io.File
 import java.math.BigInteger
-import java.time.Duration
-import java.time.LocalDate
-import java.time.LocalDateTime
-import java.time.LocalTime
-import java.time.Period
-import java.time.Year
+import java.time.*
 import java.time.temporal.ChronoUnit
 import java.time.temporal.TemporalQueries.localDate
 import java.time.temporal.TemporalQueries.localTime
-import java.util.UUID
+import java.util.*
 import kotlin.random.Random
 
 /**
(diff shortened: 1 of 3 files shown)


LeoColman:
I suppose that there's no harm on creating a `temp` file for this? But yes, it will be dependant.

But perhaps you can use any of kotlintest files?
For example, list all the files from the root directory, and then generate files from that directory and verify that all files were obtainable and that only files from the directory were obtained

kotlintest-assertions/src/jvmMain/kotlin/io/kotlintest/properties/jvmgens.kt
```diff
@@ -153,3 +148,25 @@ inline fun <reified T : Enum<T>> Gen.Companion.enum(): Gen<T> = object : Gen<T>
 
 fun Gen.Companion.regex(regex: String) = RegexpGen(regex)
 fun Gen.Companion.regex(regex: Regex) = regex(regex.pattern)
+
+/**
+ * Returns a stream of values where each value is a randomly
+ * chosen File object from given directory. Directory does not exists, an empty sequence will be return.
```
LeoColman:
```suggestion
 * chosen File object from given directory. If the directory does not exist, an empty sequence will be returned instead.
```

kotlintest-assertions/src/jvmMain/kotlin/io/kotlintest/properties/jvmgens.kt
```diff
@@ -153,3 +148,25 @@ inline fun <reified T : Enum<T>> Gen.Companion.enum(): Gen<T> = object : Gen<T>
...
+ * If recursive is true it gives files from inner directories as well recursively.
+ */
+fun Gen.Companion.file(directoryName: String, recursive: Boolean = false): Gen<File> = object : Gen<File> {
```
LeoColman, line 157: I wonder if this should be `Gen.file` or `Gen.files`. @sksamuel ?

kotlintest-assertions/src/jvmMain/kotlin/io/kotlintest/properties/jvmgens.kt
```diff
@@ -153,3 +148,25 @@ inline fun <reified T : Enum<T>> Gen.Companion.enum(): Gen<T> = object : Gen<T>
...
+   override fun constants(): Iterable<File> = emptyList()
+   override fun random(seed: Long?): Sequence<File> {
+      val fileTreeWalk = File(directoryName).walk()
+      return if (!recursive) {
```
LeoColman: I believe that this inverted condition is bad for code readability. Perhaps it's better to make it a `if (recursive)`?

kotlintest-assertions/src/jvmMain/kotlin/io/kotlintest/properties/jvmgens.kt
```diff
@@ -153,3 +148,25 @@ inline fun <reified T : Enum<T>> Gen.Companion.enum(): Gen<T> = object : Gen<T>
...
+         randomiseFiles(fileTreeWalk.maxDepth(1), seed)
+      } else randomiseFiles(fileTreeWalk, seed)
+   }
+
+   private fun randomiseFiles(files: Sequence<File>, seed: Long?): Sequence<File> {
+      val allFiles = files.toList()
+      if(allFiles.isEmpty()) return emptySequence()
+      val r = if (seed == null) Random.Default else Random(seed)
+      return generateInfiniteSequence { allFiles[r.nextInt(0, allFiles.size)] }
```
LeoColman: I believe you can use `allFiled.random(random)` instead, can't you?

LeoColman (changes requested):

LeoColman:
@sksamuel Did we change back for `random` to receive a `long` instead of a `kotlin.random`?

kotlintest-assertions/src/jvmMain/kotlin/io/kotlintest/properties/jvmgens.kt
sksamuel: I think either is fine.

Issue #851, opened 2019-06-21
Stats matchers


sksamuel:
list.shouldHaveMean(123.45)
list.shouldHaveStandardDeviation(55.123)
list.shouldHaveVariance(9.4)

alladywek:
I have a few moments regarding details of implementation:
1. what number types do you want to support?
2. what do you think about using BigDecimal as expected value? It gives more precision result and easier calculations
3. It would be great to have second optional parameter (precision: Int) for these methods where user can specify amount of digits after decimal point:
`list.shouldHaveMean(BigDecimal("0.667"), 3)`
`list.shouldHaveMean(BigDecimal("0.67"), 2)`

Problem with Double is that you can specify only 16-17 digits and other will be lost
```
fun main() {
  val double = 123456789.01234567890123456789
  println(double)                       //1.2345678901234567E8
  println(BigDecimal(double))           //123456789.01234567165374755859375
  println(BigDecimal(double.toString()))//123456789.01234567
}
```

sksamuel:
I think BigDecimal is good yes, but I think we should support doubles too. Obviously there can be precision errors, but we regularly test doubles everywhere else, so I think having both options is the best way. Perhaps with your precision parameter idea.

alladywek:
yep, I agree that both variants it is good idea, so lets summarize which methods we expect (for methods without explicit precision we can set up `precision=5` by default):
```kotlin
Collection<Double>.shouldHaveMean(mean: Double, precision: Int)
Collection<Double>.shouldHaveMean(mean: BigDecimal, precision: Int)
Collection<Double>.shouldHaveMean(mean: Double) 
Collection<Double>.shouldHaveMean(mean: BigDecimal)
DoubleArray.shouldHaveMean(mean: Double, precision: Int)
DoubleArray.shouldHaveMean(mean: BigDecimal, precision: Int)
DoubleArray.shouldHaveMean(mean: Double)
DoubleArray.shouldHaveMean(mean: BigDecimal)

Collection<Double>.shouldHaveStandardDeviation(mean: Double, precision: Int)
Collection<Double>.shouldHaveStandardDeviation(mean: BigDecimal, precision: Int)
Collection<Double>.shouldHaveStandardDeviation(mean: Double)
Collection<Double>.shouldHaveStandardDeviation(mean: BigDecimal)
DoubleArray.shouldHaveStandardDeviation(mean: Double, precision: Int)
[...]

PR #989, opened 2019-09-13
#851 - create stats matchers

alladywek:
I hope that it's enough and I will proceed with docs. I am not sure that version with value as Array instead of Collection makes much sense, but correct me if I am wrong.


sksamuel:
This is awesome work.

PR #995, opened 2019-09-14
[NEXT RELEASE] #851 add docs to matchers.md

alladywek:


LeoColman:
I'm wondering if we should add two versions of the "same" matcher to the documentation. Perhaps we can let the user discover that there's a default value set to 4? That way we can shorten the docs and avoid repetitions.

What do you think?

alladywek:
Do you prefer method signature with precision?

LeoColman:
Yes, I think it's better to keep only the method signature with precision.
```
 | `collection.shouldHaveMean(mean, precision)` | Asserts that collection has specific mean with specific precision | 
```

But we can discuss this further if you don't agree

alladywek:
I don't have objections regarding to left one signature, I think that maybe in this case we need change description, e.g.:

```
 | `collection.shouldHaveMean(mean, precision)` | Asserts that collection has mean with specific or default (4 digits after decimal point) precision |
```

What do you think?

LeoColman:
That's better :)

Issue #981, opened 2019-09-09
Possible error in shouldNotThrowAny


Zefz:
Should not `shouldNotThrowAny` return the value in the block instead of Unit?

e.g
```
var foo: Bar = shouldNotThrowAny {
            Bar()
        }
```

While `shouldNotThrowAnyUnit` should return an Unit. Maybe I am misunderstanding something, but currently these two functions behave the same?

sksamuel:
Yeah that's a bug, it should return T for () -> T.

sksamuel:
Fixed in master.

PR #996, opened 2019-09-14
#981 make shouldNotThrowAny return T instead of Unit

verkhovin:
It should be like this i believe
UPD Original issue is #981


sksamuel:
Thanks

Issue #943, opened 2019-08-09
New matcher: should be truthy


sksamuel:

LeoColman:
Do we have truthy types in Kotlin?
We can't evaluate arbitrary expressions as boolean

sksamuel:
At work we have "yes" "no" "1" "0" as boolean values in a database.
I thought it might be handy to support `"yes".shouldBeTruthy()`

LeoColman:
I see.

I'm just not sure where we should stop on this.

"true" and "false" seems correct as well.

gustavo-valvassori-codeminer42:
Are there any more conditions than theses below? (Always ignoring case)

| Condition | Acceptable values |
| --- | --- |
| shouldBeTruthy | `Yes`, `y`, `1`, `true` |
| shouldBeFalsy | `No`, `n`, `0`, `false` |

sksamuel:
I think that's acceptable.

LeoColman:
Make sure that it's case insensitive, so `YES` == `yes` == `Yes`

LeoColman:
But I'm afraid of somethings...

In JS, a date is truthy, for example. "37" is truthy

While `NaN` is falsy, `""` is falsy

How should we deal with these?

sksamuel:
We don't have to do what JS does.

sksamuel:
@gustavo-valvassori-codeminer42 the cases you listed are good.

PR #1001, opened 2019-09-17
#943 - add truthy/falsy matchers

alladywek:


kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/matchers/string/matchers.kt
```diff
@@ -281,3 +281,22 @@ fun beEqualIgnoringCase(other: String) = neverNullMatcher<String> { value ->
     "${value.show()} should not be equal ignoring case ${other.show()}"
   )
 }
+
+fun String?.shouldBeTruthy() = this should beTruthy()
+fun String?.shouldBeFalsy() = this should beFalsy()
+
+fun beTruthy() = object : Matcher<String?> {
+   override fun test(value: String?) = MatcherResult(
```
LeoColman, line 342: I think we should extract the array to be a single variable. 
This will avoid having to change different points of code

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/matchers/string/matchers.kt
```diff
@@ -281,3 +281,22 @@ fun beEqualIgnoringCase(other: String) = neverNullMatcher<String> { value ->
...
+      arrayOf("true", "yes", "y", "1").any { it.equals(value, ignoreCase = true) },
+      { """${value.show()} should be equal ignoring case one of values: ["true", "yes", "y", "1"]""" },
+      { """${value.show()} should not be equal ignoring case one of values: ["true", "yes", "y", "1"]""" }
+   )
+}
+
+fun beFalsy(): Matcher<String?> = object : Matcher<String?> {
+   override fun test(value: String?) = MatcherResult(
+      arrayOf("false", "no", "n", "0").any { it.equals(value, ignoreCase = true) },
```
LeoColman: Same thing as truthy method

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/matchers/string/matchers.kt
```diff
@@ -281,3 +281,22 @@ fun beEqualIgnoringCase(other: String) = neverNullMatcher<String> { value ->
     "${value.show()} should not be equal ignoring case ${other.show()}"
   )
 }
+
+fun String?.shouldBeTruthy() = this should beTruthy()
```
LeoColman, line 301: We should probably add some kdocs to this.
If you want, I can do it for you :D

LeoColman (changes requested):

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/matchers/string/matchers.kt
alladywek: Yeah, you are right, I missed that

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/matchers/string/matchers.kt
alladywek: No problem, I will do it

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/matchers/string/matchers.kt
LeoColman: I'd also suggest to use a list instead of an array, so we don't get tricked into a bad toString (might also suggest to add one test of the exception message)

LeoColman:
Hey, @alladywek , you've done a great work!
Do you want me to take it from here and merge?

LeoColman (approved):

LeoColman:
Oh, that was fast xD

LeoColman:
Thanks a lot for this contribution!

alladywek:
> Thanks a lot for this contribution!

Welcome, I uses Kotlintest in project, I really enjoy from using it, so thanks to you for that awesome framework =)

PR #1007, opened 2019-09-22
Implement Gen.char() - Closes #1005

### LINKED ISSUE #1005: Proposal: Gen.char()
It would be great if `Gen.char()` existed. Looks like it would be fairly trivial to implement, with `random()` of the `Gen<Char>` returning `Random.nextPrintableChar()`.

EarthCitizen:
Wrote one test to verify that seed value creates reproducible results. Writing a test that verifies different outcomes when no seed is present would be flakey with a good chance of failure.


kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
```diff
@@ -242,6 +242,17 @@ fun Gen.Companion.bool(): Gen<Boolean> = object : Gen<Boolean> {
...
 }
 
+/**
+ * Returns a stream of randomly-chosen Chars
+ */
+fun Gen.Companion.char(): Gen<Char> = object : Gen<Char> {
+  override fun constants(): Iterable<Char> = emptyList()
+  override fun random(seed: Long?): Sequence<Char> {
+    val r = if (seed == null) Random.Default else Random(seed)
+    return generateSequence { r.nextPrintableChar() }
```
LeoColman, line 252: I'm wondering if we want to limit ourselves to the "Printable Chars" only, or if we want to generate many other chars. What do you think?

kotlintest-assertions/src/jvmTest/kotlin/com/sksamuel/kotlintest/properties/CharGenTest.kt
```diff
@@ -0,0 +1,15 @@
+package com.sksamuel.kotlintest.properties
+
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.char
+import io.kotlintest.shouldBe
+import io.kotlintest.specs.FunSpec
+
+class CharGenTest : FunSpec({
```
LeoColman, line 8: Perhaps adding an exhaustive test that verifies all possible characters are obtained?

sksamuel:
Have two gens. Gen char and gen printable char

EarthCitizen:
I wanted to propose another idea, so I was trying to experiment with something, but I am having trouble figuring out how to get `kotlin.text.CharCategory` onto the classpath of `io.kotlintest.properties.gens`.

It is definitely in `kotlin-stdlib-1.3.50.jar`. I downloaded it and looked inside. Here is what I tried to get it onto the classpath:

In `kotlintest-assertions/build.gradle`:

- Added `implementation 'org.jetbrains.kotlin:kotlin-stdlib:1.3.50'` under dependencies in root, commonMain, and jvmMain
- Added `implementation kotlin("stdlib")` under dependencies in commonMain and jvmMain

In the Kotlin REPL, importing this class works fine:
```
>>> import kotlin.text.CharCategory
>>> CharCategory.UPPERCASE_LETTER.contains('T')
res38: kotlin.Boolean = true
```
Despite this, the build does not find `kotlin.text.CharCategory`. Any ideas?

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
EarthCitizen: I think that is a good idea. I am working on another proposal.

LeoColman:
Probably because you need to upgrade all the kotlin dependencies to use 1.3.50.
And that's ok!

sksamuel:
Stdlib should already be on the classpath. And for jvmMain you need
stdlib-jdk8

EarthCitizen:
@sksamuel I just noticed that CharCategory is not available for JS. I am new to Kotlin and I am slowly getting used to the complexity of multiple platforms. This was the cause of the error.

EarthCitizen:
I have another proposal for this, so I am going to close and submit another PR.

Issue #974, opened 2019-09-02
Weight choose Generators


sksamuel:
Add variations on Gen.choose, Gen.enum and so on that allows the weights to be specified.

LeoColman:
I think [this class](https://github.com/thomasnield/kotlin-statistics/blob/master/src/main/kotlin/org/nield/kotlinstatistics/Random.kt#L111)  from [Kotlin Statistics](https://github.com/thomasnield/kotlin-statistics) might help.

I had to implement a weighted random distribution recently, and I used that

sksamuel:
Wouldn't work cross platform but we could borrow implementations.

LeoColman:
The implementation itself seems very mpp-friendly

sksamuel:
Yes but they'd have to make the entire project mpp or we can copy n paste.

EarthCitizen:
An easy way of doing this might be like this:

Say we have this map of weights and gens:
```
mapOf(1 to gen1, 5 to gne5, 3 to gen3)
```
If then another list is created based on the weights:
```
listOf(1, 5, 5, 5, 5, 5, 3, 3, 3)
```
Then choose a random integer between 0 and listOfWeights.size - 1 for the lookup key.

Might not give a mathematically correct result, but I believe it would serve what is requested.

EarthCitizen:
Perhaps an easy way to do this would be the following:

Say we have a list of weights and gens:
```kotlin
val weightedGens = listOf(Pair(1, genA), Pair(3, genB), Pair(5, genC))
```
Then we create a list of indexes from above repeated by weight:
```kotlin
val repeatIndexes = listOf(0, 1, 1, 1, 2, 2, 2, 2, 2)
```
Then, we choose a random number between 0 and repeatIndexes size - 1 to select an element from repeatIndexes, and use that index to look up a gen from weightedGens.

This might not be mathematically correct, but I believe it would give the desired functionality. If it needed to work with negative numbers as well, there would need to be some offset trickery.

PR #1008, opened 2019-09-23
Rough POC for Custom Character Sets

EarthCitizen:
These changes are not intended for merge. They are simply for demonstrating a rough idea for being able to specify character ranges, and maintain current functionality. I believe it is reasonable for the project to provide predefined ranges in some way, though not in the way I did so here.

The second "test" (just for generating output), gives the following:

#### Requesting only ASCII (the default when no parameter given):
`Gen.char()`
```
%, o, ^, ^, d, l, :, y, 4, >, c, (, F, A, 1, K, x, p, O, %, R, M, V, v, &, ~, h, }, ?, 0, C, ., w, M, |, 0, P, e, w, Q, r, `, S,  , 4, u, w, C, z, 1
```

#### Requesting a mix of Arabic and Runic:
`Gen.char(CHR_ARABIC + CHR_RUNIC)`
```
ڶ٪ږظᛟرؚᚿٖؗة؍؀ؒᛃᛒۮᛡᛆزᚡٍةᛑۢᚣڽ؟سش؅ُٙᚤᛇجᚰٖᛀٳُئ؉پᛄᛋِزڷٞ
```

#### Requesting geometric shapes:
`Gen.char(CHR_GEOMETRIC_SHAPES)`
```
▷◝◛◡▽◿◅◓◐◻▦▮◗◶◢◗◅▪○◣◁◿◎▰◌◉▣◩□◱▹◒◫◵▥▫◞▰◓◮◉▩◻◫▽◞▱●◽◸
```

#### Requesting math symbols:
`Gen.char(CHR_MATH_OPERATORS)`
```
⋟∳⋛⊏∞∜⋵⋾∉⊝≡≍⋵⊵≀≳≯⋈⋥≖≴⊻∣⋺∓⋜≚≉⊐⊒⋻∁∆⊡⊍⋫⊊⋗⊯⊌⊿⊈∎≜∬∑∧∸⋫∯
```

#### Requesting Japanese symbols:
`Gen.char(CHR_KANGXI_RADICALS)`
```
⼕⼏⾸⼕⼌⾨⼺⽹⼃⼧⾕⾥⾃⾰⼕⿁⽎⽬⽟⽗⾪⼥⽐⽠⽯⼰⾆⼎⼠⽆⾧⽣⼸⾑⾟⾶⿁⽔⼸⾇⽮⼭⽀⼭⿔⼠⾨⾝⼨⽜
```

#### Requesting Cyrillic:
`Gen.char(CHR_CYRILLIC + CHR_CYRILLIC_SUPPLMNT)`
```
бԥЃӳԂӗЍԘԞԜԑдЫԂӂԃԫԍбԌӺіӯЖҽҢԜӄӿМҭԥөԞӅРѩԓҍԟԦЕҐԛХНѲӋђӈ
```

The character ranges included in this PR are not intended to be entirely accurate in their inclusions and exclusions. The hope here is to have a discussion as to whether this fits with the vision of the project and, if so, the best way to implement it.

Additionally, `Gen.string()` could also take advantage of this capability.

### DIFF
--- a/kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
+++ b/kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
@@ -22,11 +22,11 @@ fun Gen.Companion.string(minSize: Int = 0, maxSize: Int = 100): Gen<String> = ob
       "\nabc\n123\n",
       "\u006c\u0069b/\u0062\u002f\u006d\u0069nd/m\u0061x\u002e\u0070h\u0070")
 
-   override fun constants(): Iterable<String> = literals
+   override fun constants(): Iterable<String> = literals.filter { it.length in minSize..maxSize }
    override fun random(seed: Long?): Sequence<String> {
       val r = if (seed == null) Random.Default else Random(seed)
       return generateSequence {
-         r.nextPrintableString(minSize + r.nextInt(maxSize - minSize))
+         r.nextPrintableString(minSize + r.nextInt(maxSize - minSize + 1))
       }
    }
    override fun shrinker(): Shrinker<String>? = StringShrinker
--- a/kotlintest-assertions/src/jvmTest/kotlin/com/sksamuel/kotlintest/properties/StringGenTest.kt
+++ b/kotlintest-assertions/src/jvmTest/kotlin/com/sksamuel/kotlintest/properties/StringGenTest.kt
@@ -1,0 +1,34 @@
+package com.sksamuel.kotlintest.properties
+
+import io.kotlintest.inspectors.forAll
+import io.kotlintest.matchers.string.shouldHaveLength
+import io.kotlintest.matchers.string.shouldHaveMaxLength
+import io.kotlintest.matchers.string.shouldHaveMinLength
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.assertAll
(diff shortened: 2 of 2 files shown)


EarthCitizen:
After dumping all unicode characters to the terminal screen, I see one major problem. And that problem is that this technique works fine for small alphabets, but is a problem for writing systems such as Chinese, which has a LARGE number of characters. This would be WAY too much RAM to take up.

So, the best solution I am aware of is to use ranges. That also means that distribution will have to be even. I will work on a solution for this. This is similar to #974.

LeoColman:
I'll bring a point to this dicussion that might save you some work

Do we actually need that?
All letters from all possible alphabets and such?

Can we reduce this possibilities and let the user configure whih chars to generate, so the user can decide what kind of things they want?

EarthCitizen:
@Kerooker the idea was to have a *few* predefined ranges of letters (that part is optional). But, definitely, the goal is to make it possible to allow the user to specify their own characters to generate. The output from above is actually from MULTIPLE executions of generating from `Gen.char()`. So, it is not generating all possible characters from one gen request. It is multiple executions, each requesting a different character set. I will update the original comment to make that clearer.

But, overall, what I have done has a major flaw in using too much RAM. I will address this by addressing #974.

EarthCitizen:
@Kerooker I updated the original comment. Hopefully the functionality is clearer now.

PR #1009, opened 2019-09-24
Implement Gen.choose() for Weighted Distribution

### LINKED ISSUE #974: Weight choose Generators

EarthCitizen:
Port of the algorithm from Haskell QuickCheck for weighted distribution of choices.

Example:
`Gen.choose(1 to 'A', 3 to 'B')` will pick A 25% of the time and B 75% of the time.

One design choice that may or may not be what is wanted is that the function is implemented as:
`Gen.Companion.choose(a: Pair<Int, T>, b: Pair<Int, T>, vararg cs: Pair<Int, T>)`
The reason for this is because it does not make sense to have less than two choices. But, at the same time, one could give the same value for both of the first two arguments.

So, with that said, perhaps it is better just to have only one mandatory parameter even though it would not generate useful results?

Please let me know and I will change it to what ever is wanted.


EarthCitizen:
I just noticed in the test that I should should have used a linked map instead of a sorted map. Then the ordering makes sense. I will fix that.

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
```diff
@@ -490,6 +490,39 @@ fun Gen.Companion.choose(min: Long, max: Long): Gen<Long> {
   }
 }
 
+/**
+ * Returns a stream of values based on weights:
+ *
+ * <pre>Gen.choose(1 to 'A', 2 to 'B')</pre> will generate 'A' 33% of the time
```
LeoColman: I don't think IntelliJ likes `<pre>` or any other tag for Kotlin docs

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
```diff
@@ -490,6 +490,39 @@ fun Gen.Companion.choose(min: Long, max: Long): Gen<Long> {
...
+  require(weights.any { it > 0 }) { "At least one weight must be greater than zero"}
+  return object : Gen<T> {
+    // The algorithm for pick is a migration of
+    // the algorithm from Haskell QuickCheck
+    private tailrec fun pick(n: Int, l: Sequence<Pair<Int, T>>): T {
+      val (w, e) = l.first()
+      return if (n <= w) e
+      else pick(n - w, l.drop(1))
+    }
+    override fun constants(): Iterable<T> = emptyList()
```
LeoColman, line 517: I'm wondering if we should keep the constants as a list of every possible outcome, ignoring weight (but I'm not sure that's even useful)

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
```diff
@@ -490,6 +490,39 @@ fun Gen.Companion.choose(min: Long, max: Long): Gen<Long> {
...
+ * weight of zero is given.
+ */
+fun <T : Any> Gen.Companion.choose(a: Pair<Int, T>, b: Pair<Int, T>, vararg cs: Pair<Int, T>): Gen<T> {
+  val allPairs = listOf(a, b) + cs
+  val weights = allPairs.map { it.first }
+  require(weights.all { it >=0 }) { "Negative weights not allowed" }
+  require(weights.any { it > 0 }) { "At least one weight must be greater than zero"}
+  return object : Gen<T> {
+    // The algorithm for pick is a migration of
+    // the algorithm from Haskell QuickCheck
```
LeoColman, line 509: Perhaps we could add a permalink to that Haskell QuickCheck?

kotlintest-assertions/src/jvmTest/kotlin/com/sksamuel/kotlintest/properties/ChooseGenTest.kt
```diff
@@ -0,0 +1,49 @@
...
+         actualRatios[it] shouldBe (expectedRatios[it] plusOrMinus 0.02)
+      }
+    }
+  }
+
+  test("should not accept negative weights") {
+    shouldThrow<IllegalArgumentException> { Gen.choose(-1 to 'A', 1 to 'B') }
+  }
+
+  test("should not accept all zero weights") {
```
LeoColman, line 48: Perhaps adding a test that checks some zero weights (but not all)?

LeoColman (approved):
Great pull request!

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
EarthCitizen: So, remove the pre tags?

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
EarthCitizen: Will add the link.

kotlintest-assertions/src/jvmTest/kotlin/com/sksamuel/kotlintest/properties/ChooseGenTest.kt
EarthCitizen: Good idea. Will add it.

EarthCitizen:
@Kerooker What is your opinion on the 2 required parameters vs 1 required parameter?

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
LeoColman: Or try to `ctrl + Q` it and see if it renders correctly

LeoColman:
I mean, calling only `Gen.choose(1 to "b")` doesn't make much sense to me

sksamuel:
Yeah the way you did it with 3 params, last one being varargs makes sense.

EarthCitizen:
I like the 3 params, but the only downside being that someone can do `Gen.choose(1 to 'C', 1 to'C')`.

LeoColman:
> I like the 3 params, but the only downside being that someone can do `Gen.choose(1 to 'C', 2 to'C')`.

I don't think that matters. If you're concerned about it, perhaps we can add a warn that two of the elements are the same

ryandanielspmc:
@Kerooker Probably overkill.

sksamuel:
I don't think it's a big deal either way. The PR looks great.

EarthCitizen:
OK. Will will address @Kerooker 's points and fix something I saw in the tests I submitted.

kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
EarthCitizen: @Kerooker I am not 100% clear on when the constants need to be present. Can you explain that a bit?

EarthCitizen:
All issues resolved except for the question around whether constants are necessary.

sksamuel:
Constants aren't always needed. There's no edge cases to consider here so I
think its good to merge.

PR #1012, opened 2019-09-25
Fix max treated as exclusive on Gen.choose() for int and long - Fixes #1011

### LINKED ISSUE #1011: Gen.choose() for int and long - max is exclusive
`Gen.choose()` for int and long with a min and max are both treating max as exclusive.

EarthCitizen:
For `Gen.choose()` on the overloads with min and max, for int and long, the max was treated as exclusive and it should be inclusive. This is, of course, making the assumption that max is supposed to be inclusive. At least, that is the way the docs for the functions sound to me.

### DIFF
--- a/kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
+++ b/kotlintest-assertions/src/commonMain/kotlin/io/kotlintest/properties/gens.kt
@@ -468,7 +468,7 @@ fun Gen.Companion.choose(min: Int, max: Int): Gen<Int> {
     override fun constants(): Iterable<Int> = emptyList()
      override fun random(seed: Long?): Sequence<Int> {
         val r = if (seed == null) Random.Default else Random(seed)
-        return generateSequence { r.nextInt(min, max) }
+        return generateSequence { r.nextInt(min, max + 1) }
      }
 
     override fun shrinker() = ChooseShrinker(min, max)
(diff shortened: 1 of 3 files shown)


EarthCitizen:
As a side note, would it be better to have parameters for `Gen.int()` and `Gen.long()` that have default values?

Then you could just have something like this:
```kotlin
fun Gen.Companion.int(minInt: Int = Int.MIN_VALUE, maxInt: Int.MAX_VALUE) = object : Gen<Int> {
   val literals = listOf(minInt, maxInt, 0)
   override fun constants(): Iterable<Int> = literals
   override fun random(seed: Long?): Sequence<Int> {
      val r = if (seed == null) Random.Default else Random(seed)
      return generateSequence { r.nextInt(minInt, maxInt) }
   }

   override fun shrinker() = IntShrinker
}
```

EarthCitizen:
After submitting this PR, due to the Travis checks, I found the tests in `GenTest`. These tests expect the max to be exclusive, meanwhile, the tests for `Gen.string()` have the max being inclusive (if I am understanding correctly).

So, between the docs for `Gen.choose(min, max)` that seem in indicate inclusivity and `Gen.string()` which has an inclusive max, and the tests in `GenTest` that indicate exclusivity for max, which is correct?

In my opinion, it would make the most sense for min and max in all of these cases to be inclusive.

sksamuel:
I agree with you. Go inclusive.

EarthCitizen:
Everything should be working fine now.

LeoColman (approved):

LeoColman:
Thanks for the pull request!

PR #1019, opened 2019-09-29
Fixes #1017 - Incorrect comment for weighted Gen.choose()

### LINKED ISSUE #1017: Weighted Gen.choose() @throws comments incorrect
The comments for weighted `Gen.choose()` state:

> @throws IllegalArgumentException If any negative weight is given or any weight of zero is given.

This is incorrect. It should say

> or only weights of zero given

EarthCitizen:


LeoColman:
Good one, thanks!

PR #1023, opened 2019-10-01
Fix LocalDateTime test

LeoColman:
The test tries to cast a date to it's current day - 1, but that sometimes causes the day to be zero (only on the first day of every month).

This test was not correct, and this commit fixes it.


LeoColman:
@EarthCitizen, @reckter and @sksamuel FYI. Merging when green

PR #1030, opened 2019-10-02
Simplify and clean up Gen.choose() weighted ratios test

EarthCitizen:
This does not add any new features. I took a look at the original test I wrote and realized that it was a mess and way more complex than it needed to me. So, this is just a refactorization of that test.


LeoColman:
Good one

PR #1031, opened 2019-10-02
Feature Gen.char() - Closes #1005

### LINKED ISSUE #1005: Proposal: Gen.char()

EarthCitizen:
***PLEASE READ CAREFULLY BEFORE CONSIDERING MERGE:***

The idea behind this implementation of `Gen.char()` is to have a default of ASCII character generation, but to allow the user to specify `CharRange`s if they want to customize the generated values. The purpose of using `CharRange` instead of a collection of `Chars` is to perhaps make it easier to define, perhaps save a bit of memory, and perhaps a bit of performance.

If, for example, one wanted to use Chinese as the chars to generate, most likely you are going to be dealing with Unicode ranges:

https://en.wikipedia.org/wiki/CJK_Unified_Ideographs_(Unicode_block)

> List of CJK Unified Ideographs, part 1 of 4 (U+4E00..U+62FF)
> List of CJK Unified Ideographs, part 2 of 4 (U+6300..U+77FF)
> List of CJK Unified Ideographs, part 3 of 4 (U+7800..U+8CFF)
> List of CJK Unified Ideographs, part 4 of 4 (U+8D00..U+9FFF)

(not a complete set)

To have to convert all of these to complete collections, as opposed to ranges, is clumsy and inefficient in my opinion. Using ranges seems to fit well with how we think of characters as Unicode ranges.

```kotlin
val CJK_IDEOGRAPS = listOf(
    '\u4E00'..'\u62FF',
    '\u6300'..'\u77FF',
    '\u7800'..'\u8CFF',
    '\u8D00'..'\u9FFF'
)

Gen.char(CJK_IDEOGRAPHS).take(5)
```
```
娺 孿 屗 巳 天
```

When multiple ranges are given, the distribution is even across all characters via the new weighted `Gen.choose()` function. For example:
```kotlin
Gen.char('A'..'B', 'C'..'E')
```
is equivalent to doing this with the weighted `Gen.choose()`:
```kotlin
Gen.choose(1 to 'A', 1 to 'B', 1 to 'C', 1 to 'D', 1 to 'E')
```
In the actual implementation, the weights are the length of the `CharRange`, a range is chosen by weight, then a random `Char` is chosen from the `CharRange`. See the tests for proof that this works.

***ADDITIONAL NOTES***

I did not make `CharSets` public. I did not want to do so without a discussion on whether it is considered appropriate for the project to expose such an object.

And another thing, I did not include the space character in the default generated characters. Please let me know if this is not the desired behavior.


LeoColman (approved):

LeoColman:
I approve the way you did things. I really like this PR!

I would recommend you to extract some behavior to more methods to make it clearer.

Awesome!

EarthCitizen:
@Kerooker I will clean it up. Also, do you think it would be appropriate to expose a CharSets object to the users with pre-defined character sets, or should we leave it up to the users to define their own?

Another thing I should add should be a `Gen.char(<String>)` that picks from a string of chars.

LeoColman:
No, I don't believe we should allow the charsets

A String is a CharArray and I believe it's a CharList aswell. I think there should be a `Gen.char("string")` yeah

EarthCitizen:
@Kerooker Updated per feedback.

LeoColman:
I think this is ready to merge.
@sksamuel @EarthCitizen what do you think?

LeoColman:
Ok, you had your chance.

I'm merging this xD

EarthCitizen:
@Kerooker My apologies. I did not see the notification for your previous message. Yes, I am ready for this to merge.

Issue #1028, opened 2019-10-01
Add jsoup matchers


faogustavo:
skrape{it} is a Kotlin-based HTML/XML testing and web scraping library that can be used seamlessly in Spring-Boot, Ktor, Android or other Kotlin-JVM projects.

They want to add a lot of matchers and support the KotlinTest library (You can see all the expected matchers [in this issue](https://github.com/skrapeit/skrape.it/issues/9)). If you are comfortable with this, I can implement those matchers myself.

Most of them, are matchers from jsoup library and can be added to the project as an extension and be used by other projects, beside the skrape{it}.

LeoColman:
Sure, go ahead!
I believe we can add an assertions module for skrape.it.

The mantainers of the project already said that support can be mantained here (just like we do for Arrow and KTor).

I don't see a reason why not.

LeoColman:
If you need any help configuring the new module, let me know

Kerooker:
If you need any help configuring the new module, let me know

faogustavo:
Alright. Thanks @Kerooker

sksamuel:
I think kotest-assertions-skrape would be a good addition.

faogustavo:
@sksamuel as the lib just wants to add some assertations to jsoup classes, isn't it better to add a `kotest-assertations-jsoup`?

sksamuel:
Yes sorry.

LeoColman:
Oh, I didn't see that. I thought the matchers were for the lib itself. Guess I didn't read it properly =S

christian-draeger:
i think it would be cool to go for `kotest-assertations-skrapeit` because we would have the freedom to add specific assertion on skrape{it}'s [`Result` object](https://github.com/skrapeit/skrape.it/blob/master/src/main/kotlin/it/skrape/core/Result.kt) (beside the document) in the future. in this case i mean assertions regarding the http response (status code, content-type, headers etc)

sksamuel:
We could do both, no reason the jsoup specific ones should live in
a skrapeit module.

faogustavo:
Yeps, my plan is to start with the JSOUP assertations that are more generic and than add some others that are specific from the skrape.it;

I already started the JSOUP and hope to open a PR with some basic matchers soon.

faogustavo:
Hey guys.
Unfortunately, I'm a little bit busy these days and will not be able to add more matchers to jsoup. I just removed the WIP from the PR to merge the initial structure.
[...]

PR #1035, opened 2019-10-03
#1028: Add Jsoup Assertions

faogustavo:
References #1028
I'll be adding more matchers and docs in the next days.
When everything is ok, I'll remove the WIP and draft status.


kotest-assertions/kotest-assertions-jsoup/src/jvmTest/kotlin/io/kotest/assertions/jsoup/ResourceLoader.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import java.io.InputStreamReader
+
+object ResourceLoader {
+   fun getFileAsString(relPath: String): String {
+      val fileStream = javaClass.classLoader.getResourceAsStream(relPath)
+      val fileReader: InputStreamReader? = fileStream.reader()
+      return fileReader?.readText() ?: "// error: file not found or null"
+   }
+}
```
faogustavo: @Kerooker does the KotlinTest lib already have a way to do this? I put it here, but I don't think it's the best way to do.
If not, maybe we can add as a new feature from the testing core :man_shrugging:

sksamuel:
javaClass.resourceAsStream.bufferedReader.use { it.readText } or similar.
I'm on my phone.

sksamuel:
Excellent work @faogustavo

PR #1038, opened 2019-10-05
Change Json assertions to allow infix calls

LeoColman:


EarthCitizen (approved):

EarthCitizen:
Are `matchJsonResource` and `containJsonKey` supposed to be public?

BoxResin (approved):

LeoColman:
Yes, as one might want to use it with other matchers and such. The assertions are `Unit`, not matchers

Issue #970, opened 2019-09-02
Sentence generator


sksamuel:
Generates sentences by randomly putting together n words with a period.

LeoColman:
Probably important to also add punctuation, capital cases, names, numbers, apostrophe, etc

sksamuel:
Yep. Could be loads of variants.

faogustavo:
@sksamuel  @Kerooker have you seen the [FakeIt](https://github.com/moove-it/fakeit) lib? Maybe it will help with a lot of points that you want.

sksamuel:
Not sure this gen is that using, closing.

PR #1040, opened 2019-10-10
#970 Adds a sentence generator

ashishkujoy:


EarthCitizen:
I would move the tests to its own module called `GenSentenceTest`. This will help organize the tests.

EarthCitizen:
Can you give some examples of intended output? Is the intention of this to generate valid sentences? If so, it does not look like it will. The reason why I ask is because it looks like it will, with some constraints, basically generate the same output as what a bunch of random strings joined by spaces would be.

sksamuel:
qeqwe oisdo 128asd mkkjco qwieu

is a valid sentence in my eyes.

sksamuel:
Or if you want to get cleverer could use lorem ipsum

EarthCitizen:
It just seems that the result is about the same as generated strings with spaces included. If we are not concerned about the content of the sentence, I thin the same thing can be achieved with existing functions, even if we do add a `Gen.sentence()`. For example:

```kotlin
val space = Gen.from(listOf(' '))
val lower = Gen.from(('a'..'z').toList())
val upper = Gen.from(('A'..'Z').toList())
val nums = Gen.from(('0'..'9').toList())
Gen.choose(
  1 to space,
  20 to lower,
  20 to upper,
  10 to nums
).take(500)
```
This could be even simpler once `Gen.char()` is flushed out and merged.

sksamuel:
Yeah I think that's fine. It's just a useful gen to have to basically
generated strings with spaces.

EarthCitizen:
Yeah for sure. We just need to be sure that we are using existing building blocks and not re-inventing the wheel, so to speak.

@ashishkujoy What I gave is not intended to be exactly what is required. Just pointing out what we can build upon existing functionality.

EarthCitizen:
@ashishkujoy Another option (and probably better), would be something like this:

```kotlin
val lower = Gen.from(('a'..'z').toList())
val upper = Gen.from(('A'..'Z').toList())
val nums = Gen.from(('0'..'9').toList())
val wordLetters = Gen.choose(
  20 to lower,
  20 to upper,
  10 to nums
)
val word = <Gen that chooses from wordLetters X times>
Gen.list(word).next().joinToString(separator = " ")
```

LeoColman:
I think we should go for a Lorem Ipsum approach....

A sentence is composed of capital letters, periods, commas, flying commas.... Only characters help, but there is a lot of room for improvement

LeoColman:
Or sentences from books or whatever like that

PR #1041, opened 2019-10-11
Update set up with kts example

suhtai:
The set up on the reference documentation doesn't include an example for people using kts for the Gradle build files.


LeoColman:
Thanks!

PR #1043, opened 2019-10-12
Channel matchers

### LINKED ISSUE #885: Channel matchers
Channels are a great new feature of coroutines.
We should add the following matchers:

`channel.shouldReceiveWithin(duration)` // that the channel receives an element within the duration
`channel.shouldReceiveNoElementsWithin(duration)` // that the channel receives no elements within the given duration
`channel.shouldHaveSize(int)` // that the channel receives exactly int elements then closes
`channel.shouldReceiveAtLeast(int)` // that the channel receives at least int elements (can still be open after)
`channel.shouldReceiveAtMost(int)` // that the channel receives at most int elements before closing
`channel.shouldBeClosed()`
`channel.shouldBeOpen()`
`channel.shouldBeEmpty()`

junron:
Implemented matchers:

- [x] channel.shouldReceiveWithin(duration)
- [x] channel.shouldReceiveNoElementsWithin(duration)
- [x] channel.shouldHaveSize(int)
- [x] channel.shouldReceiveAtLeast(int)
- [x] channel.shouldReceiveAtMost(int)
- [X] channel.shouldBeClosed()
- [X] channel.shouldBeOpen()
- [X] channel.shouldBeEmpty()

`kotlinx-coroutines` dependency was added as `Channel` is a part of the `kotlinx` module. Relevant tests for each of the matchers were also added.

Closes #885


LeoColman:
Very nice PR!

I'm wondering if we should create a `kotlintest-assertions-coroutines` instead of adding to the default modules.

What do you think, @junron @sksamuel ?

sksamuel:
We already use coroutines as part of kotest proper, and it's available for
all platforms.
It should be fine to include in the main package.

LeoColman:
Then LGTM!

junron:
Hmm the tests are failing, similar to https://travis-ci.org/kotlintest/kotlintest/jobs/594391900. It seems to work fine when I run tests locally though.

PR #1054, opened 2019-10-25
#971 unique generator values

### LINKED ISSUE #971: Unique modifier for Generators
Filters out any value previously seen.

atmikev:
Fixes #971 

@sksamuel Not sure if this is a good idea or not, but wanted to throw it out there to see if I should keep refining it.

Also, it didn't look like it in the IDE but I think my indentation is off?

### DIFF
--- a/kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
+++ b/kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
@@ -1,6 +1,7 @@
 package io.kotest.properties
 
 import io.kotest.properties.shrinking.Shrinker
+import kotlinx.coroutines.yield
 
 /**
  * A Generator, or [Gen] is responsible for generating data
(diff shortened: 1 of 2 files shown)


atmikev:
Fixed indentation issues.

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
```diff
@@ -46,6 +47,10 @@ interface Gen<T> {
 
   fun shrinker(): Shrinker<T>? = null
 
+  fun uniqueRandoms(seed: Long? = null): Sequence<T> = sequence {
+     yieldAll(random(seed).distinct())
+  }
```
sksamuel: Can we make this an extension method on Generator and not defined inside the interface itself. It's fine to keep it in the same file just below the interface definition.

sksamuel (changes requested):

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
atmikev: Updated

sksamuel:
:+1:

PR #1062, opened 2019-10-30
[Documentation] - Update documentation on gens and Gen

3ygun:
- [x] Add missing documentation on `gens.kt` and `Gen.kt`
- [x] Fix documentation on edge cases
    - Why? (Using IntelliJ `ctrl` + `j`) Basically the old version attempts to link the full String:
    KotlintestDocumentationUpdates

### DIFF
--- a/kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
+++ b/kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
@@ -16,8 +16,8 @@ import io.kotest.properties.shrinking.Shrinker
  * The first are values that should always be included - those
  * edge cases values which are common sources of bugs. For
  * example, a generator for [Int]s should always include
- * values like zero, minus 1, positive 1, Integer.MAX_VALUE
- * and Integer.MIN_VALUE.
+ * values like zero, minus 1, positive 1, [Int.MAX_VALUE]
+ * and [Int.MIN_VALUE].
  *
  * Another example would be for a generator for enums. That
  * should include _all_ the values of the enum to ensure
@@ -44,6 +44,9 @@ interface Gen<T> {
    */
   fun random(seed: Long? = null): Sequence<T>
 
+  /**
+   * @return the [Shrinker] for this gen or `null`
+   */
   fun shrinker(): Shrinker<T>? = null
 
   /**
@@ -63,6 +66,9 @@ interface Gen<T> {
      }
   }
 
+  /**
+   * @return a new [Gen] by filtering this gen's output by the negated function [f]
+   */
   fun filterNot(f: (T) -> Boolean): Gen<T> = filter { !f(it) }
 
   /**
@@ -146,6 +152,20 @@ inline fun <T, reified U : T> Gen<T>.filterIsInstance(): Gen<U> {
    }
 }
 
+/**
+ * Create a infinite sequence of the given [generator]
+ *
+ * ```kotlin
+ * // Example
+ * generateInfiniteSequence { 42 }
+ *     .take(3)
+ *     .joinToString()
+ *     .let { println(it) }
+ * // Prints: 42, 42, 42
+ * ```
+ *
+ * @return a new [Sequence] of [T] which always returns the result of [generator]
+ */
 inline fun <T> generateInfiniteSequence(crossinline generator: () -> T): Sequence<T> =
     Sequence {
       object : Iterator<T> {
@@ -203,4 +223,7 @@ fun <T> Gen<T>.next(predicate: (T) -> Boolean = { true }, seed: Long?): T {
    return random(seed).first(predicate)
 }
 
+/**
+ * @return the result of calling [next] with the given [predicate] defaulting seed to `null`
+ */
 fun <T> Gen<T>.next(predicate: (T) -> Boolean = { true }): T = next(predicate, null)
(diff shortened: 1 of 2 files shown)


kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt
```diff
@@ -638,12 +646,23 @@ fun <K, V> Gen.Companion.map(gen: Gen<Pair<K, V>>, maxSize: Int = 100): Gen<Map<
  * Returns the next pseudorandom, uniformly distributed value
  * from the ASCII range 33-126.
  */
-fun Random.nextPrintableChar(): Char {
-  val low = 32
-  val high = 127
-  return (nextInt(high - low) + low).toChar()
-}
+fun Random.nextPrintableChar(): Char = nextInt(from = 32, until = 127).toChar()
```
3ygun, line 649: Avoid rewriting the internals of `nextInt`

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt
```diff
@@ -638,12 +646,23 @@ fun <K, V> Gen.Companion.map(gen: Gen<Pair<K, V>>, maxSize: Int = 100): Gen<Map<
...
+ * r.nextPrintableString(-10) // ""
+ * r.nextPrintableString(0)   // ""
+ * r.nextPrintableString(1)   // " ", "a", "Z", etc.
+ * r.nextPrintableString(5)   // "Ha Ha", "gens!", "[{-}]", etc.
+ * ```
+ *
+ * @param length of String
+ * @return a given length String of random printable Chars
+ */
 fun Random.nextPrintableString(length: Int): String {
```
3ygun, line 666: Tried to highlight in the example that it's not just words that are getting generated here

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt
```diff
@@ -213,7 +213,7 @@ fun Gen.Companion.numericFloats(
 /**
  * Returns a stream of values where each value is a randomly
  * chosen long. The values returned always include
- * the following edge cases: [Long.MIN_VALUE, Long.MAX_VALUE]
+ * the following edge cases: [[Long.MIN_VALUE], [Long.MAX_VALUE]]
```
3ygun, line 216: 🤔 should this have `0L` like the `Int` constants?

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
```diff
@@ -44,6 +44,9 @@ interface Gen<T> {
    */
   fun random(seed: Long? = null): Sequence<T>
 
+  /**
+   * @return the [Shrinker] for this gen or `null`
```
LeoColman, line 48: I believe this is a little bit redundant, isn't it?

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt
```diff
@@ -619,6 +622,11 @@ fun <K, V> Gen.Companion.map(genK: Gen<K>, genV: Gen<V>, maxSize: Int = 100): Ge
    }
 }
 
+/**
+ * @return a stream of values, where each value is a [Map] of [Pair]s from the given [gen]
+ *   the size of the [Map] is bounded between [0, [maxSize])
+ */
+@JvmOverloads
```
LeoColman, line 629: Why was this JvmOverloads added?

LeoColman:
I like this :+1: 

Thanks for the contribution, @3ygun

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt
3ygun: My thought was to match the definition on `map` where the `key` and `value` are separate `Gen`s 

https://github.com/kotlintest/kotlintest/blob/a3a9065bc2d69d6184f8e6e1ea22924c2c07af3b/kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt#L604-L610

And/or if the library wants to support Java having the condensed signature

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
3ygun: 1. It was the only un-commented function in `Gen` so was thinking it should have something just to appear as fully supported as everything else
1. Overall, it more quickly directs a user to look at `Shrinker` for usage/creation. This method differs IMO from the 2 proceeding methods `constants` and `random` as the return type is a Kotlintest type and not a Kotlin std lib type. Thus it's important for the reader to go understand the `Shinker` type in addition to the semantic method name. Then to avoid duplicating of the `Shrinker` documentation I just linked to it.
1. Also must admit it took me entirely too long to read shrink-er correctly. Thus I may simply have a grudge that redundancy would have helped

Could change the comment to have something like "Honey, I shrunk the `Gen`" as that might have helped with that last point 😆

3ygun:
@Kerooker no problem. #1056 took the easier change so felt I had to work a little bit

LeoColman:
Merged!

PR #1063, opened 2019-10-30
shouldBeInstanceOf and shouldBeTypeOf should return typecasted instance

swarmshine:
Both methods `shouldBeInstanceOf` and `shouldBeTypeOf` allow to use typecasted instance as argument of lambda.
In case of big nesting such approach leads to code that is hard to read. 

Suppose that we have a Parser:
```
class Parser {
    sealed class ParsingResult {
        class Error(val msg: String) : ParsingResult()
        class Model(val data: Data) : ParsingResult()
    }
    sealed class Data {
        class ValidData(val paylaod: Payload) : Data()
        class InvalidData() : Data()
    }
    sealed class Payload {
        class NumericPayload(val number: Int) : Payload()
        class TextPayload() : Payload()
    }
    fun parse(text: String): ParsingResult { ... }
}
```
Current version of assertions leads us to nesting:
```
    Parser().parse("model with valid data").shouldBeInstanceOf<ParsingResult.Model> {
        it.data.shouldBeInstanceOf<Data.ValidData> {
            it.paylaod.shouldBeInstanceOf<Payload.NumericPayload> {
                it.number.shouldBeLessThan(12)
            }
        }
    }
```
**Suggestion:**
Both method should return typecasted instance
* shouldBeInstanceOf
* shouldBeTypeOf

New version of assertions not required nesting:
```
val parsingResult = Parser().parse("model 12").shouldBeInstanceOf<ParsingResult.Model>()
val validData = parsingResult.data.shouldBeInstanceOf<Data.ValidData>()
val numericPayload = validData.paylaod.shouldBeInstanceOf<Payload.NumericPayload>()
numericPayload.number.shouldNotBeGreaterThan(10)
```
Now user can choose which way is more suitable to use.
This approach also opens another possibility to use `this` instead of `it` inside lambda: 
```
//old way
result.shouldBeInstanceOf<Some>{
  it.oneField.shouldBe(12)
  it.anotherField.shouldBe(42)
}
//new possible approach
result.shouldBeInstanceOf<Some>().apply{
  oneField.shouldBe(12)
  anotherField.shouldBe(42)
}
```
**Possible drawbacks:**
Return type of shouldBeInstanceOf changes from `Unit` to `T`.
This will break such test:
```
@Test
fun test() = Some().shouldBeInstanceOf<Some>()
```
with output error:
```
No tests were found
```
As JUnit expects test method to return void. 
I consider such cases are rare.
And benefit is worth the cost. 

Anyway upgrade from 3.3.x to 3.4.x requires code update from user side. 

Other test cases: 
```
@Test
fun test(){ 
  Some().shouldBeInstanceOf<Some>() 
}
```
will not be broken by this change.

**Alternative scenario:**
Kotlin Contracts was considered here: https://github.com/kotlintest/kotlintest/issues/601
Unforutately, as @sksamuel and @Kerooker  already mentioned earlier, Kotlin Contracts  does not support `is` condition. 
https://youtrack.jetbrains.com/issue/KT-28298

### DIFF
--- a/kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/types/matchers.kt
+++ b/kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/types/matchers.kt
@@ -28,11 +28,22 @@ import io.kotest.shouldNotBe
  * }
  *
  * ```
+ * ```
+ *
+ * val list: List<Int> = arraylistOf(1, 2, 3)
+ *
+ * // arrayList is typecasted to ArrayList<Int>
+ * val arrayList = list.shouldBeInstanceOf<ArrayList<Int>>()
+ *
+ * ```
+ * @param block Lambda that receives typecasted instance as argument for further assertions.
+ * @return The typecasted instance
  */
-inline fun <reified T : Any> Any?.shouldBeInstanceOf(block: (T) -> Unit = { }) {
+inline fun <reified T : Any> Any?.shouldBeInstanceOf(block: (T) -> Unit = { }): T {
   val matcher = beInstanceOf<T>()
   this shouldBe matcher
   block(this as T)
+  return this
 }
 
 /**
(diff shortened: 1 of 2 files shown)


LeoColman:
I like this.
Thanks for the contribution!

PR #1069, opened 2019-11-04
[#1060] Added matcher for Iterator

### LINKED ISSUE #1060: Add matcher shouldBeEmpy on Iterable/Iterator

atmikev:
Closes #1060 

Added a matcher for Iterator that checks value of `hasNext()` to determine if the iterator is empty or not.

### DIFF
--- a/kotest-assertions/src/commonMain/kotlin/io/kotest/matchers/iterator/IteratorMatchers.kt
+++ b/kotest-assertions/src/commonMain/kotlin/io/kotest/matchers/iterator/IteratorMatchers.kt
@@ -1,0 +1,19 @@
+package io.kotest.matchers.iterator
+
+import io.kotest.Matcher
+import io.kotest.MatcherResult
+import io.kotest.assertions.stringRepr
+import io.kotest.should
+import io.kotest.shouldNot
+
+fun <T> Iterator<T>.shouldBeEmpty() = this should beEmpty()
+fun <T> Iterator<T>.shouldNotBeEmpty() = this shouldNot beEmpty()
+fun <T> beEmpty(): Matcher<Iterator<T>> = object : Matcher<Iterator<T>> {
+   override fun test(value: Iterator<T>): MatcherResult {
+      return MatcherResult(
+         !value.hasNext(),
+         "Iterable should be empty but next value is ${stringRepr(value.next())}",
+         "Iterable should not be empty"
+      )
+   }
+}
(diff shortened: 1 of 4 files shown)


kotest-assertions/src/commonMain/kotlin/io/kotest/matchers/iterator/IteratorMatchers.kt
```diff
@@ -0,0 +1,19 @@
...
+import io.kotest.should
+import io.kotest.shouldNot
+
+fun <T> Iterator<T>.shouldBeEmpty() = this should beEmpty()
+fun <T> Iterator<T>.shouldNotBeEmpty() = this shouldNot beEmpty()
+fun <T> beEmpty(): Matcher<Iterator<T>> = object : Matcher<Iterator<T>> {
+   override fun test(value: Iterator<T>): MatcherResult {
+      return MatcherResult(
+         !value.hasNext(),
+         "Iterable should be empty but next value is ${stringRepr(value.next())}",
```
atmikev: @sksamuel There is a potential here for a `NoSuchElementException` to be thrown if `hasNext()` is true but `value.next()` has no value. It feels like that's a misconfigured iterator, but since its feasible, any thoughts on what to do?

LeoColman: I think that by `Iterator` specification, it's impossible by contract to have a `hasNext` return true but no `next`. I think we ignore this kind of scenario

kotest-assertions/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/iterator/IteratorMatchersTest.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import io.kotest.matchers.iterator.shouldBeEmpty
+import io.kotest.matchers.iterator.shouldNotBeEmpty
+import io.kotest.specs.WordSpec
+
+class IteratorMatchersTest: WordSpec() {
+
+   init {
+
+      "shouldBeEmpty" should {
```
LeoColman, line 11: I would suggest that instead of testing with your own implementation of `Iterator`, use ones available.

Something like `listOf<Int>().iterator().shouldBeEmpty()`. Your implementation of iterator, although tests what we want, is incorrect.

If we change the implementation in the future, this test will crash, and I don't think we want that.

What do you think?

atmikev: Sounds good. I keep forgetting some of these are already defined :-/

atmikev:
@Kerooker updated. Let me know if you have any more comments or see potential improvements!

Also, do we want a matcher for `Iterable` too? or is the one for `Iterator` enough to cover the necessary use cases?

LeoColman:
Seems good to me!

Thanks for your contribution!

PR #1073, opened 2019-11-06
Add documentation to Pitest

### LINKED ISSUE #1072: Add documentation for Pitest integration
In issue #680 we added support for Pitest.

However, there is no documentation on how to use it. We probably should have it.

LeoColman:
Note: Although at this point in time we're changing names to Kotest, this still refers to old code and thus the KotlinTest name was kept.
When we switch to Kotest we'll need to change all documentation to Kotest as well.

Closes #1072


LeoColman:
Merge if you're comfortable.

LeoColman:
@sksamuel 
I'll be merging this. If you think we should make any changes, please let me know and I can proceed to do it.

PR #1074, opened 2019-11-06
#1065 Add matchers for Regex

ashishkujoy:

PR #1080, opened 2019-11-12
[NEXT RELEASE] Adds documentation for regex matchers

ashishkujoy:
Regex Matchers were implement in #1074 , this PR adds documentation for the same.


ashishkujoy:
Hi @sksamuel, any idea by when we will merge this pull request

LeoColman:
@ashishkujoy 
The matchers aren't released yet, so we can only merge documentation when they're released as not to confuse current/future users

LeoColman:
@ashishkujoy 
I think this PR contains some commits that shouldn't be here.
Would you mind to rebase this without those commits?

ashishkujoy:
> @ashishkujoy
> I think this PR contains some commits that shouldn't be here.
> Would you mind to rebase this without those commits?

Done

PR #1087, opened 2019-11-22
Implement Gen.string() with Gen.char() - Closes #1078

### LINKED ISSUE #1078: Gen.string() Should Use the New Gen.char()
`Gen.string()` should be refactored to use the new `Gen.char()`.

EarthCitizen:
Implement `Gen.string()` using `Gen.char()`.

- Rewrote `Gen.string()` completely
- `StringShrinker` also had to be rewritten  to obey `minSize`
- Rewriting these broke a lot of tests that were expecting certain output
- Refactored / organized tests as fixes were done

### DIFF
--- a/kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt
+++ b/kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt
@@ -14,7 +14,7 @@ import kotlin.random.nextInt
 import kotlin.random.nextLong
 
 /**
- * Returns a stream of values where each value is a random string base.
+ * Returns a stream of values where each value is a random string.
  *
  */
 @JvmOverloads
--- a/kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt
+++ b/kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt
@@ -14,31 +14,21 @@ import kotlin.random.nextInt
 import kotlin.random.nextLong
 
 /**
- * Returns a stream of values where each value is a random
- * printed string.
+ * Returns a stream of values where each value is a random string base.
  *
- * The constant values are:
- * The empty string
- * A line separator
- * Multi-line string
- * a UTF8 string.
  */
 @JvmOverloads
-fun Gen.Companion.string(minSize: Int = 0, maxSize: Int = 100): Gen<String> = object : Gen<String> {
+fun Gen.Companion.string(minSize: Int = 0, maxSize: Int = 100, genChar: Gen<Char> = Gen.char()): Gen<String> = object : Gen<String> {
    init {
       require(minSize >= 0) { "minSize must be >= 0" }
       require(maxSize >= minSize) { "maxSize must be >= minSize"}
    }
-   private val genChar = Gen.char(CharSets.PRINTABLE_BASIC_LATIN)
+   private val padChar = genChar.take(1).first()
    private val genInt =
       if (minSize == maxSize) Gen.constant(minSize)
       else Gen.choose(minSize, maxSize)
-   private val literals = listOf("",
-      "\n",
-      "\nabc\n123\n",
-      "\u006c\u0069b/\u0062\u002f\u006d\u0069nd/m\u0061x\u002e\u0070h\u0070")
 
-   override fun constants(): Iterable<String> = literals.filter { it.length in minSize..maxSize }
+   override fun constants(): Iterable<String> = emptyList()
    override fun random(seed: Long?): Sequence<String> {
       return generateSequence {
          val size = genInt.next(seed = seed)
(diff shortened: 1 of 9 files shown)


LeoColman (approved):

LeoColman:
I'd suggest to add more tests to this class. We have a small test suite, but that's not necessary for this refactor.

EarthCitizen:
> I'd suggest to add more tests to this class. We have a small test suite, but that's not necessary for this refactor.

@Kerooker 

Any type of test in particular that stands out to you as missing?

LeoColman:
I mean, the tests on this gen only tests that the size is respected. I think that testing the generated characters would be cool

EarthCitizen:
@Kerooker Finally this is ready! Please review.

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt
```diff
@@ -14,39 +14,29 @@ import kotlin.random.nextInt
...
    private val genInt =
       if (minSize == maxSize) Gen.constant(minSize)
       else Gen.choose(minSize, maxSize)
-   private val literals = listOf("",
-      "\n",
-      "\nabc\n123\n",
-      "\u006c\u0069b/\u0062\u002f\u006d\u0069nd/m\u0061x\u002e\u0070h\u0070")
 
-   override fun constants(): Iterable<String> = literals.filter { it.length in minSize..maxSize }
+   override fun constants(): Iterable<String> = emptyList()
```
LeoColman, line 36: I don't agree with constant being an empty list.

At least a `""` should be constant if minSize is 0.
I would discuss if things like `\n`, `"😣"` or `" "`, but they shouldn't be as `Gen` might not generate them.

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/shrinking/StringShrinker.kt
```diff
@@ -1,14 +1,22 @@
...
-    0 -> emptyList()
-    1 -> listOf("", "a")
-    else -> {
-      val first = failure.take(failure.length / 2 + failure.length % 2)
-      val second = failure.takeLast(failure.length / 2)
-      // always include empty string as the best io.kotest.properties.shrinking.shrink
-     listOf("", first, first.padEnd(failure.length, 'a'), second, second.padStart(failure.length, 'a'))
-    }
-  }
+class StringShrinker(private val minSize: Int, private val padChar: Char) : Shrinker<String> {
```
LeoColman, line 3: If anyone uses `StringShrinker`, this code will break theirs. Do we really want to do that?

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/gens.kt
EarthCitizen: What if we create a min and max string when the object is created based on the `Gen.char()` given, and those will be the constants?

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/shrinking/StringShrinker.kt
EarthCitizen: I am not convinced it should even be public. What is the use case where someone needs access to it? To me it seems like simply having access to a customizable `Gen<String>` via `Gen.string()` would be enough for almost any use case. In in mind, if someone is doing something that `Gen.string()`, via the customizable parameters, cannot do then a generic `StringShrinker` would not work for them anyway. I would love to make this class private, and move the tests to `GenStringTests`.

As far as backwards compatibility, I had 4.0 in mind for this code.

@sksamuel What are your thoughts?

sksamuel: I've not been following this thread as I've been super busy. What's the gist and how does it fit in with what I was saying in #kotlintest-devs ?

EarthCitizen: @sksamuel 

Overall, for this PR, `Gen.string()` has been rewritten to allow passing in a `Gen<Char>`. This also has implications for the `StringShrinker`, min size, which is what was discussed in #kotlintest-devs.

`StringShrinker` has been changed to respect the min size of a generated `String` and the pad char comes from the passed-in `Gen<Char>`. Some constructor parameters were added to enable shrinking to obey the rules of the content that the `Gen<String>` creates. I have broken backwards compatibility here by changing `StringShrinker` from an object to a class and requiring constructor parameters there.

My opinion is that `StringShrinker` is not really useful outside the scope of our implementation of `Gen.string()`, that if someone has a custom `Gen<String>` that can do something that `Gen.string()` cannot (we now have custom size and custom content), then `StringShrinker` would not work for them anyway, as they would need shrinking that is specific to the rules of their custom `Gen<String>`.

Therefore, I am proposing that for 4.0, that we make `StringShrinker` private.

sksamuel: I agree.

EarthCitizen: Cool thanks. I will make the changes in this PR.

sksamuel: Could we roll out these changes to the new property test code ?

sksamuel:
Closing this in favour of #1297

PR #1092, opened 2019-11-25
Improve and document ComparableMatchers

### LINKED ISSUE #668: Comparator matchers
We should support isLessThan, isGreaterThan on any type which supports comparable.

LeoColman:
This commit aims to improve some of the available comparable matchers by adding extension infix functions.

 The new and old functions were documented to improve user's understandability of functions instead of having to navigate through the source code (CTRL + Q in IntelliJ).

 On the `shouldBeEqualComparingTo` function, I chose this name so it is not confusing to `shouldBe` and not ambiguous to `shouldCompareTo`. Although the name is a little big, I don't believe it will be a problem.

 There are no more matchers to be added to `compareTo`. All possible behavior for the compareTo matcher were added.

 Closes #668


LeoColman:
@sksamuel 
I'm merging this. If you have any issues we can change anything.

PR #1101, opened 2019-11-27
Add Gen for lines from a file

LeoColman:


LeoColman:
How to make this exhaustive? What if I want all lines from the file?

sksamuel:
Don't worry about exhaustive. I haven't fully worked out how it will work out.

LeoColman:
Maybe gens could start having a `all` method, which would generate a sequence that would contain all possible values (this may not be applicable to all gens, and that's ok)

EarthCitizen:
I have seen a lot of discussion on exhaustivity. One thought I have is that perhaps instead of thinking about how to make `Gen`s exhaustive, maybe what we really want is to have two concepts:

1. `Gen`, which we already have
2. `Series`, which we don't have

A `Series` would always be exhaustive. Instead of instead of trying to fit multiple kinds of functionality into a `Gen` (non-exhaustive randomness and exhaustivity), we let each type focus on doing one thing and doing it well. In the case of a `Series`, there would be no need for randomness.

I became familiar with the idea of a `Series` from this Haskell library:

http://hackage.haskell.org/package/smallcheck

I could see having an abstraction above `Gen` and `Series`, and functions that currently only take `Gen` could take either one. Perhaps this would be best as an ADT (sealed class).

Additionally, both of these would behave the same way in relation to the discussed size parameter:

https://github.com/kotlintest/kotlintest/issues/1095

sksamuel:
I am working on the next generation (no pun intended) of property testing
in Kotest. I very much like your idea of series.

kotest-assertions/src/jvmMain/kotlin/io/kotest/properties/jvmgens.kt
```diff
@@ -41,6 +47,29 @@ fun Gen.Companion.file(): Gen<File> = object : Gen<File> {
...
+/**
+ * Returns a stream of values where each value is a random line from [file]
+ */
+fun Gen.Companion.lines(file: File) = lines(file.inputStream())
+
+/**
+ * Returns a stream of values where each value is a random line from [fileStream]
+ *
+ */
+fun Gen.Companion.lines(fileStream: InputStream) = object : Gen<String> {
```
EarthCitizen, line 59: I would add an overload for `File`.

LeoColman:
I liked the `Series` idea, and also the nomenclature seems very fitting

kotest-assertions/src/jvmMain/kotlin/io/kotest/properties/jvmgens.kt
```diff
@@ -41,6 +47,29 @@ fun Gen.Companion.file(): Gen<File> = object : Gen<File> {
...
+   
+   private val content = fileStream.bufferedReader().readLines()
```
EarthCitizen, line 61: I don't believe there is an easy workaround, but it is unfortunate that all of the lines are read into memory. This will not be an issue 99% of the time, but could be an issue on large files.

kotest-assertions/src/jvmTest/kotlin/com/sksamuel/kotest/properties/GenLinesTest.kt
```diff
@@ -0,0 +1,33 @@
...
+            }
+            lines.shouldContainExactlyInAnyOrder("line0", "line1", "line2", "line3")
+        }
+    
+        should("Generate random different values and not sequenced") {
+            val inputStream = List(10000) { "$it" }.joinToString(separator = "\n").byteInputStream()
+            val lines = mutableListOf<String>()
+            Gen.lines(inputStream).assertAll(10000) { line ->
+                lines += line
+            }
```
EarthCitizen, line 29: Is there a reason that this has to be `assertAll`, or could it just be `val lines = Gen.lines(inputStream).take(10000)`? Same question applies to the test above as well.

kotest-assertions/src/jvmMain/kotlin/io/kotest/properties/jvmgens.kt
LeoColman: There is one, right above

EarthCitizen: I totally missed that! LOL

sksamuel:
Similar functionality has been merged in.

Issue #1102, opened 2019-11-28
Provide a private method in Gen to get Random for a given seed


ashishkujoy:
In various gens for  `Gen.random(seed)` we have repetition of 
`val r = if (seed == null) Random.Default else Random(seed)`
I think it makes sense to provide that as a private function in Gen

ashishkujoy:
Closing as this is implemented in #1103

PR #1103, opened 2019-11-28
#1102 Add a method in gen to get a Random for a given optional seed

ashishkujoy:


kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
```diff
@@ -137,6 +138,10 @@ interface Gen<T> {
       override fun random(seed: Long?): Sequence<T> = outer.random(seed) + gen.random(seed)
     }
   }
+
+  fun getRandomFor(aSeed:Long?):Random {
```
LeoColman: I don't like this as a public function and I don't like this as part of a `Gen`.

How about extracting it to outside of this interface (you can keep it in the same file) and making it internal?

LeoColman (changes requested):

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
ashishkujoy: @Kerooker Thanks  for feedback, i will make necessary changes

LeoColman:
Seems good to me.

Thanks @ashishkujoy !

PR #1117, opened 2019-12-10
#1036 - Fix duplicates issue when asserting containsExactlyInAnyOrder

### LINKED ISSUE #1036: Possible error in containExactlyInAnyOrder
In the scenarios below, the last assertion (1,1,1) passes when I would expect it to fail for not matching on 2 or 3.  Is my assumption correct, or am I misunderstanding the scope of this matcher?

```
val actual = listOf(1, 2, 3)
actual.shouldContainExactlyInAnyOrder(3, 2, 1) //Correctly passes
actual.shouldContainExactlyInAnyOrder(4, 3, 2, 1) //Correctly fails
actual.shouldContainExactlyInAnyOrder(2, 1) //Correctly fails
        
actual.shouldContainExactlyInAnyOrder(1, 1, 1) //Incorrectly passes
```

If this is indeed a bug, may I work on the PR?  I think I have a decent idea of the problem. 
 Thanks.

AdamDrennan:
Fixes [#1036](https://github.com/kotlintest/kotlintest/issues/1036)


LeoColman:
Thank you very much for this fix!

AdamDrennan:
Thanks for letting me support a great tool!

PR #1124, opened 2019-12-17
Adds matchers for java.sql.Timestamp

ashishkujoy:
Matchers added so far

1. shouldBeBefore
2. shouldNotBeBefore
3. shouldBeAfter
4. shouldNotBeAfter

### DIFF
--- a/kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/date/timestamp.kt
+++ b/kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/date/timestamp.kt
@@ -1,0 +1,45 @@
+package io.kotest.matchers.date
+
+import io.kotest.Matcher
+import io.kotest.MatcherResult
+import io.kotest.should
+import io.kotest.shouldNot
+import java.sql.Timestamp
+
+fun sameTimestamp(timestamp: Timestamp) = object: Matcher<Timestamp> {
+   override fun test(value: Timestamp): MatcherResult {
+      return MatcherResult(
+         timestamp.equals(value),
+         {"Expected $timestamp to be equal $value, but its not."},
+         {"$timestamp expected to be equal $value, but it is equal"}
+      )
+   }
+}
+
+fun afterTimestamp(timestamp: Timestamp) = object: Matcher<Timestamp>  {
+   override fun test(value: Timestamp): MatcherResult {
+      return MatcherResult(
+         value.after(timestamp),
+         {"Expected $timestamp to be equal $value, but its not."},
+         {"$timestamp expected to be equal $value, but it is equal"}
+      )
+   }
+}
+
+fun beforeTimestamp(timestamp: Timestamp) = object: Matcher<Timestamp>  {
+   override fun test(value: Timestamp): MatcherResult {
+      return MatcherResult(
+         value.before(timestamp),
+         {"Expected $timestamp to be equal $value, but its not."},
+         {"$timestamp expected to be equal $value, but it is equal"}
+      )
+   }
+}
+
+infix fun Timestamp.shouldBe(timestamp: Timestamp) = this should sameTimestamp(timestamp)
+
+infix fun Timestamp.shouldNotBe(timestamp: Timestamp) = this shouldNot sameTimestamp(timestamp)
+
+infix fun Timestamp.shouldBeAfter(timestamp: Timestamp) = this should afterTimestamp(timestamp)
+
+infix fun Timestamp.shouldBeBefore(timestamp: Timestamp) = this should beforeTimestamp(timestamp)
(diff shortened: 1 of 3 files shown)


kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/date/timestamp.kt
```diff
@@ -0,0 +1,74 @@
...
+import io.kotest.MatcherResult
+import io.kotest.should
+import io.kotest.shouldNot
+import java.sql.Timestamp
+
+fun sameTimestamp(timestamp: Timestamp) = object: Matcher<Timestamp> {
+   override fun test(value: Timestamp): MatcherResult {
+      return MatcherResult(
+         timestamp.equals(value),
+         {"Expected $timestamp to be equal $value, but its not."},
```
LeoColman: nit:
```suggestion
         {"Expected $timestamp to be equal $value, but it's not."},
```

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/date/timestamp.kt
```diff
@@ -0,0 +1,74 @@
...
+         {"$timestamp expected to be equal $value, but it is equal"}
+      )
+   }
+}
+
+fun afterTimestamp(timestamp: Timestamp) = object: Matcher<Timestamp>  {
+   override fun test(value: Timestamp): MatcherResult {
+      return MatcherResult(
+         value.after(timestamp),
+         {"Expected $timestamp to be after $value, but its not."},
```
LeoColman: nit: 
```suggestion
         {"Expected $timestamp to be after $value, but it's not."},
```

sksamuel:
Quick work @ashishkujoy

LeoColman:
@ashishkujoy 
Would you mind creating a new PR updating the documentation as well?
This one is great!

LeoColman (approved):

ashishkujoy:
> @ashishkujoy
> Would you mind creating a new PR updating the documentation as well?
> This one is great!

Sure

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/date/timestamp.kt
```diff
@@ -0,0 +1,74 @@
...
+      )
+   }
+}
+
+
+/**
+ * Assert that [Timestamp] is equal to [anotherTimestamp].
+ * @see [shouldNotBe]
+ * */
+infix fun Timestamp.shouldBe(anotherTimestamp: Timestamp) = this should sameTimestamp(anotherTimestamp)
```
LeoColman: Ok, on a further review, I believe we don't want a `shouldBe` and a `shouldNotBe` for this guy.

We already have the default `shouldBe` matcher that compares using `==`, so this is redundant and unnecessary.

Could you please take them out of this commit and from the documentation commit?

LeoColman (changes requested):

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/date/timestamp.kt
ashishkujoy: One thing here is error message which is specific to timestamp when use Timestamp.shouldBe while in the regular one error message is generic. If we don't want specific error then i can surely remove shouldBe and shouldNotBe from Timestamp as well as from Instant add under [PR](https://github.com/kotlintest/kotlintest/issues/1127)

LeoColman: The default should be message is `expected: $expected but was: $actual`
I don't think the message difference has enough relevance to justify this change

What do you think?

ashishkujoy: Yes now i also think we don't need shouldBe and shouldNotBe i will remove those methods.

ashishkujoy: Done

kotest-assertions/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/timestampTest.kt
```diff
@@ -7,17 +7,6 @@ import java.time.Instant
...
-         val nowInstance = Instant.now()
-         Timestamp.from(nowInstance) shouldBe Timestamp.from(nowInstance)
-      }
-
-      "two timestamp of different instance should be not be same" {
-         val nowInstance = Instant.now()
-         val anotherInstance = nowInstance.plusMillis(1000L)
-         Timestamp.from(nowInstance) shouldNotBe Timestamp.from(anotherInstance)
-      }
-
```
LeoColman: I think we can keep these tests

ashishkujoy: Done

sksamuel:
Is this good for merge @kerooker

LeoColman:
I think it is, yes

LeoColman (approved):

PR #1125, opened 2019-12-17
Show helpful message when comparing objects that only differ in line endings

ajalt:
Currently, if you assert something like `"a\nb" shouldBe `"a\rb"`, the assertions will fail correctly, but the message will be something like

```
expected: a[
]b

found: a[
]b
```

Which is not very helpful, since the two values appear identical in the diff.

This PR checks for this case and adds a message to the assertion error indicating the difference.

The truth assertion library [does something similar](https://github.com/google/truth/blob/3c46f24b8161a089aa6712f4f9c579428071c2e2/core/src/main/java/com/google/common/truth/Platform.java#L136).


sksamuel:
What a great idea.

PR #1132, opened 2019-12-26
Implement String UUID matcher

### LINKED ISSUE #1119: String shouldbe UUID matcher

LeoColman:
Implements String UUID matching in accordance to RFC4122. The implementation details and discussion can be found at this StackOverflow thread: https://stackoverflow.com/questions/7905929/how-to-test-valid-uuid-guid

Closes #1119

### DIFF
--- a/kotest-assertions/src/commonMain/kotlin/io/kotest/matchers/string/matchers.kt
+++ b/kotest-assertions/src/commonMain/kotlin/io/kotest/matchers/string/matchers.kt
@@ -3,9 +3,11 @@ package io.kotest.matchers.string
 import io.kotest.Matcher
 import io.kotest.MatcherResult
 import io.kotest.assertions.show.show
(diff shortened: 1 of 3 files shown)


sksamuel:
Should we have a flag that allows the users to choose which type of uuid to match against ?
And if they don't specify it, then it matches all.
Then you could have 5 implementations and the "all" just iterates over them, rather than an uber regex ?

LeoColman:
I think the difference between uuids is the way they are generated and one specific position, so the regex would stay.

Do you mean providing a `shouldBeV4uuid` or ` shouldBeUuid(V4)` ?

sksamuel:
Yeah the latter.

sksamuel:
Can we finish this and get it into master?

LeoColman:
@sksamuel Please review again

sksamuel (approved):

Issue #472, opened 2018-11-13
`map` forgets its shrinker


Chattered:
For example:

```
Gen.string().shrinker != null
Gen.string().map { it }.shrinker() == null
```

This means we don't get a free shrinker when combining Gens. Moreover, it is surprising behaviour for users since it violates the functor law for map.

If it cannot be implemented without breaking covariance, then I think shrinker() should be removed from the Gen interface, and Gen should go back to being covariant.

Kerooker:
You're correct. I believe the implementation for Gen at the moment doesn't consider mapping in a correct way. I'll see what can be done about it.

Thanks for reporting!

Kerooker:
Ok, so here's the thing:
If you're mapping a type for the same type (for example, mapping Int to Int), the shrinker should be the same. However. what if I map one value to another?

For example:

`Gen<Int>.map { it.toString }`

It's impossible to create a shrinker based on this.

What I'll do is return the original shrinker if the type is the same, and return a shrinker if the time is known. Else it will remain null

Chattered:
Unfortunately, it really weakens composability. When you present a monad interface, you expect users to be able to build up complex Gens using map, flatMap, and the various "bind" functions in the Gen interface. These compositions should always be valid according to the monad interface. 

We're using your library in our company (thanks for writing it, btw!) to do what we consider to be obvious things like:

```
Gen.bind(Gen.int(), Gen.string(), { n,str -> Pair(n,str) })
```

and we expect this to be build a composite generator. But we're surprised to lose the Shrinker.

To be pedantic, your proposal still breaks the functor laws, namely the composition law:

Gen<Int>.map { it.toString }.map { Integer.parseInt(it) }.shrinker() == null
Gen<Int>.map { Integer.parseInt(it.toString) }.shrinker() != null
```

My suggestion is still to remove Shrinker from the interface. Is it possible, instead, to pass a Shrinker to functions like "forall", once the final type is known? Perhaps this could be done implicitly, on the assumption that every type determines its shrinker, rather than letting shrinkers vary by generator.

sksamuel:
To be fair, just because there's a flatMap method, it doesn't mean Gen claims to be a Monad. Similarly for a Functor and map. 

I like your suggestion of passing the shrinker to forAll/assertAll. 
Another alternative is requiring the user to supply a fn to map the shrinker as well as the generated values.

[...]

PR #1135, opened 2019-12-27
Property test rework

### LINKED ISSUE #717: Property testing don't accept suspended lambdas
In the same fashion of https://github.com/kotlintest/kotlintest/issues/707, the following code won't work:

```
"Foo" - {
     forAll<Int> { first: Int -> 
        "Bar" { println(first) } 
    } 
}

```

### LINKED ISSUE #780: Allow min successful to be passed to property checks
This would allow for some failures.

### LINKED ISSUE #815: Must allow seeding of the random values in gens

### LINKED ISSUE #967: Option to control Shrinking
ShrinkingMode.OFF: Shrinking disabled
ShrinkingMode.UNBOUNDED: Shrinks until no smaller value can be found. May result in an infinite loop if shrinkers are not coded properly.
ShrinkingMode.BOUNDED: Shrink a maximum number of times (say 1000)

### LINKED ISSUE #969: exhaustive generation option for property testing
GenerationMode.AUTO - uses exhaustive mode where possible, otherwise defaults to random.
GenerationMode.RANDOM - forces random mode
GenerationMode.EXHAUSTIVE - forces exhaustive or throws an exception

### LINKED ISSUE #1037: An exception in a property test doesn't show the failing values
Try this,

```
      assertAll(Gen.double(), Gen.double(), Gen.double()) { a, b, c ->
         throw RuntimeException("foo")
      }
```

And you'll see it doesn't report the failing values.

sksamuel:
This isn't complete but what is there works
But if you think I'm going along the correct lines, then I'd rather merge sooner than later so other people can help out the migration effort. We can always rename / refactor again later.
What is left is completing the rest of the arbitrariness so we have what was there before in gens.
This PR also migrates all our build files to gradle kotlin files.

closes #780
closes #717 
closes #967 
closes #969 
closes #815 
closes #1037


sksamuel:
See discussion in https://github.com/kotlintest/kotlintest/issues/1121

LeoColman:
Maybe close #472

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
```diff
@@ -28,63 +28,67 @@ import io.kotest.properties.shrinking.Shrinker
  * The [Int] generator example should return a random int
  * from across the entire integer range.
  */
+@Deprecated("Deprecated and will be removed in 5.0. Migrate to the new property test classes in 4.0")
```
LeoColman, line 31: I would create a PR to remove it with "[5.0 only]" and keep it there. We'll forget about this

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/shrinking/IntShrinker.kt
```diff
@@ -3,14 +3,27 @@ package io.kotest.properties.shrinking
...
+         1, -1 -> listOf(0)
+         else -> {
+            val a = listOf(0, 1, -1, abs(failure), failure / 3, failure / 2, failure * 2 / 3)
+            val b = (1..5).map { failure - it }.reversed().filter { it > 0 }
+            (a + b).distinct().filterNot { it == failure }
+         }
+      }
+}
+
+object LongShrinker : Shrinker<Long> {
```
LeoColman, line 18: I'm unsure if LongShrinker should be at IntShrinker file

kotest-assertions/src/commonMain/kotlin/io/kotest/tuples.kt
```diff
@@ -0,0 +1,38 @@
+package io.kotest
```
LeoColman: Should this really be at the root package?

kotest-core/build.gradle.kts
```diff
@@ -0,0 +1,72 @@
...
+
+      val jvmMain by getting {
+         dependsOn(commonMain)
+         dependencies {
+            api(kotlin("stdlib-jdk8"))
+            implementation(kotlin("reflect"))
+            api("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.3")
+            implementation("org.slf4j:slf4j-api:1.7.28")
+            implementation("org.junit.jupiter:junit-jupiter-api:5.5.2")
+            implementation("io.arrow-kt:arrow-core:0.10.3")
```
LeoColman, line 65: Do we need all these `api`s? Couldn't some of them be `implementation`?

kotest-property/build.gradle.kts
```diff
@@ -0,0 +1,96 @@
...
+      }
+
+      val jvmMain by getting {
+         dependsOn(commonMain)
+         dependencies {
+            implementation(kotlin("stdlib-jdk8"))
+            implementation(kotlin("reflect"))
+            implementation("com.github.wumpz:diffutils:2.2")
+            implementation("com.univocity:univocity-parsers:2.8.3")
+            api("io.arrow-kt:arrow-core:0.10.3")
```
LeoColman, line 67: Do we need to include arrow as an api? Could it be implementation?

kotest-property/src/commonMain/kotlin/io/kotest/property/Arbitrary.kt
```diff
@@ -0,0 +1,93 @@
...
+    * Returns the values that are considered common edge case for the type.
+    *
+    * For example, for Strings this may include the empty string, a string with white space,
+    * a string with unicode, and a string with non-printable characters.
+    *
+    * The result can be empty if for type T there are no common edge cases.
+    *
+    * @return the common edge cases for type T.
+    */
+   fun edgecases(): Iterable<T>
```
LeoColman, line 34: Maybe make this default to `emptyList`?

kotest-property/src/commonMain/kotlin/io/kotest/property/Arbitrary.kt
```diff
@@ -0,0 +1,93 @@
...
+   override fun edgecases(): Iterable<T> = this@filter.edgecases().filter(predicate)
+   override fun samples(random: Random): Sequence<PropertyInput<T>> =
+      this@filter.samples(random).filter { predicate(it.value) }
+}
+
+/**
+ * Returns a new [Arbitrary] where the edge cases of the receiver are replaced with the edge
+ * cases given as input to this function. The samples are unchanged.
+ */
+fun <T> Arbitrary<T>.withEdgeCases(vararg edgecases: T): Arbitrary<T> = withEdgeCases(edgecases.asList())
```
LeoColman: To me, `withEdgeCases` mean "Get those edge cases and add these ones to them". Maybe `replaceEdgeCases` might be better here.

What I thought this could be used for would be 
```
Arbitrary.string().withEdgeCases("O'Donnel") //I know that my test specifically failed for a ', but that's not usual for everyone
```

LeoColman: But that wouldn't remove default `String` edge cases

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Models the type of inputs that can be generated from an [Arbitrary].
+ *
+ * There are two implementations of [PropertyInput]. The first is a wrapper for values without
+ * a shrinker, which is [PropertyInput.Value].
+ *
+ * The second is a wrapper for values associated with shrinks, which is [PropertyInput.ValueAndShrinker].
+ */
+sealed class PropertyInput<out T> {
```
LeoColman: Shouldn't `PropertyInput` either get it's own file or be with `Arbitrary`?
It's not exactly an `Arbitrary`-only file as stated in the docs, as `Progression` must generate it too (It comes from `Gen`)

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
```diff
@@ -0,0 +1,112 @@
...
+
+fun <T, U> PropertyInput<T>.map(f: (T) -> U): PropertyInput<U> = when (this) {
+   is PropertyInput.Value<T> -> PropertyInput.Value(f(value))
+   is PropertyInput.ValueAndShrinker<T> -> {
+      val shrinks = { this.shrinker().map { it.map(f) } }
+      PropertyInput.ValueAndShrinker(f(value), shrinks)
+   }
+}
+
+fun <T, U> Gen<T>.map(f: (T) -> U): Gen<U> = object : Gen<U> {
```
LeoColman, line 30: How would it work when I want to use a more specific `map`?

```
Arbitrary.int().map { it + 1 } 
```
Would IntelliJ suggest both options? I'm not s ure if these maps and filters should be extensions for the base `Gen` type, only for the specifics.

What do you think?

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
```diff
@@ -0,0 +1,112 @@
...
+fun <T> Gen<T>.filter(predicate: (T) -> Boolean): Gen<T> = object : Gen<T> {
+   override fun generate(random: Random): Sequence<PropertyInput<T>> =
+      this@filter.generate(random).filter { predicate(it.value) }
+}
+
+/**
+ * Returns a new [Gen] which will merge the values from this gen and the values of
+ * the supplied gen together in turn.
+ *
+ * If other words, if GenA provides 1,2,3 and GenB provides 7,8,9 then this gen will output
```
LeoColman:
```suggestion
 * In other words, if GenA provides 1,2,3 and GenB provides 7,8,9 then this gen will output
```

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
```diff
@@ -0,0 +1,112 @@
...
+         }
+   }
+}
+
+/**
+ * Returns a new [Gen]]which will return the values from this gen and once values
+ * of this gen are exhausted will return the values from the supplied gen.
+ * The supplied gen must be a subtype of the type of this gen.
+ */
+fun <T, U : T> Gen<T>.concat(other: Gen<U>): Gen<T> {
```
LeoColman, line 66: Maybe this should throw an exception if `this` is not exhaustible. Or else one might never know that values from the other gen were never obtained

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitraries/arbitraries.kt
```diff
@@ -0,0 +1,25 @@
+package io.kotest.property.arbitraries
```
LeoColman: `property` is singular, perhaps `arbitrary` should be too.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitraries/ints.kt
```diff
@@ -0,0 +1,97 @@
...
+      }
+   }
+}
+
+/**
+ * Returns an [Arbitrary] where each value is a randomly chosen positive integer.
+ * The edge cases are: [Int.MAX_VALUE]
+ */
+fun Arbitrary.Companion.positiveIntegers(iterations: Int): Arbitrary<Int> =
+   int(iterations).withEdgeCases(Int.MAX_VALUE).filter { it > 0 }
```
LeoColman: Using `filter` here might take too long to generate compatible values, and tests that use this will take longer to run

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitraries/ints.kt
```diff
@@ -0,0 +1,97 @@
...
+
+
+/**
+ * Returns an [Arbitrary] where each value is a randomly chosen negative integer.
+ * The edge cases are: [Int.MIN_VALUE]
+ */
+fun Arbitrary.Companion.negativeIntegers(iterations: Int): Arbitrary<Int> =
+   int(iterations).withEdgeCases(Int.MIN_VALUE).filter { it > 0 }
```
LeoColman: Same as above, and it should be `it < 0`

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitraries/ints.kt
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Returns an [Arbitrary] where each value is a randomly chosen natural integer.
+ * The edge cases are: [Int.MAX_VALUE]
+ */
+fun Arbitrary.Companion.nats(iterations: Int): Arbitrary<Int> = int(iterations).filter { it >= 0 }
```
LeoColman: Same as above

kotest-property/src/commonMain/kotlin/io/kotest/property/Arbitrary.kt
sksamuel: I didn't want to do that because then people might forget it's there to override.

kotest-property/build.gradle.kts
sksamuel: It will be removed when I remove arrow out of the main modules, there's another ticket for it.

kotest-core/build.gradle.kts
sksamuel: I haven't changed anything, just copied from the groovy version. We can have another ticket to review before 4.0

kotest-property/src/commonMain/kotlin/io/kotest/property/Arbitrary.kt
sksamuel: I've renamed it to setEdgeCases

kotest-assertions/src/commonMain/kotlin/io/kotest/tuples.kt
sksamuel: I guess not. Moved it.

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
sksamuel: Done.

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
sksamuel: We don't have a map for arb.

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
sksamuel: Fixed

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitraries/ints.kt
sksamuel: Same impl that we had before.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitraries/ints.kt
sksamuel: Same impl that we had before, but I will flip the > <

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitraries/arbitraries.kt
sksamuel: Done.

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/Gen.kt
sksamuel: I think the text will be changed before we go live anyway, as I don't want to keep this around 5.0. Maybe remove in 4.2.

kotest-assertions/src/commonMain/kotlin/io/kotest/properties/shrinking/IntShrinker.kt
sksamuel: That's the old file.

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
sksamuel: This method has no way of knowing. Plus it'll always hit the others because you define the iteration count in the gens now.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitraries/ints.kt
sksamuel: Plus it'll only through away 50% of values. If it was trying to find 1% of values it would be horrible.

PR #1141, opened 2020-01-03
Implement android test

LeoColman:
Executes AndroidTest through a Github action that runs an emulator and runs all android tests.

This commit only implements a very basic hello world test, but we can escalate it from here.


sksamuel:
Wow cool

sksamuel (approved):

Issue #598, opened 2019-01-20
[F] reuse specs, allow generic specs


robstoll:
I am opening this feature request due to a slack discussion. I am not a kotlintest user, so bare with me. As far as I understood, it is currently not possible to reuse specs. I would like to be able to write generic specs. Say you have Strategy A, B and C all implementing MyStrategy. 
- ideally you only need to write one spec and all strategies can reuse it.
- it should also be possible to reuse a generic spec several times in a concrete spec. 
- in case kotlintest does not allow the same test name several times (or only keep one, because the test name is put into a Map :grin:) special care has to be taken. So depending on how kotlintest is implemented, the reuse of a spec should include a prefix for the test or similar to circumvent this problem

LeoColman:
The discussion we had in Kotlinlang's Slack brought us to a hello world example of you composing specs would work.

This was asked a little while back, in #399 but nothing was very discussed over there.

The discussion we had brought us to a central point: Sometimes, behavior is the same between methods or inherited classes, and I might not want to reimplement the same test over and over. For example:

```
interface SumCalculator {
  fun add(first: Int, second: Int): Int
}

interface SubtractionCalculator {
  fun subtract(first: Int, second: Int): Int
}

class SumOnlyCalculator : SumCalculator {
  override fun add(first: Int, second: Int): Int = first + second
}

class SubtractionOnlyCalculator : SubtractionCalculator {
  override fun subtract(first: Int, second: Int): Int  = first - second

}

class ComplexCalculator : SumCalculator, SubtractionCalculator {
  override fun add(first: Int, second: Int): Int = first + second
  override fun subtract(first: Int, second: Int): Int = first - second
}

```

It's obvious that a lot of the code here should do the same thing, but there's not a very good way to do it.

One possibility is to add extension functions to Specs, and then reutilize code there, and we came up with the following to do it:

```

fun AbstractStringSpec.testSumCalculations(sumCalculator: SumCalculator) {
  "should sum correctly" {
    sumCalculator.add(1, 2) shouldBe 3
  }
}

fun AbstractStringSpec.testSubtractCalculations(subtractionCalculator: SubtractionCalculator) {
  "should subtract correctly" {
    subtractionCalculator.subtract(2, 1) shouldBe 1
  }
}

class SumOnlyCalculatorTest : StringSpec({
  testSumCalculations(SumOnlyCalculator())
})

class SubtractOnlyCalculatorTest : StringSpec({
[...]

Issue #399, opened 2018-08-31
[Q] How to compose specs?


gaydenko:
Is it possible to compose specs (that is a set of classes extending some of Spec)?

sksamuel:
So you have two separate classes that extend Spec and you want to combine them together?

LeoColman:
Closing in favor of #598

sksamuel:
This is now supported in master (for 4.0 release) via the new test factory functionality.

PR #1142, opened 2020-01-05
Reusable composable specs

sksamuel:
A design for #598 and #399

 **Goal**:

 To allow specs to be composed of smaller reusable specs. These abstractions should be easily parameterized, dynamic and mappable.

**Status**:

Currently this PR is incomplete. The basics are working but there are many more changes needed to fully convert the 3.x.y spec implementation into the 4.0.0 spec implementation. I have put this PR up for early review and feedback.

If we feel it's looking good we could merge and I can continue work on master. This way anyone else will not be doing PRs against a changing codebase. But if we feel it's not good enough I can continue to iterate.

**Implementation**:

This implementation splits specs into two types. _Class specs_ and _value specs_.

_Class specs_ are what you know and love already. Create a public class that extends one of the provided Spec classes and implement tests in the class body.

 ```kotlin
class MyTestClass: FunSpec() {
  init {
    test("foo") {
      1 + 1 shouldBe 2
    }
  }
}
```

_Value specs_ are new. They are defined as top level public vals that use a spec factory function such as `funSpec { }` or `behaviorSpec { }`
These factory functions allows tests to be defined in the same structure that the similarly named classes do. For example the factory equivalent of FunSpec is:

```kotlin
val funTests = funSpec {
   test("foo") {
      1 + 1 shouldBe 2
   }
   test("bar") {
      "a" + "b" shouldBe "ab"
   }
}
```

Callbacks such as before/after can be defined inline, along with configuration. For example:

```kotlin
val funTests = funSpec {

   beforeTest {
      println("Starting test case ${it.name}")
   }

   afterTest { testCase, result ->
      println(testCase.name + " is completed with result " + result)
   }

   assertionMode = AssertionMode.Error

   testCaseOrder = TestCaseOrder.Random

   test("my test") {
      1 + 1 shouldBe 2
   }
}
```

Specs defined as vals are in every way as functionally complete as specs defined as classes. Both types of specs will be
executed as part of the test suite so you do not need to use class based syntax if not desired.

If you don't want a class or a val to be executed then mark them as private.

```kotlin
private val skipped = funSpec { /*tests here*/ }
```

So what's the point of this new syntax? The point is to allow reusable composable specs. Since the result of a spec factory function is just a `Spec` value,
it can be passed around, filtered, mapped, copied or whatever else you want to do. It's just a value. More importantly however, it can be
included as part other specs.

```kotlin
val tests = stringSpec {
  "test 1" {
    "abc".shouldHaveLength(3)
  }
}

class MySpec : StringSpec() {
  init {
    include(tests)
    "test 2" {
      "".shouldHaveLength(0)
    }
  }
}
```

Note that you can even mix and match the spec styles when composing. The output reports will reflect the correct structure.

```kotlin
val funtests = stringSpec {
  "test 1" {
    "abc".shouldHaveLength(3)
  }
}

val stringtests = stringSpec {
  "test 1" {
    "zyx".shouldHaveLength(3)
  }
}


class MySpec : FunSpec() {
  init {
    include(funtests)
    include(stringtests)
  }
}
```

Since the spec factory functions are just that - functions - they can dynamically generate tests based on parameters.

```kotlin
// defines an interface for the Semigroup typeclass
interface Semigroup<T> {
  fun combine(a: T, b: T): T
}

object StringSemigroup : Semigroup<String> {
  fun combine(a: String, b: String): T = "$a$b"
}

object LongSemigroup : Semigroup<Long> {
  fun combine(a: Long, b: Long): T = a + b
}

// every semigroup should uphold the law that ((ab)c) == (a(bc))
fun <T> associativity(sg: SemiGroup<T>, a: T, b: T, c: T) = funSpec {
  test("semigroup should be associative") {
    sg.combine(a, sg.combine(b, c)) shouldBe sg.combine(sg.combine(a, b), c)
  }
}

class StringSemigroupTest : FunSpec() {
  init {

    include(associativity(StringSemigroup, "a", "b", "c"))

    test("combination") {
        StringSemigroup.combine("a" ,"b") shouldBe "ab"
    }
  }
}

class LongSemigroupTest : FunSpec() {
  init {

    include(associativity(LongSemigroup, 1, 2, 3))

    test("combination") {
        LongSemigroup.combine(1, 2) shouldBe 3
    }
  }
}
```

**Restrictions**

* Class specs cannot be included into other class specs (I may drop this restriction if it proves easy to implement).
* You cannot merge isolation modes. For example, if a value spec has isolation mode of per test, you cannot then include it in a spec that has the default isolation mode.
* Other parameters lose out to the outer most spec. For example, if you include testsA into a classB then settings for test case order will be honoured from classB.
* Value specs can be automatically detected and included in a test suite, so they offer an alternative syntax to classes, but only on the JVM. With JS we'd need to always include in classes.

**Nomenclature:**

The nomenclature used here is totally open for change. I am currently using terms like `SpecBuilder`, `SpecFactory` and _value spec_ and _class spec_.


LeoColman:
> If you don't want a class or a val to be executed then mark them as private.
I didn't like this. IMO we should go the other way around for `val` execution. As it will be embedded in some other specs, we don't want to make it always executable. 

For example, let's say I'm creating a fixture of an interface and placing it in it's own file. It can't be private, as other files need to see it, but it also can't be public, as it won't work on it's own.

Writing `class MySpec : CompositeSpec(myFunSpec, myFreeSpec)` might be clearer (if I want to run a composition-only spec), and keeps the same kind of syntax in JS/Native

LeoColman:
I would propose having executable `fun mySpec() = freeSpec { .. }` instead of having them as values, but I'm not sure how that would work either, as perhaps I need public functions to create the fixtures

sksamuel:
> > If you don't want a class or a val to be executed then mark them as private.
> > I didn't like this. IMO we should go the other way around for `val` execution. As it will be embedded in some other specs, we don't want to make it always executable.
> 
> For example, let's say I'm creating a fixture of an interface and placing it in it's own file. It can't be private, as other files need to see it, but it also can't be public, as it won't work on it's own.
> 
> Writing `class MySpec : CompositeSpec(myFunSpec, myFreeSpec)` might be clearer (if I want to run a composition-only spec), and keeps the same kind of syntax in JS/Native

You can make it internal, or protected. As long as its not private. Internal is enough.
But we can just decide that value specs don't get executed without being in a class if we want.

If you mix in composition by constructor parameters you lose the chance to dynamically invoke the parameters. They have to be values and not functions. The power in the composition is making them functions.

LeoColman:
> If you mix in composition by constructor parameters you lose the chance to dynamically invoke the parameters.

I don't think you do tho

```
class MyCompositeOnlySpec : CompositeSpec(comp1, comp2, comp3("foo"))

class MyCommonSpec : FreeSpec() {

init {
    include(comp1)
    "A" {
    }
}

}
```

LeoColman:
What I tried to say is that instead of having an executable `val spec`, make it easier to create a class that only executes the val spec

sksamuel:
> > If you mix in composition by constructor parameters you lose the chance to dynamically invoke the parameters.
> 
> I don't think you do tho
> 
> ```
> class MyCompositeOnlySpec : CompositeSpec(comp1, comp2, comp3("foo"))
> 
> class MyCommonSpec : FreeSpec() {
> 
> init {
>     include(comp1)
>     "A" {
>     }
> }
> 
> }
> ```

What if you want more complicated logic on what to include? With a function you can do whatever you normally do in code (read from a database to find out the test if you want!)

sksamuel:
> What I tried to say is that instead of having an executable `val spec`, make it easier to create a class that only executes the val spec

Why does it make it easier ?

sksamuel:
Also, if you don't like an executable val, that's one thing, but that's not related to why it needs to be passed in via constructor args. You could allow `include(spec)` without allowing executable vals.

PR #1153, opened 2020-01-13
Adding matcher to check a kotlin code compiles

ashishkujoy:


ashishkujoy:
As of now these matcher checks only for KotlinCode and lets the user know code compiles or not.
The underlying library does not provide support to know does compilation fail for Type error or import error.

One way to know that is to check does the message in compilation result contain word "TypeError" or "Unresolved reference" and provide matcher of top of that such compilationFailsWithTypeError and compilationFailsWithImportError. 
Or the other way is output the complete compilation message on Standard output, but that will be too much of information

Looking for feedback

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/CompilationMatchers.kt
```diff
@@ -29,10 +33,34 @@ private val compiles = object : Matcher<String> {
    }
 }
 
+/**
+ * Assert that given codeSnippet[String] compiles successfully.
+ * It includes the classpath of the calling process,
+ * so that dependencies available for calling process will be available for code snippet snippet as well.
+ * @see [String.shouldNotCompiles]
+ * */
 fun String.shouldCompiles() = this should compiles
```
sksamuel: I think shouldCompile reads better than shouldCompiles

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/CompilationMatchers.kt
```diff
@@ -0,0 +1,66 @@
...
+
+   return compilationResult
+}
+
+private val compiles = object : Matcher<String> {
+   override fun test(value: String): MatcherResult {
+      val compilationResult = compileCodeSnippet(value)
+      return MatcherResult(
+         compilationResult.exitCode == ExitCode.OK,
+         { "Expected code to compile, but it failed to compile" },
```
sksamuel: I would include the actual underlying error as well.

sksamuel (changes requested):

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/CompilationMatchers.kt
ashishkujoy: Done

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/CompilationMatchers.kt
ashishkujoy: Made compilation non verbose, redirected compilation log to a ByteArrayOutputStream instead of System.Out and uses Error message of Result in matcher.

sksamuel:
I think this should be a seperate assertions module to avoid the dependency. I'll merge and then move it out.

PR #1158, opened 2020-01-15
Provide matchers for konform

### LINKED ISSUE #1131: Integration with konform
Konform is quite nice. Should we add an assertions module that allows users to assert on the results of a konform validation ?

simonenkoi:
Provide `shouldBeValid` and `shouldBeInvalid` matchers to integrate with konform

Closes: #1131


simonenkoi:
@sksamuel I really don't like that we need to call `shouldNotBeNull` to validate konform message every time, but I can't see a better option. Are there any suggestions on how to revise it?

kotest-assertions/kotest-assertions-konform/src/jvmMain/kotlin/io/kotest/assertions/konform/matchers.kt
```diff
@@ -0,0 +1,27 @@
...
+import io.konform.validation.Validation
+import io.kotest.Matcher
+import io.kotest.MatcherResult
+import io.kotest.should
+
+infix fun <T> Validation<T>.shouldBeValid(value: T) = this should beValid(value)
+
+fun <A> beValid(a: A) = object : Matcher<Validation<A>> {
+   override fun test(value: Validation<A>): MatcherResult =
+      MatcherResult(value(a) is Valid, "$a should be valid", "$a should not be valid")
```
sksamuel: Maybe we could include the validation failure in the messsage

kotest-assertions/kotest-assertions-konform/src/jvmTest/kotlin/com/sksamuel/kotest/tests/konform/ValidatedMatchersTest.kt
```diff
@@ -0,0 +1,61 @@
...
+      validateUser shouldBeValid validUser
+      shouldThrow<AssertionError> {
+         validateUser shouldBeInvalid validUser
+      }
+   }
+
+   "shouldNotBeValid" {
+      val invalidUser = UserProfile("A", -1)
+
+      validateUser.shouldBeInvalid(invalidUser) { invalid ->
```
sksamuel: We could use contracts rather than the lambda. Or both.

sksamuel:
I agree on the null so you could add another matcher which works on the Invalid type and extracts the field for you.

invalid.shouldFailField(name) { }

With a better name if you can think of one.

kotest-assertions/kotest-assertions-konform/src/jvmTest/kotlin/com/sksamuel/kotest/tests/konform/ValidatedMatchersTest.kt
simonenkoi: I've thought about it, but not sure what should we use as a contract.
Validation can be reused for many inputs and we cannot say that it is invalid after the first check. The same for the input.
Correct me if I am wrong and didn't catch the idea

kotest-assertions/kotest-assertions-konform/src/jvmMain/kotlin/io/kotest/assertions/konform/matchers.kt
simonenkoi: Done. Since errors is a private field of Invalid, I've added a print of the whole object

simonenkoi:
> I agree on the null so you could add another matcher which works on the Invalid type and extracts the field for you.
> 
> invalid.shouldFailField(name) { }
> 
> With a better name if you can think of one.

Added Invalid<*>.shouldContainError(Any, String) method. Not sure that I need to replace String with lambda as you mentioned, but OK with it if you insist

sksamuel:
Yeah that works fine.

sksamuel (approved):

sksamuel:
I'll merge this in if you're happy with it ?

simonenkoi:
Sure. I can add the documentation, but it is not clear where to locate it. Does it make sense to place it in https://github.com/kotlintest/kotlintest/blob/master/doc/extensions.md?

sksamuel:
Let's start a new section at the bottom of
https://github.com/kotlintest/kotlintest/blob/master/doc/matchers.md,
adding in that it's an optional module.

Issue #523, opened 2019-01-07
Retry Helpers


sksamuel:
Might be handy to have a way to re-run a test once or more, based on some retry strategy.

```
class MyTest : FunSpec({
  test("foo") {
    retry(3) {
      // test here
    }
  }

  test("bar") {
    retry(exponential(16)) {
      // test here
    }
  }
}
```

This would be similar to `eventually`, but is based on a number of retries rather than a period of time.

sksamuel:
Your thoughts @Kerooker

Kerooker:
What would `exponential`  mean? I think this is good

sksamuel:
Means you try again afafter yer 2 seconds then after 4 then 8 and so on.

sksamuel:
https://github.com/jhalterman/failsafe/

Issue #698, opened 2019-02-27
Extract invocations/threads out to the library level and add support for retry and min required passes


sksamuel:
I was thinking of deprecating invocations and replacing it with a function.

In other words moving the logic _out_ of the engine and into the test itself. There's actually no reason it needs to be in the engine.

```kotlin
test("some fun spec test") {
  repeat(5) {
    // test
  }
}
```

And then we can combine `invocations`, `required`, `retry`, `threads` all into a single powerful command.

```kotlin
test("another test") {
  repeat(10).threads(3).retry(backoff(4)).required(4) {
  // test  
  }
}
```

Or

```kotlin
test("another test") {
  repeat(10) {
    retry(simple(3)) {
       // test  
    }
  }
}
```

The open question is, will this play nicely with isolation modes.
Edit: Answering my own question - we can just limit this to TestType.Test scopes to avoid any weirdness.

_Originally posted by @sksamuel in https://github.com/kotlintest/kotlintest/issues/641#issuecomment-464396770_

sksamuel:
Related to #523

sksamuel:
As part of this work, the coroutine boundary should move into the spec executor. See #710

PR #1159, opened 2020-01-16
#523 Adds a retry function

ashishkujoy:
This function repeatedly invoke the given code block till it fails with either AssertionError or any other given exception until the max retry limit or max duration limit is reached.

### DIFF
--- a/kotest-assertions/src/jvmMain/kotlin/io/kotest/Retry.kt
+++ b/kotest-assertions/src/jvmMain/kotlin/io/kotest/Retry.kt
@@ -8,7 +8,7 @@ import java.time.Duration
  * Retry the given lambda [f] in case of assertion error[AssertionError] and exception [Exception].
  * Stops trying when [maxRetry] or [maxDelay] duration is reached.
  * */
-suspend inline fun <T> retry(
+suspend fun <T> retry(
    maxRetry: Int,
    maxDelay: Duration,
    delay: Duration = maxDelay,
@@ -21,7 +21,7 @@ suspend inline fun <T> retry(
  * Retry the given lambda [f] in case of assertion error[AssertionError] and exception of type [exceptionClass].
  * Stops trying when [maxRetry] or [maxDelay] duration is reached. Or an exception other than [exceptionClass] is thrown
  * */
-suspend inline fun <T, E : Throwable> retry(
+suspend fun <T, E : Throwable> retry(
    maxRetry: Int,
    maxDelay: Duration,
    delay: Duration = maxDelay,
--- a/kotest-assertions/src/jvmMain/kotlin/io/kotest/Retry.kt
+++ b/kotest-assertions/src/jvmMain/kotlin/io/kotest/Retry.kt
@@ -1,0 +1,49 @@
+package io.kotest
+
+import io.kotest.assertions.Failures
+import kotlinx.coroutines.delay
+import java.time.Duration
+
+suspend inline fun <T> retry(
+   maxRetry: Int,
+   maxDelay: Duration,
+   delay: Duration = maxDelay,
+   multiplier: Int = 1,
+   f: () -> T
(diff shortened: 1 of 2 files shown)


kotest-assertions/src/jvmMain/kotlin/io/kotest/Retry.kt
```diff
@@ -4,6 +4,10 @@ import io.kotest.assertions.Failures
 import kotlinx.coroutines.delay
 import java.time.Duration
 
+/**
+ * Retry the given lambda [f] in case of assertion error[AssertionError] and exception [Exception].
+ * Stops trying when [maxRetry] or [maxDelay] duration is reached.
+ * */
 suspend inline fun <T> retry(
```
sksamuel: Do we want to inline it if it's suspend ?

kotest-assertions/src/jvmMain/kotlin/io/kotest/Retry.kt
```diff
@@ -0,0 +1,58 @@
...
+ * Retry the given lambda [f] in case of assertion error[AssertionError] and exception [Exception].
+ * Stops trying when [maxRetry] or [maxDelay] duration is reached.
+ * */
+suspend inline fun <T> retry(
+   maxRetry: Int,
+   maxDelay: Duration,
+   delay: Duration = maxDelay,
+   multiplier: Int = 1,
+   f: () -> T
+): T = retry(maxRetry, maxDelay, delay, multiplier, Exception::class.java, f)
```
sksamuel: What do you think about extracting the maxRetry/maxDelay/delay logic out into an interface `Retry` of which one could be `SimpleRetry` which implements this.

It means we could add `BackoffRetry` later as well.

Or do you think that's overkill ?

sksamuel:
Seems related to #698
Not sure if we want to incorporate all the settings into a single function or not.

kotest-assertions/src/jvmMain/kotlin/io/kotest/Retry.kt
ashishkujoy: I am not sure how much advantage will extracting them in a Interface will provide. Which is based on my assumption this would look like after extracting in Interface.

```kotlin
interface RetryStrategy {
   val maxRetry: Int
   fun maxDelay(): Duration
   fun delay(): Duration
}

class SimpleRetryStrategy(override val maxRetry: Int, private val maxDelay: Duration, private val delay: Duration) :
   RetryStrategy {
   override fun maxDelay(): Duration = maxDelay
   override fun delay(): Duration = delay
}

class BackoffStrategy(override val maxRetry: Int, private val maxDelay: Duration, private val delay: Duration) :
   RetryStrategy {
   var counter = 1L
   override fun maxDelay(): Duration = maxDelay
   override fun delay(): Duration {
      val nextDelay = delay.multipliedBy(counter)
      counter += 1L
      return nextDelay
   }
}
```

kotest-assertions/src/jvmMain/kotlin/io/kotest/Retry.kt
ashishkujoy: No i have copied code from eventually and had forgot to removed that

kotest-assertions/src/jvmMain/kotlin/io/kotest/Retry.kt
sksamuel: Ok I think leave it as it was. But one small change can we rename delay to timeout.

sksamuel: Sorry to be clear delay is fine but maxDelay I think is better as timeout.

ashishkujoy: Done

PR #1175, opened 2020-01-24
Move TableTesting and DataDrivenTesting to commonMain

ajalt:
The `Function.reflect` method is only available on JVM, so Data Driven Testing functions on other targets will use default parameter names, but that's still better than nothing.


sksamuel:
Awesome more mpp!

PR #1176, opened 2020-01-27
Makes properties field mandatory for shouldBeEqualToUsingFields and shouldBeEqualToIgnoringFields

ashishkujoy:


ashishkujoy:
As of now I cannot think of any shorter and meaning full name for ```shouldBeEqualToIgnoringFields``` and ```shouldBeEqualToUsingFields```. One option is ```shouldHaveEqualFields``` but that does not implies that user need to mention fields which are to use while doing equality check.

Any suggestions for alias @Kerooker @sksamuel

sksamuel:
shouldEqualUsingFields ?

But it's hardly any shorter.

ashishkujoy:
> shouldEqualUsingFields ? But it's hardly any shorter.

Without Be that sounds little off.

sksamuel:
Agreed.

PR #1177, opened 2020-01-28
Move type matchers to commonMain

ajalt:
`shouldHaveAnnotation` stays in `jvmMain` since it has a Java `Class` as a parameter. Everything else was updated to use `KClass` internally rather than converting to a Java `Class` first.


sksamuel:
Compilation error @ajalt

ajalt:
Apologies; I didn't realize that tests only run on the JVM and so don't catch JS issues.

sksamuel:
I need to improve that and run tests in commonTest. If only there was a test framework I could use.

PR #1179, opened 2020-01-29
Add Kotlin/Native support to kotest-assertions, kotest-fp, and kotest-properties

ajalt:
I set up native targets to all depend on a single source set to avoid code duplication. IntelliJ doesn't understand source sets without a target, though, so I took an approach similar to [kotlinx-io](https://github.com/Kotlin/kotlinx-io/blob/master/core/build.gradle) and make a single native target  that's an alias to the common native source set in IntelliJ. I expanded on their approach to support development on any OS. When running the build through gradle, all targets are defined normally.

I didn't add native targets to kotest-core, since I'm not sure how test runners work on native.

### DIFF
--- a/buildSrc/src/main/kotlin/Libs.kt
+++ b/buildSrc/src/main/kotlin/Libs.kt
@@ -47,6 +47,7 @@ object Libs {
       const val jdk8 = "org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:$version"
       const val core = "org.jetbrains.kotlinx:kotlinx-coroutines-core:$version"
       const val coreJs = "org.jetbrains.kotlinx:kotlinx-coroutines-core-js:$version"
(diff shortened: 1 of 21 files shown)


ajalt:
[This recent blog post](https://www.alecstrong.com/2020/01/github-actions-mpp/) details how to set up GitHub actions to publish a MPP library with K/N support. If this PR gets merged, that might be a good way to go rather than trying to get Travis working on windows/macOS

sksamuel:
Great!
So with this it'll only build the native when its on one of those environments. Like you can't build a mac binary if you're on linux I guess.

ajalt:
That's correct. Any operating system can build the linux target, but you can't cross-compile to windows or macOS. The way the blog post works is to build all targets except windows on a macOS box, and use the windows box for the windows target. Once all targets have uploaded their binaries, they will promote the snapshot to a full build.

sksamuel:
I get it. Thanks. Very exciting now we have assertions for native jvm and
js and runners for the last two.

PR #1180, opened 2020-01-30
Replace TravisCI with GitHub Actions for CI tests

ajalt:
This allows us to run unit tests for native targets.

I set up the workflow based on information from [this blog post](https://www.alecstrong.com/2020/01/github-actions-mpp/). The linux host will run JVM, JS and linux native tests, while the macOS and Windows hosts will save time by only running the native tests for their respective platforms.

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -1,0 +1,64 @@
+name: Test
+
+on:
+  pull_request:
+    paths-ignore:
+      - 'doc/**'
+      - '*.md'
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [macOS-latest, windows-latest, ubuntu-latest]
+
+    runs-on: ${{matrix.os}}
+
+    steps:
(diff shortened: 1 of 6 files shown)


ajalt:
We should also create a workflow to publish shapshots, but that requires the publishing keys, so I'm not able to do it. Here's [an example of a publish workflow](https://github.com/cashapp/sqldelight/blob/master/.github/workflows/Release.yml) and here's a [blog post that discusses it](https://www.alecstrong.com/2020/01/github-actions-mpp/)

sksamuel:
I will have a play with the publish. Do you think we can completely replace travis with this ?

ajalt:
Absolutely, and I recommend that we do. IMO, Travis has gone downhill since it was acquired, and GH Actions is a better product overall.

sksamuel:
Snapshot deploys are now running through github workflow, but only for
linux builds.
I've disabled travis and appveyor.

PR #1186, opened 2020-02-01
#1184 Add shouldHaveSingleElement matcher with predicate

### LINKED ISSUE #1184: New collections matcher
`collection.shouldHaveSingleElement { e -> ... }`

simonenkoi:
Closes: #1184


sksamuel:
Thansk for the PR. I was thinking more that this matcher would not use the lambda to test for one element that passes the predicate (but allows more), but that there's a single element and then executes the lambda. What do you think?

simonenkoi:
Yes, it makes perfect sense and seems usable. I think if it was confusing for me from the ticket's description, it can be confusing for users too. Can we think about better naming? 
Maybe it makes more sense to extend `shouldBeSingleton` matcher with a lambda parameter?

sksamuel:
Yes the ticket was poor wording by me.

shouldBeSingleElement maybe ?

simonenkoi:
@sksamuel I've added a new matcher `shouldBeSingleton { block }` to keep it consistent. Please, let me know if you insist on `shouldBeSingleElement` naming. Should I revert `shouldHaveSingleElement { predicate }` matcher changes?

sksamuel:
Lets go with your naming.

PR #1188, opened 2020-02-03
Publish native artifacts for all supported platforms

ajalt:
Currently, only the linux artifact is being published. This PR updates the workflow to use the maxOS runner to publish linux and macOS artifacts, and the windows runner to publish the windows artifact.

The build number is generated once in the build job, then downloaded in the two publish jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,25 +43,58 @@ jobs:
                name: error-report
                path: build-reports.zip
 
-   deploy:
-      needs: build
-      runs-on: ubuntu-latest
-
-      steps:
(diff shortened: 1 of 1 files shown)


ajalt:
You might also want to create a tag like `build-number-4000` on the tip of master so that generated build numbers are larger than the old snapshot numbers. Currently, the most recent snapshot was published with build number 3.

sksamuel:
Yeah I'll reset the build number so its higher.

sksamuel:
Looks like the windows publish fails.
https://github.com/kotlintest/kotlintest/actions/runs/34360070

PR #1190, opened 2020-02-03
Fix task name for windows publication workflow

ajalt:


ajalt:
Sorry about that. This looks like the correct task named based on the output of `gradlew tasks`

PR #1200, opened 2020-02-08
Fix markdown link

ilaborie:


ashishkujoy:
@ilaborie thanks for fix.

Issue #1161, opened 2020-01-17
Add alias for shouldBeEqualToUsingFields and disallow specifying NO fields


sschuberth:
Occasionally, I have to assert value-equality of Java class instances that do not implement `equals()` with kotlintest (also see [this](https://stackoverflow.com/questions/12147297/how-do-i-assert-equality-on-two-classes-without-an-equals-method)). For these cases, it would be nice to have a matcher that iterates over all fields via reflection (recursively) and asserts value equality for each.

sksamuel:
Agreed, and related to #822

LeoColman:
We do have something like that that, I think

Something similar to
`shouldBeEqualToUsingFields ... `

sschuberth:
I skimmed through https://github.com/kotlintest/kotlintest/blob/master/doc/matchers.md before but didn't find anything...

sschuberth:
But it's indeed there, yay!

https://github.com/kotlintest/kotlintest/blob/068148734769f534223be89673639f46e8430ab4/kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt#L32

LeoColman:
Hmm...
It's not documented, or it isn't released yet (and therefore not documented due to that), I wonder?

sschuberth:
Looks like it was introduced in f537d7beef9b490efdba893170a41069c001d1c1 as part of working on https://github.com/kotlintest/kotlintest/issues/711 which is included in version 3.3.1.

sksamuel:
Should we add an alias. It's a bit of a long name

sschuberth:
Also, it's only available as an extension function, not as an infix operator.

sschuberth:
Doh, and it does not work at all as I would have expected: When called without a second argument, *no* instead of *all* fields are compared. That function should probably be changed to either require at least one field to be specified, or to compare all fields if none are given.

sksamuel:
Yes that's totally a bug. Let's fix that.

LeoColman:
There's also the `shouldBeEqualToIgnoringFields`, which would fit your use case, IIRC

sschuberth:
I've actually tried it @Kerooker, but it that gave me:

    kotlin.reflect.full.IllegalCallableAccessException: java.lang.IllegalAccessException: class kotlin.reflect.jvm.internal.calls.CallerImpl$FieldGetter cannot access a member of class org.apache.maven.settings.Proxy with modifiers "private"

[...]

PR #1208, opened 2020-02-09
#1161 Updates shouldBeEqualToIgnoringFields to compare private fields as well

ashishkujoy:


kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
```diff
@@ -190,14 +184,14 @@ fun <T : Any> beEqualToIgnoringFields(
    override fun test(value: T): MatcherResult {
 
       val fieldNames = fields.map { it.name }
-      val failed = value::class.memberProperties.filterNot { fieldNames.contains(it.name) }.mapNotNull {
-         val actual = it.getter.call(value)
-         val expected = it.getter.call(other)
-         if (actual == expected) null else {
-            "${it.name}: $actual != $expected"
+      val fieldsToBeConsider: List<KProperty<*>> = value::class.memberProperties
```
sschuberth: Nit: The variable should probably be called "fieldsToBeConsider*d*".

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
```diff
@@ -190,14 +184,14 @@ fun <T : Any> beEqualToIgnoringFields(
...
+         .filterNot { fieldNames.contains(it.name) }
+         .map {
+            it.isAccessible = true
+            it
```
sschuberth: In the end, this just maps to `it` (i.e. identity) as the line `it.isAccessible = true` is ignored here. What you probably want is:

```
val fieldsToBeConsiderd = value::class.memberProperties.filter {
    !fieldNames.contains(it.name) && it.isAccessible
}
```

sschuberth (dismissed):

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
ashishkujoy: No, ```it.isAccessible = true``` is not getting ignored, it is making ever other field accessible(not considering if they are already accessible) which are not asked to ignore which checking equality

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
ashishkujoy: Done, renamed to fieldsToBeConsidered

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
sschuberth: Ah, sorry, I was misreading the assignment. So how about using `onEach` instead of a "dummy" `map` then?

ashishkujoy: done

sschuberth: You're using `forEach` now, which returns `Unit`. With `onEach` you could have kept the fluent API as it returns the modified list.

ashishkujoy: OK I was not aware of ```onEach```

kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
```diff
@@ -190,14 +184,10 @@ fun <T : Any> beEqualToIgnoringFields(
...
       val fieldNames = fields.map { it.name }
-      val failed = value::class.memberProperties.filterNot { fieldNames.contains(it.name) }.mapNotNull {
-         val actual = it.getter.call(value)
-         val expected = it.getter.call(other)
-         if (actual == expected) null else {
-            "${it.name}: $actual != $expected"
-         }
-      }
+      val fieldsToBeConsidered: List<KProperty<*>> = value::class.memberProperties.filterNot { fieldNames.contains(it.name) }
+      fieldsToBeConsidered.forEach { it.isAccessible = true }
```
sksamuel: Rather than filter it out, if you ask for a field and its private, we should throw an error ?

sschuberth: Or simply ignore it in the comparison? I was just asking myself the same.

ashishkujoy: @sksamuel Here user is not asking for including private field directly, user is only saying to ignore some fields(public) which by default resulting in comparing the private fields if there are any.

ashishkujoy: > Or simply ignore it in the comparison? I was just asking myself the same.

But it can very well happen the private field is state maintain by the object which determine its equality.
For example:

```kotlin
class Counter {
 private var currentCount = 0
fun increment(): Unit = currentCount++
fun getCurrentCount():Int = currentCount 
}
```

sksamuel: I think if you
* ask for a private field we error
* you ask for public fields but there's also a private field we error

The path should be of least surprise to the user. Silently dropping private fields is probably fine 99% of the time but we should err on the side of caution imo.

ashishkujoy: Another thought in my mind is to let user decide to include or ignore the private fields by using taking a boolean parameter ```ignorePrivateFields``` with default value as false.

sksamuel: Yep that would work.

ashishkujoy: Since the signature contain the vararg
```fun <T : Any> T.shouldBeEqualToIgnoringFields(other: T, vararg properties: KProperty<*>)```
So we can't have ```ignorePrivateFields``` as the last parameter with default value. And if put that before the vararg argument then every time we have to  user has to specify that argument. Which will break the existing user code as well that will not look clean.

So going by your suggestion I will make the change to ignore the private fields

sksamuel:
lgtm merge whenever you are ready

ashishkujoy:
@sksamuel @sschuberth should we close this now.

sksamuel:
You need to merge it first

PR #1210, opened 2020-02-09
Updated docs with kotest details #1209

sksamuel:
This is not meant to be fully comprehensive nor include all the new features but just be sufficient enough to get a beta release out.


doc/conditional_evaluation.md
```diff
@@ -0,0 +1,91 @@
...
+  ...
+}
+```
+
+`isLinux` and `isPostgreSQL` in the example are just expressions (values, variables, properties, function calls) that evaluate to `true` or `false`.
+
+
+### Focus
+
+KotlinTest supports isolating a single top level test by preceding the test name with `f:`.
```
ashishkujoy: Need to replace kotlinTest with kotest

doc/conditional_evaluation.md
```diff
@@ -0,0 +1,91 @@
...
+
+```kotlin
+"should do something".config(enabled = false) {
+  ...
+}
+```
+
+You can use the same mechanism to run tests only under certain conditions.
+ For example you could run certain tests only on Linux systems using
+ [SystemUtils](http://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/SystemUtils.html#IS_OS_WINDOWS).IS_OS_LINUX from [Apache Commons Lang](https://commons.apache.org/proper/commons-lang/).
```
ashishkujoy: here we need to add a Space after #IS_OS_WINDOWS so that link can render properly

doc/conditional_evaluation.md
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  "test 3" {
+    // this will run too
+  }
+})
+```
+
+### SkipTestException
+Sometimes you want to interrupt a test in runtime, as perhaps you don't know at compile-time if the test should be executed. For this, KotlinTest provides a way to interrupt it by throwing an exception: The `SkipTestException`.
```
ashishkujoy: Need to replace For this, KotlinTest with For this, Kotest

doc/discovery_extension.md
```diff
@@ -0,0 +1,24 @@
+### Discovery Extension
+
+_Advanced Feature_
+
+Another type of extension that can be used inside `ProjectConfig` is the `DiscoveryExtension`. This extension is designed
+ to allow customisation of the way spec classes are discovered and instantiated. There are two functions of interest that
+ can be overridden.
+
+The first is `afterScan` which accepts a list of Spec classes that were discovered by KotlinTest during the _discovery_ phase
```
ashishkujoy: Need to replace ```KotlinTest``` by ```Kotest```

sksamuel:
Thanks @ashishkujoy feedback addressed.

PR #1214, opened 2020-02-11
Add matchers to Klock. 

dragneelfps:
Related to https://github.com/kotlintest/kotlintest/issues/1156.


dragneelfps:
@sksamuel I have created this draft. I wanted to get an opinion on the approach I have used. Also, what are the matchers that should be implemented.

LeoColman:
I believe you're on the right track.
I recommend adding some matchers for numbers, like `Time.shouldHaveMinutes(5)` or something similar.

I'm not sure if you're using our current datetime matchers as inspiration, but that can be a good starting point if you feel lost! I believe you are doing things that way already tho

[Take a look](https://github.com/kotlintest/kotlintest/tree/master/kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/date)

dragneelfps:
@Kerooker I think I have added all the **required** matchers. I haven't documented the matchers like done in [existing code](https://github.com/kotlintest/kotlintest/blob/805857bf06cd185239c4c263e146713d22b699d3/kotest-assertions/src/jvmMain/kotlin/io/kotest/matchers/date/localtime.kt#L12) since, I didnt find any need for doing it, as its self documenting. But if you want me to add docs, then I can. 
I have added some tests as well. 
I dont know if publishing works on that. Need to check on that. 
Other than that, I think the PR is ready to be reviewed.

kotest-assertions/kotest-assertions-klock/src/commonMain/kotlin/io.kotest/matchers/date/date.kt
```diff
@@ -0,0 +1,136 @@
...
+fun haveSameYear(date: Date) = object : Matcher<Date> {
+   override fun test(value: Date) =
+      MatcherResult(
+         value.year == date.year,
+         { "$value should have year ${date.year}" },
+         { "$value should not have year ${date.year}" }
+      )
+}
+
+infix fun Date.shouldHaveYear(year: Long) = this shouldBe haveYear(year)
```
ashishkujoy: I think we should not take year as Long it should be Int, else while using we will have to explicitly need to convert Int to Long in tests like in example below
```kotlin
date shouldHaveSameYear 2020.toLong()
```
more preferable would be 
```kotlin
date shouldHaveSameYear 2020
```

Same would go for month, days, hours etc in this and Time matchers

dragneelfps: Yes, you are right. I will update the code.

kotest-assertions/kotest-assertions-klock/src/commonMain/kotlin/io.kotest/matchers/date/date.kt
```diff
@@ -0,0 +1,136 @@
+package io.kotest.matchers.date
+
+import com.soywiz.klock.Date
+import com.soywiz.klock.DateTime
+import io.kotest.matchers.Matcher
+import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.shouldNotBe
+
+infix fun Date.shouldHaveSameYearAs(date: Date) = this shouldBe haveSameYear(date)
```
ashishkujoy: ```shouldBe haveSameYear``` Sounds little off. 
I think ```should haveSameYear```  will sound better
Same goes for few  other matchers below

kotest-assertions/kotest-assertions-klock/src/commonMain/kotlin/io.kotest/matchers/date/date.kt
```diff
@@ -0,0 +1,136 @@
...
+fun haveDay(day: Int) = object : Matcher<Date> {
+   override fun test(value: Date) =
+      MatcherResult(
+         value.day == day,
+         { "$value should have day $day" },
+         { "$value should not have day $day" }
+      )
+}
+
+fun Date.shouldBeToday() = this shouldBe beToday()
```
ashishkujoy: Here ```Be``` is coming twice so ```this shouldBe beToday``` also sounding off.
```this should beToday```  would be better.

File: kotest-assertions/kotest-assertions-klock/src/commonMain/kotlin/io.kotest/matchers/date/time.kt [hunk 4]
```diff
@@ -0,0 +1,152 @@
...
+
+infix fun Time.shouldHaveSameHoursAs(time: Time) = this should haveSameHours(time)
+
+infix fun Time.shouldNotHaveSameHoursAs(time: Time) = this shouldNot haveSameHours(time)
+
+fun haveSameHours(time: Time): Matcher<Time> = object : Matcher<Time> {
+   override fun test(value: Time): MatcherResult =
+      MatcherResult(
+         value.hour == time.hour,
+         { "$value should have hours ${time.hour}" },
```
ashishkujoy: ```$value should have ${time.hour} hours``` will sound better compare to ```$value should have hours ${time.hour}``` same goes for other matchers

sksamuel:
If you can address the feedback @dragneelfps I will merge. This will be an excellent addition.

kotest-assertions/kotest-assertions-klock/src/commonMain/kotlin/io.kotest/matchers/date/date.kt
dragneelfps: done.

kotest-assertions/kotest-assertions-klock/src/commonMain/kotlin/io.kotest/matchers/date/time.kt
dragneelfps: done.

kotest-assertions/kotest-assertions-klock/src/commonMain/kotlin/io.kotest/matchers/date/date.kt
dragneelfps: done.

dragneelfps:
@sksamuel feedback taken and code changes done.

dragneelfps:
Build failing with: 
```
* Where:
Build file '/home/runner/work/kotest/kotest/kotest-assertions/kotest-assertions-klock/build.gradle.kts' line: 70

* What went wrong:
Could not read script '/home/runner/work/kotest/kotest/publish.gradle' as it does not exist.
```
Any help?

sksamuel:
The build was changed over the weekend so publish.gradle doesn't exist. That's ok, we can merge.

Issue #961, opened 2019-08-25
JDBC ResultSet matchers


sksamuel:
`ResultSet.shouldHaveColumnCount(n)`
`ResultSet.shouldContainColumn(label)`
`ResultSet.shouldHaveColumn(k, label)`

Any others we can think of.

arunvelsriram:
I would like to work on this.

sksamuel:
Please do!

sksamuel:
See https://github.com/joel-costigliola/assertj-db#readme

arunvelsriram:
Is this still relevant? As @sksamuel pointed out `assertj-db` is already providing assertions for databases.

sksamuel:
assertJ won't do extension matchers because it's not a kotlin library.

simonenkoi:
@sksamuel do you have any ideas on how API should look like for values assertions?
Should it be something like
```kotlin
resultSet shouldHaveColumn "name" {
   { it shouldBe "Hewson" }
   { it shouldBe "Evans" }
   { it shouldBe "Clayton" }
   { it shouldBe "Mullen" }
}
```
where the every lambda calls next value from ResultSet (analogue from https://github.com/joel-costigliola/assertj-examples/blob/master/assertions-examples/src/test/java/org/assertj/examples/db/TableAssertionExamples.java)

sksamuel:
One option:

`resultset.shouldHaveColumn("name") { it shouldBe listOf(....) }` 

just pull back all the rows at once. This is a test, so I think that's fine, and then the lambda parameter is a collection, so then all the standard collection matchers work for no extra work.

dragneelfps:
Anyone working on it? Status? Asking cuz I would like to work on it.

sksamuel:
Please do.

PR #1221, opened 2020-02-18
Add matchers for ResultSet.

dragneelfps:
Contains `java.sql.ResultSet` assertion matchers

Inspired from https://github.com/kotest/kotest/issues/961


sksamuel (approved):
Looks great. We could also add one that checks for the presence of a row based on a lamda.

dragneelfps:
> Looks great. We could also add one that checks for the presence of a row based on a lamda.

That will be interesting. Will add it

dragneelfps:
@sksamuel I have added the row assertion. Also, I changed the column value assertion to use generics which enables us to have different specific types in the lambda block of the assertion.

sksamuel:
Excellent.

PR #1237, opened 2020-02-24
1235 Add functions to call shouldContain for Array

### LINKED ISSUE #1235: We don't have Array.shouldContain
We have Array.shouldEverythingElse and Collection.shouldContain

simonenkoi:
Closes: #1235


sksamuel (commented):
Was there a reason we did C: Collection before?

simonenkoi:
@sksamuel I don't think so. Changed to keep it consistent with other methods. The only reason I can identify is not to repeat `Collection<T>` in matcher's generic and its function, but my opinion is that it worsens code readability

Issue #1150, opened 2020-01-11
Better formatted error for map comparsion


sksamuel:
expected: {"test1"=FAILED, "test2"=FAILED, "com.sksamuel.kotest.runner.junit5.JUnitTestRunnerListenerTests"=FAILED, "Kotest"=SUCCESSFUL} but was: {"test2"=FAILED, "test1"=FAILED, "com.sksamuel.kotest.runner.junit5.JUnitTestRunnerListenerTests"=SUCCESSFUL, "Kotest"=SUCCESSFUL}

is awful to read

ashishkujoy:
@sksamuel any thought on this indented error message for map comparison
<img width="910" alt="Screenshot 2020-02-24 23 19 17" src="https://user-images.githubusercontent.com/34642693/75177455-3372f100-575c-11ea-88ae-31c8317f30dd.png">

sksamuel:
Yes i like it.
Perhaps after the two maps we can display something like:
Values differed at keys "test1", "test"2.

And we should limit the map to only 10 entries or something with a ... if
we go over.

PR #1238, opened 2020-02-25
#1150 Adds shouldBeEqualTo matcher for map with formatted error for map comparison

ashishkujoy:


ashishkujoy:
@sksamuel @Kerooker I tried to keep name of matcher ```shouldBe``` but since in test there is already a generic ```shouldBe``` used, ```map.shouldBe```  was always pointing to the generic shouldBe because of that I had to name it ```shouldBeEqualTo```.
Is any workaround for this  without renaming the matcher.

sksamuel:
I think we need to override this at the shouldBe level.

If we create a new matcher like shouldBeEqualTo no one will know it's there to use.

We have a filed called compare.kt that's used inside shouldBe for comparsion. I think we should build this out better. We could have a CompareTo typeclass/interface, and MapCompareTo is one implementation. We take the type on the left and side and use that to find the correct CompareTo instance.

sksamuel:
How come you closed this @ashishkujoy

PR #1248, opened 2020-02-29
Fixed Result.shouldBeSuccess matcher

audkar:
Previous implementation was not asserting if Result actually is a Success


sksamuel:
Nice spot.

PR #1249, opened 2020-02-29
Fix the regex property generator so it produces new values

dump247:
The arb was generating a sequence of the same value; the first value for the
RandomSource seed.


sksamuel:
:+1:

PR #1251, opened 2020-03-02
Fix arb bind so that it uses one sequence for each of the input arbs

dump247:
The previous implementation starts a new sequence for the inputs for each sample
generated by the bind. This fails for input arbs that have some sequence
initialization, such as regex and distinct.

The previous implementation also skipped the edgcases defined by the input arbs.

I believe this now matches the operation of the old Gen.bind that Arb.bind is supposed to replace. I did not use Sequence.zip as that makes for a lot of references to `first.first.first.first` etc.

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/bind.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/bind.kt
@@ -3,8 +3,6 @@ package io.kotest.property.arbitrary
 import io.kotest.property.RandomSource
 import io.kotest.property.Sample
 
-fun <A, T> Arb.Companion.bind(arbA: Arb<A>, createFn: (A) -> T): Arb<T> = arbA.map(createFn)
-
 fun <A, B, T> Arb.Companion.bind(arbA: Arb<A>, arbB: Arb<B>, createFn: (A, B) -> T): Arb<T> = object : Arb<T> {
    override fun edgecases(): List<T> = emptyList()
 
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/bind.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/bind.kt
@@ -1,13 +1,27 @@
 package io.kotest.property.arbitrary
 
-fun <A, T> Arb.Companion.bind(arbA: Arb<A>, createFn: (A) -> T): Arb<T> = arb {
-   createFn(arbA.single(it))
-}
+import io.kotest.property.RandomSource
+import io.kotest.property.Sample
 
-fun <A, B, T> Arb.Companion.bind(arbA: Arb<A>, arbB: Arb<B>, createFn: (A, B) -> T): Arb<T> = arb {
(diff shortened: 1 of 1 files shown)


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/bind.kt
```diff
@@ -1,48 +1,116 @@
 package io.kotest.property.arbitrary
 
-fun <A, T> Arb.Companion.bind(arbA: Arb<A>, createFn: (A) -> T): Arb<T> = arb {
-   createFn(arbA.single(it))
-}
+import io.kotest.property.RandomSource
+import io.kotest.property.Sample
+
+fun <A, T> Arb.Companion.bind(arbA: Arb<A>, createFn: (A) -> T): Arb<T> = arbA.map(createFn)
```
dump247: Note the change here. Is this bind worth keeping since it is equivalent to map?

dump247:
I feel like these bind methods should be defined on the base interface `Gen` rather than `Arb`. Is that a reasonable change to make?

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/bind.kt
sksamuel: Probably not. Lets remove it.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/bind.kt
```diff
@@ -1,48 +1,116 @@
...
+      val b = bSamples.iterator()
+      val c = cSamples.iterator()
+
+      return sequence {
+         while (a.hasNext() && b.hasNext() && c.hasNext()) {
+            yield(Sample(createFn(a.next().value, b.next().value, c.next().value)))
+         }
+      }
+   }
 }
```
sksamuel, line 52: What's the advantage here, because it's going to propogate the random into the other arbs anyway? Is it just to avoid returning a new sequence each time that may have some performance penalty ?

dump247: There may be a performance advantage. However, the main reason I have done this is that the prior implementation results in issues for certain types of arbs. If the arb does any sort of initialization of the sequence, then the initialization occurs for each result of the bind.

For example, the following test would output duplicate values because `distinct` creates the set it collects results in each call.
```kotlin
forAll(100, Arb.bind(Arb.int(0, 10).distinct()) { it }) {
  println(it)
  true
}
```

The `regex` arb had the same issue before 711898e133bb5088c3cbe7cf250d9e8eb2ec91e2. The regex would always emit the same value to the bind because the Generex was reset each call to samples.

Another issue that I found is that `bind` was only calling `samples`, so it would never receive the edgecases from the arb. Maybe this was intended? If this is the intended behavior, it seems inconsistent to me, but I can change my PR to call samples instead of generate.

sksamuel: All seems good thanks :)

Issue #1243, opened 2020-02-27
Collection comparison matcher should print missing and unexpected elements


imanushin:
**Steps for reproduce:**
1. Write code below:
```
val expected: Set<String> = ...
val actual: Set<String> = ...

actual shouldBe expected
```
2. Use ~1000 elements for ```expected``` and ```actual``` collections
3. Execute the code and check the result

**Expected result**
System shows me, that are the difference in the collection: what elements are missing, what are unexpected. For list-based comparison - what elements aren't on the right places, etc.

**Actual result**
System shows me, that these tons of data is not the same with other tons of data.

PR #1254, opened 2020-03-02
Improve collection comparison matchers

imanushin:
Changes:

1. Added tests to see result of collections matches (e.g. exception text is verified for now).
2. Extended text
3. Minor refactoring (extract common parts, etc.).

**Important proposal:** probably it is better not to limit elements at ```joinToString``` calls (or use ```stringRepr``` for collection)?


imanushin:
Also - the following error exists in master:

```e: /home/runner/work/kotest/kotest/kotest-core/src/commonMain/kotlin/io/kotest/core/extensions/TagFilteredDiscoveryExtension.kt: (15, 31): Unsupported [This reflection API is not supported yet in JavaScript]```

sksamuel:
Looks good. I would probably increase the limit to 20 or something. I guess 10 is quite common. We just want to avoid printing out 200000 elements.

sksamuel:
Also I guess for some of them we can be even smarter. Like someList.shouldContainAll(1,2,3) should be able to print out which elements were missing.
We could merge this and continue to improve.

imanushin:
@sksamuel, to summarize, my proposal is:
1. Move code from ```getCollectionSnippet``` to ```stringRepl```. It means, that every printing function will use this parameter.
2. Increase limit to 20
3. If elements count more than 20 then append line ```and X more``` to show, that result is cut.
4. Push these changes to the current branch and only next merge all of them (to avoid partially-done work).

What do you think about this?

sksamuel:
Sounds great.

imanushin:
@sksamuel, done in this case. And will continue with pretty styling #1243 later.

However for current moment infinite ```Iterable``` leads to OOM exception, because I materialize list.

Materialization is needed to get total count, so let's start doing it.

Issue #1253, opened 2020-03-02
Add arb that generates lists


dump247:
Proposal to add a property test arb that generates lists of values.

Proposed API

```kotlin
fun <A> Arb<A>.chunked(size: IntRange): Arb<List<A>>
fun <A> Arb<A>.chunked(minSize: Int, maxSize: Int): Arb<List<A>>
```

dump247:
I am happy to implement this. I just wanted to get feedback before submitting an MR.

sksamuel:
There is `Arb.Companion.list(gen: Gen<A>, range: IntRange = 0..100)` already which I guess does hte same thing. You could delegate to this, or vice versa. I'm happy to add the new syntax.

dump247:
@sksamuel I did not see that method. Thank you for pointing it out.

Is this (and `list`) something that could or should be added to Gen rather than Arb? It's not entirely clear to me when a method should be added to Arb vs Gen, but this seems like something that would fit on Gen.

dump247:
A related question: `ListShrinker` does not take the min and max size into account when shrinking the list. I think the same thing applies to StringShrinker. Is this intentional?

sksamuel:
The shrinkers should take it into account. I guess the max doesn't matter as it's never going to shrink larger, but definitely min @dump247

sksamuel:
> @sksamuel I did not see that method. Thank you for pointing it out.
> 
> Is this (and `list`) something that could or should be added to Gen rather than Arb? It's not entirely clear to me when a method should be added to Arb vs Gen, but this seems like something that would fit on Gen.

I am thinking about tweaking the interfaces for beta2 and having Gen just be a marker interface with no methods. @Kerooker suggested this a while ago because at the moment the Exhaustive interface doesn't really do much other than be a marker itself.

sksamuel:
@dump247 what are your thoughts on that ?

dump247:
I do not really understand what marker interface means. Does that mean you remove `minIterations` and `generate` from Gen and it has no methods? Or that Gen and Exhaustive do not have any extension methods declared in kotest (everything is on Arb)?

sksamuel:
Exactly. Gen is just an interface to stop you passing Wibble to the forAll
methods. Wherever we do gen.generate at the moment, which is not in many
places, we just wrap it in a when. In other words we treat Exhaustive and
Arb as an ADT.

dump247:
[...]

PR #1256, opened 2020-03-03
Add Arb chunked extensions and Arb element vararg

dump247:
Also, fix the Arb list and set methods so they use a single sequence to generate
lists rather than constructing a new sequence for each list. A new sequence for
each list breaks for arbs that have some sequence initialization, such as distinct
and regex.


dump247:
This PR addresses issue #1253

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/collections.kt
```diff
@@ -33,11 +63,55 @@ fun <A> Arb.Companion.set(gen: Gen<A>, range: IntRange = 0..100): Arb<Set<A>> =
  * Shrinking is performed by removing elements from the list until the empty list.
  */
 @JvmOverloads
-fun <A> Arb.Companion.list(gen: Gen<A>, range: IntRange = 0..100): Arb<List<A>> = arb(ListShrinker()) { rs ->
-   val size = rs.random.nextInt(range)
-   gen.generate(rs).take(size).map { it.value }.toList()
+fun <A> Arb.Companion.list(gen: Gen<A>, range: IntRange = 0..100): Arb<List<A>> = object : Arb<List<A>> {
```
dump247, line 66: I changed `list` and `set` because the provided gen may be an arb with edgecases. In that case, something like this

```kotlin
forAll(100, Arb.list(myArbWithEdgeCases()) { testList -> 
   ...
}
```

would result in every provided testList starting with the edgecases.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/collections.kt
```diff
@@ -11,6 +14,12 @@ import kotlin.random.nextInt
  */
 fun <T> Arb.Companion.element(collection: Collection<T>): Arb<T> = arb { collection.random(it.random) }
 
+/**
+ * Returns an [Arb] whose values are chosen randomly from those in the supplied collection.
+ * May not cover all items. If you want an exhaustive selection from the list, see [Exhaustive.collection]
+ */
+fun <T> Arb.Companion.element(vararg collection: T): Arb<T> = arb { collection.random(it.random) }
```
dump247, line 21: This is an unrelated change, but I thought it would be a useful overload.

dump247: Assuming this overload is acceptable, it does seem like this is redundant with `constant`. The result of `Arb.element(a)` should be functionally equivalent to `Arb.constant(a)`. Would it be reasonable to remove `constant`?

sksamuel: At least make one a typealias. I think constant is handy for discoverability.

dump247: OK, I changed constant to call element.

Issue #1252, opened 2020-03-02
Deprecate/remove regex arb that accepts a Regex


dump247:
[Arb.Companion.regex(Regex)](https://github.com/kotest/kotest/blob/master/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/regex.kt#L18) was very confusing for me initially. I assumed that it would accept any regex, but it only works with a more restricted pattern language from the [mifmif Generex](https://github.com/mifmif/Generex) library. It also drops the regex options (e.g. case insensitive).

I would also suggest renaming [Arb.Companion.regex(String)](https://github.com/kotest/kotest/blob/master/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/regex.kt#L7) to `pattern` so it is more differentiated from a regex.

dump247:
I would be happy to make these changes. I created the issue to get some feedback first.

sksamuel:
I don't think it would be as discoverable if it was called pattern but I'm ok with it being renamed.

PR #1258, opened 2020-03-03
Refactor the regex arb to clarify that regex is not fully supported

dump247:
The Generex library supports only a very restricted subset of the java regular
expression language. I could not find any documentation on what that support is
explicitly.

issue #1252


kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/strings.kt
```diff
@@ -0,0 +1,23 @@
...
+import kotlin.random.asJavaRandom
+import kotlin.random.nextInt
+
+/**
+ * Generate strings that match the given pattern.
+ *
+ * The returned arb uses the [Generex](https://github.com/mifmif/Generex) library to generate strings. Generex
+ * supports a very restricted subset of regular expression constructs.
+ */
+fun Arb.Companion.stringPattern(pattern: String) = object : Arb<String> {
```
dump247, line 15: I can rename this back to regex if desired. I think `regex` is very misleading since the regular expression language support is very restricted and the caller does not get any indication if any unsupported regex structure is in the provided pattern. This certainly confused me up when I tried using it.

I could not find any documentation of what is supported by Generex or I would link to that in the doc.

PR #1266, opened 2020-03-05
Move blocking forAll functions into commonMain

ajalt:
There are a number of table testing functions that are defined only in jvmMain. This PR moves them to commonMain.

These functions were defined in the `io.kotlintest.tables` package. I moved them to `io.kotest.data.blocking` to fit with the rest of the table functions. They're in the `blocking` sub-package to differentiate them from the default `suspend` versions. I can move them to a different package if you'd like. I also split them up into separate files to match the changes made to the other table testing functions.

### Use cases

We should keep non-suspend versions of these functions so that they can be used by projects that use kotest-assertions but not kotest-core. Most other test runners don't support suspend functions.

For example, there are a number of projects at my workplace that use JUnit4 tests along with kotlintest-assertions for historical reasons.

For another example, Clikt makes extensive use of [table testing with a kotlin.test](https://github.com/ajalt/clikt/blob/master/clikt/src/commonTest/kotlin/com/github/ajalt/clikt/parameters/OptionTest.kt) runner. kotest-core doesn't support native targets, so I can't use it in Clikt. That project is stuck on 4.0.0.31-SNAPSHOT and can't update to the beta because of this issue.

### DIFF
--- a/kotest-assertions/src/commonMain/kotlin/io/kotest/data/blocking/forAll1.kt
+++ b/kotest-assertions/src/commonMain/kotlin/io/kotest/data/blocking/forAll1.kt
@@ -1,0 +1,18 @@
+package io.kotest.data.blocking
+
+import io.kotest.data.*
+import io.kotest.mpp.paramNames
+import kotlin.jvm.JvmName
+
+fun <A> forAll(vararg rows: Row1<A>, testfn: (A) -> Unit) {
+   val params = testfn.paramNames
+   val paramA = params.getOrElse(0) { "a" }
+   table(headers(paramA), *rows).forAll { a -> testfn(a) }
+}
+
+
+fun <A> forNone(vararg rows: Row1<A>, testfn: (A) -> Unit) {
+   val params = testfn.paramNames
+   val paramA = params.getOrElse(0) { "a" }
+   table(headers(paramA), *rows).forNone { a -> testfn(a) }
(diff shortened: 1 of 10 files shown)


sksamuel:
What about making these functions inline so they work for suspend vs non
suspend.

ajalt:
Making them inline would be nice, but that would prevent us from using `paramNames` (since you can't use reflection on an inlined lambda). If you're ok with just reporting parameters as "a", "b", "c", etc, we could go that route.

sksamuel:
So we already have inline for some in io.kotest package so we don't need to move these ones. The mistake I've made is not deprecating the old ones.
I believe we can have paramNames for jvm and a b c d for non-jvm.

ajalt:
Good points. I put the old definitions back in, and marked them all `@Deprectated`. I removed the unnecessary inline functions in the blocking package.

sksamuel:
Sorry I wasn't very clear. I don't think we need the blocking versions because we can just make the io.kotest.data forAll inline (where they are not already) so they'll work in both scenarios?

LeoColman:
The issue with inlining is compile time. Be careful with that

sksamuel:
We can't inline anyway. I just had a play and I just discovered exactly what @ajalt said about reflection. If only I had read his comment properly :)

Merging this.

PR #1267, opened 2020-03-06
Add Arb that generates random java.time.Instant

dump247:


kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
```diff
@@ -109,3 +111,34 @@ fun Arb.Companion.localDateTime(
...
+      }.random(random)
+
+      return Instant.ofEpochSecond(seconds, nanos.toLong())
+   } catch(e: IllegalArgumentException) {
+      throw NoSuchElementException(e.message)
+   }
+}
+
+/**
+ * Arberates a stream of random [Instant]
```
dump247, line 135: I can not find arberate in a dictionary :).  I copied to from the docs above.

PR #1281, opened 2020-03-12
Improve shouldBe matcher to support Regex comparison and better formatted error for map

ashishkujoy:


kotest-assertions/src/commonMain/kotlin/io/kotest/matchers/should.kt
```diff
@@ -1,26 +1,19 @@
...
 infix fun <T, U : T> T.shouldBe(any: U?) {
    when (any) {
       is Matcher<*> -> should(any as Matcher<T>)
       else -> {
          AssertionCounter.inc()
-         if (this == null && any != null) {
-            ErrorCollector.collectOrThrow(equalsError(any, this))
-         } else if (!compare(this, any)) {
-            ErrorCollector.collectOrThrow(equalsError(any, this))
+         val throwable = when (this) {
```
sksamuel, line 11: I think the first case in the when should be the check if one is null and one isn't (as they definitely can't be equal).

sksamuel (changes requested):
One small change to pull the null check into the parent when.

kotest-assertions/src/commonMain/kotlin/io/kotest/matchers/should.kt
ashishkujoy: Pulled checking of actual value is null but expected value is not null in parent

PR #1289, opened 2020-03-14
issues/1285: Fix containInOrder

ClaasJG:
Ok, I hope I managed it this time.

I also added 'shouldNotContainInOrder'.

I (ironically) had problems running some tests, but run the added tests separately.

Claas


sksamuel:
Nice job.

PR #1294, opened 2020-03-15
issues/1284: Fix containInOrder matcher message

### LINKED ISSUE #1284: Improve error messages with vararg arguments (for string Matcher shouldContainInOrder)
Hi,

i am using ' 4.0.0-BETA2 '.

Current behavior:
`"abc".shouldContainInOrder("c", "a", "b)`
reports:
`abc should include substrings [Ljava.lang.String;@4b03e083 in order`.

It should report the message:
`abc should include substrings c, a and b in order`. (Or similar, but the substrings should be visible)

Proposed fix:
`shouldContainInOrder` calls [io/kotest/assertions/show/Show.kt#L17](https://github.com/kotest/kotest/blob/525df298325779833c1e80919dc43afc963dad04/kotest-assertions/src/commonMain/kotlin/io/kotest/assertions/show/Show.kt#L17) internally which should probably have special handling for arrays.

I am not familiar with the Codebase, but when this seems a responsable fix, I am going to submit a pull request.

Thx for kotest

-Claas

ClaasJG:
This would close https://github.com/kotest/kotest/issues/1284 .

The `containInOrder` matcher now uses `io.kotest.assertions.stringRepr` instead of `io.kotest.assertions.show.show` to format the vararg parameters. This behavior is parallel to the implementation of `io.kotest.matchers.collections.containExactly`.

-Claas

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/string/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/string/matchers.kt
@@ -1,6 +1,7 @@
 package io.kotest.matchers.string
 
 import io.kotest.assertions.show.show
+import io.kotest.assertions.stringRepr
 import io.kotest.matchers.*
 import io.kotest.matchers.neverNullMatcher
 import io.kotest.matchers.string.UUIDVersion.ANY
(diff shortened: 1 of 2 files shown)


sksamuel:
Thanks for the PR.
Show is the preferred mechanism moving foward and stringRepr will be removed. Was there a reason you preferred the latter? Whatever it was, I can add it to show.

ClaasJG:
Yes, the reason is, that show had no special handling for arrays, i.e. it does not display the actual sub strings but `[Ljava.lang.String;@4b03e083` because it calls toString on the array. (See: https://github.com/kotest/kotest/issues/1284 )

sksamuel:
Make sense. I'm about to put a PR up that overhauls the way show / stringrepr works in comparisons. After that is merged, you can rebase this and use show and we should be good.

sksamuel:
Latest show stuff is up on master if you want to take a look at how it impacts this PR.

ClaasJG:
I would say https://github.com/kotest/kotest/commit/94fe709ca7361a87ebbaa37fa42e60c65cad5878 closes https://github.com/kotest/kotest/issues/1284 and makes this pr obsolete. Thanks!

PR #1315, opened 2020-03-28
Add String?.shouldBeInteger([Int]): Int

SerVB:
Hi! I frequently do something like
```kotlin
val int = str?.toIntOrNull()
int.shouldNotBeNull()

// work with int
```

So I've defined `shouldBeInteger` function in my project and now it's like
```kotlin
val int = str.shouldBeInteger()

// work with int
```
 
Is it OK to add this function to the repo?

### DIFF
--- a/doc/matchers.md
+++ b/doc/matchers.md
@@ -70,6 +70,7 @@ For Android-specific matchers, take a look [here](android_matchers.md)
 | `str.shouldContainADigit()` | Asserts that the string contains at least one digit. |
 | `str.shouldContainIgnoringCase(substring)` | Asserts that the string contains the substring ignoring case. |
 | `str.shouldContainOnlyDigits()` | Asserts that the string contains only digits, or is empty. |
+| `str.shouldBeInteger([radix])` | Asserts that the string contains an integer and returns it. |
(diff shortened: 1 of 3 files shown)


sksamuel:
Great idea, can we use contracts to smart case the type ?

SerVB:
You mean a contract for non-null like the following?

```kotlin
@OptIn(ExperimentalContracts::class)
fun String?.shouldBeInteger(radix: Int = 10): Int {
    contract {
        returns() implies (this@shouldBeInteger != null)
    }

    return when (this) {
        null -> throw failure("String is null, but it should be integer.")

        else -> when (val integer = this.toIntOrNull(radix)) {
            null -> throw failure("String '$this' is not integer, but it should be.")

            else -> integer
        }
    }
}
```

It doesn't make much sense to me because after this operation I don't work with the initial string. However, if you think it could be useful, I can add it.

sksamuel:
Not just null, but I wonder if we can use smart cast from string to int ?

a.shouldBeInt()
a % 3 shouldBe 1

SerVB:
I think smart casting to `Int` is impossible here because there is still a `String` in `a` after this function returns.

But smart cast from `String?` to `String` is possible.

sksamuel:
Ok makes sense

SerVB:
Added the contract.

PR #1316, opened 2020-03-29
Make JSON and pitest tests work

SerVB:
I've just found out that I can't run JSON tests (by the way, I'm preparing my pull request for JSON functions 😉).

I see the same for CI, for example, for my another PR: https://github.com/kotest/kotest/pull/1315/checks?check_run_id=542151737

image

I've searched for other "Executed 0 tests" entries: there are 3 more, but they seem just to not have any tests.

The solution I propose is to switch from multiplatform plugin to the jvm one. It works. Also, there are no other source sets but jvm so it makes sense I think.

However, if there is another solution, I'll be totally OK with it.


sksamuel:
But we might make the JSON matchers work for other platforms but instead of using Jackson, they use something else.

There will be a reason they're not executing.

SerVB:
Okay, thanks, I've found the problem and another module not launching tests.

sksamuel:
Nice spot.

Issue #1313, opened 2020-03-27
Any kind of `.show()` on `java.nio.file.Path` causes a StackOverflowError


netvl:
Simple example:

```kotlin
import io.kotest.assertions.show.show

println(Paths.get("a/b/c").show().value)
```

fails with the following error:

```
java.lang.StackOverflowError
	at kotlin.collections.CollectionsKt___CollectionsKt.joinTo(_Collections.kt)
	at kotlin.collections.CollectionsKt___CollectionsKt.joinToString(_Collections.kt:2330)
	at kotlin.collections.CollectionsKt___CollectionsKt.joinToString$default(_Collections.kt:2329)
	at io.kotest.assertions.show.DefaultShowKt.getCollectionSnippet(DefaultShow.kt:45)
	at io.kotest.assertions.show.DefaultShowKt.access$getCollectionSnippet(DefaultShow.kt:1)
	at io.kotest.assertions.show.DefaultShow.show(DefaultShow.kt:26)
	at io.kotest.assertions.show.DefaultShow.show(DefaultShow.kt:27)
	at io.kotest.assertions.show.ShowKt.show(Show.kt:25)
	at io.kotest.assertions.show.DefaultShowKt.recursiveRepr(DefaultShow.kt:56)
	at io.kotest.assertions.show.DefaultShowKt$getCollectionSnippet$1.invoke(DefaultShow.kt:51)
	at io.kotest.assertions.show.DefaultShowKt$getCollectionSnippet$1.invoke(DefaultShow.kt)
	at kotlin.text.StringsKt__StringBuilderKt.appendElement(StringBuilder.kt:58)
	at kotlin.collections.CollectionsKt___CollectionsKt.joinTo(_Collections.kt:2313)
	at kotlin.collections.CollectionsKt___CollectionsKt.joinToString(_Collections.kt:2330)
	at kotlin.collections.CollectionsKt___CollectionsKt.joinToString$default(_Collections.kt:2329)
	at io.kotest.assertions.show.DefaultShowKt.getCollectionSnippet(DefaultShow.kt:45)
...
```

This is probably caused by the fact that `Path` is an `Iterable<Path>`.

This problem makes many of the matchers working with `Path`s unusable, because they want to print the `Path` somehow, which invariably causes a StackOverflowError. For example, something like

```kotlin
val actual: List<Path> = ...
actual.shouldContainExactlyInAnyOrder(
    Paths.get("a/b/c"),
    Paths.get("d/e/f")
)
```

[...]

PR #1317, opened 2020-03-29
Handle show for java.nio.file.Path

daddyksen:
Suggestion for https://github.com/kotest/kotest/issues/1313

### DIFF
--- a/kotest-assertions/src/commonMain/kotlin/io/kotest/assertions/show/DefaultShow.kt
+++ b/kotest-assertions/src/commonMain/kotlin/io/kotest/assertions/show/DefaultShow.kt
@@ -1,5 +1,7 @@
 package io.kotest.assertions.show
 
+import kotlin.reflect.KClass
+
 /**
  * A default implementation of [Show] that handles arrays, collections and primitives as well as nullable values.
  */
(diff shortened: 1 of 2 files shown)


kotest-assertions/src/commonMain/kotlin/io/kotest/assertions/show/DefaultShow.kt
```diff
@@ -55,3 +57,7 @@ private fun Collection<*>.getCollectionSnippet(): Printed {
 internal fun recursiveRepr(root: Any, node: Any?): Printed {
    return if (root == node) "(this ${root::class.simpleName})".printed() else node.show()
 }
+
+private fun handleRecursiveIterable(i: Iterable<*>, kClass: KClass<out Any>) : Any {
+   return if (i.iterator().hasNext() && kClass.isInstance(i.iterator().next())) i.toString() else i.toList()
```
sksamuel, line 62: Can you add a comment to this function explaining how it works because it's not clear to me.

daddyksen: I've been thinking over it a little bit more and recognized that this solution won't work properly :(

sksamuel:
I've applied a fix to master.

Issue #1106, opened 2019-12-03
Json matcher should output errors in the intellij "diff" format


sksamuel:

sschuberth:
Has there been a regression? For me, IntelliJ IDEA 2021.2.3 (with Kotest plugin 1.1.43-IC-2021.1) does not show the usual "View assertEquals Difference" context menu on failed tests that use `shouldEqualJson()`.

sksamuel:
This is fixed again now

PR #1112, opened 2019-12-08
Update Json matcher error messages

pt2121:
Using intellij "diff" format. Closes #1106

PR #1318, opened 2020-03-29
More JSON test functions

SerVB:
Hi! Json testing kotest provides covers some of my cases and that's great! I especially like that every JSON value is a string because there is no convertion code so it looks concise.

However, many of my cases are not deterministic so I've needed to use parsers. I want to use kotest-styled functions!

Here I try to make existing functions more handy (use typealiases for self-documenting, add support for nullability with contracts and provide return values). And I add other functions devoted mainly to extract values from JSONs.

I use these changes in my project and I want to add them to the main repo.

TODO list for this PR:
- [x] Propose a PR.
- [x] Depending on the result of #1316, rebase.
- [x] Make tests for changes.
- [x] Update documentation.
- [ ] Resolve TODOs left in code.

Also, I can propose a slightly different style of JSON formatting but it isn't compatible to the current version so I will have to change some of existing tests. I have some questions:  
- Do we need to leave error messages unchanged?
- Why it has been chosen to limit length of output JSONs?
- Where can I learn more about "IntelliJ diff format" (#1106)?

What do you think? Let's discuss!

### DIFF
--- a/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
@@ -33,7 +33,7 @@ inline fun <reified T : Any?> Json?.shouldBeJsonValueOfType(): T {
             }
         }
 
-        return mapper.readValue(this, T::class.java)
+        return publishedMapper.readValue(this, T::class.java)
     } catch (thrown: Throwable) {
         throw failure("bad type of ${this.representation} - expected: ${typeOf<T>()}", cause = thrown)
     }
--- a/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
@@ -11,7 +11,8 @@ import kotlin.reflect.typeOf
  * We need to distinct `null` and `"null"`. The first one means there is no JSON at all, the second is null in JSON.
  * So we can't do just `Json?.toString()`.
  */
-val Json?.representation
+@PublishedApi
+internal val Json?.representation
     get(): String = when (this) {
         null -> null.toString()
 
--- a/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
@@ -1,6 +1,5 @@
 package io.kotest.assertions.json
 
-import com.jayway.jsonpath.JsonPath
 import io.kotest.assertions.failure
 import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.contract
@@ -22,17 +21,20 @@ val Json?.representation
 internal fun Json.limitLength() = if (this.length < 50) this.trim() else this.substring(0, 50).trim() + "..."
 
 @OptIn(ExperimentalContracts::class, ExperimentalStdlibApi::class)
-inline fun <reified T> Json?.shouldBeOfType(): T {
+inline fun <reified T : Any?> Json?.shouldBeJsonValueOfType(): T {
     contract {
-        returns() implies (this@shouldBeOfType != null)
+        returns() implies (this@shouldBeJsonValueOfType != null)
     }
 
-    return when (val decoded: Any? = JsonPath.read(this, "")) {
-        is T -> decoded
+    try {
+        if (this != null && typeOf<T>() == typeOf<String>()) {
+            check(this.length >= 2 && this.first() == '"' && this.last() == '"') {
+                "JSON string value should contain opening and closing double quotes"
+            }
+        }
 
-        else -> throw failure(
-            "bad type of ${this.representation} - " +
-                    "expected: ${typeOf<T>()} but was: ${decoded?.javaClass?.name ?: "Nothing?"}"
-        )
+        return mapper.readValue(this, T::class.java)
+    } catch (thrown: Throwable) {
+        throw failure("bad type of ${this.representation} - expected: ${typeOf<T>()}", cause = thrown)
     }
 }
--- a/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
@@ -11,19 +11,12 @@ import kotlin.reflect.typeOf
  *
  * We need to distinct `null` and `"null"`. The first one means there is no JSON at all, the second is null in JSON.
  * So we can't do just `Json.toString()`.
- *
- * TODO: Should we leave error messages unchanged?
- *       It's more logical to just add quotes to non-nullable JSONs like this but it will "break" old tests:
- *         val Json?.representation get(): String? = when (this) {
- *             null -> null.toString()
- *             else -> "\"$this\""
- *         }
  */
 val Json?.representation
     get(): String? = when (this) {
-        "null" -> "'null'"
+        null -> null.toString()
 
-        else -> this
+        else -> "\"$this\""
     }
 
 internal fun Json.limitLength() = if (this.length < 50) this.trim() else this.substring(0, 50).trim() + "..."
--- a/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
@@ -9,10 +9,10 @@ import kotlin.reflect.typeOf
  * Returns encoded JSON.
  *
  * We need to distinct `null` and `"null"`. The first one means there is no JSON at all, the second is null in JSON.
- * So we can't do just `Json.toString()`.
+ * So we can't do just `Json?.toString()`.
  */
 val Json?.representation
-    get(): String? = when (this) {
+    get(): String = when (this) {
         null -> null.toString()
 
         else -> "\"$this\""
--- a/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
@@ -1,0 +1,45 @@
+package io.kotest.assertions.json
+
+import com.jayway.jsonpath.JsonPath
+import io.kotest.assertions.failure
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
+import kotlin.reflect.typeOf
+
+/**
+ * Returns encoded JSON.
+ *
+ * We need to distinct `null` and `"null"`. The first one means there is no JSON at all, the second is null in JSON.
+ * So we can't do just `Json.toString()`.
+ *
+ * TODO: Should we leave error messages unchanged?
+ *       It's more logical to just add quotes to non-nullable JSONs like this but it will "break" old tests:
+ *         val Json?.representation get(): String? = when (this) {
+ *             null -> null.toString()
+ *             else -> "\"$this\""
+ *         }
+ */
+val Json?.representation
+    get(): String? = when (this) {
+        "null" -> "'null'"
+
+        else -> this
+    }
+
+internal fun Json.limitLength() = if (this.length < 50) this.trim() else this.substring(0, 50).trim() + "..."
+
+@OptIn(ExperimentalContracts::class, ExperimentalStdlibApi::class)
+inline fun <reified T> Json?.shouldBeOfType(): T {
+    contract {
+        returns() implies (this@shouldBeOfType != null)
+    }
+
+    return when (val decoded: Any? = JsonPath.read(this, "")) {
+        is T -> decoded
+
+        else -> throw failure(
+            "bad type of ${this.representation} - " +
+                    "expected: ${typeOf<T>()} but was: ${decoded?.javaClass?.name ?: "Nothing?"}"
+        )
+    }
+}
--- a/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonCountableElement.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonCountableElement.kt
@@ -1,0 +1,8 @@
+package io.kotest.assertions.json
+
+sealed class JsonCountableElement {
+
+    class JsonKeyValuePairs internal constructor(val count: Int) : JsonCountableElement()
+}
+
+val Int.jsonKeyValueEntries get() = JsonCountableElement.JsonKeyValuePairs(this)
--- a/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonMatchers.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonMatchers.kt
@@ -107,7 +107,7 @@ fun containJsonKey(path: JsonKey) = object : Matcher<Json> {
 }
 
 fun <T> Json?.shouldContainJsonKeyValue(path: JsonKey, value: T) = this should containJsonKeyValue(path, value)
-fun <T> Json?.shouldNotContainJsonKeyValue(path: JsonKey, value: T) = this shouldNot containJsonKeyValue(path, value)
+fun <T> Json.shouldNotContainJsonKeyValue(path: JsonKey, value: T) = this shouldNot containJsonKeyValue(path, value)
 fun <T> containJsonKeyValue(path: JsonKey, t: T) = object : Matcher<Json?> {
     override fun test(value: Json?): MatcherResult {
         val sub = value?.limitLength()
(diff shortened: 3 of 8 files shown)


SerVB:
Do you know why tests failed?..

sksamuel:
> * Do we need to leave error messages unchanged?

No.

> * Why it has been chosen to limit length of output JSONs?

I guess to stop 100mb of json being included in the output

> * Where can I learn more about "IntelliJ diff format" (#1106)?

https://github.com/JetBrains/intellij-community/blob/5422868682d7eb8511dda02cf615ff375f5b0324/java/java-runtime/src/com/intellij/rt/execution/testFrameworks/AbstractExpectedPatterns.java

This is the format that intellij recognizes. If the output messages match then you get the "click to see difference" thing.

I'm not sure that's useful for large json files though.

sksamuel:
> Do you know why tests failed?..

Looks like date test failed. Something to do with day light savings maybe ?

SerVB:
> Looks like date test failed. Something to do with day light savings maybe ?

I've checked. It seems that randomly taking datetimes 500 times can practically give not all 60 seconds (the probability of 59 different seconds is `2e-4`). Maybe make it say 5K iterations to make the probability `3e-37`? 😀 

https://github.com/kotest/kotest/blob/dc84bd1058c8a7586920d4b85a8eabe591985369/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/DateTest.kt#L77-L100

I've rerun the build without changes and now it seems fine.

SerVB:
@sksamuel could you please rerun tests?

SerVB:
@sksamuel, I've almost finished with testing and will write documentation soon. Before it, it's better to decide which terminology should we use.

Actually, all the "keys" are JSON paths. However, there was `shouldContainJsonKey` function already, so I've used "key" word for naming "paths" everywhere. Is this OK or should I rename to "paths"? If the latter, should I rename only newly introduced functions?

sksamuel:
I think the terminology should be key for a single field, and path for a.b.c.
So I guess, shouldContainJsonKey("a") or shouldContainJsonPath("a", "b", "c")
I think if we just do path then it covers both cases?

SerVB:
There is no vararg variant, but there is `json.shouldContainJsonKey("$.store.bicycle")`.  It's a single argument, but it's a path.

So okay, let's leave naming as keys...

sksamuel:
What about if my field is called "store.bicycle" ?

SerVB:
Interesting case! I've checked the current implementation using https://jsonpath.com/ and got the following behavior:

```json
{
  "store.bicycle": "in field",
  "store": {
      "bicycle": "in object"
  }
}
```

Query: `store.bicycle`.

Result: `"in object"`.

Now it seems that it's better to forbid using JSON paths as keys. It will just make behavior more clear. It can break compatibility though. However, paths instead of keys is not a documented feature, if we look at current docs: <https://github.com/kotest/kotest/blob/master/doc/matchers.md>.

So should I drop support for paths? I can do it.

sksamuel:
I would do both:

shouldContainJsonPath("store.bicycle") => in object

shouldContainJsonField("store.bicycle") => in field

sksamuel:
Let me know when you want me to review this.

SerVB:
OK. It's still WIP. I will complete the TODO list and then ask for review. One more item:

- [ ] Split functions to keys and paths.

SerVB:
@sksamuel 

So I'm done with initial tests. Current TODO list:

- [x] Update documentation to initial version.
- [ ] Split functions to keys and paths.
- [ ] Make tests for split functions.
- [ ] Make tests for all supported JSON primitives (numbers, arrays, maps) – maybe use Arb in tests, so need to provide and test Arb for JSON too.
- [ ] Use specialized functions for exceptions not just `failure` ones.
- [ ] Update documentation to final version.
- [ ] Resolve TODOs left in code.

Seems like big amount of work :) So could you please start reviewing now, to find places I'm missing?

SerVB:
If you have any suggestions how to split the changes into multiple PRs, please say them.

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
```diff
@@ -0,0 +1,40 @@
...
+import kotlin.contracts.contract
+import kotlin.reflect.typeOf
+
+/**
+ * Returns encoded JSON.
+ *
+ * We need to distinct `null` and `"null"`. The first one means there is no JSON at all, the second is null in JSON.
+ * So we can't do just `Json.toString()`.
+ */
+val Json?.representation
```
LeoColman: Does this need to stand public? We can unclutter a little bit the `Json` namespace by making this private/internal

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonCountableElement.kt
```diff
@@ -0,0 +1,8 @@
+package io.kotest.assertions.json
+
+sealed class JsonCountableElement {
+
+    class JsonKeyValuePairs internal constructor(val count: Int) : JsonCountableElement()
+}
+
+val Int.jsonKeyValueEntries get() = JsonCountableElement.JsonKeyValuePairs(this)
```
LeoColman, line 8: Are both of these supposed to be public?

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonMatchers.kt
```diff
@@ -1,81 +1,208 @@
...
+}
+
+@OptIn(ExperimentalContracts::class)
+infix fun Json?.shouldNotMatchJson(json: Json?) {
+    contract {
+        returns() implies (this@shouldNotMatchJson != null)
+    }
+
+    this shouldNot matchJson(json)
+}
```
LeoColman: I don't think this contract is correct.

If I execute
```
null shouldNotMatchJson null
```

It seems that `null == null` will work, and thus both values will be null. This isn't what the contract states

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonMatchers.kt
```diff
@@ -1,81 +1,208 @@
...
+fun matchJsonResource(resource: String) = object : Matcher<Json> {
+
+    override fun test(value: Json): MatcherResult {
+        val actualJson = mapper.readTree(value)
+        val expectedJson = mapper.readTree(this.javaClass.getResourceAsStream(resource))
+
+        return MatcherResult(
+            actualJson == expectedJson,
+            "expected: $expectedJson but was: $actualJson",
+            "expected not to match with: $expectedJson but match: $actualJson"
```
LeoColman, line 59: Maybe this sentence could improve a little bit?
What about
```suggestion
            "expected not to match with: $expectedJson but did."
```

This way we don't print the same sentence twice. Unless there's something I'm not seeing

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonMatchers.kt
```diff
@@ -1,81 +1,208 @@
...
-  override fun test(value: String): MatcherResult {
-    val sub = if (value.length < 50) value.trim() else value.substring(0, 50).trim() + "..."
-    return MatcherResult(
-        JsonPath.read<T>(value, path) == t,
-        "$sub should contain the element $path = $t",
-        "$sub should not contain the element $path = $t"
-    )
-  }
+@OptIn(ExperimentalContracts::class)
+inline infix fun <reified T> Json?.shouldContainOnlyJsonKeyAndValueOfSpecificType(path: JsonKey): T {
```
LeoColman, line 201: This function name is huge!
Can we find a way to make it shorter?

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonTypes.kt
```diff
@@ -0,0 +1,18 @@
...
+ * * JSON null: `"null"`.
+ * * JSON object: `"{"id": 42}"`.
+ * * JSON array: `"["id", 42, null]"`.
+ */
+typealias Json = String
+
+/**
+ * String representing a key for JSON value. For example, to access `42` in `"{"id": 42}"`, use `"id"` as a key.
+ */
+typealias JsonKey = String
```
LeoColman, line 18: Would the user need to import these typealias?
I don't see anything wrong in having them in Kotest, but propagating them users' code seems a little bit odd to me

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonUtil.kt
```diff
@@ -0,0 +1,7 @@
+package io.kotest.assertions.json
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.registerKotlinModule
+
+// we need to make it public because inline functions use it
+val mapper by lazy { ObjectMapper().registerKotlinModule() }
```
LeoColman:
```suggestion
private val mapper by lazy { ObjectMapper().registerKotlinModule() }
@PublishedApi
internal fun getMapper() = mapper
```

I think this works. I rather not have this mapper leak to users' code

LeoColman (changes requested):

SerVB:
Oh, it's a nice feature of marking PR as a draft 😅

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonUtil.kt
SerVB: I've changed your suggestion a bit to avoid names collision of symbols and commited as 
3ded3ce.

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonTypes.kt
SerVB: If a user just calls functions which have these aliases as receivers or parameters, there is no need to import. If `Json` is inferred when assigned to a variable, no imports are needed too.

If a user wants to explicitly declare a symbol with `Json` type, then they should import it. However, using `String` instead of `Json` doesn't require imports and is allowed by the compiler without any warnings.

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonCountableElement.kt
SerVB: Do you suggest something like the following?
```kotlin
sealed class JsonCountableElement {

    internal class JsonKeyValuePairs(val count: Int) : JsonCountableElement()
}

val Int.jsonKeyValueEntries: JsonCountableElement get() = JsonCountableElement.JsonKeyValuePairs(this)
```

I have no preference here. If you suggest to restrict visibility of `JsonKeyValuePairs`, I'll do it.

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonConverters.kt
SerVB: I've made it internal in 2c2eebb using your suggestion about PublishedApi.

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonMatchers.kt
SerVB: Good catch, fixed in a365ed5b.

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonMatchers.kt
SerVB: I'm not the author here. Git annotate shows the following commit message:

> Update Json matcher error messages (#1106) (#1112)

Your proposal seems fine to me but maybe I'm not seeing something here too...

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonMatchers.kt
SerVB: I agree but can't make up any shorter suggestions...

sksamuel:
I'd love to get this merged in soon. What can I do to help.

SerVB:
@sksamuel , I'm not sure about priorities. Here https://github.com/kotest/kotest/pull/1318#issuecomment-622124006 I've mentioned tasks, maybe you can say which tasks can be done in another PR? Anyway, any help with these tasks will be appreciated, thank you. You can even commit to this PR if you want.

By the way, what is "soon"? I can finish this PR to the end of June.

sksamuel:
I was hoping to include it in 4.1 which I want to release within a week. We can move the documentation to another PR.

SerVB:
I'm not sure I can do something within the week, sorry. I'd rather not include this PR to 4.1.

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonMatchers.kt
```diff
@@ -1,81 +1,201 @@
...
+}
 
-infix fun String.shouldMatchJson(json: String) = this should matchJson(json)
-infix fun String.shouldNotMatchJson(json: String) = this shouldNot matchJson(json)
-fun matchJson(json: String) = object : Matcher<String> {
+infix fun Json.shouldMatchJsonResource(resource: String) = this should matchJsonResource(resource)
+infix fun Json.shouldNotMatchJsonResource(resource: String) = this shouldNot matchJsonResource(resource)
 
-  override fun test(value: String): MatcherResult {
+fun matchJsonResource(resource: String) = object : Matcher<Json> {
```
klesniewski: Thanks for making JSON matchers work on nullable values. I just reported issue with it (https://github.com/kotest/kotest/issues/1504) and now I noticed you are addressing it in your PR. I have noticed however, that in your changes some matchers should work with nullable values (`shouldMatchJson`, `shouldContainJsonKey`), but some don't (for example `shouldMatchJsonResource` and `shouldNotContainJsonKey` will not work on nullable strings). Could you change it to make all those matchers consistently work with nullable values?

SerVB: I just don't change this `matchJsonResource` matcher :) But thank you for noticing, if there is demand for it, I'll do it too.

Will continue work on this PR in about a week, planning to finish to the end of June.

klesniewski: That would be great :) Sorry to piggyback to your PR, but I figured it will be easier for everyone this way, instead of doing separate small PR and having to resolve merge conflicts afterwards.

SerVB: Hey @klesniewski , I've added nullable receiver for `shouldMatchJsonResource`. Is it enough?

I'm not adding nullable receiver support for the most of negated versions as it's not clear how to treat it. So I'm not adding nullable receiver for `shouldNotMatchJsonResource` now.

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonMatchers.kt
SerVB: I think I can just omit introducing this function, but introduce the one for checking and converting type: `str?.shouldBeOfJsonType(jsonType): T`. It will be even more functional.

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/JsonCountableElement.kt
SerVB: I've decided not to introduce this API. It's not intuitive enough. I've described the API design in the documentation in the recent commit.

SerVB:
Hey! I think there is much work to be done with JSON assertions. During my progress, I've rethinked the design and found some limitations. Now I understand the changes should be split into multiple PRs.

Here is the list of JSON refactoring (as I see it):

- #1504 (already fixed via #1521) – not a breaking change.
- Make typealiases for `Json`, `JsonPath` (`= String`). Update docs (add link to JsonTypes file). Make representation function (internal) – **a breaking change for assertion messages**.
- [From Slack](https://kotlinlang.slack.com/archives/CSYLDDZUY/p1592597599127700): `"1".shouldMatchJson("1.0")  // throws: "expected 1.0 but was 1"` should be made less strict – not a breaking change.
- Make `shouldCountainKey` return `Json` – not a breaking change.
- [From GH](https://github.com/kotest/kotest/pull/1318#issuecomment-616215993): Rename current `*Key*` matchers to `*Path*` and intoduce fair matchers for keys. Add `JsonKey` typealias – **a breaking change for  `*Key*` matchers**.
- New function (as described in commits of this PR) `str?.shouldHaveSize(int)` – not a breaking change.
- New function (as described in commits of this PR) `str?.shouldBeOfJsonType(jsonType): T` – not a breaking change.
- New function (as described in commits of this PR) `str?.shouldContainOnlyJsonKey("key"): str` – not a breaking change.

I'm now closing this PR and will gradually open others. I'm not managing to do this all in June as I've said, sorry.

Will update this comment with links to PRs.

PR #1324, opened 2020-03-30
Close ScanResult (#1323)

### LINKED ISSUE #1323: Discovery.scanUris leaks resource due to not closing ScanResult
When building with Kotest 4.0.1 and Maven surefire 2.2.2 `Discovery.discover` is called once per spec class identified by Surefire.

My build started failing because the OS ran out of file descriptors when the number of specs passed a certain number. 

The file descriptors turned out to be duplicate references to jars in the classpath.

The issue is that `Discovery.scanUris` in kotest-core is failing to close the ClassGraph `ScanResult` before returning the spec instances. This causes the number of open file descriptors to increase with the number of jars in your classpath for each single spec.

This can be fixed by changing 

```kotlin
      return scanResult
         .getSubclasses(Spec::class.java.name)
         .map { Class.forName(it.name).kotlin }
         .filterIsInstance<KClass<out Spec>>()
```

..to

```kotlin
      return scanResult.use { result ->
         result.getSubclasses(Spec::class.java.name)
         .map { Class.forName(it.name).kotlin }
         .filterIsInstance<KClass<out Spec>>()
      }
```

chrned:
Started closing `ScanResult` after usage to avoid resource leakage.
Failing to close the scan result can lead to running out of file descriptors when running on *nix OS.

I did not find a repeatable and platform independent way to automatically test this fix.
I verified it manually on my Mac by doing this

```kotlin
println((ManagementFactory.getOperatingSystemMXBean() as UnixOperatingSystemMXBean).openFileDescriptorCount)
```
It showed the number of open file descriptors increasing with the number of jars on the classpath for each individual spec discovered. After the fix this issue disappeared.

This is *nix/mac specific  data and is also affected by other processes running on the OS so not suited for automatic testing.

Fixes kotest/kotest#1323

### DIFF
--- a/kotest-core/src/jvmMain/kotlin/io/kotest/core/engine/discovery/Discovery.kt
+++ b/kotest-core/src/jvmMain/kotlin/io/kotest/core/engine/discovery/Discovery.kt
@@ -86,9 +86,10 @@ object Discovery {
       }
       log("Test discovery competed in $time")
 
-      return scanResult
-         .getSubclasses(Spec::class.java.name)
-         .map { Class.forName(it.name).kotlin }
-         .filterIsInstance<KClass<out Spec>>()
+      return scanResult.use { result ->
+         result.getSubclasses(Spec::class.java.name)
+            .map { Class.forName(it.name).kotlin }
+            .filterIsInstance<KClass<out Spec>>()
+      }
    }
 }


sksamuel:
Thanks, good spot. It's strange that surefire want's to perform a discovery for every single spec.

PR #1326, opened 2020-03-30
Caching result of scanUris (#1325)

### LINKED ISSUE #1325: Disocvery caching is defeated by Surefire
When building with Kotest 4.0.1 and Maven surefire 2.2.2 `Discovery.discover` is called once per spec class identified by Surefire.

I noticed that my Kotest initialization time increased linearly by number of specs.

This is caused by `Discovery.discover` performing a full scan of the classpath for each single spec:

```
...
Test discovery competed in 669ms
Scan discovered 50 classes in the classpaths...
After filters there are 1 spec classes
classSelectors=[com.greenbird.metercloud.integration.flowserver.systemtest.flow.http.RestSourceValidationSpec]
...
Test discovery competed in 794ms
Scan discovered 50 classes in the classpaths...
After filters there are 1 spec classes
After discovery extensions there are 1 spec classes
classSelectors=[com.greenbird.metercloud.integration.flowserver.systemtest.flow.http.HttpReverseProxyErrorHandlingSpec]
...
Test discovery competed in 657ms
Scan discovered 50 classes in the classpaths...
After filters there are 1 spec classes
classSelectors=[com.greenbird.metercloud.integration.flowserver.systemtest.flow.http.SoapEchoSpec]
...
```

Discovery is caching on `DiscoveryRequest` but this does not help as each request is unique.

This can be fixed by also caching the result of `scanUris`.

chrned:
Started caching the result of `Discovery.scanUris()` since Surefire creates a unique query per spec class.

Not sure how to automatically test this. 
I verified manually that the penalty of classpath scanning is now incurred only once instead of once per spec class. This significantly improves the initialization time for Maven projects with large classpaths and many specs. In my case by  50 * ~600ms.

Fixes kotest/kotest#1325

### DIFF
--- a/kotest-core/src/jvmMain/kotlin/io/kotest/core/engine/discovery/Discovery.kt
+++ b/kotest-core/src/jvmMain/kotlin/io/kotest/core/engine/discovery/Discovery.kt
@@ -28,6 +28,11 @@ object Discovery {
    private val isSpecSubclass: (KClass<*>) -> Boolean = { Spec::class.java.isAssignableFrom(it.java) }
    private val isAbstract: (KClass<*>) -> Boolean = { it.isAbstract }
    private val isClass: (KClass<*>) -> Boolean = { it.objectInstance == null }
+   private val fromClassPaths: List<KClass<out Spec>> by lazy {
+      scanUris().apply {
+         log("Scan discovered $size classes in the classpaths...")
+      }
+   }
 
    /**
     * Returns a function that applies all the [DiscoveryFilter]s to a given class.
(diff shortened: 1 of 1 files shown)


sksamuel:
Before we cache, can we check what parameters maven surefire is sending through to the discovery mechanism.

It might be better to just load the individual class if that's what maven is asking for. I think we used to do this.

chrned:
Added optimalization for loading of individual classes. Existing test seems to cover all cases still so did not update it.

sksamuel:
Nice

PR #1339, opened 2020-04-02
Restored opentest4j and junit4 exceptions

sksamuel:
@SerVB

### DIFF
--- a/kotest-assertions/src/jvmMain/kotlin/io/kotest/assertions/jvmfailures.kt
+++ b/kotest-assertions/src/jvmMain/kotlin/io/kotest/assertions/jvmfailures.kt
@@ -54,8 +54,8 @@ fun junit5AssertionFailedError(
 ): Option<Throwable> {
    return callPublicConstructor<Throwable>(
       "org.opentest4j.AssertionFailedError",
-      arrayOf(String::class.java, Object::class.java, Object::class.java),
-      arrayOf(null, expected.value.value, actual.value.value)
+      arrayOf(String::class.java, Object::class.java, Object::class.java, Throwable::class.java),
+      arrayOf(null, expected.value.value, actual.value.value, cause)
    )
 }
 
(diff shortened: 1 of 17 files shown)


kotest-assertions/src/jvmMain/kotlin/io/kotest/assertions/jvmfailures.kt
```diff
@@ -24,19 +29,87 @@ object Failures {
...
    return throwable
 }
 
 /**
  * Creates an [AssertionError] from the given message. If the platform supports nested exceptions, the cause
- * is set to the given [cause]. If the platform supports stack traces, then the stack is cleaned of `io.kotest`
- * lines.
+ * is set to the given [cause].
+ */
+actual fun createAssertionError(message: String, cause: Throwable?): AssertionError = AssertionError(message, cause)
```
ajalt, line 42: We should use a null message when constructing junit4/5 assertions. The exception already adds the `expected: x but was: y` text, so passing in our message will cause it to be duplicated.

https://github.com/junit-team/junit4/blob/435d41f0d45cfdbc1a38e1ad4eb1d5300da533f9/src/main/java/org/junit/Assert.java#L838-L853

https://github.com/junit-team/junit5/blob/master/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java#L111-L123

ajalt (changes requested):

kotest-assertions/src/jvmMain/kotlin/io/kotest/assertions/jvmfailures.kt
sksamuel: These are helper classes though right, and we're instantiating the exceptions directly.

SerVB:
Great but could you also add a short guide how to use these functions when proposing changes to kotest? For example, I'm doing JSON enhancements and I'm not sure if I should use these functions.

kotest-assertions/src/jvmMain/kotlin/io/kotest/assertions/jvmfailures.kt
ajalt: You're right, my apologies.

ajalt (approved):

sksamuel:
> Great but could you also add a short guide how to use these functions when proposing changes to kotest? For example, I'm doing JSON enhancements and I'm not sure if I should use these functions.

No one would use this functions directly, just the framework.

sksamuel:
I'll merge once the build passes and then we're good for 4.0.2 release.

SerVB:
Yep, I want to know how to use the functions when I develop the framework...

sksamuel:
Ah ok, then they're called by the failure(XX) methods, which are used to
"build" exceptions to throw.
Anywhere we need to throw some kind of assertion exception, we use
`failure` to create the appropriate one. So the usage of these new methods
should be restricted to failure only.

PR #1343, opened 2020-04-02
Add map,filter to Exhaustive

taer:
Adds map, filter, filterNot, and flatMap to Exhaustive to be similar to Arb

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/exhaustive/Exhaustive.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/exhaustive/Exhaustive.kt
@@ -31,3 +31,35 @@ fun <A> List<A>.exhaustive(): Exhaustive<A> = object : Exhaustive<A>() {
 fun <A, B : A> Exhaustive<A>.merge(other: Exhaustive<B>): Exhaustive<A> = object : Exhaustive<A>() {
    override val values: List<A> = this@merge.values.zip(other.values).flatMap { listOf(it.first, it.second) }
 }
+
+
+/**
+ * Returns a new [Exhaustive] which takes its elements from the receiver and filters them using the supplied
+ * predicate. This gen will continue to request elements from the underlying gen until one satisfies
+ * the predicate.
+ */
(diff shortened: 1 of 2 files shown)


kotest-property/src/commonMain/kotlin/io/kotest/property/exhaustive/Exhaustive.kt
```diff
@@ -31,3 +31,35 @@ fun <A> List<A>.exhaustive(): Exhaustive<A> = object : Exhaustive<A>() {
 fun <A, B : A> Exhaustive<A>.merge(other: Exhaustive<B>): Exhaustive<A> = object : Exhaustive<A>() {
    override val values: List<A> = this@merge.values.zip(other.values).flatMap { listOf(it.first, it.second) }
 }
+
+
+/**
+ * Returns a new [Exhaustive] which takes its elements from the receiver and filters them using the supplied
+ * predicate. This gen will continue to request elements from the underlying gen until one satisfies
+ * the predicate.
```
sksamuel: the readme should be changed to say

Returns a new [Exhaustive] which takes its elements from the receiver and filters them using the supplied
 predicate. In other words this exhaustive is a subset of the elements as determined by the filter.

sksamuel (approved):

kotest-property/src/commonMain/kotlin/io/kotest/property/exhaustive/Exhaustive.kt
taer: 👍   As you can see, I just copy/pasted the ARB's docs. :)

taer: updated

LeoColman (approved):

kotest-property/src/commonMain/kotlin/io/kotest/property/exhaustive/Exhaustive.kt
sksamuel: I noticed ;)

Issue #1353, opened 2020-04-05
Integration with Testcontainers


JoonasC:
[Testcontainers](https://www.testcontainers.org/) is a tool that can dynamically spin up docker containers to be used in tests.
This can be useful for spinning up database containers in integration tests for example.
Kotest integration with this tool would be nice.

Thanks.

sksamuel:
@ashishkujoy has opened a PR at testcontainers to add support there alongside spock / junit5.
You could vote on that issue., https://github.com/testcontainers/testcontainers-java/pull/2533

sksamuel:
We've now merged this into kotest itself due to testcontainers having higher priorities.

sksamuel:
Will be included in 4.1

dkowis:
Forgive me, is there documentation for how to actually use this? I found the classes in kotest-extension-testcontainers, but I'm not sure how to actually use it in one of my specs.

Thanks!

EDIT: just found it: https://github.com/kotest/kotest/blob/master/doc/extensions.md#testcontainers

PR #1360, opened 2020-04-06
#1353 Adds pertest and perspec listener for testcontainers

ashishkujoy:


sksamuel:
So, this is superb but two things

1. I think the listener should accept vararg containers so you can have multiple containers defined in code, and the listeners start / stops them all
2. I would prefer to see this added to the main test-containers project. They already have modules for spock, junit4 and 5. So it kinda makes sense to live there. Plus, it helps with visibiltiy on Kotest if other projects are adding modules.

ashishkujoy:
> So, this is superb but two things
> 
> 1. I think the listener should accept vararg containers so you can have multiple containers defined in code, and the listeners start / stops them all
> 2. I would prefer to see this added to the main test-containers project. They already have modules for spock, junit4 and 5. So it kinda makes sense to live there. Plus, it helps with visibiltiy on Kotest if other projects are adding modules.

Sure

sksamuel:
If they decline to allow it to live in their project we can then come back
and add it here.

ashishkujoy:
Adds a PR in testcontainers https://github.com/testcontainers/testcontainers-java/pull/2533

sksamuel:
Nice job :)

kotest-extensions/kotest-extensions-testcontainer/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+         dependencies {
+            implementation(project(":kotest-runner:kotest-runner-junit5"))
+            implementation("io.mockk:mockk:1.9.3")
+         }
+      }
+   }
+}
+
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+   kotlinOptions.freeCompilerArgs += "-Xuse-experimental=kotlin.Experimental"
```
SerVB, line 53: "Use experimental" is deprecated. Better to change with OptIn like there:
https://github.com/kotest/kotest/blob/16185aa6c5c6c30f7427cfd808338391fd29f96c/kotest-core/build.gradle.kts#L33

SerVB (commented):
That's a nice feature but it seems I can't use it. I have the following [code](https://github.com/SerVB/e-shop/blob/400b4e8fba2a80888351df4e124a0f5b5b841c3e/auth/src/test/kotlin/io/github/servb/eShop/auth/testContainerEShopAuth.kt#L11-L31):

```kotlin
private fun Application.testContainerEShopAuth() {
    val container = PostgreSQLContainer<Nothing>(POSTGRES_CONTAINER_NAME).apply {
        start()
    }

    module(
        dbPort = container.getMappedPort(PostgreSQLContainer.POSTGRESQL_PORT),
        dbUser = container.username,
        dbPassword = container.password,
        dbHost = container.containerIpAddress,
        dbDb = container.databaseName
    )
}

fun BehaviorSpec.givenTestContainerEShopAuth(test: suspend GivenContext.(TestApplicationEngine) -> Unit) {
    given("test container e-shop-auth") {
        withTestApplication(Application::testContainerEShopAuth) {
            this@given.test(this)
        }
    }
}
```

I create a testcontainer for Given branch and define a function to use it multiple times. Maybe I just don't see a way to rewrite it to testcontainers listeners, so it would be great if you give an advice.

sksamuel:
It might be worth closing this and reopening it with the same code as you PRed over at the test containers repo. It doesn't look like they're going to merge it anytime soon.

ashishkujoy:
Closing this in favour of #1479

PR #1363, opened 2020-04-08
Update docs related to defaultTestCaseConfig

lj-ditrapani:
It seems defaultTestCaseConfig changed from a val to a fun.
https://github.com/kotest/kotest/blob/master/kotest-core/src/commonMain/kotlin/io/kotest/core/spec/SpecConfigurationMethods.kt#L15


sksamuel:
Ah yeah, that's because it's on the interface now.
Thanks for the PR.

Issue #1364, opened 2020-04-08
Array.shouldContainExactly wraps expected array


thebrightspark:
I am trying to assert that an `Array<String>` contains exact values, for which I am using `shouldContainExactly`. Here's a basic example:
```kotlin
import io.kotest.core.spec.style.StringSpec
import io.kotest.matchers.collections.shouldContainExactly

class ArrayContainsTest : StringSpec({
    "array should contain exactly" {
        val array1 = arrayOf("one", "two", "three")
        val array2 = arrayOf("one", "two", "three")
        array1 shouldContainExactly array2
    }
})
```

The above example fails, where I'd expect it to pass. Here's the error logged:
```
java.lang.AssertionError: Collection should be exactly [["one", "two", "three"]] but was ["one", "two", "three"]
```
Is this a bug or is this expected behaviour and there's something different I should be doing?

I did check the equivalent from AssertJ to see how it's used there (as https://github.com/kotest/kotest/pull/360 mentions), and it appears as though `shouldContainExactly` should work as per my example above:
https://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/ListAssert.html#containsExactly-ELEMENT...-

For now I'll get around it using a more manual approach:
```kotlin
import io.kotest.assertions.assertSoftly
import io.kotest.core.spec.style.StringSpec
import io.kotest.matchers.shouldBe

class ArrayContainsTest : StringSpec({
    "array should contain exactly" {
        val array1 = arrayOf("one", "two", "three")
        val array2 = arrayOf("one", "two", "three")
        assertSoftly {
            array1.forEachIndexed { index, s ->
                s.shouldBe(array2[index])
            }
        }
    }
})
```

thebrightspark:
Ah, I've just noticed the issue :)

This function:
```kotlin
infix fun <T> Array<T>.shouldContainExactly(expected: Array<T>) = asList().shouldContainExactly(expected)
```
It's passing the `expected` array into a `vararg` param for `shouldContainExactly` - this one:
```kotlin
fun <T, C : Collection<T>> C?.shouldContainExactly(vararg expected: T) = this should containExactly(*expected)
```
So it's automatically getting wrapped into another array.

sksamuel:
Nice spot. Do you want to prepare a quick PR ?

thebrightspark:
Yep was just trying to prepare that now!

ashishkujoy:
[...]

PR #1367, opened 2020-04-08
Fix shouldContainExactly for Arrays (#1364)

thebrightspark:
This is to fix the issue https://github.com/kotest/kotest/issues/1364

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/matchers.kt
@@ -123,18 +123,15 @@ fun <T, C : Collection<T>> contain(t: T) = object : Matcher<C> {
    )
 }
 
-infix fun <T> Array<T>.shouldNotContainExactly(expected: Array<T>) = asList().shouldNotContainExactly(expected.asList())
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/matchers.kt
```diff
@@ -123,18 +123,15 @@ fun <T, C : Collection<T>> contain(t: T) = object : Matcher<C> {
    )
 }
 
-infix fun <T> Array<T>.shouldNotContainExactly(expected: Array<T>) = asList().shouldNotContainExactly(expected.asList())
+fun <T> Array<T>?.shouldNotContainExactly(vararg expected: T) = this?.asList() shouldNot containExactly(*expected)
```
sksamuel, line 128: Do you think we should keep the original as well ?

thebrightspark: I wanted to, but the method signitures conflict since `expected: Array<T>` and `vararg expected: T` both resolve to an array parameter. Decided to go with the vararg since an array can simply still be passed in using the spread operator.

If we want both then one of the method will need a different name using `@JvmName` I believe.

sksamuel:
It's fine to use @JvmName

sksamuel:
:+1:

PR #1385, opened 2020-04-15
Added better quick start guide, explaining difference between the sub…

sksamuel:
…projects and modules

### DIFF
--- a/doc/reference.md
+++ b/doc/reference.md
@@ -15,16 +15,33 @@ Starting with release 4.0 **KotlinTest** was renamed to **Kotest** to avoid conf
 There is an upgrade cost. Please be prepared when you upgrade that you will need to do more work than updating the versions in your build file. Mostly this is updating imports across your project, but some function definitions have also changed to support multiplatform builds (for instance Kotlin Durations are now used instead of Java Durations).
 
 
-How to use
+Getting Started
 ----------
 
-Kotest is published to Maven Central so you can get the latest version from the little badge at the top of the readme.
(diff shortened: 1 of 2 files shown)


ashishkujoy (commented):
Looks good to me. Just a thought should we add a separate page dedicated to all type of Kotest modules such as kotest-assertion-json, kotest-assertion-klock, kotest-assertion-sql, kotest-assertion-ktor etc along with the junit runner, and assertion module, so that people will know that they need to add those dependencies in their build file.

doc/reference.md
```diff
@@ -97,7 +117,7 @@ For maven you must configure the surefire plugin for junit tests.
 </plugin>
 ```
 
-And then add the Kotest JUnit5 runner to your build.
+And then add the Kotest JUnit5 runner to your build to use the framework product.
```
LeoColman, line 120: I think we want to recommend the console runner even more than JUnit5, perhaps we should suggest it?

LeoColman (approved):

sksamuel:
> Looks good to me. Just a thought should we add a separate page dedicated to all type of Kotest modules such as kotest-assertion-json, kotest-assertion-klock, kotest-assertion-sql, kotest-assertion-ktor etc along with the junit runner, and assertion module, so that people will know that they need to add those dependencies in their build file.

Good idea. I'll add another page for that.

doc/reference.md
sksamuel: The console runner can't be used for gradle, only for the plugin. We can clarify once we release the plugin.

PR #1395, opened 2020-04-20
Resolve #1362. Add parallel testCase execution in one Spec

### LINKED ISSUE #1362: Parallel execution test cases in Spec
I think this is very good feature to give an opportunity to configurate and execute cases inside spec parallel. For example:
We have Test Spec "Lalala" where two cases - "First test case" {body}  "Second test case" {body}
I want to execute it parallel(or not - for it we need some config for spec)

In 3.4.2 i make it with it:
in class SingleInstanceSpecRunner i rewrite method execute

`override fun execute(spec: Spec, topLevelTests: TopLevelTests): Map<TestCase, TestResult> {
       var listOfCases: List<Deferred<Any>> = ArrayList()

        runBlocking {
            interceptSpec(spec) {
                topLevelTests.tests.forEach { topLevelTest ->
                    // each spec is allocated it's own thread so we can block here safely
                    // allowing us to enter the coroutine world
                    val async = this.async {
                        executor.execute(
                            topLevelTest.testCase,
                            Context(topLevelTest.testCase.description, this.coroutineContext)
                        ) { result -> results[topLevelTest.testCase] = result }
                    }
                    listOfCases = listOfCases + async
                }
                listOfCases.awaitAll()
            }
        }
        return results
    }`

And configure executor before it with config with number of threads (automatic or manual), that i need

Can you make something like that for 4.0 version ? I think it is useful

crazyk2:
My English is poor so I offer you create doc and I would pleasure edit it If you don't mind.
I tried to create tests to cover all cases and these tests are self-documented


crazyk2:
@sksamuel Will you review this request?

sksamuel:
I will look over it this weekend. It is a complicated change.

SergKhram:
@sksamuel Hello, you see it ? We very want it)))

crazyk2:
Ping

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+import io.kotest.matchers.ints.shouldBeExactly
+import io.kotest.matchers.ints.shouldBeGreaterThan
+import io.kotest.matchers.maps.shouldHaveSize
+import io.kotest.matchers.shouldBe
+import kotlinx.coroutines.delay
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.getOrSet
+
+class SpecThreadSingleInstanceTest : FunSpec({
```
sksamuel, line 21: We need a test for threads > 1 and threads < test count.

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+
+   test("lock should be locked too") {
+      delay(300)
+      shouldThrow<AssertionError> {
+         lock.isLocked shouldBe false
+      }
+   }
+})
+
+class SpecThreadWithNestedTestSingleInstanceTest : FunSpec({
```
sksamuel, line 107: I would like to see a test with multiple levels of nesting for every isolation mode.

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+
+   isolation = IsolationMode.SingleInstance
+   threadsForSpec = 3
+
+   val multipleThreadCounter =
+      PersistentThreadLocal<Int>()
+
+   afterSpec {
+      assertSoftly {
+         multipleThreadCounter.map.shouldHaveSize(2)
```
sksamuel: We have three threads so why wouldn't this be 3. Are you expecting that only root level test are parallelized ?

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+         multipleThreadCounter.map.values.sum() shouldBe 4
```
sksamuel: This doesn't assert which threads added what.
I would append characters to a string, and then assert that unique string forms are made.

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+
+      test("test should lock object") {
+         lock.lock()
+         delay(1000)
+         lock.unlock()
+      }
+
+      test("lock should be unlocked") {
+         delay(300)
+         lock.isLocked shouldBe false
```
sksamuel, line 190: Please explain this test to me.

sksamuel (commented):
I've added some comments. I also don't see any tests for instance per leaf and instance per test runners.

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
crazyk2: I don't understand your idea. I already did test ` a test for threads > 1` All tests in this file `SingleInstanceTest.kt` are test with `threadsForSpec = 3` and there is example of ` threads < test count`. For example  `SpecThreadSingleInstanceTest` threads=3 and tests = 6

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
crazyk2: `Are you expecting that only root level test are parallelized`
Exactly. We have only two root tests:
`context("First single thread context")`
`context("Second single thread context")`
so there is only two threads will be exploited.
About root test we discussed [here](https://github.com/kotest/kotest/issues/1362#issuecomment-612313721)

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
crazyk2: I agree and of course I can change for readability but there are only 4 tests and each test will be invoked only once and I assert this behavior. But if you have a will I will change this. just say a word

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
crazyk2: This test assert that all nested tests (all tests from the same context) run sequential by one thread.
If compare this tests with similar tests from `InstancePerTestTest.kt` or `InstancePerLeafTest.kt` you will see difference.

crazyk2 (commented):
> I also don't see any tests for instance per leaf and instance per test runners.

Why?
[click me](https://github.com/kotest/kotest/pull/1395/files)
Ctrl+F -> Type `Load diff` -> Click `Load diff`
 [InstancePerLeafTest](https://github.com/kotest/kotest/blob/f297b0ee018936e7d69493841207283a0cfa619a/kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/InstancePerLeafTest.kt)
[InstancePerTestTest](https://github.com/kotest/kotest/blob/f297b0ee018936e7d69493841207283a0cfa619a/kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/InstancePerTestTest.kt)

SergKhram:
ping @sksamuel @crazyk2

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
sksamuel: Where is tests = 6 in this file ?

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
sksamuel: I guess it would be possible for the first test to run twice, and the second test not to run at all, and you'd still get the answer = 2.
If you do + "a" and + "b" then you know you will have "ab". It's safer.

kotest-core/src/commonMain/kotlin/io/kotest/core/spec/configuration.kt
```diff
@@ -47,6 +47,12 @@ abstract class TestConfiguration {
     */
    var defaultTestConfig: TestCaseConfig? = null
 
+   /**
+    * Returns the number of concurrent root test cases of Spec that can be executed.
+    * Defaults to 1.
+    */
+   var threadsForSpec: Int = 1
```
sksamuel: I would prefer this to be called `threads` and the doc should read

"Returns the size of the thread pool that is used to execute root tests in this spec.
Note: This only applies to the JVM platform and has no effect on other platforms."

Because we may decide later to allow concurrency even with a single thread, for instance if you are testing coroutine code, you don't necessarily need > 1 thread.

sksamuel:
Ok this is good. I've added two new comments. If we address those and fix the merge conflicts, this can be merged and included in 4.1.

kotest-core/src/commonMain/kotlin/io/kotest/core/spec/configuration.kt
crazyk2: renaming - done.
Docs is not my area. Sorry for English.

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
crazyk2: Done

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
crazyk2: Done

kotest-tests/kotest-tests-core/src/jvmTest/kotlin/com/sksamuel/kotest/core/runtime/parallel/SingleInstanceTest.kt
crazyk2: SpecThreadSingleInstanceTest
Done

kotest-core/src/commonMain/kotlin/io/kotest/core/spec/configuration.kt
crazyk2: On my opinion `threads` for spec and `threads` for test will confuse users
```kotlin
class Test : FunSpec({
   
   threads = 3

   test("a").config(threads = 2) {
   }
}
```
First `threads` like default config for Spec class and second `threads` like config just for this test case
And users may forget about 
```kotlin
defaultTestConfig = TestCaseConfig(threads = 3)
```
are different configuration parameters

sksamuel:
Maybe we call it parallelism then.

crazyk2:
> Maybe we call it parallelism then.

We have already Project.parallelism parameter. 
I offer you to make poll.
But you will have enough time to rename but now I want to be merged and forget about it

sksamuel:
What about var testConcurrency: Int = 1

crazyk2:
My best are threadsForSpec /threadsPerSpec

SergKhram:
ping @sksamuel @crazyk2

Issue #1388, opened 2020-04-15
[Feature] Add config option alwaysRun to callback


crazyk2:
I want add a feature to discuss
As I said I used TestNG and there is one feature. You can add option `alwaysRun` to callback.
Now this parameter is indefined  in Kotest.
`afterTest` and  `afterProject` will be invoked even if `beforeTest` and `beforeProject` are failed.
And opposite situation with `beforeSpec` and `afterSpec`. if `beforeSpec` is failed then `afterSpec` will not be invoked.

But what if I don't want to execute `afterTest` if `beforeTest` is failed. For example I connect to Specific DB in `beforeTest` and release connect in `afterTest`. If 1 step is failed I don't want to run final step.
Of course I can add condition in afterTest `if (connection != null && connection.areOpen()) connection.close()`.
But user have to have choice or the same behavior or clear documentation

sksamuel:
The signature of the callback methods will not change now after 4.0 has been released. We could possibly add a configuration value into ProjectConfig to allow you to always/never run callbacks if something fails ?

crazyk2:
For different callbacks in the same project I would want to have different behavior. but ProjectConfig applied to all project.
And if [this](https://github.com/kotest/kotest/issues/1386#issuecomment-614176133) feature could be implement I don't see barrier to implement `alwaysRun` feature with default value.  I think it is not so hard work and `Interface`'s defaults help with it.
Just a brief that print `beforeTest` callback names
```kotlin
TestConfiguration{
    abstract fun beforeTest(name:String = "default", f: BeforeTest)
  }
Spec{
   override fun beforeTest( name:String, f: BeforeTest) {
      listener(object : TestListener {
         override val name = if (name == "default") "beforeTest " + this@Spec::class.description().name else name
         override suspend fun beforeTest(testCase: TestCase) {
            f(testCase)
         }
      })
   }
}
interface TestListener : Listener {
   val name : String
   get() = "default"
}

//callbacks.kt
suspend fun TestCase.invokeBeforeTest(): Try<TestCase> = Try {
   val listeners = spec.resolvedTestListeners() + Project.testListeners()
   listeners.forEach {
      println(it.name)
      it.beforeTest(this)
   }
   this
}
abstract class TestFactoryConfiguration : TestConfiguration() {
   override fun beforeTest(name :String, f: BeforeTest) {
      beforeTests = beforeTests + f
   }
}
class ExampleTest : FunSpec({

   beforeTest("should print it") {
      "void"
[...]

PR #1397, opened 2020-04-20
Resolve #1387. Catch all exceptions in project listeners.

### LINKED ISSUE #1387: Before\AfterProject catch only one Exception
[This](https://github.com/kotest/kotest/blob/master/kotest-core/src/jvmMain/kotlin/io/kotest/core/engine/KotestEngine.kt#L98), [this](https://github.com/kotest/kotest/blob/master/kotest-core/src/jvmMain/kotlin/io/kotest/core/engine/KotestEngine.kt#L108), [this](https://github.com/kotest/kotest/blob/master/kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt#L19) and [this](https://github.com/kotest/kotest/blob/master/kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt#L27)
work and handle only for first exception in `afterProject` and `beforeProject` callbacks.
In `kotest-tests-core` and other projects I can register several callbacks and I want to have guarantee that all of them will be invoked and if callback has some assertions I will be notified about all exceptions during my test run because test run may take hours (for example Selenium tests).
I did with `afterProject` callbacks by myself but with `beforeProject` can't without big changes of code.

crazyk2:
Add listener name as part of #1388
Welcome to discuss


kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
```diff
@@ -13,22 +13,41 @@ import io.kotest.core.test.TestResult
 import io.kotest.fp.Try
 import io.kotest.mpp.log
 
+
+fun List<ProjectListener>.resolveName() =
```
sksamuel, line 17: A better name might be groupByName()

kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
```diff
@@ -13,22 +13,41 @@ import io.kotest.core.test.TestResult
...
-fun List<Listener>.afterProject(): Try<Unit> = Try {
-   log("invokeAfterAll")
-   filterIsInstance<ProjectListener>().forEach { it.afterProject() }
-}.mapFailure { AfterProjectListenerException(it) }
+fun List<Listener>.afterProject(): Try<List<RuntimeException>> = Try {
+   log("invokeAfterProject")
+   filterIsInstance<ProjectListener>()
+      .resolveName()
+      .map { it.first to Try { it.second.afterProject() } }
+      .filter { it.second.isFailure() }
```
sksamuel, line 34: Please add a method to Try.kt called collectFailures, which does the filter and map for you in a more elegant way.

`fun List<Try<A>>.collectFailures()`

kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
```diff
@@ -13,22 +13,41 @@ import io.kotest.core.test.TestResult
...
+      .map { AfterProjectListenerException(it.first, (it.second as Try.Failure).error) }
+}.mapFailure { AfterProjectListenerException("afterProjectsInvocation", it) }
 
 /**
  * Invokes the beforeProject listeners, and prints project config using [dumpProjectConfig].
  */
-fun List<Listener>.beforeProject() = Try {
-   log("invokeBeforeAll")
+fun List<Listener>.beforeProject(): Try<List<RuntimeException>> = Try {
```
sksamuel: Should this return be Try<list<BeforeProjectListenerException>> ?

kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
```diff
@@ -13,22 +13,41 @@ import io.kotest.core.test.TestResult
...
+      }
+
 /**
  * Invokes any afterProject functions from the given listeners.
  */
-fun List<Listener>.afterProject(): Try<Unit> = Try {
-   log("invokeAfterAll")
-   filterIsInstance<ProjectListener>().forEach { it.afterProject() }
-}.mapFailure { AfterProjectListenerException(it) }
+fun List<Listener>.afterProject(): Try<List<RuntimeException>> = Try {
```
sksamuel: Should this return value by Try<List<AfterProjectListenerException>>

kotest-extensions/kotest-extensions-allure/src/jvmMain/kotlin/io/kotest/extensions/allure/AllureTestListener.kt
```diff
@@ -28,6 +28,9 @@ fun TestCase.description() = spec::class.findAnnotation<io.qameta.allure.Descrip
 @AutoScan
 object AllureTestListener : TestListener, ProjectListener {
 
+   override val name: String
+      get() = "AllureTestListener"
```
sksamuel: We can do this without the get() ?

sksamuel (changes requested):
This is awesome.

kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
crazyk2: grouping is only first step. And why this step is needed. JUnit can't register dynamic test with the same name if you has several equaled listener names you should add some different part to name. This is why it called resolveName or make listeners name to be differrent

kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
crazyk2: I don't understand question but try to answer. Earlier beforeProject return just `Try` and don't use success result of it only failed result. It works only until first error. Now we wrap each before\after call with each own `Try` and filter only Failed results because Success results are not interested for us. So return should be only List of exception of known type. 
And Kotlin compiler doesn't allow to use just list without type `Try<List>` 
image

kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
crazyk2: Answered above

kotest-extensions/kotest-extensions-allure/src/jvmMain/kotlin/io/kotest/extensions/allure/AllureTestListener.kt
crazyk2: Definitely

kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
crazyk2: To be honest I don't know how to do it.
If you could do refactoring it it would be nice

kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
sksamuel: Github hid some of the message because of the < > tags. I was asking if the return type should be List of AfterProjectListenerException rather than List of RuntimeException

sksamuel:
> You delete my dependency. I used it

Add it back in.

PR #1400, opened 2020-04-22
Updates package-lock.json to use 1.2.5 version of minimist

ashishkujoy:
This PR upgrades version minimist and others transitive dependencies which fixes the security issues which were introduced by minimist 0.0.8.


sksamuel:
Lgtm

Issue #1381, opened 2020-04-13
Error in beforeTest listener is swallowed


OliverCulleyDeLange:
Kotest 4.0.2
 I run a test which throws an exception in my `beforeTest{}` setup steps, however the process finishes with exit code 0, and no error is printed to console. 

Example: 
```
class TestingSpec : StringSpec() {
    init {
        beforeTest {
            throw ExceptionInInitializerError("Fail")
        }

        "test live data updated when doThing()" {
            true shouldBe false
        }
    }
}
```

Gives:

```
 "C:\Program Files\Android Studio\jre\bin\java.exe" -ea -Didea.test.cyclic.buffer.size=1048576 "-javaagent:C:\Program Files\Android Studio\lib\idea_rt.jar=55570:C:\Program Files\Android Studio\bin" -Dfile.encoding=UTF-8 -classpath "..." com.intellij.rt.execution.junit.JUnitStarter -ideVersion5 -junit5 uk.co.oliverdelange.wcr_android_kt.TestingSpec
~~~ Kotest Configuration ~~~
-> Parallelism: 1 thread(s)
-> Default test timeout: 600000ms
-> Default test order: TestCaseOrder
-> Default isolation mode: IsolationMode
-> Global soft assertations: False
-> Write spec failure file: False
-> Fail on ignored tests: False
-> Spec execution order: LexicographicSpecExecutionOrder
-> Extensions
  - io.kotest.core.extensions.SystemPropertyTagExtension
  - io.kotest.core.extensions.RuntimeTagExtension
  - io.kotest.core.extensions.IgnoredSpecDiscoveryExtension
  - io.kotest.core.extensions.TagFilteredDiscoveryExtension



Process finished with exit code 0```

Similar to https://github.com/kotest/kotest/issues/307 but thats closed.

sksamuel:
There's plenty of tests around this, I will investigate.

sksamuel:
This is your exact scenario. https://github.com/kotest/kotest/blob/master/kotest-runner/kotest-runner-junit5/src/jvmTest/kotlin/com/sksamuel/kotest/runner/junit5/StringSpecEngineKitTest.kt#L253

OliverCulleyDeLange:
Interesting, the plot thickens. So i ran that test, in the kotest repo to see if its a platform issue (java version or windows related), and it passes ok. 

I then tried to run my failing test through `gradle test` and it appeared to be fine, the test runs as expected and fails due to exception. 

[...]

PR #1402, opened 2020-04-23
#1381 Makes LinkageError as fatal error

ashishkujoy:
This fixes issues where beforeTest swallow exceptions like ExceptionInInitializerError, NoClassDefFoundError etc


sksamuel:
Good spot !

PR #1404, opened 2020-04-23
Allowing to use specific dates to generate date

ca-r0-l:


sksamuel:
Looks great, but can we keep the old method

`fun Arb.Companion.localDate(minYear: Int = 1970, maxYear: Int = 2030): Arb<LocalDate> = object : Arb<LocalDate>()`

and deprecate it, and then forward to the new method.

ca-r0-l:
> Looks great, but can we keep the old method
> 
> `fun Arb.Companion.localDate(minYear: Int = 1970, maxYear: Int = 2030): Arb<LocalDate> = object : Arb<LocalDate>()`
> 
> and deprecate it, and then forward to the new method.

yes, I'll do it!

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
```diff
@@ -45,23 +47,40 @@ fun Arb.Companion.period(maxYear: Int = 10): Arb<Period> = arb(listOf(Period.ZER
...
    override fun edgecases(): List<LocalDate> {
-      val yearRange = (minYear..maxYear)
+      val yearRange = (minDate.year..maxDate.year)
       val feb28Date = LocalDate.of(yearRange.random(), 2, 28)
 
       val feb29Year = yearRange.firstOrNull { Year.of(it).isLeap }
       val feb29Date = feb29Year?.let { LocalDate.of(it, 2, 29) }
 
-      return listOfNotNull(feb28Date, feb29Date, LocalDate.of(minYear, 1, 1), LocalDate.of(maxYear, 12, 31))
+      return listOfNotNull(feb28Date, feb29Date, LocalDate.of(minDate.year, minDate.month, minDate.dayOfMonth), LocalDate.of(maxDate.year, maxDate.month, maxDate.dayOfMonth))
```
richard-clements, line 76: If the min date is past these edge cases, or the max date is before, then it could cause a false date to be supplied.
There should be a check that the edge case still meets the criteria before returning

Issue #1406, opened 2020-04-24
Arb.choice should take Arb instances and not combine edgecases into arb.values().


jaymoid:
Currently `Arb.choice` takes Gen instances as parameters and combines the edgecases and the arb values. Such that you get the edgecases first when you call `.values()` on the resulting Arb instance. In addition the edgecases in the passed in Arbs are not available via the `edgecases()` function on the resulting Arb, instead you get an empty list.

As this might might not be expected, and invoking `choice` on an Exhaustive is counter intuitive, after discussing with @sksamuel in slack, it's been suggested that we:
* Change `Arb.choice` to accept `vararg Arb<A>`
* Combine the edgecases from the provided Arbs, and make available via resulting `Arb.edgecases`
* Expect values from the provided Arb instances to be available via the resulting `.values()`
* deprecate the existing one and replace it with the below
* Use @JvmName to keep the same function name with the varargs
Something like:
```
fun <A> Arb.Companion.choice(vararg arbs: Arb<A>): Arb<A> = arb (arbs.flatMap(Arb<A>::edgecases)){ rs ->
    val iters = arbs.map { it.values(rs).iterator() }
...
```

I'm happy to have a go at this.

PR #1407, opened 2020-04-25
#1406 Adding a Arb.Choice function that returns an Arb instance, deprecated…

jaymoid:
… Arb.Choice that returns a Gen.

The compiler didn't complain about conflicting JVM names, so I didn't need to add the the JvmName annotation, can add it if required? 

https://github.com/kotest/kotest/issues/1406


sksamuel:
nice job

PR #1408, opened 2020-04-25
Added Byte.shouldBeBetween(min, max) matcher; added bounds to Arb.byt…

sksamuel:
…e; added Arb.byteArray

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/bytes/byte.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/bytes/byte.kt
@@ -1,0 +1,14 @@
+package io.kotest.matchers.bytes
+
+import io.kotest.matchers.Matcher
+import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.shouldBe
+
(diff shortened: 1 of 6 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/bytes/byte.kt
```diff
@@ -0,0 +1,14 @@
...
+import io.kotest.matchers.Matcher
+import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.shouldBe
+
+fun Byte.shouldBeBetween(lower: Byte, upper: Byte) = this shouldBe between(lower, upper)
+fun between(lower: Byte, upper: Byte) = object : Matcher<Int> {
+   override fun test(value: Int) = MatcherResult(
+      value in lower..upper,
+      "$value should be between ($lower, $upper) inclusive",
+      "$value should be between ($lower, $upper) inclusive"
```
ashishkujoy: Message should be change to 
"$value should not be between ($lower, $upper) inclusive"

sksamuel: fixed.

ashishkujoy (approved):

PR #1409, opened 2020-04-25
Fix edge cases for long/int arbs to only include values in the given …

### LINKED ISSUE #1405: Number generators have wrong edge cases
Number generators (like `Arb.long()`, `Arb.int()`) accept a range as a parameter but using `MAX_VALUE`, `MIN_VALUE` as edgecases.

I guess more correct approach is to use `range.start` and `range.endInclusive` as edgecases.

https://github.com/kotest/kotest/blob/d03c9c9f0525071ac85ffdb3b8067acf24d75da6/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/longs.kt#L11

This part should look like:
```
fun Arb.Companion.long(range: LongRange = Long.MIN_VALUE..Long.MAX_VALUE): Arb<Long> {
   val edgecases = listOf(0, range.start, range.endInclusive)
   return arb(LongShrinker, edgecases) { it.random.nextLong(range) }
}
```

sksamuel:
Fix edge cases for long/int arbs to only include values in the given ranges 

closes #1405

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/ints/IntMatchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/ints/IntMatchers.kt
@@ -3,9 +3,27 @@ package io.kotest.matchers.ints
 import io.kotest.matchers.Matcher
 import io.kotest.matchers.MatcherResult
 import io.kotest.matchers.should
+import io.kotest.matchers.shouldBe
 import io.kotest.matchers.shouldNot
 
+/**
+ * Verifies that the given integer is between a and b inclusive.
+ */
+fun Int.shouldBeBetween(a: Int, b: Int) = this shouldBe between(a, b)
+
+/**
+ * Verifies that the given integer is NOT between a and b inclusive.
+ */
+fun Int.shouldNotBeBetween(a: Int, b: Int) = this shouldNot between(a, b)
+
+/**
+ * Verifies that the given integer is between a and b inclusive.
+ */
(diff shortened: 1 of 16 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/ints/IntMatchers.kt
```diff
@@ -7,9 +7,11 @@ import io.kotest.matchers.shouldNot
 
 fun beBetween(a: Int, b: Int) = between(a, b)
 fun between(a: Int, b: Int): Matcher<Int> = object : Matcher<Int> {
-  override fun test(value: Int) = MatcherResult(value in a..b,
-    "$value is not between ($a, $b)",
-    "$value is between ($a, $b)")
+  override fun test(value: Int) = MatcherResult(
+     value in a..b,
+    "$value should be between ($a, $b)",
```
ashishkujoy, line 30: Should we mention explicitly that check is inclusive of both a and b like we have mentioned for byte matchers.
https://github.com/kotest/kotest/blob/c1b1b1256b0894e01f065fe1ead4e73d73ed6ba6/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/bytes/byte.kt#L8-L12

sksamuel: Done.

ashishkujoy (approved):

PR #1413, opened 2020-04-26
Reflection equality improvements

sschuberth:
Please have a look at the individual commit messages for the details.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
@@ -73,9 +73,9 @@ fun <T : Any> T.shouldNotBeEqualToUsingFields(other: T, vararg properties: KProp
 /**
  * Matcher that compares values using specific fields
  *
- * Verifies that two instances not equal using only some specific fields. This is useful for matching
- * on objects that contain unknown values, such as a database Entity that contains an ID (you don't know this ID, and it
- * doesn't matter for you, for example)
+ * Verifies that two instances are equal considering only some specific fields. This is useful for matching on objects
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
```diff
@@ -139,7 +141,6 @@ fun <T : Any> beEqualToUsingFields(other: T, vararg fields: KProperty<*>): Match
  * Note: Throws [IllegalArgumentException] in case [properties] parameter is not provided.
  */
 fun <T : Any> T.shouldBeEqualToIgnoringFields(other: T, vararg properties: KProperty<*>) {
```
sksamuel, line 141: Do you think this method should allow specifying no fields? Equal to ignoring nothing? That would seem to be a strange way to use this method.

sschuberth: Well, that's what has been [recommend](https://github.com/kotest/kotest/issues/1161#issuecomment-575724787) to me for me use case. From a strictly logical perspective this sounds alright, but I agree it feels a bit unnatural.

But if neither `shouldBeEqualToUsingFields` nor `shouldBeEqualToIgnoringFields` offer an easy way to compare *all* fields, and preferably also with an option to compare `private` fields, what's the alternative?

sksamuel:
What about adding shouldBeEqualToUsingAllFields ?

sksamuel:
Or if we have to allow empty fields on one of them, I think
shouldBeEqualToUsingFields
makes more sense than the ignored one.

sschuberth:
> I think shouldBeEqualToUsingFields makes more sense than the ignored one.

Yes, that's how I [expected it to work](https://github.com/kotest/kotest/issues/1161#issuecomment-575675811) in the first place. I'll make that change.

sksamuel:
lgtm

PR #1414, opened 2020-04-26
Adding orNull

sksamuel:


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/null.kt
```diff
@@ -0,0 +1,10 @@
+package io.kotest.property.arbitrary
+
+import io.kotest.property.Arb
+
+fun <A> Arb<A>.orNull(): Arb<A?> = arb(this.edgecases().plus(null as A?)) { rs ->
+   sequence {
+      while (true)
+         yield(if (rs.random.nextBoolean()) null else this@orNull.next(rs))
```
LeoColman: every other value from the arb will be a null.
Should we consider a more realistic scenario and reduce the amount of nulls generated?

sksamuel:
We had a similar discussion on slack.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/null.kt
sksamuel: We had a similar discussion on slack

sksamuel:
Nice.

PR #1415, opened 2020-04-27
For issue: https://github.com/kotest/kotest/issues/790

kelly-yinn:
This PR implements that we can use now(zoneId) in source code with the now listener. 
If this implementation is acceptable, I'd like to implement enable using now(clock) in the same way.
@Kerooker would you mind help review this PR, thank you!


LeoColman:
Hey, @ylt1534, thanks for the contribution!

PR #1417, opened 2020-04-27
Add support to isEqual for ZonedDateTime

albertattard:
The `ZonedDateTime` [equals()](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#equals-java.lang.Object-) method also compares the time zone.  It will return `false` if the two instances have a different time zone.

The [isEqual()](https://docs.oracle.com/javase/8/docs/api/java/time/chrono/ChronoZonedDateTime.html#isEqual-java.time.chrono.ChronoZonedDateTime-) method, on the other hand, returns `true` if the instant equals the instant of the specified date-time.  This works across time zones.


sksamuel:
I think we should call this shouldBeEqualIgnoringTimezone or something like that so it's clear what it's comparing.

albertattard:
Would you like me to rename it to `shouldBeEqualIgnoringTimezone` before you merge it?

It is a bit tricky as the time zone is not ignored.  Maybe `shouldHaveSameInstant` is more appropriate.

sksamuel:
Yes that seems a better name.

Issue #1418, opened 2020-04-27
`isolation` instead of `isolationMode`


LeoColman:
When configuring tests using the inline features, `isolation` is a parameter, however intuition would suggest that `isolationMode = IsolationMode.InstancePerTest` is expected.

The old `isolation` parameter should be deprecated in favor of `isolationMode`

PR #1419, opened 2020-04-28
#1418 Depreciate `isolation` in favor of new `isolationMode` parameter for inline test configuration

BrianOdisho:
Resolves issue #1418


sksamuel:
Thanks for the PR. Can't merge while the build is red. @Kerooker the reason I called it isolation and not isolationMode is because the latter then clashes with the fun isolationMode() in the parent.

You might need to do some trickery to get it to work.

crazyk2:
How about tests and docs?

BrianOdisho:
> Thanks for the PR. Can't merge while the build is red. @Kerooker the reason I called it isolation and not isolationMode is because the latter then clashes with the fun isolationMode() in the parent.
> 
> You might need to do some trickery to get it to work.

Thanks for the additional context @sksamuel ! I’ll look into this more today.

sksamuel:
Since I don't think we can reuse isolationMode because of the function I vote we close this and stick with the status quo.

BrianOdisho:
Agreed I’m not seeing a way to resolve this without a breaking change to the api.

PR #1423, opened 2020-04-29
Document usage with mocks

### LINKED ISSUE #583: Add documentation for usage with `Mockk` and other mock libraries
Recently, in `Mockk`'s slack and `KotlinTest`'s slack, some users have reported troubles when using `Mockk` with KotlinTest, due to the mockk context not being cleared after tests.

This is due to the isolation mode being incorrect for the use case, as the user normally wants the class to be instantiated once per test (`InstancePerTest` isolation).

As `Mockk` is a very used mocking library for Kotlin, among other libraries, maybe we should create a specific part in the documentation for explaining this, or a paragraph in isolation mode.

Users don't associate directly that the "Bug" in their mocks using KotlinTest is due to isolation mode. Usually because the default mode for JUnit is single instance per test, and we don't do that, so writing the same test "bugs" their code because of it.

What do you think, @sksamuel ?

Also summoning @oleksiyp

LeoColman:
Closes #583

### DIFF
--- a/doc/mocks.md
+++ b/doc/mocks.md
@@ -1,0 +1,149 @@
+Mocking and Kotest
+==================
+
+Kotest itself has no mock features. However, you can plug-in your favourite mocking library with ease!
+
+Let's take for example [mockk](https://mockk.io):
+
+```kotlin
+class MyTest : FunSpec({
+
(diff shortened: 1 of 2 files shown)


sksamuel:
lgtm

Issue #1422, opened 2020-04-29
Make the arb.choice function take derived classes


ataronet:
I am using the new Arb.choice function (Kotest version 4.0.5) and when I use derived classes I need to cast them. For example
```
    Arb.choice(
        Arb.constant(RuntimeException(message)) as Arb<Exception>,
        Arb.constant(Exception(message)),
        Arb.constant(IllegalArgumentException(message)) as Arb<Exception>,
        Arb.constant(IllegalStateException(message)) as Arb<Exception>
    )
```
I would like to know if it is possible to declare their types internally as anything that derive from T so I won't need to cast them.
Thank you.

sksamuel:
Yes that makes sense. do you want to prepare a PR ?

ataronet:
Happy too but I will only get to it on the weekend.
Here is the fix that I am now testing for my code 
`fun <A> Arb.Companion.choice(vararg arbs: Arb<out A>): Arb<A> = arb(arbs.flatMap(Arb<out A>::edgecases))` 
I just changed the signature of your implementation to add `out`

jaymoid:
Ah this is my fault/bug, I prepared a fix, but don't want to step on any toes if you wanted to raise a PR (https://github.com/kotest/kotest/compare/master...jaymoid:arbVarianceFix)

sksamuel:
If you're ready with it james make sense to take yours.

sksamuel:
Can you try the latest snapshot as this should be fixed.

ataronet:
I will! 
btw spent few hours trying to set up my environment and it is failing to build. I think it'd be great to have instructions on "how to set up your env" and how to run tests etc.
I was hoping to fix it myself but maybe next time.
Will let you know if your fix works

ataronet:
let me know which snapshot you want me to try. I could not find any newer version after 4.0.5 which I have

sksamuel:
Snapshots are 4.1.x and are linked on github main page.

ataronet:
Can you be more specific about the main page? I wasn't able to find any snapshot on this page https://github.com/kotest/kotest, and not in here: https://github.com/kotest/kotest/releases. what is the link to the snapshots?

sksamuel:
Badge at the top under the logo.

ataronet:
👍 Thank you, it works!!

PR #1428, opened 2020-04-30
Fix for covariance in Arb.choice #1422

jaymoid:


jaymoid:
_should_ fix https://github.com/kotest/kotest/issues/1422

PR #1433, opened 2020-05-09
Reworked dsl to share scopes

sksamuel:


crazyk2:
Which problem did you solve?

sksamuel:
There is a bug in intellij which caused it to get into a loop when
highlighting the files in MPP mode. This avoids that bug.

crazyk2:
Could you share link (bug or discussion of problem)?

sksamuel:
https://github.com/kotest/kotest/issues/5511

PR #1437, opened 2020-05-11
Deprecates `isolation` variable; Adds `isolationMode`

### LINKED ISSUE #1418: `isolation` instead of `isolationMode`

LeoColman:
With this commit, I've deprecated the `isolation` spec variable, in favour of a variable that's closer to the `isolationMode()` function.

Closes #1418


sksamuel:
lgtm

PR #1449, opened 2020-05-21
Fix native publishing.

ajalt:
Currently, native code is unusable because only one native target is defined. Each build job will overwrite the artifacts of previous jobs.

This PR fixes this by defining a separate target for each platform. The targets each depend on a common "native" source set, so they all use the same code. When running in IntelliJ, the "native" source set is defined to be the current platform so that the code is treated as native rather than common code.


sksamuel:
:+1:

Issue #1435, opened 2020-05-10
Add property test global config with before / after prop test callbacks


ataronet:
I am using the `checkAll` function with some `Arb` and a lambda for the test. I need something like `beforeTest` but for every iteration of the `checkAll`. Thank you.

sksamuel:
Can't you just put it inside the check all functions ?

ataronet:
Of course but when someone adds a new test they will have to remember to add the setup cleanup stage as well.
I can try to add the functionality if it is not available but I might need some help to set up the environment. Last time I could not build the project on my laptop.

sksamuel:
Oh so something before every single prop test.

ataronet:
Yes. Exactly! We only use prop tests and normally we need to clean things in between their iterations.

sksamuel:
The quick fix is to write your own checkAll which delegates to the built in
one. You'd need to write them for all arities though.
The long term fix is to add some property test project wide configuration
that allows you to specify a beforePropertyTest and afterPropertyTest
callbacks.

ataronet:
@sksamuel @Kerooker I have a fix for that but I don't have the permission to push my changes. Please advise how I can commit my code so I can create a PR. I created a dedicated branch but still can't push my changes.

ashishkujoy:
@ataronet You need to fork kotest, push to that fork and then create a PR from the fork.

PR #1452, opened 2020-05-22
Fixed issues 1435  Add before and after test listeners to property ba…

ataronet:
Add listeners to property base tests. You can add before and after for each iteration of the test.
See the file PropListenersTest for example how to use it. 
https://github.com/kotest/kotest/issues/1435

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
@@ -17,7 +17,7 @@ data class PropTest(
    val maxFailure: Int = 0,
    val shrinkingMode: ShrinkingMode = ShrinkingMode.Bounded(1000),
    val iterations: Int? = null,
-   val listeners: MutableList<PropTestListener> = mutableListOf()
+   val listeners: List<PropTestListener> = listOf()
 )
 
 fun PropTest.toPropTestConfig() =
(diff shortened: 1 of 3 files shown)


kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
```diff
@@ -25,13 +26,37 @@ fun PropTest.toPropTestConfig() =
...
+) {
+   fun beforeTest(before: suspend () -> Unit) {
+      listeners.add(object : PropTestListener {
+         override suspend fun beforeTest() {
+            before()
+            super.beforeTest()
+         }
+      })
+   }
+   fun afterTest(after: suspend () -> Unit) {
```
sksamuel: I don't like the idea of having this class have mutable state.
Kotest is built along FP principles so we should keep this class immutable to avoid surprises for users. 
The only mutable classes we have are either internal and not visible to users, or in the project config god object.
I know that means the nice DSL methods can't exist here but that's a trade off that we must accept.

ataronet: Ok. If you like having the list in there (and not in a different object), I had another solution with List (and listOf).
Edit: I made the change. Let me know what you think.

sksamuel:
:+1:

PR #1467, opened 2020-05-28
Adjust Exhaustive default iterations

### LINKED ISSUE #1456: checkAll(Gen) for Exhaustive with large lists does not work
4.0.5

I got an exception that I didn't understand when using `.exhaustive()` of a list with 2500 elements.
```
java.lang.IllegalArgumentException: Require at least 1000 iterations to cover requirements
```
After debugging I found that this check is in [proptest.kt](https://github.com/kotest/kotest/blob/master/kotest-property/src/commonMain/kotlin/io/kotest/property/internal/proptest.kt#L17). I think the message contains the wrong value: The "at least" number shown should be the result of `genA.minIterations()`. Or the text must be changed to "at most iterations". That's the bug I wanted to report.

As a workaround I just pass the size of the list to `checkAll(Int, Gen)`, so this is not a blocker.

But I also would like to suggest that `io.kotest.property.checkAll(Gen)` does not use `PropertyTesting.defaultIterationCount` but at least the minimum of that and `genA.minIterations()`. For exhaustive generators the full set may be iterated multiple times but I think that is often not useful, so exactly the size of the generator is what I expect from `checkAll`.

Another thing for me is the missing KDoc. It is hard for me to understand what the parameter `iterations` means for a generator. Maybe it should be named `generations`.

LeoColman:
This commit aims to change how we calculate the default iterations for a given gen. When dealing with Arbs, it's ok to assume a number, as any iterations can be used for an Arb.

However, when talking about Exhaustives, the default iterations might not be enough for them to execute, and thus they fail.

I changed how we compute that default to assume that Exhaustives should have at least their min iterations as default iterations.

Fixes #1456

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/PropertyTest6.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/PropertyTest6.kt
@@ -3,6 +3,7 @@
 package io.kotest.property
 
 import io.kotest.matchers.shouldBe
+import io.kotest.property.PropertyTesting.computeDefaultIteration
 import io.kotest.property.arbitrary.default
 import io.kotest.property.internal.proptest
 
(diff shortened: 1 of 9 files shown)


sksamuel (approved):
What about adding a test for the compute function directly.

Otherwise lgtm.

LeoColman:
Sure, I'll write a few

sksamuel (approved):

sksamuel:
Can be merged in if you're happy with it.

sksamuel:
This will probably work after we merge in #1475 or #1476

PR #1474, opened 2020-05-31
Reduced amount of platform specific code

sksamuel:
Removed a lot of the common duplication by moving a default implementation into the common code, and having the js / native modules just return a reference to the default, and the jvm provide a platform specific implementation.


ajalt (approved):
LGTM! The more common code, the less target-specific code, the better.

PR #1475, opened 2020-05-31
Moved native code into duplicated sourcesets

sksamuel:


sksamuel:
Would something like this be a solution. Platform specific sourcesets, each one only active if on the appropriate platform. With a minimal amount of duplicated common code, I think this is an acceptable compromise (if it's a valid approach!)

PR #1479, opened 2020-06-01
Adds support for testcontainers

ashishkujoy:


kotest-extensions/kotest-extensions-testcontainers/src/jvmTest/kotlin/io/kotest/extensions/testcontainers/StartableTestLifecycleAware.kt
```diff
@@ -0,0 +1,17 @@
+package io.kotest.extensions.testcontainers
+
+
+import org.testcontainers.lifecycle.TestDescription
+import org.testcontainers.lifecycle.TestLifecycleAware
+import java.util.*
+
+internal class StartableTestLifecycleAware : TestStartable(), TestLifecycleAware {
```
sksamuel, line 8: What is the purpose of this class

kotest-extensions/kotest-extensions-testcontainers/src/jvmTest/kotlin/io/kotest/extensions/testcontainers/TestcontainerKotestIntegrationPerSpec.kt
```diff
@@ -0,0 +1,17 @@
+package io.kotest.extensions.testcontainers
+
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.shouldBe
+
+class TestcontainerKotestIntegrationPerSpec : StringSpec({
+   val testStartable = TestStartable()
+   listeners(testStartable.perSpec())
```
sksamuel, line 8: In master I have changed listener to return the type passed in.
So we could do

```
val startable = listener(TestStartable().perSpec())
```

Another option is 
```
val startable = TestStartable().apply { listener(this) }
```

sksamuel: We could add a function to `TestConfiguration` like this:
```
   fun <T : TestListener> T.register(): T {
      listener(this)
      return this
   }
```

Then you could just do 

```
val startable = TestStartable().register()
```

But it does mean that you would have to have some way of deciding if the test startable is a per spec or per test. Perhaps

```
val startable = TestStartable().perSpec.register()
```

kotest-extensions/kotest-extensions-testcontainers/src/jvmTest/kotlin/io/kotest/extensions/testcontainers/StartableTestLifecycleAware.kt
ashishkujoy: This class is used for testing in two places. One in ```TestLifecycleAwareListenerTest``` and second in ```StartableTestLifecycleAwareTest```.
This class extends ```TestLifecycleAware``` in its before test i am storing the ```TestDescription``` passed to it. Later in  ```TestLifecycleAwareListenerTest```  i am checking that test description passed to it has fileSystemFriendly name and in ```StartableTestLifecycleAwareTest``` i am checking that if a startable is also a ```TestLifecycleAware``` then we are calling its beforeTest and afterTest in both perSpec and perTest variants.

kotest-extensions/kotest-extensions-testcontainers/src/jvmTest/kotlin/io/kotest/extensions/testcontainers/TestcontainerKotestIntegrationPerSpec.kt
ashishkujoy: ```val startable = TestStartable().perSpec.register()```
Looks good to me.

sksamuel:
If you want to make those changes you can.
Merge when you're happy, it looks good.

PR #1481, opened 2020-06-02
Resolve #1469. Add reverse order to after callbacks execution

### LINKED ISSUE #1469: callback order execution and inheritance
master
I use inheritance with my test instead composition and faced with problem of order execution.
```kotlin
open class BaseDbTest : FunSpec() {

    protected lateinit var mainJdbcTemplate: JdbcTemplate
    protected lateinit var virtualJdbcTemplate: JdbcTemplate

    init{
        val pg = ServiceContainer("pg", ".*database system is ready to accept connections.*")

        beforeSpec{
            println("beforeSpec pg.start()")

            pg.start()
            mainJdbcTemplate = DbFactory.create(mainConfig)
            virtualJdbcTemplate = DbFactory.create(virtualConfig)
        }
        afterSpec {
            println("afterSpec pg.stop()")
            pg.stop()
        }
    }
}

class VersionTest : BaseDbTest() {

    init {

        val importer = ServiceContainer("info-importer", ".*Start executing ImportHostedService.*")

        afterTest {
            println("importer.stop()")
            importer.stop()
        }

        afterSpec {
            println("afterSpec clearDB()")
            clearDB()
        }
       
       test("1"){
         "some test"
      }
}
    fun clearDB() =
            mainJdbcTemplate.update(
                    "DELETE FROM public.\"Chains\";\n" +
                            "DELETE FROM public.\"Exchanges\";" 
            )
```
As known the class start initialization from parents to class itself and all listeners are stored in list and execution of listeners are 
```kotlin
suspend fun Spec.invokeAfterSpec(): Try<Spec> = Try {
   log("invokeAfterSpec $this")

   autoCloseables.forEach { it.close() }

   val listeners = resolvedTestListeners() + Project.testListeners()
   listeners.forEach { it.afterSpec(this) }
   this
}
```
so their order in list are important. And now situation is  parents add all listeners then child add all listener. 
So my `clearDB` function will be called after `pg.stop()` and DB won't be cleared

For test
```kotlin
open class Parent : FunSpec() {
    init {
        beforeSpec {
            println("parent beforeSpec")
        }
        beforeTest {
            println("parent beforeTest")
        }
        afterTest {
            println("parent afterTest")
        }
        afterSpec {
            println("parent afterSpec")
        }
    }
}

class Child : Parent() {
    init {
        beforeSpec {
            println("child beforeSpec")
        }
        beforeTest {
            println("child beforeTest")
        }
        afterTest {
            println("child afterTest")
        }
        afterSpec {
            println("child afterSpec")
        }
        test("1"){}
    }
}
```
And order will be
ParentBeforeSpec
ChildBeforeSpec
ParentBeforeTest
ChildBeforeTest

ParentAfterTest
ChildAfterTest
ParentAfterSpec
ChildAfterSpec

But order should be 

ChildAfterTest
ParentAfterTest
ChildAfterSpec
ParentAfterSpec

JUnit, TestNG guarantee right order

crazyk2:


sksamuel:
@ashishkujoy do you think reversing the order will have any side effects ?

ashishkujoy:
@sksamuel Chances of that is very low. That will only matter when more than one listener is used and tests depend on the sequence in which listener are getting invoked.

sksamuel:
We can reverse them, but we still can't offer any guarantee on the overall
execution order due to project level listeners.

sksamuel:
I'm going to close this. We can't offer any guarantees as to the order.
If you need guarantees, then you can wrap a listener in a composite listener. That would work.

Issue #1488, opened 2020-06-04
Parent scopes are not coroutine scopes


sksamuel:
Eg, the following should compile

```
     context("a context") {
         launch { delay(100) }
        should("a test") {
           1.shouldBeLessThan(2)
           launch { delay(100) }
        }
    }
```

PR #1489, opened 2020-06-05
#1488 Makes ContainerScope extends CoroutineScope

ashishkujoy:
This PR makes ContainerScope extends CoroutineScope which enable user to launch a coroutine in ContainerScopes such as ExpectScope, WhenScope, ShouldSpecContextScope etc.


sksamuel:
Nice job. Lgtm

PR #1498, opened 2020-06-07
Initial WIP for 826: detailed data class diffs

sksamuel:


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/DataClassEq.kt
```diff
@@ -0,0 +1,87 @@
...
+      return dataClassConstructorParams.map { membersByName[it.name] }.filterNotNull()
+   }
+
+/**
+ * Prints a detailed diff of data class instances.
+ *
+ * HELP! Where should this class live, it contains reflection code and should only be reachable if the platform
+ * supports data class diff (See Eq.kt)
+ */
+internal object DataClassEq : Eq<Any> {
```
sksamuel, line 29: All reflection code should live in kotest-commons in an existing `interface Reflection`.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
```diff
@@ -19,5 +19,17 @@ fun <T> eq(actual: T, expected: T): Throwable? = when {
    actual is Regex && expected is Regex -> RegexEq.equals(actual, expected)
    actual is String && expected is String -> StringEq.equals(actual, expected)
    actual is Number && expected is Number -> NumberEq.equals(actual, expected)
+   // HELP! How to get project config here? I was thinking...
+   // Project.showDetailedDataClassDiff() && actual.isDataClass() ...
```
sksamuel: ProjectConfig is only part of kotest-core (the test runner) so we can't use it in the assertion library.
I would just have a top level var like I do for `var shouldRemoveKotestElementsFromStacktrace: Boolean = sysprop("kotest.failures.stacktrace.clean", "true") == "true"`
It's unlikely that people are overriding equals, so I would default to true.

kotest-assertions/kotest-assertions-shared/src/jsMain/kotlin/io/kotest/assertions/eq/isDataClass.kt
```diff
@@ -0,0 +1,8 @@
+package io.kotest.assertions.eq
+
+/**
+ * Determine if the object's class is is a data class.
+ *
+ * Not available on this platform.
+ */
+actual fun Any?.isDataClass(): Boolean = false
```
sksamuel: The aforementioned reflections interface has `un <T : Any> isDataClass(kclass: KClass<T>): Boolean` so you can invoke that with (myobj::class)

kotest-core/src/commonMain/kotlin/io/kotest/core/config/Project.kt
```diff
@@ -202,6 +203,9 @@ object Project {
    fun setGlobalAssertSoftly(g: Boolean) {
       globalAssertSoftly = g
    }
+
+   fun showDetailedDataClassDiff(): Boolean = showDetailedDataClassDiff
```
sksamuel: This would need to come out as project config is for the runner config only.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
sksamuel: It might be worth starting an AssertionsConfig object, and putting the var in there.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/DataClassEq.kt
```diff
@@ -0,0 +1,87 @@
...
+      val noOfDifferences = dataClassDiff.differences.size
+      return dataClassDiff.differences.mapIndexed { index, (property, difference) ->
+         val isLastProperty = index + 1 == noOfDifferences
+         val diffMsg = when (difference) {
+            is StandardDifference -> difference.differenceError.message
+            is DataClassDifference -> formatDifferences(difference, indentStyle + isLastProperty)
+         }
+         buildString {
+            append(indentStyle.map { if (it) "   " else "│  " }.joinToString(separator = ""))
+            append(if (isLastProperty) '└' else '├')
```
sksamuel, line 82: the glyphs are nice

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/DataClassEq.kt
```diff
@@ -0,0 +1,87 @@
...
+      )
+   }
+
+   private fun formatDifferences(dataClassDiff: DataClassDifference, indentStyle: List<Boolean> = emptyList()): String {
+      val noOfDifferences = dataClassDiff.differences.size
+      return dataClassDiff.differences.mapIndexed { index, (property, difference) ->
+         val isLastProperty = index + 1 == noOfDifferences
+         val diffMsg = when (difference) {
+            is StandardDifference -> difference.differenceError.message
+            is DataClassDifference -> formatDifferences(difference, indentStyle + isLastProperty)
```
sksamuel, line 78: I would have a counter on the recursion and if you hit say 5 levels, you abort and just show the standard diff. To avoid stack overflows on circular references.

kotest-common/src/commonMain/kotlin/io/kotest/mpp/Reflection.kt
```diff
@@ -1,5 +1,6 @@
 package io.kotest.mpp
 
+import kotlin.reflect.KCallable
```
sksamuel: You're only using this in the jvm implementation right, so you could remove the import ?

kotest-common/src/jvmMain/kotlin/io/kotest/mpp/reflection.kt
```diff
@@ -23,6 +24,13 @@ object JvmReflection : Reflection {
 
    override fun paramNames(fn: Function<*>): List<String>? = fn.reflect()?.parameters?.mapNotNull { it.name }
 
+   override fun <T : Any> primaryConstructorMembers(klass: KClass<T>): List<Property> {
+      val dataClassConstructorParams = klass::constructors.get().last().parameters
```
sksamuel: I think there is an actual method called primaryConstructor. Getting the last one might not be sufficnet.

sksamuel (commented):
Looking good. I'd add an object to io.kotest.assertions called AssertionsConfig and in there have a var to disable this functionality (which defaults to true or the value of a sys prop).

kotest-common/src/jvmMain/kotlin/io/kotest/mpp/reflection.kt
jaymoid: You're right, not sure how I missed that 🤦  Yeah I was worried about the edgecases with getting the last one.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/DataClassEq.kt
jaymoid: Added this, but I've set it to 10 from a selfish point of view because we have some pretty deeply nested at work, and because 10 is where things start to look weird in the output. Can happily decrease, or maybe add a configurable sysprop?

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/failures.kt
```diff
@@ -47,6 +47,17 @@ fun failure(expected: Expected, actual: Actual): Throwable {
...
+   return stacktraces.cleanStackTrace(
+      Exceptions.createAssertionError(
+         detailedDiffMessage + clueContextAsString() + intellijFormatError(expected, actual),
+         null,
+         expected,
+         actual
+      )
+   )
+}
+
```
jaymoid: A little torn here, I could have added an overloaded param to the `failure` function above, but seems like 99% of the time it wouldn't be called. Any ideas? [Edit]I removed this function and added the extra param, it seemed tidier[/Edit]

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
jaymoid: Good idea, I've added this. I've brought in the other sysprops that refer to kotest.assertions... - I think this is a granular commit so could be easily removed if it's the wrong direction.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/DataClassEq.kt
sksamuel: 10 is good

PR #1517, opened 2020-06-18
Use the SPDX compliant license identifier "Apache-2.0" in POM files

mnonnenmacher:
Using an SPDX compliant license identifier makes it easier for other
tools to correctly determine the license of this project.

See: https://spdx.org/licenses/

### DIFF
--- a/publish-mpp.gradle.kts
+++ b/publish-mpp.gradle.kts
@@ -65,7 +65,7 @@ publishing {
 
             licenses {
                license {
-                  name.set("The Apache 2.0 License")
+                  name.set("Apache-2.0")
                   url.set("https://opensource.org/licenses/Apache-2.0")
                }
             }


LeoColman:
Thanks!

LeoColman:
I'm curious though. 
Is there a reason for this identifier to not be used? For instance, in [Maven Publish Plugin](https://docs.gradle.org/current/userguide/publishing_maven.html) it's also in another format

mnonnenmacher:
I just pushed again because I used the wrong email in the commit message.

> Is there a reason for this identifier to not be used? For instance, in Maven Publish Plugin it's also in another format

No, not really, for automated parsing the short identifier is always the best option. I think before SPDX there simply was no standard to follow. You could use the full name "Apache License 2.0" if you prefer that for human readability. I think Apache itself mostly uses "Apache License, Version 2.0", but that name is not used in the SPDX standard.

In the Maven POM reference they recommend to use an SPDX identifier, but don't do it in their own example :)
http://maven.apache.org/pom.html#Licenses

mnonnenmacher:
Since you are curious, you can see all the different spellings for the Apache license we found so far here:
https://github.com/oss-review-toolkit/ort/blob/bd88d4bba90c3ed444555fe651954f15d58a3a8b/spdx-utils/src/main/resources/declared-license-mapping.yml#L21

I think this shows how beneficial it is to follow a standard.

LeoColman:
Thanks!
LGTM

PR #1518, opened 2020-06-19
Rename reference from kotlintest to kotest

koperagen:


ashishkujoy:
@DisPony Thanks for the correction.

PR #1520, opened 2020-06-19
Native thread local fix

sksamuel:


ajalt (commented):
This should solve the problem.

A grep turns up a couple more mutable globals in common code:

- io.kotest.property.PropertyTesting
- io.kotest.core.spec.style.scopes.DslState
- io.kotest.assertions.BasicErrorCollector
- io.kotest.core.config.Project

I don't know if those globals are actually mutated.

sksamuel:
DslState and Project are only used by the framework which isn't available in native builds.
BasicErrorCollector will have the same issue so I'll apply the same fix.

ajalt (approved):
With these changes, Clikt's test suite passes on mingwX64, so this LGTM

sksamuel:
Did you try with my 2nd commit or the first?

ajalt:
> 
> 
> Did you try with my 2nd commit or the first?

The second commit.

PR #1523, opened 2020-06-20
add Dependabot

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


sksamuel:
Does this work with gradle kotlin files now ?

sullis:
> Does this work with gradle kotlin files now ?

Ugh.  I added dependabot.yml to one of my Gradle projects yesterday and it doesn't look like Dependabot can handle **build.gradle.kts**

sksamuel:
Yeah I didn't know if they had just added it or not (was hoping!).

PR #1538, opened 2020-07-01
Add some Iterable matchers

### LINKED ISSUE #1321: Expand some matchers to Iterable
When working with an Iterable, we sometimes might want to assert that it's empty, or that it contain a specific element.

The current way to to this is using `toList`, as in `iterable.toList().shouldBeEmpty()`

Should we expand the collection matchers to a higher level?

LeoColman:
Closes #1321


LeoColman:
I think I got all of them.

sksamuel (approved):

sksamuel:
nice job

PR #1546, opened 2020-07-06
add data class backed data tests that execute in a separate scope

### LINKED ISSUE #1537: Data driven tests should utilise data classes and automatically register test names (ala rollup)
inspired by the common practice in [go](https://github.com/golang/go/wiki/TableDrivenTests) - I would like to suggest to use data classes as the recommended way for data/table driven tests: 

It is already possible to do manually 

1. but a short helper method would make the code more concise (see below)
2. I'd also suggest to change the description to use this pattern as the default
3. by using the helper method it's hopefully also possible to create gutter icons in the intellij plugin to run an individual row

```kotlin
import io.kotest.core.spec.style.FreeSpec
import io.kotest.core.spec.style.scopes.FreeScope
import io.kotest.core.test.TestContext

class DataDrivenDataClass : FreeSpec({
    "the very manual way" - {
        data class TestCase(val a: String)
        listOf(
            TestCase("a"),
            TestCase("b")
        ).forEach {
            it.a - {
                it.a.length shouldBe 2
            }
        }
    }

    "nicer" - {
        data class TestCase(val a: String)
        forAll(listOf(
            TestCase("a"),
            TestCase("b")
        ), TestCase::a) {
            it.a.length shouldBe 2
        }
    }
})

suspend fun <T> FreeScope.forAll(testCases: List<T>, name: (T) -> String, body: (TestContext).(T) -> Unit) {
    testCases.forEach { tc ->
        val name1 = name(tc)
        name1 {
            body(this, tc)
        }
    }
}
```

zeitlinger:
Fixes #1537

I'll add documentation once the feature is accepted.

### DIFF
--- a/kotest-core/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/FreeScope.kt
+++ b/kotest-core/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/FreeScope.kt
@@ -52,4 +52,12 @@ class FreeScope(
       invocationTimeout,
       test
    )
+
+   suspend fun <T> forAll(testCases: Map<String, T>, testfn: suspend (FreeScope).(T) -> Unit) {
+      testCases.forEach { entry ->
+         entry.key - {
+            testfn(this, entry.value)
+         }
+      }
+   }
 }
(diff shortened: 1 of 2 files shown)


sksamuel:
I'm closing this, as per our discussion in #1537 we can approach this in a style agnostic way.

PR #1549, opened 2020-07-08
Ensure exhaustive isn't build with a empty list.

### LINKED ISSUE #1548: checkAll and forAll with empty list as Exhaustive hang forever
**Kotest Version: 4.0.3**

How to reproduce?

```kotlin
checkAll(emptyList<String>().exhaustive()) { a ->
    println("Some string $a")
}
````
I have been explore the code base and the problem is that `proptest` will `take(n)` on a sequence generator that always produce empty sequence as it itens, it will never materialize the sequence.

Change `Gen.generate` as below solve the problem:
```kotlin
fun generate(rs: RandomSource): Sequence<Sample<A>> = when (this) {
      is Arb -> this.edgecases().asSequence().map { Sample(it) } + this.values(rs)
      is Exhaustive ->
         if (this.values.isEmpty()) emptySequence()
         else generateSequence { this.values.map { Sample(it) } }.flatten()
}
```
Let me know if I could open up a pull request.

cleidiano:
Close #1548


LeoColman (approved):

LeoColman:
Thanks!

PR #1553, opened 2020-07-10
Replace runBlocking with future where submitted to an executor

nomisRev:
Hey,

As discussed on Slack with @sksamuel, `runBlocking` can be replaced by a simple `Future` backed by `startCoroutine`.

This draft proposes a simple small implementation to do so.
[KotlinX](https://github.com/Kotlin/kotlinx.coroutines/tree/master/integration/kotlinx-coroutines-jdk8) also offers an integration module for JDK8 which includes similar functionality.

Side-notes:
 The `runBlocking` in the platform runners look fine, I wasn't sure about the one in `ExecutorExecutionContext.kt` but that might possibly also be quite easily replaced.

I'm very curious to try this on the Arrow Fx repository since we've been having stress-related issues lately.


sksamuel:
I'll have to get the tests to pass with this.
Remind me why replacing runBlocking with running on the main thread was better?

nomisRev:
When using `runBlocking` to run a `TestCase`, it'll launch a blocking `EventLoop` to later trampoline back to the original context.
An `EventLoop` requires locks, parking/unparking, SynchronisedQueues etc which is quite the overhead.

Since you're already using `runBlocking` to run the `suspend` `TestCase` on a `Future`, you can completely remove the `runBlocking` (and its overhead) by running `suspend` `TestCase` directly in a `Future`. Which is completely non-blocking, and lock-free.
Doing so will give you the same control over the running `TestCase` on your `Executors` as now.

EDIT: you should be able to edit/push to my fork :)

kotest-core/src/jvmMain/kotlin/io/kotest/core/engine/SpecRunner.kt
```diff
@@ -7,10 +7,10 @@ import io.kotest.core.test.TestCase
 import io.kotest.core.test.TestResult
 import io.kotest.fp.Try
 import io.kotest.mpp.log
-import kotlinx.coroutines.runBlocking
-import java.util.concurrent.ExecutionException
-import java.util.concurrent.Executors
-import java.util.concurrent.TimeUnit
+import java.util.concurrent.*
```
ibrahimyilmaz: Minor: Your indentations and `*` import way show that your IDE configuration is different from the others. Maybe you can validate this with other developers of this project.

nomisRev: Thanks for catching this!
You're correct, I had an incorrect configuration compared to normally after a fresh install on a new machine.

kotest-core/src/jvmMain/kotlin/io/kotest/core/engine/SpecRunner.kt
```diff
@@ -47,9 +51,7 @@ abstract class SpecRunner(val listener: TestEngineListener) {
 
       val futures = testCases.map { testCase ->
          executor.submit {
-            runBlocking {
```
nomisRev: I'm not sure I understand the error in the `ubuntu` build since it failed with.

> e: /home/runner/work/kotest/kotest/kotest-core/src/jvmMain/kotlin/io/kotest/core/engine/SpecRunner.kt: (52, 13): Unresolved reference: runBlocking
e: /home/runner/work/kotest/kotest/kotest-core/src/jvmMain/kotlin/io/kotest/core/engine/SpecRunner.kt: (53, 16): Suspension functions can be called only within coroutine body

But the relevant code is removed in this PR. Do I have a missunderstanding of how CI checks run in Kotest?

sksamuel:
CI just just ./gradlew check

sksamuel:
Since this PR is out of date with the latest code, I tried to adapt it myself, quite simple change.
https://github.com/kotest/kotest/commit/bcb375f7a7c6ee484832aab7607746ad9d6a7b04

But like before, the tests still don't pass. I'm not sure if changing this out is a simple drop in replacement. There must be some other semantics changing too.

nomisRev:
Hey @sksamuel,

You're correct, the `Future`'s weren't not properly scheduled on the `ExecutorService`.
They were just being submitted as `Runnable`, but not properly awaited by the `executor.shutdown()`.

This still small additional change schedules them properly on the `ExecutorService`, so they're now properly awaited by `executor.shutdown()`.

I had some issues before locally when running `./gradlew check` since my Xcode setup was not ready for Kotlin Native. Test locally run great, up-to-the `testcontainers` extensions. Which don't seem to be running at all, I looked into the code but had no idea how those tests could be negatively effected but the rest not. So I decided to push this already, and see what happens on actual CI.

Sorry, for my inactivity on this PR. All my time was/is consumed by the nearing release of Arrow Fx Coroutines.

I wasn't sure how to best organize this code, so I quickly duplicated it.

sksamuel:
Something seems to be up with the build since it's stalled for 2 hours.
I'll take your code and apply it locally and see what happens.

nomisRev:
Awesome, I'll try to keep a better eye on this PR! (Please feel free to poke me on Kotlin Slack, if I've seemed to forget about it!)

sksamuel:
Just getting around to looking at this again. I want to remove the executors completely from spec runners and allow people to provide their own dispatcher. 

Is it the run blocking that's the root of this problem here?

nomisRev:
> Is it the run blocking that's the root of this problem here?

Yes, since that requires spinning up an `EventLoop` to create a dispatcher on the current `Thread`. Providing a `default` Dispatcher` would solve that problem afaik.

sksamuel:
Inspired by your work the latest code in master is now run blocking free except for the main thread that starts the engine. Once 4.4 RC1 is released I would be eager to see your results.

I will close this PR now as the equivalent changes are merged.

nomisRev:
Awesome! Thanks @sksamuel 👏 👏 👏

PR #1557, opened 2020-07-13
Updates method signature of assertSoftly to take object under test as…

ashishkujoy:
Updates method signature of assertSoftly to take object under test as parameter instead of invoking assertSoftly as extension function on it.


LeoColman:
I believe there's a necessary change in the documentation as well, but I suppose that can be marked as `[NEXT RELEASE]`.

However, one thing I would suggest: Add a test to that behavior that should work but slipped through. The one you reported in Slack

LeoColman (approved):

PR #1561, opened 2020-07-16
Removed junit markers

sksamuel:


LeoColman (approved):
Burying JUnit for good, huh?

sksamuel:
These were unused after we released 4.1.0

PR #1568, opened 2020-07-17
Feature/tag expressions

sksamuel:


doc/tags.md
```diff
@@ -64,9 +87,67 @@ RuntimeTagExtension.excluded += MyOtherTag
...
+```
+
+| Property | Spec Created | Callbacks | Outcome |
+|----------|--------------|---------|----------|
+| kotest.tags=Linux                  | yes | yes | A and B are executed because all tests inherit the `Linux` tag from the annotation |
+| kotest.tags=Linux & Mysql          | yes | yes | A is executed only because all tests have the `Linux` tag, but only A has the `Mysql` tag |
+| kotest.tags=!Linux                 | no  | no  | No tests are executed, and the MyTestClass is not instantiated because we can exclude it based on the tags annotation |
+| kotest.tags=!UnitTest              | yes | no  | No tests are executed because all tests inherit `UnitTest` from the tags function. MyTestClass is instantiated in order to retrieve the tags defined in the class. The beforeSpec callback is not executed because there are no active tests. |
+| kotest.tags=Mysql                  | yes | yes | A is executed only, because that is the only test marked with `Mysql` |
+| kotest.tags=!Mysql                 | yes | yes | B is executed only, because that is the only test marked with `Mysql` |
```
ashishkujoy: | kotest.tags=!Mysql                 | yes | yes | B is executed only, because that is the only test marked with `Mysql` |
It should be: 
| kotest.tags=!Mysql                 | yes | yes | B is executed only, because that is the only test not marked with `Mysql` |

sksamuel: Fixed

PR #1571, opened 2020-07-18
Upgrade build to use kotlin 1.4 m3

sksamuel:
Master will become the Kotest 4.2 development branch, which will target Kotlin 1.4.
All further 4.1 fixes to go on release/4.1.x


settings.gradle.kts
```diff
@@ -9,8 +17,8 @@ include("kotest-assertions:kotest-assertions-json")
 include("kotest-assertions:kotest-assertions-ktor")
 include("kotest-assertions:kotest-assertions-jsoup")
 include("kotest-assertions:kotest-assertions-konform")
-include("kotest-assertions:kotest-assertions-compiler")
-include("kotest-assertions:kotest-assertions-klock")
+//include("kotest-assertions:kotest-assertions-compiler")
+//include("kotest-assertions:kotest-assertions-klock")
```
ashishkujoy, line 21: @sksamuel Is there any specific reason for not including these two modules?

sksamuel: They don't seem to be compatible with 1.4 m3. We can re-enable them when they have new releases. The master branch is just being used for development atm, so I thought it was ok to temporarily remove them.

ashishkujoy: ohh

Issue #1563, opened 2020-07-16
Android: Kotest-assertions uses java.nio.file.Path causing failure on Android < 26


JustinTullgren:
**Kotest Version**: 4.1.1
**Kotest Library**: kotest-assertions-core-jvm
**Environment**: Android Api <26 (Android 8 (O)reo)

Hi,

I was very happy to see the assertions library was broken out into its own library and attempted to use it in an "Android" test (which requires the test to be run on an android phone) so we could be consistent with our unit tests at least for assertions.

The assertions work fine for Android O and above (android api 26).

However, the `platformShow` actual implementation in the _jvmMain_ source uses `java.nio.file.Path` which is unfortunately not available on older android devices. Printing messages for failures results in runtime errors of class not found.

Would it be possible to use the `java.io.File#getPath` to print instead of calling toString on the `java.nio.file.path`?

Here is the [page](https://developer.android.com/reference/java/nio/file/Path) which shows that `Path` was added in api 26, meaning it cannot be used on older devices for testing.

Here is the [page](https://developer.android.com/reference/java/io/File#getPath()) which shows you can get the file path on older android devices back to Api 9 (very old).

```
@file:JvmName("platformjvm")
package io.kotest.assertions.show

import java.nio.file.Path

actual fun <A : Any> platformShow(a: A): Show<A>? = when (a) {
   is Path -> PathShow as Show<A>
   else -> null
}
```
 ([actual platform show](https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/show/platformShow.kt#L4))

```
package io.kotest.assertions.show

import java.nio.file.Path

object PathShow : Show<Path> {
   override fun show(a: Path): Printed = a.toString().printed()
}
```
([pathshow](https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/show/PathShow.kt#L3))

`expect fun <A : Any> platformShow(a: A): Show<A>?` ([expect platform show](https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/show/Show.kt#L32))

sksamuel:
Create the PR and I'll merge it and include it in 4.1.2

LeoColman:
Sure! 
It seems really easy to make a change. 
[...]

PR #1572, opened 2020-07-20
#1563 Bug fix for Android API < 26 Instrumented Tests. ClassNotFoundException

JustinTullgren:
#1563 Bug fix for Android API < 26 Instrumented Tests. The referenced type `java.nio.file.Path` does not exist on Android JVMs for versions < 26 which results in Android Instrumented Tests failing while running on an Android Device. 

This PR checks for the `java.nio.file.Path` type via pure reflection to support the existing PathShow object and adds support for `java.io.File` as well. 

An existing UNCHECKED_CAST warning is still present and a new one is added to support the different types via reflection.

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/show/FileShow.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/show/FileShow.kt
@@ -1,0 +1,7 @@
+package io.kotest.assertions.show
+
+import java.io.File
+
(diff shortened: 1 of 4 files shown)


JustinTullgren:
@sksamuel @LeoColman I went with the use of reflection to support the existing use of Path types without breaking Android Tests. I made the assumption, hopefully correctly, that you do not want an android specific actual implementation for platformShow which would reside in another module and that reflection was acceptable.

sksamuel (approved):

sksamuel:
Great job

PR #1575, opened 2020-07-20
Fix beforeTest/afterTest only being executed for immediate child scopes

### LINKED ISSUE #1574: beforeTest of parent scope not executed before tests of child scope
When nesting test cases the `beforeTest` of the parent scope is not executed before each test of the child scope.

Simple example:

```kotlin
class TestSpec : DescribeSpec({
    // This beforeTest is executed as expected
    beforeTest {
        println("beforeTest 1")
    }

    describe("some scope") {
        // This beforeTest is not executed before any test of the "some deeper scope" scope
        beforeTest {
            println("beforeTest 2")
        }

        it("test A") {
            println("test A")
        }

        it("test B") {
            println("test B")
        }

        describe("some deeper scope") {
            it("test C") {
                println("test C")
            }

            it("test D") {
                println("test D")
            }
        }
    }
})
```

Test run output (`IsolationMode.SingleInstance`):
```
beforeTest 1

beforeTest 1
beforeTest 2
test A

beforeTest 1
beforeTest 2
test B

beforeTest 1
beforeTest 2

beforeTest 1
test C

beforeTest 1
test D
```

Test run output (`IsolationMode.InstancePerLeaf`):
```
beforeTest 1

beforeTest 1
beforeTest 2
test A

beforeTest 1

beforeTest 1
beforeTest 2
test B

beforeTest 1

beforeTest 1
beforeTest 2

beforeTest 1
test C

beforeTest 1

beforeTest 1
beforeTest 2

beforeTest 1
test D
```

Test run output (`IsolationMode.InstancePerTest`):
```
beforeTest 1
beforeTest 1

beforeTest 1
beforeTest 2
test A

beforeTest 1

beforeTest 1
beforeTest 2
test B

beforeTest 1

beforeTest 1
beforeTest 2
beforeTest 1
beforeTest 1
beforeTest 2

beforeTest 1
test C

beforeTest 1
beforeTest 1
beforeTest 2

beforeTest 1
test D
```

As you can see no matter which `IsolationMode` is used, there is no scenario where `beforeTest 1` and `beforeTest 2` is actually executed in the right order just before `test C` or `test D` is executed.

___

### Environment
OS: MacOS `10.15.3`
IDE: Android Studio `4.0.1`
Kotlin: `1.3.72`
Gradle: `6.1.1`
Kotest: `kotest-runner-junit5-jvm`: `4.1.1` & `kotest-runner-console-jvm`: `4.1.1`
Kotest IntelliJ Plugin: `1.0.5`

reevn:
## Description

Fixes #1574.

Normally `beforeTest` and `afterTest` in the `*Scope` classes need to be executed for every test case of every child scope, no matter how deeply nested they are.

Currently, this only happens if the parent of the test case in question is the immediate parent (verified via `description.isParentOf(testCase.description)`). The problem is that this needs to be executed even if the parent is an ancestor and not just an immediate parent.

## Fix
Changed the `ContainerScope` to execute `beforeTest` and `afterTest` if the scope is an ancestor of the test case in question, not just a parent.

## Tests
Extended the three test classes that verify `beforeTest` / `afterTest` to include tests that verify the behaviour in question. The added tests would fail without this fix.

### DIFF
--- a/kotest-core/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
+++ b/kotest-core/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
@@ -56,7 +56,7 @@ interface ContainerScope : CoroutineScope {
    fun beforeTest(f: BeforeTest) {
       lifecycle.addListener(object : TestListener {
          override suspend fun beforeTest(testCase: TestCase) {
-            if (description.isParentOf(testCase.description)) f(testCase)
+            if (description.isAncestorOf(testCase.description)) f(testCase)
          }
       })
    }
@@ -67,7 +67,7 @@ interface ContainerScope : CoroutineScope {
    fun afterTest(f: AfterTest) {
       lifecycle.addListener(object : TestListener {
          override suspend fun afterTest(testCase: TestCase, result: TestResult) {
-            if (description.isParentOf(testCase.description)) f(Tuple2(testCase, result))
+            if (description.isAncestorOf(testCase.description)) f(Tuple2(testCase, result))
          }
       })
    }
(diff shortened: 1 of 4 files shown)


sksamuel (approved):

sksamuel:
good job, will be included in 4.1.3

Issue #1543, opened 2020-07-05
Complex structures deep comparison


ov7a:
I'm using kotest 4.1.1.

I think it would be nice to have deep recursive comparison in `eq` assertion by default. It is already checking for type and doing a smart comparison for upper lever. I think the same should be done for nested elements as well.

Consider this code snippet:
```
arrayOf(1,2,3) shouldBe arrayOf(1,2,3) //works fine
mapOf("1" to arrayOf(1,2,3)) shouldBe mapOf("1" to arrayOf(1,2,3)) //fails
```
Two arrays are equal, but if we wrap them in a map, they suddenly don't. That happens because `eq` checks for type, and compares `actual` and `expected` as Maps by simple `==`. In case of two arrays, the `DefaultEq` with `makeComparable` are used (BTW, why `makeComparable` instead of calling `contentEquals` for arrays?). 

That behaviour is a bit counter-intuitive. I think that can be solved if smart comparisons will be used for all container elements. As a bonus, a more detailed assertion message can be generated. Consider this code snippet:
```
val actual = mapOf(
    "1" to mapOf(
        "2" to mapOf(
            "3" to mapOf(
                "4" to "value"
            )
        )
    )
)

val expected = mapOf("1" to mapOf("2" to mapOf("3" to mapOf("4" to "value2"))))
actual shouldBe expected
```
It fails with `Values differed at keys 1`. With recursive comparison it can be changed to full path like `Values differed at key 1.2.3.4`. It would also be very very handy when two jsons are compared:
```
val actual = """{"1": {"2": {"3":{"4":"value"}}}}"""
val expected = """{"1": {"2": {"3":{"4":"value2"}}}}"""
actual shouldMatchJson expected
```
Right now it fails with no details at all.

ashishkujoy:
Fixed in release/4.1.x branch, will be released with 4.1.3

ashishkujoy:
Released with 4.1.3

ov7a:
@ashishkujoy Wow, thank you!

Unfortunately, it seems like there is a bug related to nulls:
```
val mapWithNull = mapOf("null" to null)
val mapWithNullCopy = mapOf("null" to null)
mapWithNull shouldBe mapWithNullCopy //fails with "null cannot be cast to non-null type kotlin.Any"
```

I also have [a more complex testCase](https://github.com/ov7a/kotlin-asserts-comparison/blob/master/src/test/kotlin/common/TestData.kt#L34):
```
[...]

PR #1577, opened 2020-07-20
#1543 Improve MapEq to compare map of complex value

ashishkujoy:


kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/com/sksamuel/kotest/eq/MapEqTest.kt
```diff
@@ -0,0 +1,103 @@
...
+            "a" to arrayOf(1, 2, 3)
+         )
+      )
+      val map2 = mapOf(
+         mapOf("a" to "c") to mapOf(
+            "a" to arrayOf(1, 2, 3)
+         )
+      )
+      MapEq.equals(map1, map2).shouldNotBeNull()
+   }
```
sksamuel, line 146: Should we add some tests that test the error output?

ashishkujoy: Done

sksamuel (approved):

PR #1586, opened 2020-07-22
Issue #1458 Test name case control setting

### LINKED ISSUE #1458: Allow case control in test's reports
When you use a test method, case is not handled. I.e.: `should("Happen something")` will be rendered as `should Happen something`.

The proposed feature is to allow case control on reports by using a report setting. This could be something like: `case = AS_IS | SENTENCE | LOWERCASE`

Being rendered in the reports this way:

* AS_IS: `should Happen something`
* SENTENCE: `Should happen something`
* LOWERCASE: `should happen something`

This setting should only affect the firsts letter of the test text and the first letter of the test method.

jaguililla:
Closes #1458


kotest-core/src/commonMain/kotlin/io/kotest/core/config/Project.kt
```diff
@@ -242,6 +258,19 @@ object Project {
...
 }
 
+/**
+ * Test naming strategies to adjust test name case.
+ *
+ * @property AS_IS For: should("Happen something") yields: should Happen something
+ * @property SENTENCE For: should("Happen something") yields: Should happen something
+ * @property LOWERCASE For: should("Happen something") yields: should happen something
+ */
+enum class TestNameCaseOptions {
```
sksamuel, line 268: I think TestNameCase is a better name because there are not multiple options controlled by this flag, but that single setting.

kotest-core/src/commonMain/kotlin/io/kotest/core/test/TestName.kt
```diff
@@ -31,18 +32,40 @@ data class TestName(val prefix: String?, val name: String, val focus: Boolean, v
...
+         when (Project.testNameCase()) {
+            TestNameCaseOptions.SENTENCE -> flattened.capitalize()
+            TestNameCaseOptions.LOWERCASE -> flattened.uncapitalize()
+            else -> flattened
+         }
+      }
+      else {
+         when (Project.testNameCase()) {
+            TestNameCaseOptions.SENTENCE -> "${withPrefix.capitalize()}${flattened.uncapitalize()}"
+            TestNameCaseOptions.LOWERCASE -> "${withPrefix.uncapitalize()}${flattened.uncapitalize()}"
```
sksamuel: This would leave mySILLYTESTNAME as mySILLYTESTNAME which doesn't seem intuitive with lowercase.

jaguililla: Right! Would START_LOWERCASE be clearer? Any suggestion?

sksamuel: Yes, or INITIAL_LOWER or something.
And I would keep LOWERCASE as a fourth option that just does .tolowercase on it all.

sksamuel: One final thing, can you call them Lowercase, Sentence etc as that fits in with the enum style used everywhere else.

Eg,

enum class TestCaseOrder {
   // the order in which the tests are defined in code
   Sequential,
   // a randomized order
   Random,
   // ordered a-z
   Lexicographic
}

jaguililla: Ok, I've applied the suggested changes

sksamuel (approved):

sksamuel:
Nice job

jaguililla:
Thanks! glad to be helpful!

PR #1589, opened 2020-07-22
Add new scoped callbacks

### LINKED ISSUE #1584: Add new scoped callbacks
`beforeTest`/`afterTest` is being executed additional times for containers (i.e. `describe` block) before then being executed again for the tests in the container.

Example test class
```kotlin
class SomeSpec: DescribeSpec({
  beforeTest {
    println("beforeTest 1")
  }

  afterTest {
    println("afterTest 1")
  }

  describe("some scope") {
    beforeTest {
      println("beforeTest 2")
    }

    it("test A") {
      println("test A")
    }

    it("test B") {
      println("test B")
    }

    describe("some deeper scope") {
      it("test C") {
        println("test C")
      }

      it("test D") {
        println("test D")
      }
    }
  }
})
```

Output (4.2.0.393-SNAPSHOT, as another listener fix is included there):

```
beforeTest 1

// test case "A" start
beforeTest 1
beforeTest 2
test A
afterTest 1
// test case "A" end

// test case "B" start
beforeTest 1
beforeTest 2
test B
afterTest 1
// test case "B" end

beforeTest 1
beforeTest 2

// test case "C" start
beforeTest 1
beforeTest 2
test C
afterTest 1
// test case "C" end

// test case "D" start
beforeTest 1
beforeTest 2
test D
afterTest 1
// test case "D" end

afterTest 1
afterTest 1
```

**Basically everything not in-between the start/end comments is not just unnecessary, it messes up certain test cases where the typical "mock in beforeTest and cleanup in afterTest" pattern is used.**

___

### Scenario

- mocks are created at the top level scope outside of `beforeTest 1`
- `beforeTest 1` defines behaviour for mocks/listeners and creates a fresh instance of the class under test
- `afterTest 1` is supposed to clear all the mocks/reset the listeners

As you can see in the beginning of the output, `beforeTest 1` is executed twice in a row without `afterTest 1` being executed in between. This leads to problems as `beforeTest 1` now created two instances of the class under test without mocks/listeners being reset in-between.

I think `afterTest 1` not being executed here is a bug, but I also don't really understand why listeners like before**_Test_** are executed additionally for containers before executing them again for the actual tests.

I already identified which lines in the project would need to be changed for every execution outside the `start/end` comments to not occur, but that fix seems more like fighting the symptom and not the root cause.

Maybe the real problem is that containers are considered tests (and therefore add to the test count etc.), which should also not be the case?

reevn:
### Description

Closes #1584.

Currently, the only way to execute code before every test is via `beforeTest`. This callback is executed before containers as well, so there is currently no way to only execute code just before "leaf" tests (`TestCase`s with type `TestType.Test`).

Adding more callbacks that are similar to `beforeTest` will offer more flexibility for people who do not want to execute multiple `beforeTest`s before something is actually tested.

Please refer to the issue for discussion about the new callbacks.

### Solution
Added six new listener callbacks:
- `beforeContainer`/`afterContainer` for code that is only executed before/after a test case with type `Container`
- `beforeEach`/`afterEach` for code that is only executed before/after a test case with type `Test`
- `beforeAny`/`afterAny` for code that is executed before/after any test case. This is the exact same behaviour as the current `beforeTest`/`afterTest`.

### Questions to reviewers
~There is a lot of duplicate code in here (the checks for the `TestType` before executing one of the new callbacks). Is there any interest in trimming this down and just delegating the new functions to `beforeTest` once they pass the `TestType` verification or do we want to keep these callbacks as standalone implementations?~

This was addressed.


kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
```diff
@@ -60,26 +60,48 @@ suspend fun List<Listener>.beforeProject(): Try<List<BeforeProjectListenerExcept
...
     Try {
-        val listeners = spec.resolvedTestListeners() + Project.testListeners()
-        listeners.forEach {
-            it.beforeTest(this)
+        spec.resolvedTestListeners() + Project.testListeners()
+    }.fold({
+        Try.Failure(it)
+    }, { listeners ->
+        Try {
+            listeners.forEach {
```
sksamuel, line 74: If a user has created a TestListener manually, then it will be invoked here regardless of the test type I think.

If that's correct (I've yet to look at the rest of the PR) then I think here is where we should do the filtering, not when they are registered via the DSL.

kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
```diff
@@ -60,26 +60,48 @@ suspend fun List<Listener>.beforeProject(): Try<List<BeforeProjectListenerExcept
...
     Try {
-        val listeners = this.config.listeners + spec.resolvedTestListeners() + Project.testListeners()
-        listeners.forEach {
-            it.afterTest(this, result)
+        this.config.listeners + spec.resolvedTestListeners() + Project.testListeners()
+    }.fold({
+        Try.Failure(it)
+    }, { listeners ->
+        Try {
+            listeners.forEach {
```
sksamuel, line 96: same as above

kotest-core/src/commonMain/kotlin/io/kotest/core/spec/Spec.kt
```diff
@@ -91,6 +91,54 @@ abstract class Spec : TestConfiguration(), SpecConfigurationMethods {
...
+         override suspend fun beforeContainer(testCase: TestCase) {
+            if (testCase.type == TestType.Container) f(testCase)
+         }
+      })
+   }
+
+   override fun afterContainer(f: AfterContainer) {
+      listener(object : TestListener {
+         override suspend fun afterContainer(testCase: TestCase, result: TestResult) {
+            if (testCase.type == TestType.Container) f(Tuple2(testCase, result))
```
sksamuel: So this filters out the test case properly, but we'd have to do it in the other places someone can register a test too.

reevn:
One possibility of getting rid of all the boilerplate involved in adding these new callbacks that are very similar to `beforeTest` and `afterTest` would be to have the default implementations of the new callbacks in the `TestListener` interface just delegate to `beforeTest`/`afterTest` once they pass the `TestType` check.

```kotlin
interface TestListener {
  suspend fun beforeContainer(testCase: TestCase) = if (testCase.type == TestType.Container) {
    beforeTest(testCase)
  } else {
    Unit
  }
}
```

This would spare us the need to override them in all the `TestListener` implementations in the project, but I think this would lead to confusion and open us up for more bugs, like for example:

A user wants to implement their own `TestListener` to add at the project level:

```kotlin
class MyListener : TestListener {
    override suspend fun beforeTest() {
        // some custom user code for beforeTest
    }

    override suspend fun beforeContainer() {
        // some custom user code for beforeContainer
        super.beforeContainer()
    }
}
```

The `super.beforeContainer` would call the overridden `beforeTest` without the user expecting it. There might also be the same kind of unintended behavior in the project itself when other features are added. For this reason I've treated the new callbacks as completely stand-alone, even though it includes a lot of boilerplate.

kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/time/ConstantNowExtensions.kt
```diff
@@ -105,6 +106,30 @@ class ConstantNowTestListener<Time : Temporal>(now: Time) :
    override suspend fun afterTest(testCase: TestCase, result: TestResult) {
       resetNow()
    }
+
+   override suspend fun beforeContainer(testCase: TestCase) {
+      if (testCase.type == TestType.Container) changeNow()
```
sksamuel: We probably only need to call changenow resetnow on the beforeAny / beforeTest cases as that will cover it for every test.

kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/system/wireListeners.kt
```diff
@@ -54,6 +55,38 @@ class SystemOutWireListener(private val tee: Boolean = true) : TestListener {
...
+   override suspend fun afterEach(testCase: TestCase, result: TestResult) {
+      if (testCase.type == TestType.Test) restoreStandardOutput()
+   }
+
+   override suspend fun beforeAny(testCase: TestCase) {
+      redirectStandardOut()
+   }
+
+   override suspend fun afterAny(testCase: TestCase, result: TestResult) {
+      restoreStandardOutput()
```
sksamuel: We probably only need to call redirectStandardOutrestoreStandardOutputresetnow on one of the beforeAny / beforeTest cases as that will cover it for every test.

kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/system/SystemPropertiesExtensions.kt
```diff
@@ -192,6 +193,30 @@ class SystemPropertyTestListener(newProperties: Map<String, String?>, mode: Over
...
+   override suspend fun afterEach(testCase: TestCase, result: TestResult) {
+      if (testCase.type == TestType.Test) resetSystemProperties()
+   }
+
+   override suspend fun beforeAny(testCase: TestCase) {
+      changeSystemProperties()
+   }
+
+   override suspend fun afterAny(testCase: TestCase, result: TestResult) {
+      resetSystemProperties()
```
sksamuel, line 194: We probably only need to call changeSystemProperties resetSystemProperties on one of the beforeAny / beforeTest cases as that will cover it for every test.

kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/system/SystemEnvironmentExtensions.kt
```diff
@@ -183,6 +184,30 @@ class SystemEnvironmentTestListener(environment: Map<String, String>, mode: Over
...
+   override suspend fun afterContainer(testCase: TestCase, result: TestResult) {
+      if (testCase.type == TestType.Container) resetSystemEnvironment()
+   }
+
+   override suspend fun beforeEach(testCase: TestCase) {
+      if (testCase.type == TestType.Test) changeSystemEnvironment()
+   }
+
+   override suspend fun afterEach(testCase: TestCase, result: TestResult) {
+      if (testCase.type == TestType.Test) resetSystemEnvironment()
```
sksamuel: We probably only need to call changeSystemEnvironment /  resetSystemEnvironment on one of the beforeAny / beforeTest cases as that will cover it for every test.

kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/system/SecurityManagerExtensions.kt
```diff
@@ -62,6 +63,30 @@ class SecurityManagerTestListener(securityManager: SecurityManager?) : SecurityM
    override suspend fun afterTest(testCase: TestCase, result: TestResult) {
       resetSecurityManager()
    }
+
+   override suspend fun beforeContainer(testCase: TestCase) {
+      if (testCase.type == TestType.Container) changeSecurityManager()
+   }
+
+   override suspend fun afterContainer(testCase: TestCase, result: TestResult) {
+      if (testCase.type == TestType.Container) resetSecurityManager()
```
sksamuel: We probably only need to call changeSecurityManager /  resetSecurityManager on one of the beforeAny / beforeTest cases as that will cover it for every test.

kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/locale/TimezoneExtensions.kt
```diff
@@ -58,6 +59,30 @@ class TimeZoneTestListener(timeZone: TimeZone) : TimeZoneListener(timeZone), Tes
...
+   override suspend fun beforeContainer(testCase: TestCase) {
+      if (testCase.type == TestType.Container) changeTimeZone()
+   }
+
+   override suspend fun afterContainer(testCase: TestCase, result: TestResult) {
+      if (testCase.type == TestType.Container) resetTimeZone()
+   }
+
+   override suspend fun beforeEach(testCase: TestCase) {
+      if (testCase.type == TestType.Test) changeTimeZone()
```
sksamuel: We probably only need to call resetTimeZone /  changeTimeZone on one of the beforeAny / beforeTest cases as that will cover it for every test.

kotest-extensions/kotest-extensions-testcontainers/src/jvmMain/kotlin/io/kotest/extensions/testcontainers/StartablePerTestListener.kt
```diff
@@ -35,4 +36,54 @@ class StartablePerTestListener<T : Startable>(val startable: T) : TestListener {
...
+      withContext(Dispatchers.IO) {
+         testLifecycleAwareListener.afterTest(testCase, result)
+         startable.stop()
+      }
+   }
+
+   override suspend fun beforeAny(testCase: TestCase) {
+      withContext(Dispatchers.IO) {
+         testLifecycleAwareListener.beforeTest(testCase)
+         startable.start()
```
sksamuel: This will call the .start() method 3 times, once for beforeTest, beforeAny and beforeEach / beforeContainer

kotest-extensions/kotest-extensions-testcontainers/src/jvmMain/kotlin/io/kotest/extensions/testcontainers/StartablePerSpecListener.kt
```diff
@@ -38,6 +39,28 @@ class StartablePerSpecListener<T : Startable>(val startable: T) : TestListener {
...
+      withContext(Dispatchers.IO) {
+         testLifecycleAwareListener.beforeTest(testCase)
+      }
+   }
+
+   override suspend fun beforeEach(testCase: TestCase) {
+      if (testCase.type != TestType.Test) return
+
+      withContext(Dispatchers.IO) {
+         testLifecycleAwareListener.beforeTest(testCase)
```
sksamuel: These shouldn't be implemented as the StartablePerSpecListener doesn't reset on each test.

kotest-extensions/kotest-extensions-spring/src/jvmMain/kotlin/io/kotest/spring/SpringListener.kt
```diff
@@ -32,20 +33,52 @@ object SpringListener : TestListener {
...
+   override suspend fun afterContainer(testCase: TestCase, result: TestResult) {
+      if (testCase.type == TestType.Container) postProcessTest(testCase)
+   }
+
+   override suspend fun beforeEach(testCase: TestCase) {
+      if (testCase.type == TestType.Test) preProcessTest(testCase)
+   }
+
+   override suspend fun afterEach(testCase: TestCase, result: TestResult) {
+      if (testCase.type == TestType.Test) postProcessTest(testCase)
```
sksamuel: These shouldn't be implemented as the beforeTest will take care of it

kotest-extensions/kotest-extensions-koin/src/jvmMain/kotlin/io/kotest/koin/KoinListener.kt
```diff
@@ -17,13 +19,41 @@ class KoinListener(
...
    override suspend fun afterTest(testCase: TestCase, result: TestResult) {
       stopKoin()
    }
+
+   override suspend fun beforeContainer(testCase: TestCase) {
+      if (testCase.type == TestType.Container) startKoinIfNotStarted()
+   }
+
+   override suspend fun afterContainer(testCase: TestCase, result: TestResult) {
+      if (testCase.type == TestType.Container) stopKoin()
```
sksamuel: These shouldn't be implemented as the beforeTest will take care of it

kotest-extensions/kotest-extensions-allure/src/jvmMain/kotlin/io/kotest/extensions/allure/AllureTestListener.kt
```diff
@@ -51,6 +49,38 @@ object AllureTestListener : TestListener, ProjectListener {
...
+   override suspend fun afterTest(testCase: TestCase, result: TestResult) {
+      stopAllureTestCase(testCase, result)
+   }
+
+   override suspend fun beforeContainer(testCase: TestCase) {
+      if (testCase.type == TestType.Container) startAllureTestCase(testCase)
+   }
+
+   override suspend fun afterContainer(testCase: TestCase, result: TestResult) {
+      if (testCase.type == TestType.Container) stopAllureTestCase(testCase, result)
```
sksamuel: These shouldn't be implemented as the beforeTest will take care of it

kotest-core/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
```diff
@@ -72,4 +71,79 @@ interface ContainerScope : CoroutineScope {
...
    }
 
+   /**
+    * Registers a [BeforeContainer] function that executes before every test with type [TestType.Container] in
+    * this scope.
+    */
+   fun beforeContainer(f: BeforeContainer) {
+      lifecycle.addListener(object : TestListener {
+         override suspend fun beforeContainer(testCase: TestCase) {
+            if (testCase.type == TestType.Container && description.isAncestorOf(testCase.description)) {
```
sksamuel: We can probably move the check to the engine.

sksamuel (changes requested):

sksamuel:
> One possibility of getting rid of all the boilerplate involved in adding these new callbacks that are very similar to `beforeTest` and `afterTest` would be to have the default implementations of the new callbacks in the `TestListener` interface just delegate to `beforeTest`/`afterTest` once they pass the `TestType` check.
> 
> ```kotlin
> interface TestListener {
>   suspend fun beforeContainer(testCase: TestCase) = if (testCase.type == TestType.Container) {
>     beforeTest(testCase)
>   } else {
>     Unit
>   }
> }
> ```
> 
> This would spare us the need to override them in all the `TestListener` implementations in the project, but I think this would lead to confusion and open us up for more bugs, like for example:
> 
> A user wants to implement their own `TestListener` to add at the project level:
> 
> ```kotlin
> class MyListener : TestListener {
>     override suspend fun beforeTest() {
>         // some custom user code for beforeTest
>     }
> 
>     override suspend fun beforeContainer() {
>         // some custom user code for beforeContainer
>         super.beforeContainer()
>     }
> }
> ```
> 
> The `super.beforeContainer` would call the overridden `beforeTest` without the user expecting it. There might also be the same kind of unintended behavior in the project itself when other features are added. For this reason I've treated the new callbacks as completely stand-alone, even though it includes a lot of boilerplate.

I've added a bunch of feedback which is along these lines.

kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/time/ConstantNowExtensions.kt
reevn: Thanks for pointing this out. I've mistakenly thought about the extension listeners differently and somehow thought that the functions executed are dependent on what the user writes in their tests, but that is obviously not the case 🙂

kotest-core/src/commonMain/kotlin/io/kotest/core/runtime/callbacks.kt
reevn: Addressed in [`3526470` (#1589)](https://github.com/kotest/kotest/pull/1589/commits/3526470a33381cff30a5f2a9b34ccfc14c575ba5)

kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/time/ConstantNowExtensions.kt
reevn: Addressed in [`2ed277b` (#1589)](https://github.com/kotest/kotest/pull/1589/commits/2ed277b0b70c6e9e285e8abfca44e6c18aff7715) & [`9b20d34` (#1589)](https://github.com/kotest/kotest/pull/1589/commits/9b20d342b938d45446fd4b8fa237793e2317bb50)

sksamuel (approved):

sksamuel:
Once you're happy with it, move it out of draft (a conflict to fix) and then I will merge.

reevn:
Ready to go from my side 👍

Issue #1573, opened 2020-07-20
Allow to disable SpringListener on final class warning


fauscik:
Hi,

I would like to be able to disable _Using SpringListener on a final class. If any Spring annotation fails to work, try making this class open._ warning which appears quite often in console logs and causes unnecessary noise in the output.
Currently it's printed using _println_ so I'm unable to disable it. Would it be possible to log it using some logger or allow to disable this warning?

Thanks

sksamuel:
To clarify, you have the *SpringListener *in a spec, and you are using it
with a normal final class and it works, you just don't want the warning?

fauscik:
Yes, simple as that.

sksamuel:
The simplest way would be to add

var ignoreWarning: Boolean = false

Then you can do SpringListener.ignoreWarning = true

Or change the object to be a class, so you can do

listener(SpringListener(false))

fauscik:
Ok, thanks for advice @sksamuel . If you don't mind, I'll go with some PR then.

sksamuel:
Yes a PR would be welcome.

PR #1590, opened 2020-07-23
ISSUE-1573: Allow to disable SpringListener warning on a final class

fauscik:
Hi,

https://github.com/kotest/kotest/issues/1573
As discussed I've introduced `ignoreWarning` to provide possibility to disable warning about using SpringListener on final classes. I decided to go with the property instead of changing to object to class to not break current behaviour.

Thanks

### DIFF
--- a/kotest-extensions/kotest-extensions-spring/src/jvmMain/kotlin/io/kotest/spring/SpringListener.kt
+++ b/kotest-extensions/kotest-extensions-spring/src/jvmMain/kotlin/io/kotest/spring/SpringListener.kt
@@ -1,11 +1,12 @@
 package io.kotest.spring
 
-import io.kotest.core.test.TestCase
-import io.kotest.core.test.TestResult
-import io.kotest.core.spec.Spec
 import io.kotest.core.extensions.ConstructorExtension
 import io.kotest.core.listeners.TestListener
 import io.kotest.core.spec.AutoScan
+import io.kotest.core.spec.Spec
+import io.kotest.core.test.TestCase
+import io.kotest.core.test.TestResult
+import io.kotest.mpp.sysprop
 import net.bytebuddy.ByteBuddy
 import net.bytebuddy.description.modifier.Visibility
 import net.bytebuddy.dynamic.loading.ClassLoadingStrategy
@@ -20,7 +21,7 @@ import kotlin.reflect.full.primaryConstructor
 
 object SpringListener : TestListener {
 
-   var ignoreWarning: Boolean = false
+   var ignoreSpringListenerOnFinalClassWarning: Boolean = false
 
    // Each Spec needs its own context. However, this listener is a singleton, so we need
    // to keep this map to separate those contexts instead of making this class non-singleton, thus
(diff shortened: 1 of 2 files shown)


kotest-extensions/kotest-extensions-spring/src/jvmMain/kotlin/io/kotest/spring/SpringListener.kt
```diff
@@ -20,6 +20,8 @@ import kotlin.reflect.full.primaryConstructor
 
 object SpringListener : TestListener {
 
+   var ignoreWarning: Boolean = false
```
LeoColman: Which warning are we ignoring? Better to place it in the variable name.

Also, could you please add it as a system property as well?

LeoColman (changes requested):

kotest-extensions/kotest-extensions-spring/src/jvmMain/kotlin/io/kotest/spring/SpringListener.kt
sksamuel: A system property is better yes. @fauscik that's my fault for suggesting the var, but I like keeping the var as well.

I would just do something like if (!ignoreWaring || sysprop("kotest.listener.spring.ignore.warnings") == true) { }

sksamuel (changes requested):

kotest-extensions/kotest-extensions-spring/src/jvmMain/kotlin/io/kotest/spring/SpringListener.kt
fauscik: I think system property makes sense as well as var, so I will keep both.

fauscik: I have added system prop, but I'm wondering, does it make sense to add dependecy on `kotest-common` to have possibility to use `sysprop` instead of `System.getProperty`?

sksamuel: kotest-common should be being pulled in anyway, as this depends on kotest-core

sksamuel: But I don't think it matters because sysprop is for multi platform and spring is jvm only

sksamuel (approved):

kotest-extensions/kotest-extensions-spring/src/jvmMain/kotlin/io/kotest/spring/SpringListener.kt
```diff
@@ -56,7 +59,9 @@ object SpringListener : TestListener {
          val klass = this::class.java
 
          return if (Modifier.isFinal(klass.modifiers)) {
-            println("Using SpringListener on a final class. If any Spring annotation fails to work, try making this class open.")
+            if (!ignoreSpringListenerOnFinalClassWarning || !sysprop("kotest.listener.spring.ignore.warning").toBoolean()) {
```
sksamuel: Does sysprop not return a null without a default ?

fauscik: It does return null, but `toBoolean` is defined for nullable `String?` and it returns false.

fauscik: Ok, but it's since kotlin 1.4. I will change it to support also 1.3

sksamuel:
:+1:

PR #1592, opened 2020-07-24
Added kotest impl of ExtensionContext #1503

sksamuel:


kotest-extensions/kotest-extensions-junit5extensions/src/jvmMain/kotlin/io/kotest/extensions/junit5/KotestExtensionContext.kt
```diff
@@ -0,0 +1,86 @@
...
+
+class ExtensionStore : ExtensionContext.Store {
+
+   private val map = mutableMapOf<Any?, Any?>()
+
+   override fun get(key: Any?): Any? = map[key]
+
+   override fun <V : Any?> get(key: Any?, requiredType: Class<V>?): V? {
+      return map[key] as? V
+   }
```
ashishkujoy: Type casting will here. I have tried out with following test case.

```kotlin
class ExtensionStoreTest : FunSpec({
   data class Person(val name: String, val age: Int)
   data class Vehicle(val name: String, val model: String)

   test("should give null when value for given is not of given type in store") {
      val extensionStore = ExtensionStore()
      val person = Person("foo", 20)

      extensionStore.put("foo", person)

      val vehicle = extensionStore.get("foo", Vehicle::class.java)

      vehicle shouldBe null
   }
})
```

I think here expectation was to get null, but we get type case error.

```bash
java.lang.ClassCastException: class io.kotest.extensions.junit5.ExtensionStoreTest$1$Person cannot be cast to class io.kotest.extensions.junit5.ExtensionStoreTest$1$Vehicle (io.kotest.extensions.junit5.ExtensionStoreTest$1$Person and io.kotest.extensions.junit5.ExtensionStoreTest$1$Vehicle are in unnamed module of loader 'app')
	at io.kotest.extensions.junit5.ExtensionStoreTest$1$1.invokeSuspend(ExtensionStoreTest.kt:16)
	at io.kotest.extensions.junit5.ExtensionStoreTest$1$1.invoke(ExtensionStoreTest.kt)
	at io.kotest.core.runtime.ExecutionsKt$executeWithBehaviours$2$1.invokeSuspend(executions.kt:16)
	at io.kotest.core.runtime.ExecutionsKt$executeWithBehaviours$2$1.invoke(executions.kt)
	at io.kotest.core.runtime.ExecutionsKt.wrapTestWithGlobalAssert(executions.kt:43)
	at io.kotest.core.runtime.ExecutionsKt$executeWithBehaviours$2.invokeSuspend(executions.kt:15)
	at io.kotest.core.runtime.ExecutionsKt$executeWithBehaviours$2.invoke(executions.kt)
```

sksamuel: I checked the junit default implementation and they throw when the type is not right.

I've added your test, changed it to expect a throw and tidied up the code, including adding in namespace support.

kotest-extensions/kotest-extensions-junit5extensions/src/jvmMain/kotlin/io/kotest/extensions/junit5/KotestExtensionContext.kt
```diff
@@ -0,0 +1,104 @@
...
+   override fun getStore(namespace: ExtensionContext.Namespace): ExtensionContext.Store {
+      return ExtensionStore(namespace)
+   }
+}
+
+class KotestTestInstances(private val instance: Spec) : TestInstances {
+   override fun getInnermostInstance(): Any = instance
+   override fun getEnclosingInstances(): MutableList<Any> = mutableListOf(instance)
+   override fun getAllInstances(): MutableList<Any> = mutableListOf(instance)
+   override fun <T : Any?> findInstance(requiredType: Class<T>?): Optional<T> = Optional.of(instance as T)
```
ashishkujoy: I think here as well we can potentially get a class cast exception.

sksamuel: Thanks, fixed.

PR #1595, opened 2020-07-24
#1543 Adds IterableEq to check equality of iterable by comparing each elements

ashishkujoy:


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
```diff
@@ -15,15 +17,32 @@ interface Eq<T> {
...
  */
-fun <T> eq(actual: T, expected: T): Throwable? = when {
-   actual is Map<*, *> && expected is Map<*, *> -> MapEq.equals(actual, expected)
-   actual is Throwable && expected is Throwable -> ThrowableEq.equals(actual, expected)
-   actual is Regex && expected is Regex -> RegexEq.equals(actual, expected)
-   actual is String && expected is String -> StringEq.equals(actual, expected)
-   actual is Number && expected is Number -> NumberEq.equals(actual, expected)
-   shouldShowDataClassDiff(actual, expected) -> DataClassEq.equals(actual as Any, expected as Any)
-   else -> DefaultEq.equals(actual as Any, expected as Any)
+fun <T: Any?> eq(actual: T, expected: T): Throwable? {
```
ibrahimyilmaz, line 20: Why do we prefer to  T nullable `<T: Any?>` in the template instead of making `T?` in the parameters?

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
```diff
@@ -15,15 +17,32 @@ interface Eq<T> {
...
+   // if we have null and non null, usually that's a failure, but people can override equals to allow it
+   return when {
+      actual == null && expected == null -> null
+      actual == null && expected != null && actual != expected -> actualIsNull(expected)
```
ibrahimyilmaz, line 25: `actual != expected` in both case adds extra complexity to when  statement.
Also when one of them is null and another is not, why do we check this equality?

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
```diff
@@ -15,15 +17,32 @@ interface Eq<T> {
...
-   actual is Map<*, *> && expected is Map<*, *> -> MapEq.equals(actual, expected)
-   actual is Throwable && expected is Throwable -> ThrowableEq.equals(actual, expected)
-   actual is Regex && expected is Regex -> RegexEq.equals(actual, expected)
-   actual is String && expected is String -> StringEq.equals(actual, expected)
-   actual is Number && expected is Number -> NumberEq.equals(actual, expected)
-   shouldShowDataClassDiff(actual, expected) -> DataClassEq.equals(actual as Any, expected as Any)
-   else -> DefaultEq.equals(actual as Any, expected as Any)
+fun <T: Any?> eq(actual: T, expected: T): Throwable? {
+   // if we have null and non null, usually that's a failure, but people can override equals to allow it
+   return when {
```
ibrahimyilmaz, line 22: reference equality `===` case is missing.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
```diff
@@ -15,15 +17,32 @@ interface Eq<T> {
...
+      actual == null && expected == null -> null
```
ibrahimyilmaz, line 24: reference equality could be used for early return
`actual === expected`

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
```diff
@@ -0,0 +1,21 @@
+package io.kotest.assertions.eq
+
+import io.kotest.assertions.Actual
+import io.kotest.assertions.Expected
+import io.kotest.assertions.failure
+import io.kotest.assertions.show.show
+
+object IterableEq : Eq<Iterable<*>> {
+   override fun equals(actual: Iterable<*>, expected: Iterable<*>): Throwable? {
+      val actualAsList = actual.toList()
```
ibrahimyilmaz: if performance matters, `while` statement can be used instead of `toList`. Also it is simpler than `toList`,`zip` and `any` combo.

```
while (actual.hasNext() && expected.hasNext()) {
        int comparison = actual.next().compareTo(expected.next());
        if (comparison != 0) {
            return failure(Expected(expected.show()), Actual(actual.show()))
        }
    }
return null
```

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/should.kt
```diff
@@ -11,28 +13,11 @@ infix fun <T, U : T> T.shouldBe(expected: U?) {
...
-            errorCollector.collectOrThrow(actualIsNull(expected))
-         } else if (actual != null && expected == null && actual != expected) {
-            errorCollector.collectOrThrow(expectedIsNull(actual))
-         } else if (actual != null && expected != null) {
-            // for two non-null values, we use the [eq] typeclass.
-            val t = eq(actual, expected)
-            if (t != null)
-               errorCollector.collectOrThrow(t)
-         }
+         eq(actual, expected)?.let { errorCollector.collectOrThrow(it) }
```
ibrahimyilmaz: Minor: this could be 
`eq(actual, expected)?.let(errorCollector::collectOrThrow)`

ashishkujoy: Done

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
sksamuel: I prefer the code base to be more FP style, and a while loop with an early return is not. This is definitely faster, but micro optimizations in a test framework are less important than more robust code imo.

sksamuel:
This should be merged into master not release/4.1.x

ashishkujoy:
> This should be merged into master not release/4.1.x

Will need to merge https://github.com/kotest/kotest/pull/1577 as well in master.

ashishkujoy:
Opened https://github.com/kotest/kotest/pull/1614 for merging in master

PR #1609, opened 2020-08-04
Fixed ByteArray generator not creating any values

### LINKED ISSUE #1608: Sequence is empty error when using Arb.byteArrays
Using Arb.bytesArrays() throws a `java.util.NoSuchElementException`: Sequence is empty.

It looks like the cause is this line:
https://github.com/kotest/kotest/blob/aa26b494dbb80294a728ee422ea3a9da615b5c98/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/bytearrays.kt#L16

It should be `generateSequence` instead of `sequence`.

tlusk:
Arb.byteArrays() was not working correctly and was creating a sequence
with no values.

The unit tests were still passing because the forAll inspector was being
run on an empty list and therefore wasn't actually testing any values.

This fixes #1608


sksamuel (approved):

sksamuel:
Thanks

PR #1614, opened 2020-08-08
Adds Iterable eq for checking equality of iterables

ashishkujoy:


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
```diff
@@ -15,15 +17,35 @@ interface Eq<T> {
...
-   shouldShowDataClassDiff(actual, expected) -> DataClassEq.equals(actual as Any, expected as Any)
-   else -> DefaultEq.equals(actual as Any, expected as Any)
+fun <T: Any?> eq(actual: T, expected: T): Throwable? {
+   // if we have null and non null, usually that's a failure, but people can override equals to allow it
+   return when {
+      actual === expected -> null
+      actual == null && expected == null -> null
+      actual == null && expected != null && actual != expected -> actualIsNull(expected)
+      actual != null && expected == null && actual != expected -> expectedIsNull(actual)
+      actual !=null && expected != null -> when {
```
sksamuel, line 27: What is the null changes for ?

ashishkujoy: So this is related to nested equality changes.
Earlier these null checks were in `should` function but there we were throwing error for inequality. While checking each element of Map or Iterables I am now using `eq`,  there I need to have all these null checks else I was getting null pointer exception. So I have moved null checks from `should` to `eq`.

So from:
https://github.com/kotest/kotest/blob/cb23f758ca42669b75962ab642120a762fc2df74/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/should.kt#L11-L26

It changes to.
        https://github.com/ashishkujoy/kotest/blob/4065ea2b6908f2063befcdda52d50bd1c22333f1/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/should.kt#L14-L16

sksamuel:
I see. Did the error messages change at all ?

ashishkujoy:
Messages are still the same as they were before these changes.

sksamuel:
Ok lgmt merge whenever

PR #1626, opened 2020-08-12
Fix forAll(1) suspend parameters

aballano:
As is done in propertyTest2 and others:

* Added suspend modifier to PropertyContext
* Removed suspend modifier for inline functions

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/propertyTest1.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/propertyTest1.kt
@@ -87,63 +87,63 @@ suspend inline fun <reified A> checkAll(
 )
 
 @JvmName("forAllExt")
-suspend fun <A> Gen<A>.forAll(property: PropertyContext.(A) -> Boolean) =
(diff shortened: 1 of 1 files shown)


kotest-property/src/commonMain/kotlin/io/kotest/property/propertyTest1.kt
```diff
@@ -87,63 +87,63 @@ suspend inline fun <reified A> checkAll(
...
+   crossinline property: PropertyContext.(A) -> Boolean
 ) = forAll(iterations, PropTestConfig(), property)
 
 suspend inline fun <reified A> forAll(
    config: PropTestConfig,
-   crossinline property: suspend PropertyContext.(A) -> Boolean
+   crossinline property: PropertyContext.(A) -> Boolean
 ) = forAll(PropertyTesting.defaultIterationCount, config, property)
 
 suspend inline fun <reified A> forAll(
```
aballano, line 143: A quick check made me also realise that it might not be needed to have the suspend modifier here neither since it's inline 🤔

sksamuel:
:+1:

PR #1629, opened 2020-08-14
Add matchers for Kotlinx Time

LeoColman:
This commit implements most of the basic matchers with LocalDate + LocalDateTime for the new `kotlinx.time` types.

As the kotlinx library improves, we should follow and try to improve on our side as well.

As the matchers are very similar to our current datetime matchers, I don't believe we need to add extra documentation at the matchers page, but can do if you find it necessary.


LeoColman:
I couldn't find a way to add these to the `ConstantNowListener`, as mocking is not very easy on multiplatform kotlin.

I created https://github.com/Kotlin/kotlinx-datetime/issues/33 to try to find a solution, let's see what comes from there :D

sksamuel:
Some test failures, otherwise looks good.

sksamuel:
Merging in

PR #1633, opened 2020-08-16
Update arbitrary number generator docs

mattmikolay:
I noticed some of the docs describing the arbitrary number generators are outdated. Feel free to double check my edits, but this PR will fix the descriptions for the integer and double generators.


sksamuel:
Nice thanks.

PR #1648, opened 2020-08-25
Added some new tests

crazyk2:


LeoColman:
Extra tests are always nice!
Good one :)

PR #1672, opened 2020-08-31
Make Set comparisons use .contains() instead of a linear search

### LINKED ISSUE #1671: Set to Set comparisons have poor performance in 4.2.x
`someSet shouldBe someOtherSet` scales poorly as of 4.2.0. 

This also affects Map comparisons, since one of the first steps is comparing their key sets.

Take this assertion:

```Kotlin
List(1000) { it }.toSet() shouldBe List(1000) { it }.reversed().toSet()
```

On my machine, it takes about 80 seconds to complete.

This is because the comparison is being done via linear search for each individual element, which I believe means it's n^2 complexity [source](https://github.com/kotest/kotest/blob/60a46ed1805f6f7a0ee60e71086d298d501d8470/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt#L22):

```Kotlin
   private fun equalsIgnoringOrder(actual: Set<*>, expected: Set<*>): Boolean {
      return actual.all { elementInActualSet ->
         expected.any { elementInExpectedSet ->
            eq(elementInActualSet, elementInExpectedSet) == null
         }
      }
   }
```

If we just use `Set.contains(...)` instead:
```Kotlin
   private fun equalsIgnoringOrder(actual: Set<*>, expected: Set<*>): Boolean {
      return actual.all { elementInActualSet ->
         expected.contains(elementInActualSet)
      }
   }
```
Then the complexity becomes O(n) and runtime drops down to about 3ms.

RBusarow:
Closes #1671

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
@@ -21,9 +21,7 @@ object IterableEq : Eq<Iterable<*>> {
 
    private fun equalsIgnoringOrder(actual: Set<*>, expected: Set<*>): Boolean {
       return actual.all { elementInActualSet ->
-         expected.any { elementInExpectedSet ->
-            eq(elementInActualSet, elementInExpectedSet) == null
-         }
+         expected.contains(elementInActualSet)
       }
    }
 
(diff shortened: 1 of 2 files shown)


sksamuel:
:+1:

ashishkujoy:
@RBusarow that linear search was added to be able to use `eq` function for inner elements check.
To support equality check for complex structure for example.
```kotlin
setOf(arrayListOf(1, 2, 3), 1, listOf(arrayListOf(1, 2, 3))) shouldBe setOf(arrayListOf(1, 2, 3), 1, listOf(arrayListOf(1, 2, 3)))
```

It was my mistake I have not added any test for that. Can u please check if this passes with your change my hunch is i won't pass.

RBusarow:
@ashishkujoy That test does pass.  

`Set.contains(...)` may use `.hashCode()` as an optimization under the hood to fetch the element, but ultimately it still winds up calling `.equals(...)` before returning true/false.

For instance, Java's [HashSet.contains](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/HashSet.html#contains(java.lang.Object)) says:
```
Returns true if this set contains the specified element. More formally, returns true if and only if this set contains an element e such that Objects.equals(o, e).
```

ashishkujoy:
Thanks for confirmation, @RBusarow

sksamuel:
So what @ashishkujoy is talking about is our Eq typeclass, not the equals method in java.lang.Object.
We use the `eq` typeclass as a way to compare instances and generate useful error messages (rather than boolean).
For this change however, we are only interested in true/false so we're ok to use the equals method.

sksamuel:
I've added your test case ashish to our test suite.

PR #1674, opened 2020-09-01
Fix and remove double negative in empty directory assertions

litpho:
Current (reversed) situation:
File.shouldBeNonEmptyDirectory to check if a directory is empty
File.shouldNotBeNonEmptyDirectory to check if a directory is not empty

I suggest removing the negation in the name, leading to:
File.shouldBeEmptyDirectory to check if a directory is empty
File.shouldNotBeEmptyDirectory to check if a directory is not empty
and
Path.shouldBeEmptyDirectory to check if a directory is empty
Path.shouldNotBeEmptyDirectory to check if a directory is not empty

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/file/FileMatchersTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/file/FileMatchersTest.kt
@@ -11,6 +11,8 @@ import io.kotest.matchers.file.haveExtension
 import io.kotest.matchers.file.shouldBeADirectory
 import io.kotest.matchers.file.shouldBeAFile
 import io.kotest.matchers.file.shouldBeAbsolute
+import io.kotest.matchers.file.shouldBeEmptyDirectory
+import io.kotest.matchers.file.shouldBeNonEmptyDirectory
 import io.kotest.matchers.file.shouldBeRelative
 import io.kotest.matchers.file.shouldBeSymbolicLink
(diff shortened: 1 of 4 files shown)


sksamuel:
Name change makes sense. We have to keep old ones in and deprecate them though.

litpho:
> Name change makes sense. We have to keep old ones in and deprecate them though.

I'll be happy to do that. I assume the old names should refer to the old (incorrect) meaning so no existing client code breaks?

sksamuel:
I would fix the old ones so they are correct (flip the is empty check or point to the new ones), and deprecate them in favour of the new names.

sksamuel:
Lovely, will merge once green.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/file/FileMatchersTest.kt
```diff
@@ -149,10 +151,17 @@ class FileMatchersTest : FunSpec() {
...
+    test("directory should be empty (deprecated)") {
+       val dir = Files.createTempDirectory("testdir").toFile()
+       dir.shouldNotBeNonEmptyDirectory()
+       dir.resolve("testfile.txt").writeBytes(byteArrayOf(1, 2, 3))
+       dir.shouldBeNonEmptyDirectory()
+    }
+
     test("directory should be empty") {
-       val dir = Files.createTempDirectory("testdir")
+       val dir = Files.createTempDirectory("testdir").toFile()
```
litpho, line 162: Tested the Path extension methods rather than the File extension methods

PR #1687, opened 2020-09-04
Add Spec EnabledIf

### LINKED ISSUE #1683: EnabledIf annotation for specs

LeoColman:
This commit adds the @EnabledIf annotation for specs. The way that Kotlin handles annotations isn't very fexible, and thus
using this annotation is verbose and repetitive. We must reference a class, as we can't reference just a single method as we do for
Test EnabledIf.

This currently only works on JVM, as does @Ignored.

Closes #1683

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/annotation/specAnnotations.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/annotation/specAnnotations.kt
@@ -1,5 +1,8 @@
 package io.kotest.core.annotation
 
+import io.kotest.core.spec.Spec
+import kotlin.reflect.KClass
+
 /**
  * Attach tag to [io.kotest.core.spec.Spec] and a spec excluded by a tag expression won't be instantiated.
  * An unannotated spec will still be instantiated to order to check if root tests are included.
@@ -16,3 +19,17 @@ annotation class Tags(vararg val values: String)
 @Target(AnnotationTarget.CLASS)
 @Retention(AnnotationRetention.RUNTIME)
 annotation class Ignored
+
+/**
+ * Attach to [io.kotest.core.spec.Spec], and the logic inside [enabledIf] will be executed
+ * to define if a Spec will be instantiated or executed.
+ * 
+ * [SpecEnabaledIf] must contain a no-arg constructor as it will be used via reflection.
+ */
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class EnabledIf(val enabledIf: KClass<out EnabledCondition>)
+
+interface EnabledCondition {
+    fun enabled(specKlass: KClass<out Spec>): Boolean
+}
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io.kotest.engine/extensions/IgnoredSpecDiscoveryExtension.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io.kotest.engine/extensions/IgnoredSpecDiscoveryExtension.kt
@@ -7,10 +7,11 @@ import io.kotest.mpp.hasAnnotation
 import kotlin.reflect.KClass
 
 /**
- * Filters out any [Spec] annotated with @[Ignored].
+ * Filters out any [Spec] annotated with @[Ignored]
  */
 object IgnoredSpecDiscoveryExtension : DiscoveryExtension {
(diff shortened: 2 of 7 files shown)


LeoColman:
This is still missing some docs. 
I'll get to it if the syntax is agreed upon.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/annotation/specAnnotations.kt
```diff
@@ -16,3 +19,17 @@ annotation class Tags(vararg val values: String)
...
+
+/**
+ * Attach to [io.kotest.core.spec.Spec], and the logic inside [enabledIf] will be executed
+ * to define if a test will be instantiated or executed.
+ * 
+ * [SpecEnabaledIf] must contain a no-arg constructor as it will be used via reflection.
+ */
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class EnabledIf<T : SpecEnabledIf>(val enabledIf: KClass<T>)
```
sksamuel: I was thinking something along the lines of

`annotation class EnabledIf(val test: (KClass<*>) -> Boolean)`

The function could be replaced with a type as well.

LeoColman:
That doesn't compile

LeoColman:
Unless I looked at it wrong. Annotations must be constants at compile time

sksamuel:
What about:

annotation class EnabledIf(val test: KClass<EnabledCondition>)

abstract class EnabledCondition {
   abstract fun test(kClass: KClass<out Spec>): Boolean
}

LeoColman:
I tried that as well, but it won't work when you try to use it on a class.

Unless it supports subclasses, that `KClass<EnabledCondition>` is useless, and thus requires the generic type

LeoColman:
Hmm... I might actually be able to add it as a contravariant generic. It may work
Let me tweak it

LeoColman:
@sksamuel take a look now

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/annotation/specAnnotations.kt
```diff
@@ -16,3 +19,17 @@ annotation class Tags(vararg val values: String)
 @Target(AnnotationTarget.CLASS)
 @Retention(AnnotationRetention.RUNTIME)
 annotation class Ignored
+
+/**
+ * Attach to [io.kotest.core.spec.Spec], and the logic inside [enabledIf] will be executed
+ * to define if a test will be instantiated or executed.
```
sksamuel: to define if a spec will be instantiated or executed

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/annotation/specAnnotations.kt
```diff
@@ -16,3 +19,17 @@ annotation class Tags(vararg val values: String)
...
+ * 
+ * [SpecEnabaledIf] must contain a no-arg constructor as it will be used via reflection.
+ */
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class EnabledIf(val enabledIf: KClass<out SpecEnabledIf>)
+
+abstract class SpecEnabledIf {
```
sksamuel: I think EnabledCondition is better.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io.kotest.engine/extensions/IgnoredSpecDiscoveryExtension.kt
```diff
@@ -1,16 +1,24 @@
...
 
 /**
- * Filters out any [Spec] annotated with @[Ignored].
+ * Filters out any [Spec] annotated with @[Ignored] or specs which [@EnabledIf] evaluates to false.
  */
 object IgnoredSpecDiscoveryExtension : DiscoveryExtension {
    override fun afterScan(classes: List<KClass<out Spec>>): List<KClass<out Spec>> {
-      return classes.filterNot { it.hasAnnotation<Ignored>() }
+      return classes
+         .filterNot { it.hasAnnotation<Ignored>() || !it.isEnabled() }
```
sksamuel: I think it's clearer if we add another discovery extension. Separation of concerns.

sksamuel (changes requested):
Overall looks great, just a couple of tweaks needed.

LeoColman:
@sksamuel 
Please take a look if this pleases you

sksamuel (approved):

Issue #1646, opened 2020-08-25
Arb flatmap with A -> Arb<B> transformation type


myuwono:
**Which version of Kotest are you using**

Kotest Version: 4.1.3

The provided flatMap type signature of `fun <A, B> Arb<A>.flatMap(fn: A -> List<B>): Arb<B>` is not exactly very pleasant to use currently. I've been using another flatMap with this signature `fun <A, B> Arb<A>.flatMap(fn: (A) -> Arb<B>): Arb<B>`  in one of our kotlin projects internally. So far it seems to work nicely and doesn't seem to suffer from the `sequence.flatMap` issues described in https://github.com/kotest/kotest/issues/1500 (if i understand the problem correctly).

Here's my current implementation

```kotlin
import io.kotest.property.Arb
import io.kotest.property.RandomSource
import io.kotest.property.Sample
import io.kotest.property.arbitrary.next

fun <A, B> Arb<A>.flatMap(fn: (A) -> Arb<B>): Arb<B> = object : Arb<B>() {
    override fun edgecases(): List<B> = this@flatMap.edgecases().flatMap { fn(it).edgecases() }
    override fun values(rs: RandomSource): Sequence<Sample<B>> =
        this@flatMap.generate(rs).zip(generateSequence { rs.random.nextLong() }) { sample, nextLong ->
            Sample(fn(sample.value).next(RandomSource.seeded(nextLong)))
        }
}
```

@sksamuel what do you think? I'm keen to contribute this to the library if this implementation makes sense to you?

sksamuel:
By using .next() inside the implementation this really boils down to a signature like Arb.<A>.flatMap(f: A -> B) since you are pulling a single value from the arb returned by the function., which is invoked every time. 

Yes it would probably work for most arbs, but not for things like Arb.distinct() where each invoking fn() each time would throw away memory of previous values.

Given an Arb is effectively just an infinite sequence, I'm not sure how flatMap can really work with nested Arbs.

myuwono:
Thanks @sksamuel 

> By using .next() inside the implementation this really boils down to a signature like Arb..flatMap(f: A -> B) since you are pulling a single value from the arb returned by the function., which is invoked every time.

[...]

PR #1688, opened 2020-09-04
#1646 introduce single emission via value(rs) to Arb

myuwono:
- mainly just to make sure backward compatibility
- fixed flatMap
- fixed merge


kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
```diff
@@ -101,15 +112,15 @@ abstract class Exhaustive<out A> : Gen<A>() {
...
-      values.shuffled(it.random).asSequence()
-   }
+   fun toArb(): Arb<A> = Arb.of(values)
 
    companion object
 }
 
 data class RandomSource(val random: Random, val seed: Long) {
+
+   fun sequence(): Sequence<RandomSource> = generateSequence { seeded(random.nextLong()) }
```
sksamuel: Since Random is mutable does it seem more intuitive to pass the same rs around like before.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
```diff
@@ -9,17 +9,29 @@ import io.kotest.property.Shrinker
...
+/**
+ * Returns a sequence from values generated from this arb.
+ * Edgecases will be ignored.
+ */
+fun <A> Arb<A>.randomSequence(rs: RandomSource = RandomSource.Default): Sequence<A> {
+   val valuesIterator = this.values(rs).iterator()
+   return if (valuesIterator.hasNext()) {
+      generateSequence { valuesIterator.next() }.map { it.value }
+   } else {
+      rs.sequence().map { value(it) }.map { it.value }
```
sksamuel: i think this code is in two places

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
```diff
@@ -29,6 +41,7 @@ fun <A> Arb<A>.next(rs: RandomSource = RandomSource.Default): A = single(rs)
 fun <A> arb(edgecases: List<A> = emptyList(), f: (RandomSource) -> Sequence<A>) = object : Arb<A>() {
```
sksamuel, line 32: Should we deprecate this too

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
```diff
@@ -9,17 +9,29 @@ import io.kotest.property.Shrinker
 import io.kotest.property.map
 import io.kotest.property.sampleOf
 
+/**
+ * Returns a sequence from values generated from this arb.
+ * Edgecases will be ignored.
+ */
+fun <A> Arb<A>.randomSequence(rs: RandomSource = RandomSource.Default): Sequence<A> {
```
sksamuel: This exists to do the logic of which method to use - value or values. Given that its just a workaround might it be clearer to have this in the arb class itself so it's easy to see that this should be used and not values when adding new arbs.

myuwono: Awesome I'll make that change

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
myuwono: Yeah i think we should

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
```diff
@@ -82,6 +74,19 @@ abstract class Arb<out A> : Gen<A>() {
    @Deprecated("implement one value at a time using value(rs)", ReplaceWith("value(rs)"))
    open fun values(rs: RandomSource): Sequence<Sample<A>> = emptySequence()
 
+   /**
+    * Returns a sequence from values generated from this arb.
+    * Edgecases will be ignored.
+    */
+   fun randomSamples(rs: RandomSource = RandomSource.Default): Sequence<Sample<A>> {
```
sksamuel: random is kinda implied by the fact this is an arb and you pass in the rs.
Maybe on the next pr rename to just samples ?

myuwono: awesome - come to think of it, even `value(rs)` really is `sample(rs: RandomSeed): Sample<A>`. I can rename that too in the subsequent PR if that makes more sense.

sksamuel:
Yes I suppose it is, sample(rs) seems good to me, whatever you prefer.

sksamuel:
So all looks good. The only thing I'm not struck on is the sequence on RandomSource. We don't need a sequence of randoms, since the same random can be re-used, so it seems to be pretending to be an immutable value when it's not. I guess you added it to simply the implementation of some arbs so you can just map.

myuwono:
> The only thing I'm not struck on is the sequence on RandomSource. We don't need a sequence of randoms, since the same random can be re-used, so it seems to be pretending to be an immutable value when it's not. I guess you added it to simply the implementation of some arbs so you can just map.

Yeah 100%. I don't think that was correct either. I've removed the the sequence from random source altogether. 

I've moved the sample creation under `Arb.samples(rs)` which is much more ergonomic now.

```kotlin
   fun samples(rs: RandomSource = RandomSource.Default): Sequence<Sample<A>> {
      val valuesIterator = values(rs).iterator()
      return if (valuesIterator.hasNext()) {
         generateSequence { valuesIterator.next() }
      } else {
         generateSequence { sample(rs) } // here
      }
   }
```

I've pushed another commit to rename `randomSamples(rs)` to `samples(rs)` and `value(rs)` to `sample(rs)`. 

I believe this PR should be ready to merge when you give the green light.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
myuwono: added deprecation notice.

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
myuwono: awesome, yeah agreed - i believe this wasn't intentional. removed.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
myuwono: fixed, i've put that under `Arb.samples(rs)` as suggested. It's much nicer

Issue #1679, opened 2020-09-01
Add property test gens for kotlinx datetime


sksamuel:
**Please describe the feature you'd like to see including any solutions in mind if you have any**

PR #1690, opened 2020-09-05
Add property test gens for kotlinx datetime #1679

sksamuel:
@myuwono as well


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
```diff
@@ -40,7 +49,7 @@ fun <A> arb(edgecases: List<A> = emptyList(), f: (RandomSource) -> Sequence<A>)
  * supplied edge case values, and provides values from sequence returning function.
  */
 @Deprecated(
-   "This function will no longer accept Sequence<A>. Use Arb.create or arb with (RandomSource -> A) for compatibility"
+   "This function will no longer accept Sequence<A>. Use Arb.create or arb with (RandomSource -> A) for compatibility. This function will be removed in 4.5."
```
ashishkujoy, line 52: Should we consider adding `replaceWith` argument to `@Deprecated` that will help IDE to do code assistance in replacing it with new api.

[Reference](https://blog.8bitzen.com/posts/18-05-2018-kotlin-deprecated-replace-with/#:~:text=Anyway%20in%20Kotlin%20there%20is,to%20replace%20it%20with%3A%20replaceWith%20.)

ashishkujoy (approved):

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
sksamuel: We can't because the required code is different. Users need to modify their logic.

PR #1702, opened 2020-09-07
Updated a bunch of arbs to use the new non-sequence builders

sksamuel:
@myuwono


myuwono:
Thanks heaps for this! I wonder if you mind me contributing to this branch?

sksamuel:
I don't like keeping branches around to get stale. I'll merge this in and you can make another building off master then.

PR #1704, opened 2020-09-07
Loosen result matchers

### LINKED ISSUE #1703: Loosen nullability on Result matchers
Currently all failure `Result` matchers are defined on `Result<Any>`. This makes it impossible to use them when expected type is `Result<T?>`. This is backwards compatible as it loosens the type.

https://github.com/kotest/kotest/blob/4354170a27d06583477b7ea974bcc537f39cd48c/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/result/matchers.kt#L43-L50

https://github.com/kotest/kotest/blob/4354170a27d06583477b7ea974bcc537f39cd48c/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/result/matchers.kt#L59-L64

MiSikora:
I was not sure about 56da94d. If you don't like it I'll revert the commit.

Closes #1703.


sksamuel:
:+1:

PR #1712, opened 2020-09-09
move more arbs to arbitraries and introduce withEdgecases

myuwono:
@sksamuel moved more arbs to the new builder.

i've also introduced `withEdgecases`.


myuwono:
closes #1713

PR #1718, opened 2020-09-17
update function signature for Arb.bind/5

phanimahesh:
It requires Arbs instead of Gens, must be a mistake since there's no reason to restrict it.


sksamuel (approved):

sksamuel:
:+1:

phanimahesh:
@sksamuel Can you please release a new version as well? I would rather not include a hack in my codebase to circumvent this.

sksamuel:
You can use a snapshot in the interim. The badge on the home page has the
version numbers

PR #1721, opened 2020-09-19
Add data driven tests with automatic nesting support

### LINKED ISSUE #1537: Data driven tests should utilise data classes and automatically register test names (ala rollup)

sksamuel:
Closes #1537

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
@@ -37,9 +37,24 @@ interface ContainerScope : CoroutineScope {
    val defaultConfig: TestCaseConfig
 
    /**
+    * Registers a test to this scope, chosing the most specific type if this container supports
+    * both nested tests and terminal tests.
+    */
+   suspend fun addTest(
+      name: DescriptionName.TestName,
+      test: suspend TestContext.() -> Unit,
+   ) {
+      TODO()
+   }
+
+   /**
     * Adds a new test case to this scope with type [TestType.Container].
     */
-   suspend fun addContainerTest(name: DescriptionName.TestName, xdisabled: Boolean, test: suspend TestContext.() -> Unit) {
+   suspend fun addContainerTest(
+      name: DescriptionName.TestName,
+      xdisabled: Boolean,
+      test: suspend TestContext.() -> Unit,
+   ) {
       addTest(name, xdisabled, defaultConfig, TestType.Container, test)
    }
 
@@ -51,7 +66,7 @@ interface ContainerScope : CoroutineScope {
    }
 
    /**
-    * Registerd a new test case to this scope with the given test type.
+    * Registers a new test case to this scope with the given test type.
     */
    suspend fun addTest(
       name: DescriptionName.TestName,
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
@@ -40,12 +40,10 @@ interface ContainerScope : CoroutineScope {
     * Registers a test to this scope, wrapping the test in the most appropriate scope.
     * For example, invoking this method on a [DescribeSpec] will choose the `it` scope.
     */
-   suspend fun addTest(
+   abstract suspend fun addTest(
       name: String,
       test: suspend TestContext.() -> Unit,
-   ) {
-      TODO()
-   }
+   )
 
    /**
     * Adds a new test case to this scope with type [TestType.Container].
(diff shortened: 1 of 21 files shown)


ashishkujoy:
Looks good to me.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
```diff
@@ -36,10 +36,25 @@ interface ContainerScope : CoroutineScope {
 
    val defaultConfig: TestCaseConfig
 
+   /**
+    * Registers a test to this scope, wrapping the test in the most appropriate scope.
+    * For example, invoking this method on a [DescribeSpec] will choose the `it` scope.
```
zeitlinger, line 41: `[DescribeSpec]` cannot be resolved (import missing)

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
```diff
@@ -36,10 +36,25 @@ interface ContainerScope : CoroutineScope {
...
+    */
+   suspend fun addTest(
+      name: String,
+      test: suspend TestContext.() -> Unit,
+   ) {
+      TODO()
```
zeitlinger: should this delegate to another `addTest` - IIUC, this can fail at runtime.

zeitlinger: I tried to make it abstract, which shows that this is not implemented everywhere.

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/datatest/DataTest.kt
```diff
@@ -0,0 +1,123 @@
...
+import io.kotest.engine.KotestEngineLauncher
+import io.kotest.engine.listener.TestEngineListener
+import io.kotest.matchers.shouldBe
+
+class DataTest : FunSpec() {
+   init {
+
+      test("free spec should support data tests") {
+         val results = mutableMapOf<DescriptionName.TestName, TestStatus>()
+         val listener = object : TestEngineListener {
```
zeitlinger: extract `TestEngineListener`, since it's repeated in all tests.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
zeitlinger: AFAIK, if the test style does not support adding as a sub-test, it should be added to the current level.

sksamuel: Fixed.

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/datatest/DataTest.kt
sksamuel: Done

sksamuel:
I think this is good to merge now.

zeitlinger:
I found another `TODO()` - not introduced in this PR, but also in a method called `addTest`: https://github.com/kotest/kotest/blob/2d5f8bae55af2faaa7d4af4426508b23131379ef/kotest-framework/kotest-framework-api/src/jvmMain/kotlin/io/kotest/core/spec/style/AnnotationSpec.kt#L28

sksamuel:
Thanks, will address that separately.

zeitlinger:
great to have this in the next release :smile: 

Even though it's already merged, I'd like to understand:

`WordSpecWhenScope` and `GivenScope` return an error on `addTest` - does this mean data driven tests won't work there?

sksamuel:
Yes, because adding it to a given scope doesn't make any sense, as you would just have `given`, `when` without a `then`. You can add it to the `WhenScope` instead.

```
class Foo : BehaviorSpec {
  init {
    given("foo")
       When("boo") {
          forAll(...) { }
       }
  }
}
```

zeitlinger:
makes sense - what is the reason for `WordSpecWhenScope`?

sksamuel:
Same thing - when scope is for

```
when("xxx") {
  "yyyy" should {
       "test name" {
       }
   }
}
```

PR #1722, opened 2020-09-20
fix Arb.distinct() indefinitely hanging

### LINKED ISSUE #1701: Arb.distinct() does not resolve for take(n) where n > number of distinct items
because internally it calls sequence.distinct on an Arbitrary which is guaranteed to yield an infinite sequence.

myuwono:
closes #1701


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
```diff
@@ -64,28 +64,48 @@ fun <A, B> Arb<A>.flatMap(f: (A) -> Arb<B>): Arb<B> = object : Arb<B>() {
...
    }
 
-   override fun sample(rs: RandomSource): Sample<A> =
-      samples(rs).distinctBy { selector(it.value) }.first()
+   override fun values(rs: RandomSource): Sequence<Sample<A>> = generateSequence { sample(rs) }
+
+   private fun sampleFromPopulation(rs: RandomSource): List<A> =
+      this@distinctBy.generate(rs).map { it.value }.take(minSampleSize).toList()
+
+   private val sampledElements: MutableList<A> = mutableListOf()
```
myuwono: I don't think this is a good idea because it makes tests less repeatable. Another approach would be not to cache it or to always generate this list with a static RandomSeed, i.e.

```kotlin
private val sampledElements: List<A> by lazy {
   this@distinctBy.generate(RandomSource.seeded(1234L)).map { it.value }.take(minSampleSize).toList()
}
``` 

Summoning @sksamuel for guidance

sksamuel:
I think you can have the same elements, because the selector just ensures that the population is distinct, not that the next element you pick is distinct from the ones previously.

Like, listOf(1,1,1,2,2,2).distinctBy { it }.random() could keep emitting 1.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
sksamuel: I don't think this addresses the underlying concern, that this function can hang because you may never get enough random values.

One way around that is to take a parameter which is how many times you'll attempt to get the next element say `maxUniqueAttempts` or something and error if you exceed it.

myuwono: In terms of hanging - it should not really hang anymore but i'm not entirely convinced of what we're trying to solve here. This function looks more like a `Arb.resampleToUniformRandom()` at this point which is essentially sampling from an underlying distinct distribution. i.e. `Arb.of(1, 2, 3 ,4, 5, 6, 7, 8)`

sksamuel:
Right.  Not sure of the use of distinct anymore. Could just deprecate
distinct.

myuwono:
agreed - i'll put a deprecation to it

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
myuwono: I've marked this function as deprecated. The primary problem with `distinct` is that it is terminal and stateful. Hence in terms of actually implementing the behaviour that we are after, i.e. sampling without replacement, we need to rethink the return type. Arb<A> is not capable of modelling this, as it is a population, it is infinite, and stateless. We may need to think of a type that is terminal. `Exhaustive<A>` is backed by `List<A>` under the hood so that might be an option. However, as you've rightly mention, Exhaustive itself is probably not suitable either. There's nothing in its type that says generate(rs) will emit only unique elements without replacement and terminates when there's nothing left for it to sample from.

myuwono:
We're not considering this anymore, hence the fix is no longer necessary. Distinct will be phased out. A separate PR will be raised to deprecate distinct

PR #1730, opened 2020-09-26
deprecate distinct

myuwono:


magneticflux-:
What is the replacement for this? Is it not supported at all anymore?

I'm currently using this feature to automatically generate (unique/distinct-named) nested tests for properties, like this issue does for data: https://github.com/kotest/kotest/issues/1537

PR #1735, opened 2020-09-28
Update documentation to include JSON module

KeremAslan:
The documentation does not mention that the JSON matches are in the JSON module.


sksamuel:
:+1:

PR #1724, opened 2020-09-23
Fixed the empty allure result for tests with the failed afterTest block

SergKhram:
If we have the failed block afterTest and use the AllureListener - in the report we don't see anything about the failed test:
image (6)
image (7)

Because AllureTestReporter as the listener can't start his afterTest block (in listeners.forEach we have error in test's 'afterTest' -> other listeners can't start their 'afterTest' blocks)

PR #1736, opened 2020-09-28
Fix first failure in `beforeTest` blocks

SergKhram:
The same problem, but with `beforeTest`: https://github.com/kotest/kotest/pull/1724

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/lifecycle.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/lifecycle.kt
@@ -68,16 +68,15 @@ suspend fun TestCase.invokeAllBeforeTestCallbacks(): Try<TestCase> =
    }.fold({
       Try.Failure(it)
    }, { listeners ->
-      Try {
-         listeners.forEach {
+      listeners.map {
+         Try {
             if (type == TestType.Container) it.beforeContainer(this)
             if (type == TestType.Test) it.beforeEach(this)
             it.beforeAny(this)
             it.beforeTest(this)
+            this
          }
-
-         this
-      }
+      }.find { it.isFailure() } ?: Try { this }
    })
 
 /**
--- a/kotest-runner/kotest-runner-junit5/src/jvmTest/kotlin/com/sksamuel/kotest/runner/junit5/DiscoveryTest.kt
+++ b/kotest-runner/kotest-runner-junit5/src/jvmTest/kotlin/com/sksamuel/kotest/runner/junit5/DiscoveryTest.kt
@@ -37,7 +37,7 @@ class DiscoveryTest : FunSpec({
          .build()
       val engine = KotestJunitPlatformTestEngine()
       val descriptor = engine.discover(req, UniqueId.forEngine("testengine"))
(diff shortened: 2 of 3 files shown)


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/lifecycle.kt
```diff
@@ -69,11 +69,24 @@ suspend fun TestCase.invokeAllBeforeTestCallbacks(): Try<TestCase> =
       Try.Failure(it)
    }, { listeners ->
       Try {
+         var currentStatus: TestStatus? = null
```
ashishkujoy: Redundant variable this is not getting used.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/lifecycle.kt
```diff
@@ -69,11 +69,24 @@ suspend fun TestCase.invokeAllBeforeTestCallbacks(): Try<TestCase> =
...
+               if (currentStatus == null || currentStatus!! != TestStatus.Failure) {
+                  currentStatus = TestStatus.Failure
+                  currentException = e
+               }
+            }
+         }
+
+         currentException?.let { it ->
+            throw Error(it)
          }
```
ashishkujoy: I think a cleaner way to handle this might look something like this.
```kotlin
listeners.map {
         Try {
            if (type == TestType.Container) it.beforeContainer(this)
            if (type == TestType.Test) it.beforeEach(this)
            it.beforeAny(this)
            it.beforeTest(this)
            this
         }
      }.find { it.isFailure() } ?: Try { this }
```

1) Its avoiding putting try-catch block and using existing Try.
2) Its avoiding mutation.

SergKhram: @ashishkujoy  with this decision we have problems:
[ERROR] C:\projects\totest\io\kotest\core\spec\lifecycle.kt:[71,9] Type mismatch: inferred type is Try`<TestCase>`? but Try.Failure? was expected
[ERROR] C:\projects\kotest\io\kotest\core\spec\lifecycle.kt:[79,38] Type mismatch: inferred type is Try`<TestCase>` but Try.Failure was expected

ashishkujoy: Interesting i don't see that error in my local i made that change and pushed to my fork. Here is the github action link. All test passes, the only failure(expected failure) is for publishing the artifacts to repository.

https://github.com/ashishkujoy/kotest/runs/1180868149
https://github.com/ashishkujoy/kotest/commit/7f75d89742c06f016864a4400af486fcb28920c4

ashishkujoy:
@SergKhram Could you please add a test for this fix.

SergKhram:
> @SergKhram Could you please add a test for this fix.

Done, check it, pls

ashishkujoy (approved):

ashishkujoy:
@SergKhram Thanks for the contribution.

Issue #1545, opened 2020-07-05
Option to strip whitespace from test names


lgtout:
This feature idea arises from the limitations of StringSpec style.  And I don't think what's described below is already covered by any other existing styles.

The motivation is this:

The developer doesn't want to nest tests in a tree, because understanding such tests can require much visual scanning up and down through the tree of nested tests, paying very close attention to indentation, in order to figure out what the state of the SUT is when the test's action is finally performed.

However, the developer likes the readability of long test names (only the name) being broken into multiple nested segments in the IDE, each segment on a separate line immediately after the previous segment.  The same is desired for appearance in the IntelliJ plugin tool window, and in the run output window.

In the IDE it might boil down to something like this:

```
"Given blah blah blah blah blah blah" +
    "When blah blah blah" +
        "Then blah blah blah"  {
    TODO()
}
"Given blah blah blah blah blah blah" +
    "When blah blah blah" +
        "Then shake shake shake" {
    TODO()
}
```

Note how the second test name differs only in the third line/segment.
Also, note that my using the "Given/When/Then" words in the segments is irrelevant to the style.  The style would be insensitive to the actual content of each segment.

In the plugin/run tool window it might look like this:

```
Given blah blah blah blah blah blah 
    When blah blah blah
        Then blah blah blah
Given blah blah blah blah blah blah 
    When blah blah blah
        Then shake shake shake
```

Or, like this:

```
Given blah blah blah blah blah blah 
    When blah blah blah
        Then blah blah blah
        Then shake shake shake
```

Even with the latter/second display structure, understanding the state of the SUT at any point within it is much easier than when reading the actual test code, despite the nesting, because one doesn't have the problem of a large spread-out tree with nested code separating name segments.

I know little about how Kotest is implemented, but I imagine this style would require it to  understand that the unique name of a test is a combination of all its name "segment"s.  It should not conclude, for example, that two names map to the same test because the first segment of both are the same.

sksamuel:
The problem with this is that the inner tests do not exist until the lambdas execute at runtime, so there's no way to know the final layout until all tests have finished.

If this was added as an option, the framework would need to execute all tests in a spec before showing any output. Whether that is important or not depends on whether you care about seeing the tests appear in intellij/gradle as they are executing, or whether you're only really interested in the final output.

sksamuel:
[...]

PR #1745, opened 2020-10-02
#1545 Remove whitespace from test display name based on system props

ashishkujoy:
This PR adds functionality to strip extra whitespace from test name when system property `kotest.framework.testname.multiline` is set as true.

As a result we be able to get visual nesting in stringspec.
```kotlin
class MyTest : StringSpec({
    """Given: given statement
                 When: when statement
                          Then: then statement""" {
                                 // assertions
                          }
    
})
```

**Note:** Test report will have name as:  `Given: given statement When: when statement Then: then statement`


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/Description.kt
```diff
@@ -101,7 +103,13 @@ sealed class Description {
    /**
     * Returns the name of this description formatted for display.
     */
-   fun displayName() = name.displayName
+   fun displayName(): String {
+      return if (sysprop(KotestEngineSystemProperties.allowMultilineTestName, "FALSE").toUpperCase() == "TRUE") {
+         val spaceCharacter = " "
+         val emptyCharacter = ""
+         name.displayName.split(spaceCharacter).filterNot { it == emptyCharacter }.joinToString(spaceCharacter)
```
sksamuel: What about adding a test that this works on \n only. So

"""this
is
my
test"""

ashishkujoy: @sksamuel 
here is no `\n` in this multiline string. 
So something of below screenshot wont work. It does not have \n or " " character so it will appear as `3thirdtestlastline`

Screenshot 2020-10-02 at 10 38 58 PM

However this will work find as it has " " character in it. It will appear as `3 third test last line` 
Screenshot 2020-10-02 at 10 42 21 PM

Which one of the above two case are you referring?

sksamuel: I'm referring to the top screenshot. That should still display as "3 third test last line". Instead of splitting on " " you should split on \s

ashishkujoy: Got it thanks @sksamuel

ashishkujoy: Done

sksamuel (approved):

PR #1752, opened 2020-10-04
propagated coroutine context from extensions to tests #1725

### LINKED ISSUE #1725: Coroutine context from extensions is not propagated to tests
4.2.5

I am trying to use a custom extension or `aroundTest` call that surrounds the test execution with a `withContext(...)`, augmenting the coroutine context.

Expected: the added coroutine context is accessible from the tests and other setup logic that is invoked as `suspend`

Actual: the coroutine context is not propagated to the test, somewhere on the call stack it gets replaced with a different context

It seems that `ExecutorExecutionContext` is the place where this replacement happens. It seems to me that the call to `runBlocking`:

https://github.com/kotest/kotest/blob/24769598e4158afbbcbd57a28443cf8ff94cdf36/kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/ExecutorExecutionContext.kt#L44-L46

should actually pass the `CoroutineContext` from the `Continuation`:

```kotlin
runBlocking(cont.context.minusKey(ContinuationInterceptor)) {
    ...
```

It seems prudent to remove the `ContinuationInterceptor` (i.e. the `CoroutineDispatcher`) from the context, because the documentation for `runBlocking` states that otherwise it would just continue the coroutine on the given dispatcher and block the thread until it's done, but the purpose of this class seems to be that we want to interrupt the thread if a test ends up in an endless loop.

**Example Code:**

```kotlin
class FooContext(val value: Any) : AbstractCoroutineContextElement(Key) {
    companion object Key : CoroutineContext.Key<FooContext>
}

/** Gets the contextual foo value from a test */
suspend fun foo(): Any? = coroutineContext[FooContext]?.value

class FooContextSpec : FreeSpec({
    aroundTest { (testCase, execute) -> 
        val fooValue = 42
        withContext(FooContext(fooValue)) {
            execute(testCase)
        }
    }
    
    "should have contextual value" - {
        foo().shouldNotBeNull()
    }
})
```

**Background:**

I often create extensions that produce contextual objects around a scope (spec / container / test), which then need to be accessed in individual tests or before/after callbacks.

For example, consider an extension that launches and tears down an embedded database around some test scope. The tests need some handle to that DB, which might be a randomly allocated TCP port or some object from the DB's client library.

In simple cases, such scenarios can be addressed similar to kotest's `tempfile()`: a simple function that is called directly from the spec, which does the setup immediately and registers an `afterSpec` callback to do the cleanup. However, doing this in an extension and making the value available via some sort of context would improve composability and reduce boilerplate.

Many test frameworks have a facility for these contextual values. In JUnit 5, there is the `ExtensionStore` which can be used to resolve test method parameters. In Spek, there is `memoized`. Since kotest is using coroutines throughout, I thought that a custom coroutine context element would be the intended way to pass these contextual values.

sksamuel:
Closes #1725

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/TestCaseExecutor.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/TestCaseExecutor.kt
@@ -78,15 +78,23 @@ class TestCaseExecutor(
       start: Long,
       extensions: List<TestCaseExtension>,
    ): TestResult {
-      return when {
-         extensions.isEmpty() -> executeIfActive(testCase) { executeActiveTest(testCase, context, start) }
-         else -> extensions.first().intercept(testCase) {
-            // the user's intercept method is free to change the context of the coroutine
-            // to support this, we should switch the context used by the test case context
-            val newContext = context.withCoroutineContext(coroutineContext)
-            intercept(it, newContext, start, extensions.drop(1))
+
+      val innerExecute: suspend (TestCase, TestContext) -> TestResult = { tc, ctx ->
+         executeIfActive(tc) { executeActiveTest(tc, ctx, start) }
+      }
+
+      val execute = extensions.foldRight(innerExecute) { extension, execute ->
+         { testCase, context ->
+            extension.intercept(testCase) {
+               // the user's intercept method is free to change the context of the coroutine
+               // to support this, we should switch the context used by the test case context
+               val newContext = context.withCoroutineContext(coroutineContext)
+               execute(it, newContext)
+            }
          }
       }
+
+      return execute(testCase, context)
    }
 
    /**
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/TestCaseExecutor.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/TestCaseExecutor.kt
@@ -16,6 +16,7 @@ import io.kotest.core.test.TestStatus
 import io.kotest.core.test.TestType
 import io.kotest.core.test.resolvedInvocationTimeout
 import io.kotest.core.test.resolvedTimeout
+import io.kotest.core.test.withCoroutineContext
 import io.kotest.fp.Try
 import io.kotest.mpp.log
 import io.kotest.mpp.replay
@@ -24,6 +25,7 @@ import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.withTimeout
 import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.coroutineContext
 
 data class TimeoutException constructor(val duration: Long) : Exception("Test did not complete within ${duration}ms")
 
@@ -76,9 +78,15 @@ class TestCaseExecutor(
       start: Long,
       extensions: List<TestCaseExtension>,
    ): TestResult {
+      println("Intercept context $context")
       return when {
          extensions.isEmpty() -> executeIfActive(testCase) { executeActiveTest(testCase, context, start) }
-         else -> extensions.first().intercept(testCase) { intercept(it, context, start, extensions.drop(1)) }
+         else -> extensions.first().intercept(testCase) {
+            // the user's intercept method is free to change the context of the coroutine
+            // to support this, we should switch the context used by the test case context
+            val newContext = context.withCoroutineContext(coroutineContext)
+            intercept(it, newContext, start, extensions.drop(1))
+         }
       }
    }
 
@@ -122,7 +130,8 @@ class TestCaseExecutor(
       start: Long,
    ): TestResult {
 
-      log("Executing active test $testCase")
+      println("Executing active test with context $context")
+      log("Executing active test $testCase with context $context")
       listener.testStarted(testCase)
 
       return testCase
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/TestCaseExecutor.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/TestCaseExecutor.kt
@@ -78,7 +78,6 @@ class TestCaseExecutor(
       start: Long,
       extensions: List<TestCaseExtension>,
    ): TestResult {
-      println("Intercept context $context")
       return when {
          extensions.isEmpty() -> executeIfActive(testCase) { executeActiveTest(testCase, context, start) }
          else -> extensions.first().intercept(testCase) {
@@ -130,7 +129,6 @@ class TestCaseExecutor(
       start: Long,
    ): TestResult {
 
-      println("Executing active test with context $context")
       log("Executing active test $testCase with context $context")
       listener.testStarted(testCase)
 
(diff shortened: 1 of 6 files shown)


sksamuel:
@tkrullmann

tkrullmann:
Nice work, thanks! I'll have a look at it.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/TestCaseExecutor.kt
```diff
@@ -78,7 +80,12 @@ class TestCaseExecutor(
    ): TestResult {
       return when {
          extensions.isEmpty() -> executeIfActive(testCase) { executeActiveTest(testCase, context, start) }
-         else -> extensions.first().intercept(testCase) { intercept(it, context, start, extensions.drop(1)) }
+         else -> extensions.first().intercept(testCase) {
```
tkrullmann: This looks a bit quirky, each lambda created here would always store the entire list of extensions that follow it, even though it only calls the next one. This results in quadratic memory usage over the number of extensions (`n * (n+1) / 2` elements), but could be a lot if there are a lot of extensions, and it is done for each test.

I know it was there before, so it doesn't have anything to do with this change, feel free to ignore it :-)

Consider rewriting the method using a `foldRight` instead of recursing inside the lambda:

```kotlin
      val initialExecute: suspend (TestCase, TestContext) -> TestResult = { testCase, context ->
         executeIfActive(testCase) { executeActiveTest(testCase, context, start) }
      }
      val execute = extensions.foldRight(initialExecute) { extension, execute ->
            { testCase, context ->
               extension.intercept(testCase) {
                  // the user's intercept method is free to change the context of the coroutine
                  // to support this, we should switch the context used by the test case context
                  val newContext = context.withCoroutineContext(coroutineContext)
                  execute(it, newContext)
               }
            }
      }
      return execute(testCase, context)
```

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/TestCaseExecutor.kt
```diff
@@ -78,7 +80,12 @@ class TestCaseExecutor(
...
+            // the user's intercept method is free to change the context of the coroutine
+            // to support this, we should switch the context used by the test case context
+            val newContext = context.withCoroutineContext(coroutineContext)
```
tkrullmann: Maybe I'm missing something, but I don't really see why handing down the augmented `coroutineContext` is necessary... from my understanding the fact that `TestContext` extends `CoroutineScope` is just for convenience, so you can `launch` etc. directly in the test without creating a new scope.

The extensions don't receive a `TestContext` parameter, so they can't access or modify the `TestContext.coroutineContext`. And `TestContext.coroutineContext` gets replaced anyway by the current coroutine scope's context just before the actual test is called (which seems to happen in `TestCaseExecutor.executeAndWait`).

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/TestContext.kt
```diff
@@ -44,3 +45,14 @@ interface TestContext : CoroutineScope {
     */
    suspend fun registerTestCase(nested: NestedTest)
 }
+
+/**
+ * Returns a new [TestContext] which uses the given [coroutineContext] with the other methods
+ * delegating to the receiver context.
+ */
+internal fun TestContext.withCoroutineContext(coroutineContext: CoroutineContext) = object : TestContext {
```
tkrullmann: For a large number of extensions the delegation chain will be quite long, consider a small optimization here:

```kotlin
private class TestContextWithCoroutineContext(
   val delegate: TestContext,
   override val coroutineContext: CoroutineContext
) : TestContext by delegate {
   override fun toString(): String = "TestCaseContext [$coroutineContext]"
}

/**
 * Returns a new [TestContext] which uses the given [coroutineContext] with the other methods
 * delegating to the receiver context.
 */
internal fun TestContext.withCoroutineContext(coroutineContext: CoroutineContext): TestContext =
   if (this is TestContextWithCoroutineContext) {
      TestContextWithCoroutineContext(delegate, coroutineContext)
   } else {
      TestContextWithCoroutineContext(this, coroutineContext)
   }
```

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/ExecutorExecutionContext.kt
```diff
@@ -42,7 +44,10 @@ object ExecutorExecutionContext : TimeoutExecutionContext {
          scheduler.shutdown()
 
          try {
-            runBlocking {
+            // we use the context from the caller, in order to allow context params to propogate down
+            // into the test case from test extensions
+            // see https://github.com/kotest/kotest/issues/1725
+            runBlocking(context) {
```
tkrullmann: As mentioned in the original issue description, we should remove the `ContinuationInterceptor` context key here, which contains the `CoroutineDispatcher` element.

```kotlin
runBlocking(context.minusKey(ContinuationInterceptor)) {
   // ...
}
```

According to the documentation of `runBlocking` if we give it a context that includes a `CoroutineDispatcher`, the coroutine will continue to run on the given dispatcher, and the `runBlocking` will just wait for it to finish. Since we want to be able to interrupt this thread, we would just interrupt the waiting but not the actual execution of the test.

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/runners/SingleInstanceSpecRunner.kt
```diff
@@ -23,6 +23,7 @@ import io.kotest.mpp.log
 import kotlinx.coroutines.coroutineScope
 import java.util.concurrent.ConcurrentHashMap
 import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.coroutineContext
```
tkrullmann: seems to be unused

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/ExecutorExecutionContext.kt
sksamuel: Thanks, that's what I was missing.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/TestCaseExecutor.kt
sksamuel: I guess it was because I wanted the TestContext to be "correct", in that it represented the currently used coroutine context, rather than the one it was started with, which has perhaps been overwritten by the user (using withContext). 

If TestContext isn't exposed to the user until executeAndWait then I guess it doesn't matter, and it can rely on the final switch just before the test is executed. 

But if we add some kind of further extensions later that expose TestContext, them it would have an outdated coroutine context reference.

I think I'm right on the above anyway.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/TestCaseExecutor.kt
sksamuel: This is much nicer.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/TestContext.kt
sksamuel: Done.

tkrullmann (approved):

PR #1756, opened 2020-10-05
Wrote the test where we override severity for testCase and check allure for it

SergKhram:


SergKhram:
@sksamuel @ashishkujoy check, pls

PR #1757, opened 2020-10-05
Restored Robolectric extension build and refactored build.gradle.kts

alexfacciorusso:
It seems the Robolectric extension was excluded from the project build. I've added it again and corrected some dependencies.


alexfacciorusso:
I would consider the work on moving the Robolectric extension and updating versions etc to the main repo done.

I’ve tried to create an Android submodule in the tests module in order to have a proof that the extension works, but it is not passing giving me runtime errors regarding some internal Robolectric classes not found.

Although, the extension seems to be perfectly working if I use an external project linked with the library produced by kotest, so I think it’s mostly a problem regarding the definitely weird nature of adding an android project as a sub-sub-module of a multiplatform project

I was thinking that we might temporarily remove the android project from the repo so the test is not failing anymore, and we then could consider the PR mergeable

kotest-extensions/kotest-extensions-robolectric/src/jvmMain/kotlin/io/kotest/extensions/robolectric/ContainedRobolectricRunner.kt
```diff
@@ -1,10 +1,10 @@
-package io.kotest.experimental.robolectric
+package io.kotest.extensions.robolectric
 
 import org.junit.runners.model.FrameworkMethod
 import org.robolectric.RobolectricTestRunner
 import org.robolectric.internal.bytecode.InstrumentationConfiguration
 
-class ContainedRobolectricRunner : RobolectricTestRunner(PlaceholderTest::class.java) {
+public class ContainedRobolectricRunner : RobolectricTestRunner(PlaceholderTest::class.java) {
```
sksamuel: Why add public here ?

alexfacciorusso: Committed that by mistake, I'll remove

PR #1758, opened 2020-10-05
Added retry to null probability test

alexfacciorusso:
the `null probability values can be specified` is non-deterministic, and sometimes it might fail. Added a `retry` in case that happens.


sksamuel:
lgtm will merge once green

PR #1759, opened 2020-10-06
Make line break error message compatible with IntelliJ diff

ajalt:
This commit slightly changes the error message shown when a string differs from its expected value only based on line endings. The message now works with IntelliJ's visual diff, but still shows the line endings escaped in logs.


sksamuel:
:+1:

PR #1761, opened 2020-10-08
If we have the failed test on init - we don't see it in the allure report

SergKhram:


SergKhram:
@sksamuel How we can resolve this problem ?
Circular dependency between the following tasks:
:kotest-extensions:kotest-extensions-allure:compileKotlinJvm
+--- :kotest-extensions:kotest-extensions-allure:jvmJar
|    +--- :kotest-extensions:kotest-extensions-allure:compileKotlinJvm (*)
|    \--- :kotest-extensions:kotest-extensions-allure:jvmMainClasses
|         \--- :kotest-extensions:kotest-extensions-allure:compileKotlinJvm (*)
\--- :kotest-framework:kotest-framework-engine:jvmJar
     +--- :kotest-framework:kotest-framework-engine:compileKotlinJvm
     |    \--- :kotest-extensions:kotest-extensions-allure:jvmJar (*)
     \--- :kotest-framework:kotest-framework-engine:jvmMainClasses
          \--- :kotest-framework:kotest-framework-engine:compileKotlinJvm (*)

SergKhram:
I will rewrite this code with another way ( don't check this decision now )

SergKhram:
@sksamuel check it pls

sksamuel:
Hi,

I don't think we should add new listener methods in the `TestListener` interface without careful consideration.

There is `notifySpecFinished` in `FinalizeSpecListener` which will carry a throwable if there was an error (including in instantiation) which might be suitable for what you need.

If you want to be specifically notified that a spec failed to be created, then we could consider adding a `SpecInstantiationListener` which has methods for failure and success and is invoked as part of the createAndInitializeSpec function.

And if we do want to add that, it should be a standalone interface and not part of the TestListener super interface.

SergKhram:
@sksamuel i see only finalizeSpec in FinalizeSpecListener and it doesn't have Throwable parameter. 
And we don't go to the 'invokeFinalizeSpecListeners'(where we use finalizeSpec  from FinalizeSpecListener ) , because we failed on createInstance in this tree
      notifySpecStarted(kclass)
         .flatMap { invokePrepareSpecListeners(kclass) }
         .flatMap { createInstance(kclass) }
         .flatMap { runTestsIfAtLeastOneActive(it) }
         .flatMap { checkClosedTestCases(it) }
         .flatMap { invokeFinalizeSpecListeners(kclass, it) }
         .fold(
            { notifySpecFinished(kclass, it, emptyMap()) },
            { notifySpecFinished(kclass, null, it) }
         )


Or you want i need to add my new fun for failure (afterSpecInitFailure) to FinalizeSpecListener and add my executeAllAfterSpecInitFailure to notifySpecFinished for case w throwable ?

sksamuel:
That whole code block should be re-written as invokeFinalizeSpecListeners
should always be invoked.
Perhaps we should add a throwable parameter to the method in
FinalizeSpecListener.

SergKhram:
@sksamuel  About  what block did you speak ? Do you want to replace notifySpecFinished with invokeFinalizeSpecListeners ? Or what

sksamuel:
The code block you pasted.

Let's create a new listener.

```
interface SpecInstantiationListener {
   fun specInstantiated(spec: Spec) {}
   fun specInstantiationError(kclass: KClass<out Spec>, t: Throwable) {}
}
```

You could invoke this in the spec runner around line 40s.

sksamuel:
Can you make a separate PR to add the listener support.
And then update this PR to use it. It's easier to work out what each commit has changed later this way.

SergKhram:
> Can you make a separate PR to add the listener support.
> And then update this PR to use it. It's easier to work out what each commit has changed later this way.

@sksamuel done, check it pls

sksamuel (approved):

sksamuel:
:+1:

PR #1767, opened 2020-10-13
Created the new listener for spec instantiation

SergKhram:


SergKhram:
@sksamuel check it pls

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/listeners/TestListener.kt
```diff
@@ -20,6 +20,7 @@ interface TestListener :
    AfterInvocationListener,
    PrepareSpecListener,
    FinalizeSpecListener,
+   SpecInstantiationListener,
```
sksamuel: Don't include it here, I want to define all new listeners separately.

sksamuel (commented):
We will need to invoke this in the engine as part of this PR.

SergKhram:
Check it pls > We will need to invoke this in the engine as part of this PR.

SergKhram:
@sksamuel check it pls

SergKhram:
@sksamuel done > We will need to invoke this in the engine as part of this PR.

sksamuel (approved):

PR #1774, opened 2020-10-15
Added some testcontainers modules(Kafka, Selenium, ElasticSearch)

SergKhram:


sksamuel:
Why would you add these to the main build. This would mean every consumer of the extension now gets these modules, whether they want them or not.

SergKhram:
> Why would you add these to the main build. This would mean every consumer of the extension now gets these modules, whether they want them or not.

Ok

Issue #1773, opened 2020-10-14
forAll function for more than seven params in a row


driver733:
As of now (v. 4.3.0) it is impossible to use the [forAll](https://github.com/kotest/kotest/blob/0a271b4c9479ddf6c1e94623cb2459f8e97e26c1/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/data/forAll7.kt#L6) function (in data driven tests) with more than seven params in one `row`. The [docs](https://kotest.io/data_driven_testing/) say that a new data class should be created for such scenarios, however, it's very inconvenient. It would be great if it was possible to use the `forAll` function with more params in one `row`. I think that it should be increased to 22, in accordance with the maximum number of fields a `row` can be created with. (class [Row22](https://github.com/kotest/kotest/blob/624a36a1936e370ae720cca2a7878af47f27604b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/data/rows.kt#L110))

ashishkujoy:
I think 22 is big number and i doubt will people really be needing that.

driver733:
@ashishkujoy From my personal experience, sometimes I tend to go above 15.  I agree that 20 is big, but it would be better to have some extra space.  What do you think?

ashishkujoy:
I think that many of the user will not need that support and atleast I have never seen people going beyond 5-6 arguments .

And since Data class support can help user to achieve that if they need it, I don't see much value in adding more variant of `forAll`.

@sksamuel What's your thought on this.

sksamuel:
What about expanding out our Tuple values.
They are data classes, so would work with the new data testing, but also
absolves the user from having to create the data classes themselves.

Effectively,

forAll(
  row(1,2,3)
)

can become

forAll(
  Tuple3(1,2,3)
  Tuple3(1,2,3)
)

And you get all the nice new benefits.

ashishkujoy:
That seems a better alternative to me.

driver733:
@sksamuel Just to clarify: with your solution _all_ of the `Tuple` classes will work with the `forAll` function, right?

For instance, will it be possible to use the `Tuple22` class the following way?

```
forAll(
  Tuple22("a", ... , "x")
)
```

sksamuel:
Yes, because they're just data classes.

ashishkujoy:
Added Variants of Tuples upto Tuple22

driver733:
[...]

PR #1777, opened 2020-10-17
#1773 Adds variant of tuples upto 22

ashishkujoy:
`forAll` support upto 7 arguments which makes it difficult for user to use it for more than 7 properties until they wrap them in data class. Tuples being data class will be handy for these use cases.


LeoColman:
That must have been tiring xD

LeoColman (approved):

ashishkujoy:
> That must have been tiring xD

Actually not. Just wrote some code to generate that.

PR #1778, opened 2020-10-17
Added annotation and sysprop for filtering spec class instantiation

SergKhram:
For filtering we need to add the annotation SpecClassExecutionFilter w value -> and use 'kotest.framework.spec.class.filter' for list. example: [desc,something]


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/filter/Filter.kt
```diff
@@ -1,3 +1,6 @@
 package io.kotest.core.filter
 
+
 interface Filter
+
+annotation class SpecClassExecutionFilter(val filter: String)
```
SergKhram, line 6: Or we can use not the String - something from new Interface (and the users can implement this for their needs as they want)

Issue #1763, opened 2020-10-12
Provide for more powerful test filter


sksamuel:
Existing TestFilter on launcher only filters by description.
Allow for filtering on TestCase instance, or Spec class, or Spec instance..

sksamuel:
Not sure why I created this now. Closing.

PR #1779, opened 2020-10-17
Added the annotation and sysprop for filtering spec class instantiation

SergKhram:
For filtering we need to add the annotation SpecClassExecutionFilter w value -> and use 'kotest.framework.spec.class.filter' with list. example: some,something

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/filter/SpecClassFilter.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/filter/SpecClassFilter.kt
@@ -1,0 +1,7 @@
+package io.kotest.core.filter
+
+interface SpecClassFilter: Filter {
+
+}
+
+annotation class SpecClassExecutionFilter(val filter: String)
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/KotestEngineSystemProperties.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/KotestEngineSystemProperties.kt
@@ -61,4 +61,6 @@ object KotestEngineSystemProperties {
    *  If set -> filter testCases by this severity level and higher, else running all
    */
    const val severityPrefix = "kotest.framework.test.severity"
+
+   const val specClassExecutionFilter = "kotest.framework.spec.class.filter"
 }
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/active.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/active.kt
@@ -1,6 +1,7 @@
 package io.kotest.core.internal
 
 import io.kotest.core.config.configuration
+import io.kotest.core.filter.SpecClassExecutionFilter
 import io.kotest.core.filter.TestFilter
 import io.kotest.core.filter.TestFilterResult
 import io.kotest.core.test.TestCase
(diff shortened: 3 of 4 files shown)


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/filter/SpecClassFilter.kt
```diff
@@ -0,0 +1,7 @@
+package io.kotest.core.filter
+
+interface SpecClassFilter: Filter {
+
+}
+
+annotation class SpecClassExecutionFilter(val filter: String)
```
SergKhram, line 7: Or we can use not the String - something from new Interface (and the users can implement this for their needs as they want)

SergKhram:
@sksamuel check it pls

SergKhram:
@sksamuel @ashishkujoy check it pls

sksamuel:
I will review it this week.

SergKhram:
@sksamuel ping

sksamuel:
Can you explain what this will be used for @SergKhram

SergKhram:
@sksamuel it may be about https://github.com/kotest/kotest/issues/1763 (partly) - for spec instantiation filtering : for example we want to run tests with before test blocks on some TeamCity build. We have 3-4 with different configuration ( with mock and without ) and we don't want to execute tests where we use mock in build without mock - we can filter it and don't make steps from before test , because we filtered it before test class instantiated.

sksamuel:
Isn't this what tags are for?

SergKhram:
Ok , but if we make the table with body generation , and in current build generation  body ( in table ) failed , because it can't generate and mustn't-> we need to have some filter , that affect on need we instantiate class or not

For example we need to generate bodies for some api request , each case the different body -> in test we named just send this body in request . But in generation this body we use some things , that must work in one build and mustn't work in another one . We add tag to filter case execution , but we failed on generation , because table instantiate in class instantiation. We need to filter it

SergKhram:
@sksamuel 

Ok , but if we make the table with body generation , and in current build generation  body ( in table ) failed , because it can't generate and mustn't-> we need to have some filter , that affect on need we instantiate class or not

For example we need to generate bodies for some api request , each case the different body -> in test we need just send this body in request . But in generation this body we use some things , that must work in one build and mustn't work in another one . We add tag to filter case execution , but we failed on generation , because table instantiate in class instantiation. We need to filter it

sksamuel:
Can you show me a small complete example.

SergKhram:
@sksamuel 
**Test:** 
```kotlin
class SomeSpec: FreeSpec() {
...
val data = table(    //will fail here, on initialization table in AUTONOMOUS build , where doesn't deploy some things (see below)
        headers("body", "testName"),
        row(generateEntRequestBody(), "Ent"), 
        row(generateOrgRequestBody(), "Org"),
        row(generateSubRequestBody(), "Sub")
    )
init {
       forAll(table) { body, testName ->
         "Get $testName customer info".config(tags = setOf(INTEGRATION)) {  //we want to filter execution by tag (configure it in the each 
 TeamCity build
              val customerId = createCustomer(body)
              ...
         }
      }
    }
}
```

```kotlin
**in other kt file (for example prepare.kt)**
fun generateEntRequestBody(): String {
     val model = SomeModel().apply{
     someField = getSomeDictionaryValueFromForeignDBByName("Name")   //but this db doesn't  and mustn't deploy in this build
     ....
     }
     ...
}
```

SergKhram:
@sksamuel ping

SergKhram:
@sksamuel ping

sksamuel:
You could add `@Tags(INTEGRATION)` to the SomeSpec class and then it won't be instantiated.

SergKhram:
@sksamuel  how does it work? 
I use -Dkotest.tags.include=MOCK
And expected that test with @Tags(INTEG) will not execute
But it executed

In debug i see:

because ```tags.map { it.name }.contains(ident)``` return false -> ```if (tags.map { it.name }.contains(ident)) true else null``` return null -> it will be active in the future , because ```isPotentiallyActive(tags.toSet()) ?: true```. As the result we will instantiate class


and as the result it instantiated the class and execute step in it

SergKhram:
@sksamuel fixed in https://github.com/kotest/kotest/pull/1823

PR #1783, opened 2020-10-19
Feature/assertions livedata

alexfacciorusso:
Created assertions and extension for LiveData testing.

## What has been added?
- Matchers in `kotest-matchers-livedata`
- Listener in `kotest-extensions-livedata`

## TODO: 
- [x] Fix the publish script to allow for Android libraries
- [ ] ~Documentation~ (can do in another PR)
- [x] More matchers
- [x] API for observables-like testing (to be used in future for Flows etc)


kotest-assertions/kotest-assertions-livedata/build.gradle.kts
```diff
@@ -0,0 +1,86 @@
...
+      showStandardStreams = true
+      events = setOf(
+         TestLogEvent.FAILED,
+         TestLogEvent.PASSED
+      )
+      exceptionFormat = TestExceptionFormat.FULL
+   }
+}
+
+//apply(from = "../../publish-mpp.gradle.kts")
```
alexfacciorusso: I need to edit the Publish script since it doesn't support android libraries for now (to add an if to it to add conditionally the java/java-library plugins since they're not supported by the `com.android.library` one).

kotest-assertions/kotest-assertions-livedata/src/androidMain/kotlin/io/kotest/assertions/livedata/getOrAwaitValue.kt
```diff
@@ -0,0 +1,39 @@
+package io.kotest.assertions.livedata
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.Observer
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+
+internal fun <T> LiveData<T>.getOrAwaitValue(
```
alexfacciorusso: Taken from Google examples; I was thinking of instead borrowing the TestObserver from https://github.com/jraska/livedata-testing/blob/master/testing/src/main/java/com/jraska/livedata/TestObserver.java and rewriting it in Kotlin since it's more complete and will allow for history-based assertions.

LeoColman:
I wonder if it would be better to call it `kotest-assertions-android` and just bundle everything in there

alexfacciorusso:
> I wonder if it would be better to call it `kotest-assertions-android` and just bundle everything in there

I get what you mean, but what if, as a user, I want to use the extension/listener but I am not interested in the matchers?

alexfacciorusso:
@sksamuel @LeoColman The PR's code is done, I just need to write the documentation. What do you think of the concept of `Spectator`? We'll use that also for the Kotlin `Flow` testing assertions I can write next, so we have a common interface for testing that kind of things

alexfacciorusso:
Any news on this? I'd love to have confirmation so I can try fixing the problem preventing this to build (it seems related to the javadoc task not being present in Android library projects) and then I can focus on the Kotlin Flow extension!
@sksamuel

LeoColman:
I don't think I like `Observer` and `Spectator` to "leak" into Kotest User's code, they should be private to our project

alexfacciorusso:
@LeoColman Spectator is a Kotest concept/protocol. It abstracts observers for LiveData, Flows etc :) I've created it in order to make an abstraction for testing those observable-like classes

alexfacciorusso:
It basically acts as a bridge between external classes and Kotest code, and it serves exactly the function to not make external code leak in the Kotest code, so that users will always have a `spectate` function extending the class to test and use that instead of using directly observers etc to test that kind of classes (as I said, LiveData, Kotlin Flow, Rx, etc)

sksamuel:
@alexfacciorusso I know you're still waiting on my review. I have some questions so we can discuss on slack.

PR #1784, opened 2020-10-19
Correct example code and output for asClue

mindhaq:
The example would not compile as is, and the output is different as well, switching expected and actual.


sksamuel:
Some really bad typos in the original! Thanks for the fix.

PR #1787, opened 2020-10-19
Performance improvements for Exceptions on the JVM

U1F984:
For our usecase we have lots of exceptions being created by `shouldBe` and similar. We noticed that a lot of time was being spent in class loading. Since it is unlikely the availability of the JUnit 4/5 Assertion classes will change during runtime we can cache the class and constructor lookup. 

Our test suite now runs in 45 seconds instead of 85, which I would consider a worthwhile improvement.


sksamuel:
Nice improvement. Will merge once green.

PR #1793, opened 2020-10-20
Implement inline listeners and extensions in project config

### LINKED ISSUE #1764: Allow for inline registration of listeners and extensions in project config

LeoColman:
Closes #1764


sksamuel:
The Pr is fine, but I've had this thought for a while that listeners/extensions should be registered in the same way (instead of two functions). I've not yet come up with a single name for the function that works, but something like `install`, `register`, `callback`, or something?

LeoColman:
Perhaps. 
Do we want to do this now? And should we deprecate the other functions if we do

sksamuel:
The reason I brought it up is because I don't think we should add 2 new methods if we plan to deprecate them very shortly after.

LeoColman:
This is now conflicting. I'll close this until we decide on what we discussed above.

sksamuel:
We can resurrect this. I'm happy to merge it.

sksamuel:
If we can just fix the conflicts we're good.

sksamuel:
We can close this or fix the conflicts.

PR #1798, opened 2020-10-24
Updated data driven tests to use stable identifiers #1795

### LINKED ISSUE #1795: Unexpected behavior when using new data-driven tests with non-data classes
Kotest 4.3.0

Given the following test:
```kotlin
package com.example

import io.kotest.core.datatest.forAll
import io.kotest.core.spec.IsolationMode
import io.kotest.matchers.ints.shouldBeLessThanOrEqual

data class Working(val data: Int)

class NotWorking(val data: Int)

class TestSpec : DescribeSpec({
    describe("DataSpec") {
        describe("working") {
            forAll(
                Working(1),
                Working(2),
            ) { d ->
                d.data shouldBeLessThanOrEqual 1
            }
        }
        describe("not-working") {
            forAll(
                NotWorking(1),
                NotWorking(2),
            ) { d ->
                d.data shouldBeLessThanOrEqual 1
            }
        }
    }
}) {

    override fun isolationMode() = IsolationMode.InstancePerLeaf
}
```

It behaves unexpectedly in the not-working block: it only runs the first test case, and completely ignores the following ones:
image

While I understand this isn't how to use the data-driven tests (we should use data classes), I think Kotest shouldn't fail silently in that case. The failure is less obvious when using `data class TestData(val someValue: Int, val someNonDataClass: Foo)`, where mere having a non-data class as part of the test data triggers the error as well.

I'm also curious what specifically causes this behavior? Is `toString()` for a given data row required to be stable and return the same string? I was playing with returning incrementing integer values in `toString()` and noticed that registering each row case calls `toString()` ~3 times.

sksamuel:
Closes #1795

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/names.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/names.kt
@@ -5,19 +5,27 @@ import io.kotest.mpp.reflection
 import kotlin.reflect.KClass
 import kotlin.reflect.KType
 
-/**
- * Each test name must be unique. We can use the toString if we determine the instance is stable.
- *
- * An instance is considered stable if it is a data class where each parameter is either a data class itself,
- * or a primitive type.
- *
- * Note: If the user has overridden toString() and the returned value is not stable, tests may not appear.
- */
-internal fun stableIdentifier(t: Any, counter: Int): String {
-   return if (isStable(t::class)) {
-      t.toString()
-   } else {
-      t::class.bestName() + " $counter"
+internal class Identifiers() {
+
+   private var names = mutableListOf<String>()
+
+   /**
+    * Each test name must be unique. We can use the toString if we determine the instance is stable.
+    *
+    * An instance is considered stable if it is a data class where each parameter is either a data class itself,
+    * or one of the [primitiveTypes].
+    *
+    * Note: If the user has overridden toString() and the returned value is not stable, tests may not appear.
+    */
+   fun stableIdentifier(t: Any): String {
+      val name = if (isStable(t::class)) {
+         t.toString()
+      } else {
+         t::class.bestName()
+      }
+      val count = names.count { it == name }
+      names.add(name)
+      return if (count == 0) name else "$name ($count)"
    }
 }
 
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/names.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/names.kt
@@ -1,0 +1,44 @@
+package io.kotest.core.datatest
+
+import io.kotest.mpp.bestName
+import io.kotest.mpp.reflection
+import kotlin.reflect.KClass
+import kotlin.reflect.KType
+
+/**
+ * Each test name must be unique. We can use the toString if we determine the instance is stable.
+ *
+ * An instance is considered stable if it is a data class where each parameter is either a data class itself,
+ * or a primitive type.
+ *
+ * Note: If the user has overridden toString() and the returned value is not stable, tests may not appear.
+ */
+internal fun stableIdentifier(t: Any, counter: Int): String {
+   return if (isStable(t::class)) {
+      t.toString()
+   } else {
+      t::class.bestName() + " $counter"
+   }
+}
+
+internal fun isStable(type: KType) = when (val classifier = type.classifier) {
+   is KClass<*> -> isStable(classifier)
+   else -> false
+}
+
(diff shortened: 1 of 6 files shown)


simonenkoi (approved):
LGTM, thank you for your quick response

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/names.kt
```diff
@@ -0,0 +1,44 @@
...
+internal fun isStable(type: KClass<*>): Boolean {
+   return when (type) {
+      String::class -> true
+      Int::class -> true
+      Long::class -> true
+      Double::class -> true
+      Float::class -> true
+      Byte::class -> true
+      Short::class -> true
+      Boolean::class -> true
```
simonenkoi: Can we simplify it somehow? 
As an option, we can have a set of classes that we consider as stable and check if `type` is in the set or call `javaPrimitiveType != null`  or `klass.java.isPrimitive`

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/names.kt
```diff
@@ -0,0 +1,44 @@
...
+import kotlin.reflect.KClass
+import kotlin.reflect.KType
+
+/**
+ * Each test name must be unique. We can use the toString if we determine the instance is stable.
+ *
+ * An instance is considered stable if it is a data class where each parameter is either a data class itself,
+ * or a primitive type.
+ *
+ * Note: If the user has overridden toString() and the returned value is not stable, tests may not appear.
```
simonenkoi: Can we somehow notify users about this behavior? For example, if the generated identifier is not unique we can log a warning

simonenkoi (commented):
there are a few nitpicks to consider

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/names.kt
sksamuel: I didn't know about klass.java.isPrimitive so I will use that instead.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/names.kt
sksamuel: Excellent idea, or maybe even not support it.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/names.kt
sksamuel: Ah that's not available on multiplatform, since it's Java only.

sksamuel: I moved the stable types to a set.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/names.kt
simonenkoi: As another option, we can concatenate counter to the identifier
Something like `StableIdentifier`, `StableIdentifier (1)`, `StableIdentifier (2)` for repeated values

sksamuel: So add the counter even for stable values when repeated.

sksamuel: image

image

simonenkoi: If I were a user of such an API, it would be the most intuitive outcome from all the discussed options. 
Will it work the same way for other isolation modes? Do we guarantee the order?  I guess it can be confusing if identifiers are flicky between launches. What will happen in case `Foo(name=ham) (1)` fails and I want to relaunch the failed test only? Will it be named `Foo(name=ham)`? 
What I care about is to identify the place of failure and reproduce it when I have a test report only.

However, I cannot imagine the real-world scenario for this case and consider it as a failover for human mistake. We might want to wait for users' feedback until considering these questions.

sksamuel: Good questions. The plugin is not able to re-launch a single test, and probably won't ever be (as it involves knowing runtime information). The order of execution is not guaranteed if using different isolation modes, but the ordering of names is as all the tests are guaranteed at once.

PR #1813, opened 2020-10-28
BOM for Kotest

grodin:
I haven't updated the Github actions to publish the BOM, and there aren't any tests obviously, because it's just changes to the gradle build.

The Github actions changes necessary is just to execute the task ":kotest-bom:publishBomPublicationToDeployRepository" so it should just be a copy-and-paste job, but I don't have much experience with Github actions and in any case, I'd be unable to test if I'm right, so I just left it as is.

### DIFF
--- a/kotest-bom/build.gradle.kts
+++ b/kotest-bom/build.gradle.kts
@@ -5,53 +5,27 @@ plugins {
 
 version = Ci.publishVersion
 
-val bomProjectName = project.name
+val bomProject = project
+val excludeFromBom = listOf("kotest-examples", "kotest-tests")
+fun projectsFilter(candidateProject: Project) =
+   excludeFromBom.all { !candidateProject.name.contains(it) }
+      && candidateProject.name != bomProject.name
 
-rootProject.subprojects.filter { project ->
-   project.name != bomProjectName
-}.forEach {
-   evaluationDependsOn(it.path)
-}
-
+rootProject.subprojects.filter(::projectsFilter).forEach { bomProject.evaluationDependsOn(it.path) }
 
 dependencies {
    constraints {
       rootProject.subprojects.filter { project ->
-         project.name != bomProjectName &&
-            project.plugins.hasPlugin(MavenPublishPlugin::class)
-      }.forEach {
-
-         println(it)
-         it.publishing.publications.filterIsInstance<MavenPublication>()
-            .forEach { publication ->
-               println(publication.artifactId)
-            }
-      }
+         project.tasks.findByName("publish")?.enabled == true &&
+            projectsFilter(project)
+      }.forEach { api(project(it.path)) }
    }
 }
 
 publishing {
    publications {
-      create<MavenPublication>("kotest-bom") {
+      create<MavenPublication>("Bom") {
          from(components["javaPlatform"])
-      }
-   }
-}
-
-val listPublishingTasks by tasks.registering {
-   doLast {
-      println("List Publishing artifacts")
-      rootProject.subprojects.filter { project ->
-         project.tasks.findByName("publish")?.enabled == true &&
-            project.name != bomProjectName
-      }.forEach { project ->
-         println(project)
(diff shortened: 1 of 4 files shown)


kotest-bom/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+
+version = Ci.publishVersion
+
+val bomProject = project
+val excludeFromBom = listOf("kotest-examples", "kotest-tests")
+fun projectsFilter(candidateProject: Project) =
+   excludeFromBom.all { !candidateProject.name.contains(it) }
+      && candidateProject.name != bomProject.name
+
+rootProject.subprojects.filter(::projectsFilter).forEach { bomProject.evaluationDependsOn(it.path) }
```
sksamuel, line 20: What does evaluationDependsOn do, can you add a comment ?

kotest-bom/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+
+dependencies {
+   constraints {
+      rootProject.subprojects.filter { project ->
+         project.tasks.findByName("publish")?.enabled == true &&
+            projectsFilter(project)
```
sksamuel: Can you do a quick comment

kotest-bom/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+      }.forEach { api(project(it.path)) }
+   }
+}
+
+publishing {
+   publications {
+      create<MavenPublication>("Bom") {
```
sksamuel: What is "Bom" here?

sksamuel (approved):
It looks good, but a couple of comments on things I'm not familiar with would help.

kotest-bom/build.gradle.kts
grodin: It's just an arbitrary name for the publication which doesn't propagate to anything published as far as I can tell. I'll change it to KotestBom for completeness.

kotest-bom/build.gradle.kts
grodin: Done

sksamuel:
So to publish the BOM its just a matter of running that `publishBomPublicationToDeployRepository` task. Is that a built in task created through a plugin?

kotest-bom/build.gradle.kts
grodin: I've added a comment but I wanted to say something here as well. We don't want to end up with non-publishing subprojects contributing to the BOM (it won't really matter, as a BOM just says what version to use if you do end up using a given dependency) so we filter out those.

`evalutationDependsOn(":subproject:)` declares that the current project has a dependency on  `:subproject` and configuration of the current project triggers configuration of `:subproject`.

The reason this is necessary is that we use the `constraints` block to declare a dependency constraint and not a hard dependency on the project.

I'm not sure how well I've explained it actually. Is that reasonably clear?

sksamuel: I don't really understand the bit about constraints and why you did that.

grodin:
Actually, that should change to `publishKotlinBomPublicationToDeployRepository` as I've changed the name to `KotlinBom`.

Essentially, whenever you create a publication `Foo` in a gradle project, the `maven-publish` plugin generates a task `publishFooPublicationToDeployRepository`. Normally, these publications are created by the `java` or `java-library` plugins from the project name. When you use the `java-platform` plugin you've got to do it manually.

sksamuel:
> Actually, that should change to `publishKotlinBomPublicationToDeployRepository` as I've changed the name to `KotlinBom`.
> 
> Essentially, whenever you create a publication `Foo` in a gradle project, the `maven-publish` plugin generates a task `publishFooPublicationToDeployRepository`. Normally, these publications are created by the `java` or `java-library` plugins from the project name. When you use the `java-platform` plugin you've got to do it manually.

Ah that makes sense.

PR #1822, opened 2020-10-30
Changed invocationTimeout to duration in ProjectConfig

SergKhram:


SergKhram:
@sksamuel check it pls

sksamuel:
We can't change a public API that hundreds / thousands of projects are using.

sksamuel:
We can add durations back in once they are no longer experimental.

Issue #1820, opened 2020-10-30
Better tags for spec level includes/excludes


sksamuel:
The current @Tags on a spec is powerful but can be confusing. A spec annotated with @Tags("Foo") will not necessarily be excluded if you run with kotest.tags=Bar because root level tests may be annotated with Bar and so it must instantiate the spec to check. Only if a tag is explicitly excluded does the @Tags on a spec help.

This ticket is to:
* Add a new annotation (name whatever, say @Tagged) which, when an include is set, is required, and when an exclude is set, must not be present, for that spec to be instantiated. 
* deprecate @Tags.

kopper:
To give you some context let me describe our use-case first. As many other projects we have unit and integration tests. We would like to be able to run only one set of them. Right now we are using naming convention *Test for uni tests and *IT for integration and maven surefire plugin to run unit tests while maven failsafe plugin to run integration tests. I expect us to define more test types in the future and classify some as "slow" to avoid running them too frequently.

In order to save time/CPU we'd like to avoid instantiating specs that are not of the type that was requested to run. I understand that in order to determine all tags associated with the test we need to instantiate the spec.

What would work in our case is to have distinction between test type and test tag, for example:

```
@TestType("UnitTest")
class MyTest : FunSpec({
  tags(SlowTest, FlakyTest)
}
```

and then `-Dkotest.type=IntegrationTest` wouldn't even instantiate the spec.

sksamuel:
Your use case is the perfect example of what I was trying to convey.

Although rather than introduce two types of tags/things, I think just having a different annotation may be easier, because then it expands out to any use case. For example @jschneidereit uses a different tag per team.

```kotlin
@RequiresTag(UnitTest) // tag must be present to be instantiated
class MyTest1 : FunSpec({
  test("a").config(tag = FlakyTest) { }
  test("b").config(tag = SlowTest) { }
}

@Tags(FooTag) // a tag that is simply added to the other tags of the tests
class MyTest2 : FunSpec({
  test("c").config(tag = FlakyTest) { }
  test("d").config(tag = SlowTest) { }
}
```

MyTest1 will now only be instantiated when kotest.tags explicitly includes UnitTest.

So under the following behaviors.

kotest.tags=UnitTest => a,b (but MyTest2 is instantiated to search inside)
kotest.tags=UnitTest & FlakyTest => a (but MyTest2 is instantiated to search inside)
kotest.tags=FlakyTest => a, c
[...]

PR #1823, opened 2020-10-30
Fixed the incorrect condition for filtering instantiating spec class when we use @Tags annotation

SergKhram:


SergKhram:
I think, that what we use in "kotest.tags.include" - we will execute only tests, with data from this list
Because if we have , for example, 5 different tags and in each teamcity (or other CI) build have use different test tags - > easy to include one of it and execute only tests with this tag rather than exclude all another tags

sksamuel:
This makes sense but will break the existing behavior.
Instead, what about implementing https://github.com/kotest/kotest/issues/1820 ?

SergKhram:
#1820 I will see how to realize it better this weekend

SergKhram:
@sksamuel  check it, pls

SergKhram:
@sksamuel check it, pls

sksamuel:
I get an email when you make a change you know.

sksamuel (commented):
This looks good, but before I merge, I want to sit down and think it over. I'll spend time on it this week and merge next weekend.

SergKhram:
@sksamuel ping

SergKhram:
@LeoColman check pls)

SergKhram:
@sksamuel can we merge it ?

SergKhram:
@sksamuel  ??

SergKhram:
@sksamuel )

sksamuel:
This has been handled in https://github.com/kotest/kotest/pull/2285/

Issue #1831, opened 2020-11-04
Wrong behavior when combining assertSoftly and shouldNotBe


PeekAndPoke:
**Which version of Kotest are you using**
4.3.1

**Desription**
Under 'certain' circumstances the combination of `assertSoftly` and `shouldNotBe` show a wrong behavior.

I have an example:

```
enum class SimpleEnum {
    First,
    Second
}
data class WithSimpleEnum(val enumValue: SimpleEnum = SimpleEnum.First)
class AssertSoftlySpec : StringSpec({
    "asserting me softly" {
        val source = WithSimpleEnum(enumValue = SimpleEnum.First)
        val result = WithSimpleEnum(enumValue = SimpleEnum.Second)
        assertSoftly {
            withClue("simple strings") {
                "a" shouldBe "b"
                "a" shouldNotBe "b"
            }
            withClue("more complex with data class and enums") {
                source shouldBe result
                source shouldNotBe result
            }
        }
    }
})
```

The will output:

```
The following 4 assertions failed:
1) simple strings
expected:<"b"> but was:<"a">
	at de.peekandpoke.ultra.AssertSoftlySpec$1$1.invokeSuspend(AssertMeSoftlySpec.kt:26)
2) more complex with data class and enums
expected:<Second> but was:<First>
	at de.peekandpoke.ultra.AssertSoftlySpec$1$1.invokeSuspend(AssertMeSoftlySpec.kt:31)
3) more complex with data class and enums
expected:<WithSimpleEnum(enumValue=Second)> but was:<WithSimpleEnum(enumValue=First)>
	at de.peekandpoke.ultra.AssertSoftlySpec$1$1.invokeSuspend(AssertMeSoftlySpec.kt:31)
4) more complex with data class and enums
expected:<Second> but was:<First>
	at de.peekandpoke.ultra.AssertSoftlySpec$1$1.invokeSuspend(AssertMeSoftlySpec.kt:32)
```

So the situation is: I was **not** able to produce the behaviour with scalar types like strings.

But as you can see, in this part **both assertions fail**, while this should be impossible (or my logic-chip is broken)

```
withClue("more complex with data class and enums") {
    source shouldBe result
    source shouldNotBe result
}
```

ashishkujoy:
Fixed in master

sksamuel:
4.3.2 is released with this fix.

PR #1833, opened 2020-11-06
#1831 Fixes multiple error in case of data class assertion inside assertsoftly

ashishkujoy:


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/DataClassEq.kt
```diff
@@ -64,9 +65,12 @@ internal object DataClassEq : Eq<Any> {
...
             }
-         else
-            runCatching { actualPropertyValue shouldBe expectedPropertyValue }
-               .fold(onSuccess = { null }, onFailure = { t -> Pair(prop, StandardDifference(t)) })
+         else {
+            eq(actualPropertyValue, expectedPropertyValue)
+               .toOption()
+               .map { Pair(prop, StandardDifference(it)) }
+               .orNull()
+         }
```
sksamuel, line 73: Can you explain what the change is doing ?

ashishkujoy: Earlier we are using `actualPropertyValue shouldBe expectedPropertyValue`

shouldBe internally add the error to errorCollector due to which we are getting multiple error message when data class contains enum, now i replaced that with eq function call which check equality and return error using which we are creating the difference message for data class without adding those extra errors in errorCollector.

ashishkujoy: Is it good to merge?

sksamuel: Ok I see.

sksamuel (approved):

PR #1838, opened 2020-11-09
Added inline version for intanceOf alias

### LINKED ISSUE #1836: Add inline version of instanceOf to simplify usage
I would like to add a version of instanceOf that looks like

    inline fun <reified T: Any> instanceOf(): Matcher<Any?> = instanceOf(T::class)

So that the call site goes from

    instanceOf(SomeType::class)
    // to
    instanceOf<SomeType>()

It is a minor addition and would just remove the need for `::class`. This is already done `beInstanceOf` and `beOfType`. This would just do the same thing for the `instanceOf` alias.

dmstocking:
Added a version of `instanceOf` that is inlined with a reified generic
type. This was done to remove the need for `::class` from the caller.
This type of function is already implemented for `beInstanceOf` and
`beOfType`. This commit just also adds it to `instanceOf`.

Also updated the tests to test every matcher, even aliases. I don't
suspect anyone would accidentally change the implementation of the
aliased functions, but you never know.

Fixes #1836


sksamuel:
Thanks. There's always `.shouldBeInstanceOf<T>` too but this makes the infix version nicer to use.

PR #1844, opened 2020-11-13
Revert Kotlin API version to 1.3

### LINKED ISSUE #1840: Kotlin API Version changed to 1.4 in 4.3.1
When upgrading from version 4.3.0 to 4.3.1 our tests broke because of #1801: in `kotest-assertions/kotest-assertions-shared/build.gradle.kts` the `apiVersion` was changed to 1.4. While kotest 4.4 will require Kotlin 1.4 anyway, upgrading to Kotlin 1.4 in a minor release seems like an accident (especially since it is not mentioned in the changelog).

U1F984:
Fixes #1840


sksamuel:
Can you change this PR to point to release/4.3.x branch, as master is 4.4 dev.

Issue #1860, opened 2020-11-24
String shrinking is not being executed 


sksamuel:
See ShrinkingTest.kt for an example.

Sebruck:
I have a similiar issue, here a short reproducer:


My code

```kotlin
import io.kotest.core.spec.style.StringSpec
import io.kotest.property.PropTestConfig
import io.kotest.property.PropertyTesting
import io.kotest.property.ShrinkingMode
import io.kotest.property.forAll

class PropertyExample: StringSpec({
    "Shrinking happens" {
        PropertyTesting.shouldPrintShrinkSteps = true
        val config = PropTestConfig(shrinkingMode = ShrinkingMode.Unbounded)
        forAll<Int>(config) { a ->
            a < 1000
        }
    }
}
```

Output:
```
Property failed after 4 attempts

	Arg 0: 2147483647

Repeat this test by using seed 776132071643096315

Caused by AssertionFailedError: expected:<true> but was:<false>
Expected :true
Actual   :false
<Click to see difference>

java.lang.AssertionError: Property failed after 4 attempts

	Arg 0: 2147483647

Repeat this test by using seed 776132071643096315
```

So as you can see, no shrinking happened, even though I have set the ShrinkingMode to Unbounded.

In the Kotest code I can see that an IntShrinker exists and is used for the Integers Arb, but somehow it is not executed:
```kotlin
fun Arb.Companion.int(range: IntRange = Int.MIN_VALUE..Int.MAX_VALUE): Arb<Int> {
   val edgecases = listOf(0, 1, -1, Int.MAX_VALUE, Int.MIN_VALUE).filter { it in range }
   return arbitrary(edgecases, IntShrinker(range)) { it.random.nextInt(range) }
}
```

sksamuel:
4.3.2 is released with this fix.

PR #1861, opened 2020-11-24
fixes "ListShrinker in action"

tieskedh:
"ListShrinker in action" did not test for shrink-result

Question, do we want the number of shrinks hardcoded?
This will fail on an improvement and is not the thing being tested.
The other option would be to test for "=>" but that means that it can never be used for something else then the final result.

The best option (I think) would be  to create an util-function which would test for the final result, so we could call something like:

```
 test("ListShrinker in action") {
         shouldThrowAny {
            checkAll(PropTestConfig(seed = 123132), Arb.list(Arb.int(0..100))) { list ->
               list.shouldHaveAtMostSize(3)
            }
         }.message should containShrinkResult("[0, 1, 51, 24]")
      }
```

Or "shouldmatch" with an util function.
(I'm not that familiar with regex, though)


sksamuel:
Looks good but tests failing.

tieskedh:
@sksamuel 
That's because shrinking is not working, see #1860
This means this test **should** be failing (for now)

tieskedh:
can't edit, as I removed repo after accidently pushing incorrect content.
reopening with passing test.

PR #1868, opened 2020-11-25
Fix for "Unexpected ToleranceMatcher behavior [...]"

### LINKED ISSUE #1832: Unexpected ToleranceMatcher behavior at infinite doubles
4.3.1

Mathematically I would expect

    checkAll(Arb.positiveDoubles()) { eps ->
      Double.NEGATIVE_INFINITY shouldBe (Double.NEGATIVE_INFINITY plusOrMinus eps)
      Double.POSITIVE_INFINITY shouldBe (Double.POSITIVE_INFINITY plusOrMinus eps)
    }

to succeed. But Kotest fails with

    Property test failed for inputs

    0) 1.0

    Caused by java.lang.AssertionError: -Infinity should be equal to -Infinity within tolerance of 1.0 (lowest acceptable value is -Infinity; highest acceptable value is -Infinity) at
	[...]
	io.kotest.property.internal.ProptestKt$proptest$$inlined$forEach$lambda$1.invokeSuspend(proptest.kt:32)
	io.kotest.property.internal.ProptestKt$proptest$$inlined$forEach$lambda$1.invoke(proptest.kt)

sielenk-yara:
Fixes #1832

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/doubles/Tolerance.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/doubles/Tolerance.kt
@@ -14,14 +14,14 @@ import kotlin.math.abs
  * ```
  */
 infix fun Double.plusOrMinus(tolerance: Double): ToleranceMatcher {
-   require(tolerance >= 0)
+   require(tolerance >= 0 && tolerance.isFinite())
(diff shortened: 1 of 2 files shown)


sksamuel:
Awesome. The tests are failing, I haven't checked why.

sielenk-yara:
Only on Ubuntu, with a timeout.
The one I added ran successfully there, I found it in the log.

LeoColman:
LGTM
Thanks for the fix!

PR #1870, opened 2020-11-25
Fix for "Problems with the double generator"

### LINKED ISSUE #1869: Problems with the double generator (Arb.Companion.double).
4.3.1

The edge cases are incomplete:
Missing values are `-0.0` (negative zero), `-Double.MIN_VALUE`, `-Double.MAX_VALUE` and, since `1e300` is there, maybe also `-1e300`.

The random values are far from exhaustive:
They are generated with `nextDouble()` and are therefore ["[...] uniformly distributed between 0 (inclusive) and 1 (exclusive)"](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.random/-random/next-double.html).
A better choice would be `nextDouble(-Double.MAX_VALUE, Double.MAX_VALUE)`.

This in turn leads to problems with `Arb.Companion.negativeDoubles()` since, at the moment, there are exactly two values that can pass the filter `{ it < 0.0 }`. And trying to get 1000 values from that generator freezes the tests.

More details about what a nasty piece of work floating point numbers are can be found in 
- the [Java specification](https://docs.oracle.com/javase/specs/jls/se7/html/jls-4.html#jls-4.2.3) or
- the [IEEE 754 Standard for Floating-Point Arithmetic](https://en.wikipedia.org/wiki/IEEE_754)

See also #530, #1216

sielenk-yara:
Fixes #1869

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/doubles.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/doubles.kt
@@ -11,16 +11,20 @@ import kotlin.math.round
 fun Arb.Companion.double(): Arb<Double> {
    val edgecases = listOf(
       0.0,
+      -0.0,
       1.0,
       -1.0,
       1e300,
+      -1e300,
       Double.MIN_VALUE,
+      -Double.MIN_VALUE,
       Double.MAX_VALUE,
+      -Double.MAX_VALUE,
+      Double.POSITIVE_INFINITY,
       Double.NEGATIVE_INFINITY,
-      Double.NaN,
-      Double.POSITIVE_INFINITY
+      Double.NaN
    )
-   return arb(DoubleShrinker, edgecases) { it.random.nextDouble() }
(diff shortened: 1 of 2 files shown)


sielenk-yara:
Again a timeout in the ubuntu build with the new test passing:

    [...]
    com.sksamuel.kotest.property.arbitrary.DoubleTest

      Test Numeric Doubles should generate negative values by default[jvm] PASSED
      Test Double should generate corner cases, positive and negative values[jvm] PASSED
    [...]
    com.sksamuel.kotest.timeout.InlineTimeoutSuccessPrecedenceTest > test case config timeout should take precedence FAILED
        java.lang.InterruptedException
            at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:83)
            at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:59)
    [...]

LeoColman (approved):

LeoColman:
Thanks for this fix!

sksamuel:
Test failures are unrelated so merging in.

PR #1873, opened 2020-11-28
#1864 Makes scenario create test instead of container test

### LINKED ISSUE #1864: AssertionMode.Error doesn't work on FeatureSpec
version: io.kotest:kotest-runner-junit5:4.3.1
kotlin version: 1.4.10
Android Studio version: 4.1.1

assertion mode doesn't work on FeatureSpec, and `Tests passed` always equal to 0 even if I do assertion

```
class FeatureSpecTest : FeatureSpec({
  assertions = Error

  feature("Test feature") {
    scenario("test scenario") {
    }
  }
})
```

image

In do assertion case: 

```
class FeatureSpecTest : FeatureSpec({
  assertions = Error

  feature("Test feature") {
    scenario("test scenario") {
      val a = 1
      a shouldBe 1
    }
  }
})
```
`Tests passed` still equal to 0
image

ashishkujoy:
Since we do not provides any other function inside `scenario` block to register a `test case` neither do we can call `scenario` inside `scenario` block, so it makes sense for `scenario` function to register a `test` instead of `container test`. As a result of that kotest will be able to fail test when assertion mode is `error` and there is no assertion inside `scenario` block.

closes #1864

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/FeatureScope.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/FeatureScope.kt
@@ -61,11 +61,11 @@ class FeatureScope(
    }
 
    suspend fun scenario(name: String, test: suspend TestContext.() -> Unit) {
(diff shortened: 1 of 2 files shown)


sksamuel:
lgtm; merge once green.

PR #1876, opened 2020-12-02
numeric double edge cases - follow up to the double generator edge cases fix

sielenk-yara:
further updates along the changes in #1870:
- the `numericDoubles` generator should have its limits as edge cases
- pulled shared code into variables - maybe to become public `Exhaustive`s?


sksamuel:
Exhaustives of "edge cases" is a good idea.

PR #1924, opened 2020-12-18
:bug: improved the json assert error messages so intellij recognizes them as a diff

enolive:
ok, this is a follow up to the issues #1106 and #1112.

The message format is still not correct and is not recognized by intellij as something that should go inside a diff dialog.

I've updated the error messages so it is really recognized. This seems to be the pattern that IntelliJ is expecting:

https://github.com/JetBrains/intellij-community/blob/master/java/java-runtime/src/com/intellij/rt/execution/testFrameworks/AbstractExpectedPatterns.java


enolive:
I am closing this for now -- the diff window actually works fine in junit runner, but not in kotest. Will investigate it further

PR #1925, opened 2020-12-18
adds partial shrinkers

tieskedh:
Adds  a way to compose shrinkers out of multiple other shrinkers.

example:

```kotlin
val OrganisationShrinker = createShrinker<Organisation> {
    build(
        select { organisationID.value } shrinkWith StringShrinker,
        Organisation::name shrinkWith StringShrinker,
        Organisation::type shrinkWith StringShrinker,
        Organisation::thumbnail shrinkWith StringShrinker,
        Organisation::clientId shrinkWith IntShrinker(0..100)
    ) { orgId, name, type, thumb, clientId ->
        copy(
            OrganisationID(orgId),
            name,
            type,
            clientId,
            thumb,
            baseUrl,
            primaryColor
        )
    }
}
```

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/PartialShrink.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/PartialShrink.kt
@@ -1,0 +1,292 @@
+package io.kotest.property
+
+import io.kotest.property.arbitrary.arbitrary
+import io.kotest.property.arbitrary.next
+import kotlin.reflect.KProperty1
+
+fun <T> Arb<T>.shrinkWith(
+   shrinker: Shrinker<T>
+) = arbitrary(edgecases(), shrinker) { next(it) }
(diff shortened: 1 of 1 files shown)


kotest-property/src/commonMain/kotlin/io/kotest/property/PartialShrink.kt
```diff
@@ -0,0 +1,292 @@
+package io.kotest.property
+
+import io.kotest.property.arbitrary.arbitrary
+import io.kotest.property.arbitrary.next
+import kotlin.reflect.KProperty1
+
+fun <T> Arb<T>.shrinkWith(
```
sksamuel, line 7: Can we call this `withShrinker`

sksamuel (changes requested):
Can we add some tests so I can see how this works.

PR #1926, opened 2020-12-18
Fix "ListShrinker in action" - test

tieskedh:
List shrinker in action did not test shrinking


tieskedh:
hmm...   `Test ListShrinker in action[jvm] PASSED`
> Received a failure event for test with unknown id '33.1093'. Registered test ids: '[33.1, 33.1090, :kotest-tests:kotest-tests-core:jvmTest]'

sksamuel:
There's a timing issue unrelated to your PR.

Issue #1929, opened 2020-12-21
Increase arity of checkall / forall property tests to 12


sksamuel:
**Please describe the feature you'd like to see including any solutions in mind if you have any**

PR #1933, opened 2020-12-22
Increase arity of checkall / forall property tests to 10 #1929

sksamuel:


sksamuel:
These tests are failing because of a test on master that is now fixed.

PR #1937, opened 2020-12-22
#1932 Set assertion mode back to hard even in case any exception is thrown from assertion block

### LINKED ISSUE #1932: assertSoftly: If one test in spec fais in assertSoftly with non-assert exception - other tests do nothing and shown as "green"
Assume following test:
```

internal class InvalidAssertSoftlyBehaviorTest : StringSpec() {
    init {
        "a"{
            assertSoftly {
                throw Exception()
            }
        }

        "b" {
            1 shouldBe 0
        }
    }
}

```

If you strart just `b` - it fails, as it is expected,

but if you start whole test - `a` will be failed, but `b` will OK!!

In debug I can see that `b` is still in `Soft` assertion mode and it's assert never thrown!

ashishkujoy:
Currently in case if any exception is getting thrown from assertion block of `assertSoftly` the errorCollection mode was not changing back to `hard` as result any assertion error in subsequent test were not getting reported. This PR make sure that even in case of exception the assertion mode is get set back to `hard` so that subsequent test does not effected by current test.

Fixes #1932


sksamuel:
Nice find.

PR #1940, opened 2020-12-23
Allow null clues

### LINKED ISSUE #1939: Allow null clues
Clues are declared as `Any` rather than `Any?`. I'm not sure why this is helpful.

Notably, I find myself wanting to write `foo.asClue { foo.shouldBeNull() }` sometimes. Is there a better way to do this? `foo?.asClue { foo.shouldBeNull() }` skips the assertion, and `foo!!.asClue { foo.shouldBeNull() }` moves the failure to `!!` itself.

glasser:
Fixes #1939.

Helpful for e.g. `foo.asClue { foo.shouldBeNull() }`


LeoColman:
Great! Thank you 👍🏻

PR #1944, opened 2020-12-24
[doc] Fixed Get Started button color for dark theme

alexfacciorusso:
The Get started button text and the header text are changing their colour in the dark theme, but their background isn't, meaning their readability goes down (in the case of the button, the text gets completely unreadable).

Fixed that by assigning static colours to those classes.

|           | Before | After |
|--------|--------|-------|
| Light | localhost_3000_ | localhost_3000_ (2) |
| Dark | localhost_3000_ (1) | localhost_3000_ (3) |

### DIFF
--- a/documentation/src/pages/styles.module.css
+++ b/documentation/src/pages/styles.module.css
@@ -40,6 +40,7 @@
 .heroSlogan {
    font-size: 60px;
    line-height: normal;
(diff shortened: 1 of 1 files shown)


sksamuel:
Who uses dark mode :p

PR #1946, opened 2020-12-27
Add functions for data tests with custom names

malteesch:
The [documentation](https://kotest.io/docs/framework/data-driven-testing.html#custom-test-names) shows the ability to assign custom names to data tests.
In `4.3.2` this is only possible with the `forNone` function so I added the `forAll` functions which accept custom names.


sksamuel:
👍🏼

PR #1955, opened 2020-12-31
not null matcher should show the value that was meant to be null, closes #1942

### LINKED ISSUE #1942: not null matcher should show the value that was meant to be null
https://github.com/kotest/kotest/blob/757529896b9d117240b281e69926953022d6ca9d/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/nulls/matchers.kt#L104

swanandvk:
https://github.com/kotest/kotest/blob/757529896b9d117240b281e69926953022d6ca9d/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/nulls/matchers.kt#L104

closes #1942


sksamuel (approved):

sksamuel:
👍🏼

glasser:
The second half of the change doesn't look right? Won't $value be null?

sksamuel:
Good spot @glasser 
Fixed in master.

PR #1964, opened 2021-01-04
fix Arb.set hanging when underlying arb cardinality was insufficient

### LINKED ISSUE #1931: Arb.set with a range hangs the test if the given gen inside the set cannot produce enough values for the range
Found in Kotest version 4.3.2

Example
```
  "test set of enum with less than 3 elements hangs" {
                checkAll(Arb.set(Arb.enum<Enum>(), 1..3)) {
                    it shouldContain Enum.SingleValue
                }
            }

enum class Enum {
        SingleValue // I also tried this with 2 elements in the enum and it hangs. 3 elements would not hang.
    }
```

myuwono:
fixes #1931


sksamuel:
nice 👍🏼

PR #1968, opened 2021-01-05
Add data driven in root scope

### LINKED ISSUE #1967: Allow data driven tests to register in root scope

jschneidereit:
Should fix #1967 - I only added the overload to one spec since all of the data driven specs are disabled right now


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/datatest.kt
```diff
@@ -31,20 +25,13 @@ suspend fun <T> ContainerScope.forNone(data: List<Pair<String, T>>, test: suspen
...
-}
+suspend fun <T : Any> ContainerScope.forAll(vararg ts: T, test: suspend (T) -> Unit) = forAll(ts.toList(), test)
 
-suspend fun <T : Any> ContainerScope.forAll(vararg ts: T, test: suspend (T) -> Unit) {
-   val idents = Identifiers()
-   ts.forEach { t ->
-      val name = idents.stableIdentifier(t)
-      addTest(createTestName(name), false) { test(t) }
-   }
+suspend fun <T : Any> ContainerScope.forAll(vararg data: Pair<String, T>, test: suspend (T) -> Unit) = forAll(data.toList(), test)
```
jschneidereit, line 31: Tomorrow morning I need to verify that this won't get confused with the `List<T>` overload

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/datatest/datatestconcurrent.kt
```diff
@@ -0,0 +1,42 @@
+package io.kotest.core.datatest
```
sksamuel: I would rather these live in the same file so avoid import clashes. That way when you import forAll from this package, you can use it in both root scope and container scope in the same spec.

sksamuel (changes requested):
Looks good. I would like to see the test added to the other spec types too, so that we know it works on all (where applicable).

sksamuel (approved):

PR #1971, opened 2021-01-05
Enhances email arb; Creates domain arb

### LINKED ISSUE #1941: Add more configurations to Email Generator
The current email generator doesn't allow customization of domain and TLD. Sometimes it's desirable, as for example, you want to test that people from a "bad domain" will be denied to register in your website.

We should support more customization in that sense, and deprecate the non-customizable approach.

### LINKED ISSUE #1969: Add domain arbitrary

LeoColman:
This commit will improve how we generate random emails, in a way that it
 will become closer to real email addresses.

 In addition, this commit also creates a Domain arbitrary. Originally
 used only in emails, generating random valid domains can be useful for
 many usecases.

 Closes #1941
 Closes #1969

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/domain.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/domain.kt
@@ -1,0 +1,164 @@
+package io.kotest.property.arbitrary
+
+import io.kotest.property.Arb
+
+/**
+ * Generates random domains
+ *
+ * https://en.wikipedia.org/wiki/Domain_name#Domain_name_syntax
+ * https://whogohost.com/host/knowledgebase/308/Valid-Domain-Name-Characters.html
+ */
+fun Arb.Companion.domain(
+   tlds: List<String> = top120TLDs,
+   labelArb: Arb<String> = _labelArb
+): Arb<String> = arbitrary { rs ->
+   val amountOfLabels = rs.random.nextInt(1, 4)
+
+   val labels = labelArb.take(amountOfLabels).joinToString(separator = ".", postfix = ".")
+
+   labels + tlds.random(rs.random)
+}
+
+private val _labelArb = arbitrary {
+   buildString(63) {
+      append(validCharsArb.filter { it != '-' }.single(it))
+
+      append(validCharsArb.take(it.random.nextInt(0, 61), it).joinToString(""))
+
+      append(validCharsArb.filter { it != '-' }.single(it))
+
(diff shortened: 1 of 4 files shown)


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/domain.kt
```diff
@@ -0,0 +1,164 @@
...
+ * https://en.wikipedia.org/wiki/Domain_name#Domain_name_syntax
+ * https://whogohost.com/host/knowledgebase/308/Valid-Domain-Name-Characters.html
+ */
+fun Arb.Companion.domain(
+   tlds: List<String> = top120TLDs,
+   labelArb: Arb<String> = _labelArb
+): Arb<String> = arbitrary { rs ->
+   val amountOfLabels = rs.random.nextInt(1, 4)
+
+   val labels = labelArb.take(amountOfLabels).joinToString(separator = ".", postfix = ".")
```
sksamuel, line 17: Do we want domains to end with a dot ?

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/domain.kt
```diff
@@ -0,0 +1,164 @@
...
+      }
+   }
+}
+
+private val validCharsArb = arbitrary { (('a'..'z') + ('A'..'Z') + ('0'..'9') + '-').random(it.random) }
+
+/**
+ * https://www.scoutdns.com/100-most-popular-tlds-by-google-index/
+ */
+private val top120TLDs = listOf(
```
sksamuel, line 43: top120 ?

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/domain.kt
LeoColman: They don't. 
It's just the label (parts before the TLD). So we take 1 to 3 labels (a.b.c) and then add a dot and a tld (a.b.c.com)

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/domain.kt
LeoColman: From position 120 and onwards the amount of indexed pages decreases faster than from 1 to 120, that's why I cut the line at 120. But I think it could be any amount here, the spreadsheet has 1510 indexed TLDs

sksamuel (approved):

PR #1972, opened 2021-01-06
Output from reporters should be single threaded #1895

### LINKED ISSUE #1895: Output from reporters should be single threaded
Given a parallel test suite, I'd like to have the my console logs not be interleaved - test output should be thread safe.

sksamuel:
Closes #1895


sksamuel:
@jschneidereit

jschneidereit (approved):

PR #1994, opened 2021-01-12
Add bind 11, 12, 13, 14

i-walker:
TODO:
- [x] Tests -- I didn't know which Tests I should add or simply copy the BindTests for jvm and add them for an arity 11 and 12, 13, 14?


sksamuel:
Yes please, just add something in jvmTest that binds values for 11 and 12 like the others.

sksamuel (approved):

sksamuel:
👍🏼

PR #2002, opened 2021-01-18
Fix Small Typo in "Isolation Mode" Docs

caramelomartins:
Hello,

I was reading the documentation for isolation mode and I found "exected" in the place of what I believe should be "executed". I thought about contributing with this small fix. If this is not a typo, I apologize for opening a PR for this. 🙂

Thanks,
Hugo


sksamuel:
Thanks for the clean up.

Issue #2004, opened 2021-01-20
Replace jackson and js native with kotlinx-serialization json parsing in kotest-assertions


jschneidereit:
The kotlinx-serialization library can do [arbitrary json parsing](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#subtypes-of-json-elements) and is a cross platform project by default. Replacing Jackson with kotlinx-serialization would make [kotest-assertions](https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-json/build.gradle.kts) a cross platform module by default.

jschneidereit:
Might be blocked by https://github.com/Kotlin/kotlinx.serialization/issues/204 unless we're ok with temporarily reading resources right into memory `this.javaClass.getResourceAsStream(resource)` see `io.kotest.assertions.json.resources.kt`

sksamuel:
I don't think a test needs to be streaming. Loading into memory seems fine ?

jschneidereit:
Fixed in #2006

PR #2006, opened 2021-01-20
replace jackson and js native with kotlinx-serialization json parsing

jschneidereit:


jschneidereit:
Eventually should fix #2004

kotest-assertions/kotest-assertions-json/build.gradle.kts
```diff
@@ -1,6 +1,7 @@
 plugins {
    id("java")
    kotlin("multiplatform")
+   kotlin("plugin.serialization") version "1.4.10"
```
sksamuel: Is there no 1.4.21 version ?

sksamuel:
You said on slack this was ready.  If so can you mark it as ready for review.

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -15,11 +15,8 @@ enum class CompareOrder {
 /**
  * Compares two json trees, returning a detailed error message if they differ.
  */
-fun compare(expected: JsonNode, actual: JsonNode, mode: CompareMode, order: CompareOrder) =
-   compare(emptyList(), expected, actual, mode, order)
-
 fun compare(
-   path: List<String>,
+   path: List<String> = emptyList(),
```
sksamuel: not sure we need the default here if it's only called internally.

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -182,6 +181,8 @@ fun compareDouble(path: List<String>, expected: JsonNode.DoubleNode, actual: Jso
    return when {
       actual is JsonNode.DoubleNode -> compareDoubles(path, expected.value, actual.value)
       actual is JsonNode.LongNode -> compareDoubles(path, expected.value, actual.value.toDouble())
+      actual is JsonNode.FloatNode -> compareDoubles(path, expected.value, actual.value.toDouble())
```
sksamuel, line 184: Is float actually used if we're just parsing strings? Like how does it know that 0.2 should be a float and not a double ?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/wrappers.kt
```diff
@@ -0,0 +1,28 @@
...
+import kotlinx.serialization.json.floatOrNull
+import kotlinx.serialization.json.intOrNull
+import kotlinx.serialization.json.longOrNull
+
+fun JsonElement.toJsonNode(): JsonNode = when (this) {
+   JsonNull -> JsonNode.NullNode
+   is JsonObject -> JsonNode.ObjectNode(entries.map { it.key to it.value.toJsonNode() }.toMap())
+   is JsonArray -> JsonNode.ArrayNode(map { it.toJsonNode() })
+   is JsonPrimitive -> when {
+      intOrNull != null -> JsonNode.IntNode(intOrNull!!)
```
sksamuel, line 20: What is intOrNull here? Looks like an imported function ?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
jschneidereit: It would be a breaking change to not have the default since I removed the overload, but I can remove it if you'd like!

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/wrappers.kt
jschneidereit: Yeah these are extension methods on the `JsonPrimitive` type that unwrap the value for you

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
sksamuel: This is brand new in 4.4 and can be marked internal anyway, so please add internal and feel free to break.

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/wrappers.kt
sksamuel: Oh wow, so you're in a receiver and using extensions. That's cool.

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
jschneidereit: So for just json strings the answer is no. I guess we have two options here: remove the case here since it appears to always actually parse as a double, or keep the float option to eventually expand this assertions library to include the other serialization kinds that kotlinx-serialization supports.

sksamuel: I guess if they provide a float node as part of their ADT then it won't hurt to include it here.

sksamuel (approved):

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
jschneidereit: 👍 sounds good to me

sksamuel:
If this is ready to merge let me know and I will merge it.

Issue #2007, opened 2021-01-20
assertSoftly loses stacktraces if only one assertion fails


ghost:
**Which version of Kotest are you using**
4.3.1

If only one assertion fails inside a assertSoftly block, then the stack trace of the logged failure will be lost. A sample test is shown here
```
import io.kotest.assertions.assertSoftly
import io.kotest.core.spec.style.FunSpec
import io.kotest.matchers.shouldBe
import kotlin.random.Random
class ATest : FunSpec({
    val foo = Random(1)
    val x = foo.nextInt()
    test("asserts") {
        assertSoftly {
            x shouldBe  5
//            x shouldBe  6
    }}
})
```

If run, the stack traces will be useless. If you uncomment the other assertion, then the assertSoftly logs the message as expected.

It looks like the `UserStackTraceConverter.getUserStacktrace(throwable.stackTrace)` is called. That method strips the top of the stack until it finds a stacktrace after a io.kotest class. If multiple fail, then the list is wrapped in a `MultiAssertion`, which includes a kotest classs to strip away.

The fix looks to be as simple as simple as replacing https://github.com/kotest/kotest/blob/a84d8405babb455a97c79403f04e2bfdcb3fdbdf/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/ErrorCollector.kt#L104
```
      val t = if (failures.size == 1) failures[0] else MultiAssertionError(failures)
```
with 
```
      val t = MultiAssertionError(failures)
```

PR #2008, opened 2021-01-20
Handle only one failure in assertSoftly

taer:
Fix for https://github.com/kotest/kotest/issues/2007


sksamuel:
Does this end up with a nested clause in the stack output?
Also can you see if the tests would pass.

taer:
The error is from a seemingly unrelated failure.
```
  java.lang.NoSuchMethodError: org.springframework.test.context.TestContext.computeAttribute(Ljava/lang/String;Ljava/util/function/Function;)Ljava/lang/Object;
```

sksamuel (approved):

PR #2025, opened 2021-01-31
Add missing highlights in documentation

attilapapai:
Also, it seems to me the `Edit this page` (bottom of the page) is broken on all pages. Couldn't find which configuration drives its value but whenever I click on it shows a `Not Found` message.


sksamuel (approved):

sksamuel:
Yeah I'm not sure how to fix the edit pages thing. Thanks for the other fixes.

PR #2036, opened 2021-02-03
Avoid creating extra lambdas in blocking forAll

ajalt:
This is a very minor optimization that simplifies the code slightly and avoids allocating an extra lambda function. This optimization is already in place for the suspending versions of these functions.


sksamuel:
Thanks

Issue #2044, opened 2021-02-04
Improvements on Eventually given Until deprecation


jschneidereit:
@xerus2000 pointed out some issues with my changes made in #2022  and those should be fixed:

- [x] `EventuallyPredicate` should be a `typealias` instead of a `fun interface`
- [x] Differentiate between a failure of the predicate and an exception in `eventually`
- [x] Add an overload that has no `f` for eventually so you can use `eventually` like you could use `until` (and fix the deprecation replace message)

xeruf:
I think we shouldn't have normal predicates, just matchers and a good way for handling them, because then we can leverage their elaborate error message generation.

xeruf:
By changing to soft assertion mode within the eventually block I think we can catch these more gracefully.

jschneidereit:
I'll look at switching to soft, but I think that probably should be up to the tester to opt in to. What if they don't want every assertion to run if one fails? 

I also think the predicate is fine, consider a producer that returns a kotlin `Result` type and I want to gate on `.isSuccess`

xeruf:
> I also think the predicate is fine, consider a producer that returns a kotlin `Result` type and I want to gate on `.isSuccess`

Totally easy with matchers: `result.isSuccess shouldBe true`

PR #2046, opened 2021-02-05
fix EventuallyPredicate, add a producerless version, and inform user …

jschneidereit:
…about predicate failures for #2044

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/timing/eventually.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/timing/eventually.kt
@@ -58,6 +58,18 @@ suspend fun <T> eventually(duration: Duration, exceptionClass: KClass<out Throwa
    eventually(EventuallyConfig(duration = duration, exceptionClass = exceptionClass), f = f)
 
 /**
+ * Runs a predicate until the predicate returns true as long as the specified duration hasn't passed
+ */
+suspend fun eventuallyPredicate(
+   duration: Duration,
+   interval: Interval = 25.milliseconds.fixed(),
+   retries: Int = Int.MAX_VALUE,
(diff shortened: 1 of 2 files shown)


xeruf:
yeah, might as well call it `until` again ^^

Either way, you need to rework or remove this, as it is incorrect:
```kt
suspend fun until(duration: Duration, interval: Interval = 1.seconds.fixed(), f: suspend () -> Boolean) =
   eventually(duration, interval, f = f)
```
It will return immediately after the first try if that doesn't throw an exception (which until lambdas are not supposed to do) regardless of the returned boolean value.

sksamuel:
> It will return immediately after the first try if that doesn't throw an exception (which until lambdas are not supposed to do) regardless of the returned boolean value.

That's how eventually has always worked - it looked for exceptions.

Now you can additionally evaluate the predicate, combining the power of both eventually and until.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/timing/eventually.kt
```diff
@@ -57,6 +57,18 @@ suspend fun <T> eventually(duration: Duration, poll: Duration, f: SuspendingProd
 suspend fun <T> eventually(duration: Duration, exceptionClass: KClass<out Throwable>, f: SuspendingProducer<T>): T =
    eventually(EventuallyConfig(duration = duration, exceptionClass = exceptionClass), f = f)
 
+/**
+ * Runs a predicate until the predicate returns true as long as the specified duration hasn't passed
+ */
+suspend fun eventuallyPredicate(
```
sksamuel: Can we make this work without requiring the name overload ?

xeruf:
> That's how eventually has always worked - it looked for exceptions.

Yes, but `until` never worked like that, so this alias was bogus.

sksamuel:
4.4.1 has been released with until restored and documented

PR #2054, opened 2021-02-06
Feature/#2051

sksamuel:
Fixes #2054


myuwono (approved):

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/collections.kt
```diff
@@ -40,17 +42,24 @@ fun <T> Arb.Companion.of(vararg collection: T): Arb<T> = element(*collection)
 fun <A> Arb.Companion.set(gen: Gen<A>, range: IntRange = 0..100): Arb<Set<A>> {
    check(!range.isEmpty())
    check(range.first >= 0)
-   val maxIterations = 1000
+   // we may generate duplicates, but we don't know if the underlying gen has sufficient cardinality
+   // to satisfy our range, so we can try for a while, but must not try for ever
+   // the slippage factor controls how many times we will accept a non unique element before giving up,
+   // which is the number of elements in the target set * slippage
+   val slippage = 10
```
myuwono: that's a good idea

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/collections.kt
```diff
@@ -40,17 +42,24 @@ fun <T> Arb.Companion.of(vararg collection: T): Arb<T> = element(*collection)
...
       val genIter = gen.generate(it).iterator()
       val targetSize = it.random.nextInt(range)
+      val maxMisses = targetSize * slippage
       val set = mutableSetOf<A>()
       var iterations = 0
-      while (iterations++ < maxIterations && set.size < targetSize && genIter.hasNext()) {
+      while (iterations < maxMisses && set.size < targetSize && genIter.hasNext()) {
+         val size = set.size
          set.add(genIter.next().value)
+         if (set.size == size) iterations++
```
myuwono, line 65: i think we need to clear the counter when there's an increment in size. one way is to leverage the return value of set.add

```kotlin
val isElementAdded = set.add(genIter.next().value)
iterations = if (isElementAdded) 0 else iterations + 1
```

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/collections.kt
myuwono: that would probably means that it's going to yield a 10x number of iterations before it gives up which can be a bit slow when the set size is big 🤔 . Maybe I'm overthinking it, I don't think that's going to happen very often.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/collections.kt
sksamuel: Thanks, good call.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/collections.kt
sksamuel: Yeah, I think because I wasn't resetting it. I was kind of thinking, if you ask for k then 10 * k "misses" is a good value.

myuwono: We maybe can use lower values or even static value if we're resetting the iterations on successful sample. Another alternative is to make this configurable, i don't know how much value does that add 😓 .

sksamuel: I guess it can't be static - if you ask for 10,000 elements, and your arb produces 20,000 then you probably want to run for more than 20k, as 20k will likely have repeats. That's why I settled on an order of magnitude increase. It's not perfect, and could be configurable on another PR ?

sksamuel: Because we don't know how many elements the arb can provide, so we're in the blind as to how many times to try. 1000 seems reasonable on smaller sets, but will break on bigger sets (hence this issue).

myuwono: yeah that makes sense, i like slippage factor better tbh.

sksamuel: I've made slippage a parameter.

myuwono: champ! thanks @sksamuel

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/collections.kt
myuwono: oh are we not resetting the counter?

PR #2073, opened 2021-02-07
Updated non deterministic docs

sksamuel:


documentation/docs/assertions/eventually.md
```diff
@@ -0,0 +1,129 @@
...
+        userRepository.size() shouldBe 1
+      }
+    }
+  }
+}
+```
+
+#### Exceptions
+
+By default, `eventually` will ignore any exception that is thrown inside the function (note, that means it won't catch `Error`).
```
jschneidereit: I think this will only ignore `AssertionError`s thrown by default

sksamuel: Yes good catch.

PR #2076, opened 2021-02-08
use factor for exponential interval

### LINKED ISSUE #2075: ExponentialInterval should use a factor for its duration calculation
tl;dr: `ExponentialInterval` should use the function `base * factor^iter` instead of `base^iter` to produce a less severe curve and a more expected starting 0th value

---

In my last PR #2053 I added a cap to the exponential intervals due to the extremely steep nature of `ExponentialInterval`'s exponential curve.  Consider the interval `2.seconds.exponential(cap = Duration.INFINITE)` which currently will produce a curve of duration values that look like this:

```
0 : 1.00ms
1 : 2.00s
2 : 66.7m
3 : 92.6d
4 : 185185d
5 : 3.70e+8d
```

Here you can see how much of a bandage the "cap" fix in #2053 was - basically we allow one retry and then immediately hit the cap.

Instead of applying the power function to the base directly, I think what might be needed here is a factor to power and multiply the time value against.

```
0 - 2.00s
1 - 4.00s
2 - 8.00s
3 - 16.0s
4 - 32.0s
5 - 64.0s
```

Using a factor (my example uses 2) also has the benefit of having the first delay being what one expects, instead of a fraction of the specified delay (1ms vs 2s).

jschneidereit:
Fixes #2075

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/until/ExponentialInterval.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/until/ExponentialInterval.kt
@@ -3,20 +3,31 @@ package io.kotest.assertions.until
 import kotlin.math.pow
 import kotlin.time.Duration
 import kotlin.time.hours
-import kotlin.time.milliseconds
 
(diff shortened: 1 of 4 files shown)


sksamuel:
Failing test, but otherwise looks good.

Issue #2083, opened 2021-02-11
assertSoftly breaks eventually


xeruf:
**Which version of Kotest are you using**
4.4.1

```kt
@ExperimentalTime
class SimpleTest: StringSpec({
    "eventually a second passes" {
        val target = System.currentTimeMillis() + 1000
        eventually(2.seconds) {
            System.currentTimeMillis() shouldBeGreaterThan target
        }
    }
    "softly, no second passes eventually" {
        val target = System.currentTimeMillis() + 1000
        assertSoftly {
            eventually(2.seconds) {
                System.currentTimeMillis() shouldBeGreaterThan target
            }
        }
    }
})
```
both tests should pass, but the second doesn't: Due to the soft assert mode, the exception isn't propagated to any catch block but instead rethrown at some later point.

I think soft assertions should use some sort of scope that is overriden by the eventually instead of the special handling by the Matchers they get currently.

sksamuel:
You can move the assertSoftly inside the eventually and it will work.

And we can think of a long term solution.

xeruf:
I didn't even mean to use an `assertSoftly` there, it was a leftover that puzzled me for a while since I did not expect the respective test to fail.

sksamuel:
assertSoftly activates a flag that tells the matchers not to throw exceptions but to collect them. So eventually isn't seeing the exceptions to know to loop.

I think we could have eventually opt into this mechanism too so it knows to ignore that setting.

PR #2091, opened 2021-02-12
#2083 Update docs to highlight eventually does work properly with assert softly

ashishkujoy:


sksamuel:
We can also add a check inside eventually that collection mode is _not_ set to soft.

sksamuel:
Ah I see you just did that :)

PR #2094, opened 2021-02-12
Updates Throwable eq to check throwable cause as well while checking equality

ashishkujoy:


sksamuel (approved):

sksamuel:
failing test @ashishkujoy

PR #2096, opened 2021-02-13
[4.5] Arb composition through Arb.bind / Arb.flatMap should preserve edgecases

### LINKED ISSUE #2095: Arb composition through Arb.bind / Arb.flatMap should preserve edgecases
**Description**

When the edgecases in one of the arbs in bind / flatmap were empty, the resulting edgecases is empty.

```kotlin
val arbA = Arb.int().withEdgecases(1, 2, 3)
val arbB = Arb.string(1).withEdgecases(emptyList())

val result = Arb.bind(arbA, arbB) { a, b -> a to b }

result.edgecases() // empty list 
```

This is due to the cartesian product between any list with an empty list is an empty list. This is sometimes less than ideal especially if we need to test a system which depends on multiple inputs, each with its own set of edgecases. We need to find a way to make sure these edgecases are preserved in binds and flatMaps. 

The proposition is for each arbs that doesn't provide any edgecase, we should randomly generate a value for it which is conjured from `Arb<A>.sample(rs: RandomSource): A`. The final expectation is for edges to always exist (either provided / random), and that the cartesian product between them is guaranteed to be nonempty. i.e.

```kotlin
val arbA = Arb.int().withEdgecases(1, 2, 3)
val arbB = Arb.string(1).withEdgecases(emptyList())

val result = Arb.bind(arbA, arbB) { a, b -> a to b }

result.edges().all(RandomSource.seeded(1234L)) 
// [(1, "a"), (2, "%"), (3, "Q")]
```

4.4.1

myuwono:
fixes #2095


kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
```diff
@@ -68,6 +70,11 @@ abstract class Arb<out A> : Gen<A>() {
     */
    abstract fun edgecases(): List<A>
 
+   open fun edges(): EdgeCases<A> =
```
myuwono: i haven't annotated edgecases() as deprecated as i realized this `edges()` can theoretically coexists with the current `edgecases(): List<A>`, e.g. making it internal or some sort. what's the best course of action here @sksamuel ?

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/BindTest.kt
```diff
@@ -502,11 +532,392 @@ class BindTest : StringSpec({
...
+
+   "Arb.bind(a,b,c,d,e) should compute the cartesian product of edges" {
+      val arbA = Arb.string(1)
+      val arbB = Arb.string().withEdgecases("a", "b")
+      val arbC = Arb.string(1)
+      val arbD = Arb.string().withEdgecases("a", "b")
+      val arbE = Arb.string(1)
+      val edges = Arb.bind(arbA, arbB, arbC, arbD, arbE) { a, b, c, d, e -> a + b + c + d + e }.edges()
+      edges.values(RandomSource.seeded(1234L)) shouldContainExactlyInAnyOrder listOf(
+         "Uakaz", "%bcaP", "?aCb-", "KbGbw"
```
myuwono: 🎉

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
```diff
@@ -24,7 +26,7 @@ import io.kotest.property.arbitrary.of
 sealed class Gen<out A> {
 
    fun generate(rs: RandomSource): Sequence<Sample<A>> = when (this) {
-      is Arb -> this.edgecases().asSequence().map { Sample(it) } + this.samples(rs)
+      is Arb -> this.edges().values(rs).asSequence().map { Sample(it) } + this.samples(rs)
```
myuwono: here we call `.edges().values(rs)` providing a random source `rs` to seed those `EdgeCases.Random`

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
```diff
@@ -104,10 +129,15 @@ fun <A, B> Arb<A>.distinctBy(selector: (A) -> B) = object : Arb<A>() {
  */
 fun <A, B : A> Arb<A>.merge(other: Gen<B>): Arb<A> = object : Arb<A>() {
    override fun edgecases(): List<A> = when (other) {
-      is Arb -> this@merge.edgecases().zip(other.edgecases()).flatMap { listOf(it.first, it.second) }
```
myuwono, line 107: this is wrong. these edgecases need to be concatenated instead of zipped. I've fixed this too.

kotest-property/src/commonMain/kotlin/io/kotest/property/EdgeCases.kt
```diff
@@ -0,0 +1,30 @@
+package io.kotest.property
+
+import io.kotest.fp.NonEmptyList
+import io.kotest.fp.nel
+
+sealed class EdgeCases<out A> {
```
myuwono: Not sure if this is the right place for this or should this be in Gen.kt

sksamuel:
Quality code, but on reflection I'm starting to wonder if having edge cases defined as random is kind of inverting what edge cases are. An arb has edge cases + randoms, but now its edge cases that might be random + random. 

Would it be better to just handle this inside bind ? If a particular set of edge cases are empty, then ask for a single random value and use that? With the way you changed arb to generate single values, that should be very easy?

myuwono:
@sksamuel right, yeah that makes sense. we just have one caveat in this case. The main ingredient that I need is the random seed. Essentially we want to propagate this through to the edgecases. This is because essentially, the same random seed that is used to generate the samples will now be utilized to generate the edgecases to ensure repeatability. - of course static edges just simply one that generates the same list over any rs.

essentially, this is the absolute minimum that is required. lmk what you think

```kotlin
interface EdgeCases<out A> {
    fun values(rs: RandomSeed): NonEmptyList<A>
}
```

you can imagine static edges are just this, that generates a static list
```kotlin
val staticEdges = EdgeCases { NonEmptyList.of(1, 2, 3)  } 
```

sksamuel:
Hmmm yes. Going off on a tangent, should edgecases have a degree of
uncertainty be built in. Rather than being strictly hard coded, they could
have elements of random.

For example, for an integer arb, we hard code 0, that's static, maybe min
and max int, but another two edge cases would be a negative and positive
number, but doesn't matter what they are.
So kind of like arbs provide static, semi random, then random.

And the semi random is where the seed could come in, which bind could of
course use.

myuwono:
> but another two edge cases would be a negative and positive number, but doesn't matter what they are. So kind of like arbs provide static, semi random, then random.

Yeah that's exactly what i was thinking when i put together the implementation code. In a sense `EdgeCases.Random` is indeed semi-random when it goes through a process of binding (applicative or monadic, i.e. flatmap) with some static edges.

I'll ping you in slack to discuss further about the implementation detail.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
```diff
@@ -28,6 +31,17 @@ fun <A> Arb<A>.next(rs: RandomSource = RandomSource.Default): A = single(rs)
  */
 fun <A> Arb<A>.filter(predicate: (A) -> Boolean): Arb<A> = object : Arb<A>() {
    override fun edgecases(): List<A> = this@filter.edgecases().filter(predicate)
+
+   // sample the random edgecase generator 10 times and sift those values that satisfies the given predicate
+   // otherwise fallback to Arb's sample.
+   override fun edgecases(rs: RandomSource): List<A> =
+      generateSequence { this@filter.edgecases(rs) }
+         .take(10)
```
myuwono, line 39: I'm not convinced about this magic number. filter is kind of weird for this anyway, and iterating it multiple times is like expecting the random sample to somehow resolve something that is outside of the edgecases. maybe i should just take it once.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
```diff
@@ -55,6 +71,19 @@ fun <A, B> Arb<A>.map(f: (A) -> B): Arb<B> = object : Arb<B>() {
  */
 fun <A, B> Arb<A>.flatMap(f: (A) -> Arb<B>): Arb<B> = object : Arb<B>() {
    override fun edgecases(): List<B> = this@flatMap.edgecases().flatMap { f(it).edgecases() }
+
+   override fun edgecases(rs: RandomSource): List<B> =
+      NonEmptyList
```
myuwono, line 76: @sksamuel this is the primary machinery for bind, where it falls back to taking one random sample accordingly.

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
```diff
@@ -68,12 +68,17 @@ abstract class Arb<out A> : Gen<A>() {
     */
    abstract fun edgecases(): List<A>
```
sksamuel, line 69: Should we deprecate this version of edge cases ?

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
sksamuel: Once or twice maybe yeah.
I think once you start with filters, edge cases starts to break down anyway.

sksamuel (approved):

sksamuel:
@myuwono shall we wait to merge this because of #2112

myuwono:
yeah i'll decline this for now

PR #2099, opened 2021-02-14
Fix withEnvironment() to be case-insensitve on Windows

sschuberth:
On Windows, environment variables names are case-insensitive. The
OverrideMode in withEnvironment() did not take this into account before.
Instead of making OverrideMode itself conditionally ignore case, simply
let it operate on a map with case-insensitive keys on Windows.

Note that the original comment about using "toMap()" was bogus, as all
consumers of "originalEnvironment" anyway already take an immutable map as
their parameter.


sksamuel:
Nice thanks. Will be included in 4.4.2

sschuberth:
Sorry, I just had to do a fixup: https://github.com/kotest/kotest/pull/2100

PR #2100, opened 2021-02-14
SystemEnvironmentExtensions: Fix checking for Windows

sschuberth:
This is a fixup for 034ef9c.


sksamuel:
I thought the other one had gone green before I merged. My bad, thanks for the fix.

sschuberth:
> I thought the other one had gone green before I merged. My bad, thanks for the fix.

It actually was green, but this particular feature is not covered with a (Windows-specific) test yet...

sksamuel:
Ah right.

PR #2104, opened 2021-02-15
Merge ExpectSpec and ShouldSpec into FunSpec

### LINKED ISSUE #2093: Merge FunSpec and ShouldSpec?
Is there any architectural reason to not have a single style where both `test` and `should` can be used? I actually have tests where that could be convenient.

xeruf:
Fixes #2093

### DIFF
--- a/doc/styles.md
+++ b/doc/styles.md
@@ -6,15 +6,13 @@ Others were created just for Kotest.
 
 | Test Style | Inspired By |
 | --- | --- |
-| [Fun Spec](#fun-spec) | ScalaTest |
+| [Fun Spec](#fun-spec) | ScalaTest & more |
 | [Describe Spec](#describe-spec) | Javascript frameworks and RSpec |
-| [Should Spec](#should-spec) | A Kotest original |
 | [String Spec](#string-spec) | A Kotest original |
 | [Behavior Spec](#behavior-spec) | BDD frameworks |
 | [Free Spec](#free-spec) | ScalaTest |
 | [Word Spec](#word-spec) | ScalaTest |
 | [Feature Spec](#feature-spec) | Cucumber |
-| [Expect Spec](#expect-spec) | A Kotest original |
 | [Annotation Spec](#annotation-spec) | JUnit |
 
 There are no functional differences between the styles.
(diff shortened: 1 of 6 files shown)


sksamuel (commented):
Please un-deprecate should spec and expect spec.
If we merge this, we can't deprecate specs that other people have used since version 1.0.
Perhaps in 5.0 we can deprecate some specs.

We need to consider if we are just complicating further the already extensive list of specs. Do you think people will want to mix test / should / expect in the same file?

xeruf:
Well, this gets rid of a lot of duplication and simplifies choosing a spec - since there are less choices now :) I don't see how this complicates anything, quite the opposite.
and yes, depending on the wording, it might be helpful to use multiple in one. Either way, is there any advantage to separating them?

As for the deprecation - why? It is a simple find-replace operation to update these.

sksamuel:
I'm not _against_ merging them, I just need to be convinced that changing things is an improvement :) I certainly welcome the discussion.

Lots of people use kotest and if we deprecate some specs we're asking them to make name updates without clear justification why. I'm definitely against deprecations prior to 5.0 on something as fundamental as the spec names. We have to remember as well that in version 4.0 - that only came out last year - we renamed the project from kotlintest to kotest, and expected everyone to update every single package name. People have 1000s or 10000s of tests, and they just want it to work without needing to massage their tests every time there's an upgrade. There's also people who build on top of this functionality with their own extensions. So caution is key here.

I also think there's an elegance in should spec having should, expect spec having expect and so on. Otherwise where do you draw the line, just go and add all the keywords to a single spec.

xeruf:
> I also think there's an elegance in should spec having should, expect spec having expect and so on. Otherwise where do you draw the line, just go and add all the keywords to a single spec.

If it's implemented nicely, yes. But currently, there's a lot of annoying copy-paste between the specs. I could also imagine keeping ShouldSpec and ExpectSpec but having FunSpec somehow inherit from them.
In the end it would indeed be possible to allow to mix and match all specs, and maybe we can find some clean way for that without convoluting everything.

sksamuel:
The copy n paste is annoying if you're maintaining the spec classes, but it's only kotest doing that and they barely ever get touched. So I wouldn't cite that as a big reason to change it up.

PR #2105, opened 2021-02-16
Run eventually one final time if iterations == 1 and delay > interval

### LINKED ISSUE #2089: Eventually sometimes only checks a single time despite short  scheduled intervals
4.4.1

```java
java.lang.AssertionError: Eventually block failed after 200ms; attempted 1 time(s); FibonacciInterval(base=20.0ms, offset=0, max=120m) delay between attempts
```
I regularly encountered this now. Per se it isn't a big problem if the intervals aren't followed exactly, if the tests keep my processors busy it doesn't really matter. However, the eventually should definitely do at least a second check at the end of its interval.

sksamuel:
Fixes #2089 

@xerus2000 @jschneidereit


xeruf (approved):
Haven't tested it in practice, but it looks good :+1:

PR #2108, opened 2021-02-17
[4.5] #2071 Adds an extension for wiremock

ashishkujoy:
closes: #2071

Issue #2071, opened 2021-02-07
Wiremock Integration


sksamuel:
https://github.com/tomakehurst/wiremock

LeoColman:
Probably better to use a listener

pinwheel101:
@LeoColman Could you tell me why listener can be a better option compare to extension. Actually I cannot use kotest becuase of the absence of @ClassRule annotation so I don't know how can I use wiremock in kotest.

If someone give me an example how to use Wiremock with kotest, it would be really helpful to use kotest

pinwheel101:
@sksamuel I am looking foward to release the wiremock extension 👍

sksamuel:
> @LeoColman Could you tell me why listener can be a better option compare to extension. Actually I cannot use kotest becuase of the absence of @ClassRule annotation so I don't know how can I use wiremock in kotest.

A TestListener in kotest is read only, a TestExtension can modify tests.
This particular "extension" only needs to listen to test lifecycle events, so @LeoColman is write, I should have said listener.

pinwheel101:
@sksamuel Thank you for the explanation. I understand now. Listener means TestLinstner in Kotest :)

PR #2109, opened 2021-02-17
#2071 Adds docs for wiremock integration [Merge after releasing #2108]

ashishkujoy:


sksamuel:
Merging this in because wiremock is now in it's own it's own repo and released.

Issue #2114, opened 2021-02-20
Add a factory function to Matcher to simplify writing matchers


dalewking:
There is a repetitive boilerplate when writing matchers. For example consider this one core matcher on Int:

    fun beLessThanOrEqualTo(x: Int) = object : Matcher<Int> {
      override fun test(value: Int) = MatcherResult(value <= x, "$value should be <= $x", "$value should not be <= $x")
    }
    
There are several bits of boilerplate in there that can be eliminated. What I propose is adding this to the Matcher interface:

    companion object {
        operator fun <T> invoke(tester: (T) -> MatcherResult) = object: Matcher<T> {
            override fun test(value: T) = tester(value)
        }
    }

This would allow the matcher above to be simplified to this:

    fun beLessThanOrEqualTo(x: Int) = Matcher<Int> {
        MatcherResult(it <= x, "$it should be <= $x", "$it should not be <= $x")
    }

You could also go a step further and add this to the companion object

        operator fun <T> invoke(tester: (T) -> Boolean, failure: (T) -> String, negatedFailure: (T) -> String) =
           Matcher<T> { MatcherResult(tester(it), failure, negatedFailure) }
        }

This would  allow simplifying further to just:

    fun beLessThanOrEqualTo(x: Int) =
        Matcher<Int>({ it <= x }, { "$it should be <= $x" }, { "$it should not be <= $x" })

Although if one were redesigning the interface entirely, Matcher<T> should really just be an alias for (T) -> MatcherResult and the other interface functions would be extension functions

dalewking:
Actually now that Matcher has a companion object you could use an extension function for this. Here is what I added to our codebase:

    /**
     * Convenience function for creating custom matchers
     * @param tester Function that generates a MatcherResult for a value being tested
     */
    inline operator fun <T> Matcher.Companion.invoke(crossinline tester: (T) -> MatcherResult) =
        object : Matcher<T> {
            override fun test(value: T) = tester(value)
        }

But it would be better to have in the companion object as using an extension function requires a separate import of the invoke function.

sksamuel:
I like the idea of adding the invoke to the companion object @dalewking . Do you want PR that up?

PR #2122, opened 2021-02-22
Add factory function to simplify matchers

dalewking:
Addresses #2114.

I started to look at converting the existing matchers to use the factory function but I didn't want to take on fixing over a hundred instances


sksamuel:
👍🏼

PR #2126, opened 2021-02-23
[4.5] implement probabilistic edgecases

### LINKED ISSUE #2112: Probabilistically use edge cases with random data instead of a cartesian product of all edge cases
Current status: Given 2 arbs with 5 edge cases each, a test that uses these two arbs would require 25 (5*5) iterations before going into testing random values. This scales poorly with a large amount of arbs, or when combining arbs. For example - if we instead had 4 of these arbs, we'd need a minimum of 625 iterations. 6 of these arbs = 15,000. 8 = almost 400,000.

This makes using complex objects or a objects with a large amount of edgecases infeasible. A potential solution to this would be to probabilistically pull in edgecases, rather than exhaustively. For example - whenever generating a value, pull out an edgecase 10% of the time and a random value 90% of the time.

This would achieve the task of checking the more problematic values more frequently, and testing combinations of these edge cases more frequently, without having the iteration count for tests skyrocket.

Additional context on the kotlinlang slack with @sksamuel and @myuwono : https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1613715625006600

myuwono:
fixes #2112 and #2095

there's a few material behavioural changes that would require your review @sksamuel. 
- With the advent of probabilistic edgecases for arbs, the minimum iterations for arbs no longer makes sense. 
- I've added an extra config to `PropTestConfig` which allows users to modify edgecase generation probability. this is also configurable via `kotest.proptest.arb.edgecases-generation-probability`. the default `edgecaseProbability` is 0.1 (i.e. 10% edgecases, 90% samples) as discussed.
- There's a determinism factor introduced in bind which is 0.9 by default. I've made that configurable, but that might also be an overkill. Happy to hardcode that to 0.9 if you don't think that's necessary.
- I've introduced `generateEdgecase(rs): A`. I'm wondering if i should / should not deprecate the `edgecases()`. I'm thinking that we should, because essentially the builders already exposes a way for users to do the same thing.

There's quite a lot in here, so please take your time and not rush this. I'm keen to make careful stride with this PR.


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
```diff
@@ -43,18 +50,26 @@ fun <A> Arb<A>.filterNot(f: (A) -> Boolean): Arb<A> = filter { !f(it) }
...
+   override fun generateEdgecase(rs: RandomSource): B = f(this@map.generateEdgecase(rs))
+
    override fun values(rs: RandomSource): Sequence<Sample<B>> {
       return this@map.values(rs).map { Sample(f(it.value), it.shrinks.map(f)) }
    }
 
-   override fun sample(rs: RandomSource): Sample<B> = Sample(f(this@map.sample(rs).value))
+   override fun sample(rs: RandomSource): Sample<B> =
+      this@map.sample(rs).let {
+         Sample(f(it.value), it.shrinks.map(f))
```
myuwono, line 70: that looked like a bug, map previously discarded the shrinker

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
```diff
@@ -66,8 +79,18 @@ abstract class Arb<out A> : Gen<A>() {
    /**
     * Edgecase values for this type A.
     */
+   @Deprecated("implement one value at a time using generateEdgecase(rs) or use Arb<T>.withEdgecases(...) instead")
    abstract fun edgecases(): List<A>
 
+   /**
+    * Returns a random edgecase value using the supplied random source
+    */
+   open fun generateEdgecase(rs: RandomSource): A =
```
sksamuel: I'm not so sure this should be a non empty list.
With this implementation, if you ask for an edge case and the arb doesn't provide any, we return a random value, which is perhaps misleading given the name of the function?.
Perhaps this can just return null and then the generate functionality can then defer to the normal samples method.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/arbs.kt
sksamuel: nice spot.

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
myuwono: Awesome, yeah totally agree that edgecases can be empty as well. Before we go full throttle with nulls, I wonder if we can find a better type to model that behaviour?

Looking at `generateEdgecases(rs): A?` Currently, as a library user I don't know if I'd necessarily agree with the design or if that's less or more confusing. Firstly, unlike option null doesn't nest so we can't really encode e.g. edgecases of `Arb<A?>`. Secondly if we have `Arb<Option<A>>` that is fine but you'd get `Option<A>?` in the edgecases type..? 

if we analyze this further, I'd argue the type signature of `generateEdgecases(rs): A?` is ambiguous. It doesn't really imply empty edgecases per se. Rather, it suggests that the edgecase may be null in one iteration and not null in another. I'd personally not put such design if i were given a choice. I'd pursue something like `Edgecases<A>` which is a sealed class and it may either be empty, or probabilistic. I.e `edges(): Edgecases<A>`.

```kotlin
sealed class Edgecases<A> {
    object Empty : Edgecases<Nothing>()
    data class Probabilistic(generate: (RandomSource) -> A) : Edgecases<A>()
}
```

This way users of kotest would know straight up with no ambiguity. And clearly there's no way to generate anything from an Empty edgecases.

In terms of building edgecases we might do something like

```kotlin
edgecasesOf("foo", "bar")

// or
edgecases { rs -> ... }

// or
emptyEdgecases()
```

And that will yield the correct type under the hood.

Out of curiosity, i can't remember why we decided not to pursue the approach of having a seperate sealed class?

sksamuel:
I'm totally with you on Options vs nulls, but most people in the Kotlin
ecosystem are not coming from an FP background like us. I don't think it's
Kotest's mission to evangelize FP, even though I use it internally and
encourage other contributors to. In public APIs, it must be nulls
especially as that's how the Kotlin stdlib does it. It's different in Scala
or Haskell where standard apis are using options too.

We could use an ADT but boils down to the same thing. If you're asking for
a single value, ie, give me an edge case if you have one, the familiar way
in Kotlin is null or value.

I'm trying to think if this can be improved. The whole thing driving this
is bind. Originally we had edgecases vs samples. Now we really just have
"values" which are sometimes samples and sometimes an edgecase and
edgecases did not have access to a randomsource so that was introduced.
Just thinking out loud without working this through, if the generate method
itself was overridden by bind, then it would have the RS in order to
generate a sample for the components that didn't have edgecases...

myuwono:
@sksamuel yeah that makes sense. I'm totally with you in terms of the external facing interface should use nulls. 

I've experimented with overriding generate in my earlier experiment but in my opinion that would yield a different problem i.e. https://github.com/kotest/kotest/issues/2112#issuecomment-782745397 - i'll go take another look if that would work with the new addition of edgecase probability.

My suggestion was around using the ADT instead of exposing any internals and having these data classes hide most of the intricate details. It might pay dividends as it's more extendable than nulls.

```kotlin
sealed class Edgecases<A> {
    object Empty : Edgecases<Nothing>()
    data class Probabilistic(generate: (RandomSource) -> A) : Edgecases<A>()
}
```

and
```kotlin
edgecasesOf("foo", "bar")

// or
edgecases { rs -> ... }

// or
emptyEdgecases()
```

However if you have a strong feeling against it, let's go with `generateEdgecase(rs): A?` instead (or maybe overriding generate if i can get it to work).

sksamuel:
My strong opinions are on:
- nulls > options in public apis
- it should be easy to add a fixed list of edge cases when implementing an
Arb without requiring ADTs

If generate edge cases is some internal thing, then we can use whatever
want - it could move out to an extension method so it's clear it's not in
user land.

kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt
```diff
@@ -68,6 +84,11 @@ abstract class Arb<out A> : Gen<A>() {
     */
    abstract fun edgecases(): List<A>
 
+   internal open fun edges(): Option<Edgecase<A>> =
```
myuwono, line 87: @sksamuel this is now an internal function as recommended; i've decoupled all the edgecases related machineries to the data class so that all the configurations can be injected last. The user-facing api should stay the same.

renpillay:
@sksamuel What do you think of this PR? Would like to see this released as it would unblock us from using edgecases.

sksamuel:
Sorry just been super busy at work last 2 weeks. I'll get to this as soon
as I can.

PR #2128, opened 2021-02-24
Fix IncorrectDereferenceException when calling assertions on backgrond thread on native

ajalt:
This commit adds a test that runs basic assertions on a background thread on kotlin native. Prior to this commit, that test would fail because not all top-level `val`s are annotated with `@SharedImmutable` or `@ThreadLocal`. I annotated all top-level `val`s I found in `dektopMain` sourceSets, which fixes the errors.


sksamuel:
Excellent find thank you.

Issue #2011, opened 2021-01-21
Generate HTML reports from JUnit XML


caramelomartins:
When using JUnit XML, we can generate XML results from tests that are able to produce output with nested tests. If we use Kotest/JUnit XML with Gradle, then we can disable the XML results from Gradle and keep _only_  the results from JUnit XML. Unfortunately, Gradle generates its HTML reports with the results it has in-memory, which can't don't support nested tests, and it doesn't seem to be able to fetch results from a different XML. It has a `TestReport` task that is able to generate HTML reports from binary results, supposedly, but not from XML results.

With this in mind, I believe it would be interesting to offer a listener, similar to what JUnit XML does, that would generate HTML reports based on XML results produced by JUnit XML. In this way, we could have custom HTML reports for Kotest that would support nested tests based on the XML generated by JUnit XML.

sksamuel:
Would you like to make a start on it, and then we can all contribute to improving it ?

caramelomartins:
Hey @sksamuel. Yes, I'd like to make a start of it (at least the HTML, even if it doesn't have much styling at first). I'm going to try to use some spare time this week on this. :crossed_fingers:

sksamuel:
Awesome.

caramelomartins:
Hey! Just wanted to get back to you on this. I've got some minor progress:

![image](https://user-images.githubusercontent.com/2028058/108105192-36a97000-7084-11eb-96b3-e5fc67d2bdaa.png)

Not pretty or particularly useful yet but it does produce something. :smile:

caramelomartins:
We can now go from having a summary of all the test classes to seeing tests in each of the test classes. Still without any styles.

![image](https://user-images.githubusercontent.com/2028058/108606713-899c6380-73b3-11eb-9f90-c5e77ffe959b.png)

![image](https://user-images.githubusercontent.com/2028058/108606727-a3d64180-73b3-11eb-8b3b-2974cdacb89c.png)

I've noticed something odd which is that none of the test results have any duration in the XML of `JunitXmlReporter`. Is this expected?

sksamuel:
Looking good so far.

No, the durations should be set, assuming they take longer than 1ms

sksamuel:
We could expand the junit XML too, since the format is kind of limiting.

caramelomartins:
> We could expand the junit XML too, since the format is kind of limiting.

[...]

PR #2130, opened 2021-02-27
Add HTMLReporter to Generate HTML Based on XML 

caramelomartins:
This PR has the aim of tackling #2011 and adds a listener named `HTMLReporter` which will produce HTML reports based on the XML generated by the `JunitXmlReporter`. It has the following results:

image

image

It currently generates some very basic HTML reports but they have the essential functionality of showing the test classes that have failed and the tests that have failed. 

The tests still need some working on but I thought it would be important to open this PR early to get some early feedback because it is getting quite big, due mostly to the fact that I had to add an entire module containing the `HTMLReporter`.


kotest-extensions/kotest-extensions-htmlreporter/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
+plugins {
+   id("java")
+   kotlin("multiplatform")
+   id("java-library")
+   id("org.jetbrains.kotlin.plugin.spring") version Libs.kotlinVersion
```
sksamuel: Is spring needed ?

kotest-extensions/kotest-extensions-htmlreporter/src/jvmMain/kotlin/io.kotest.extensions.htmlreporter/HtmlReporter.kt
```diff
@@ -0,0 +1,199 @@
...
+import org.jdom2.input.SAXBuilder
+import org.jdom2.output.Format
+import org.jdom2.output.XMLOutputter
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+import java.nio.file.StandardOpenOption
+
+class HtmlReporter(
```
sksamuel, line 14: The project listener that locates the test data, and the class that generates HTML could be separated since they are different things. The former could invoke the latter.

sksamuel (changes requested):

kotest-extensions/kotest-extensions-htmlreporter/build.gradle.kts
caramelomartins: I don't believe so, this must be a typo. I'll check it and remove it.

kotest-extensions/kotest-extensions-htmlreporter/src/jvmMain/kotlin/io.kotest.extensions.htmlreporter/HtmlReporter.kt
caramelomartins: I don't mind that. What would we call these? `HtmlReporter` and `HtmlGenerator`?

sksamuel: HtmlWriter maybe for the thing that does the html generator and HtmlReporter yeah for the listener class.

caramelomartins:
@sksamuel I've refactored this code a bit. I moved the logic to generate the HTML to a separate class called `HtmlWriter`. I've also improve the code's readability by creating some data classes that we can use to group the data for each of the test classes and then we can call `HtmlWriter` based on that, instead of passing around XML elements. I believe this merits some tests but I'm having trouble thinking up of ways to write tests for HTML code produce...it won't be trivial. Let me know what you think about how it is now, thank you!

sksamuel (approved):

caramelomartins:
Thanks for the approve @sksamuel. Are you happy if I merge this?

sksamuel:
Yep.

caramelomartins:
I cannot do that, I'm sorry, because I don't seem to have permissions to merge PRs in Kotest's repository.

sksamuel:
Yes of course, I'm thinking I'm at work :)

PR #2131, opened 2021-02-28
haveCauseOfType shows exception type instead of cause type

jvmusin:
`haveCauseOfType` shows the type of the exception itself, but not the type of the `cause` exception.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/throwable/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/throwable/matchers.kt
@@ -35,8 +35,8 @@ inline fun <reified T : Throwable> haveCauseInstanceOf() = object : Matcher<Thro
     null -> resultForThrowable(null)
     else -> MatcherResult(
       cause is T,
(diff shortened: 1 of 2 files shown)


ashishkujoy:
@jvmusin Thanks for the fix, can you please also add a test for the same.

jvmusin:
@ashishkujoy Yes sure

jvmusin:
@ashishkujoy Please checkout the changes.

I've notices some more mistakes in the messages and added tests. Not sure if the tests are written well, tell me if something's wrong, I'll try to fix that.

jvmusin:
I also think that we should provide a bit more detailed descriptions for `haveCauseInstanceOf` matcher like these:
Throwable cause should be of type `class name` ***or it's descendant***, but instead got `class name`.
Throwable cause should not be of type `class name` ***or it's descendant***.

I can add them.

ashishkujoy (approved):

ashishkujoy:
> I also think that we should provide a bit more detailed descriptions for `haveCauseInstanceOf` matcher like these:
> Throwable cause should be of type `class name` _**or it's descendant**_, but instead got `class name`.
> Throwable cause should not be of type `class name` _**or it's descendant**_.
> 
> I can add them.

Sure

jvmusin:
Done

PR #2133, opened 2021-03-01
Fix typo in haveClassAnnotations matcher

jvmusin:
The new method with a correct name now does the job. The one with a typo marked as deprecated.
The old method also has ReplaceWith in it's Deprecated annotation to let users replace the old method with a new one using alt+enter (and one more enter).


sksamuel:
👍🏼

PR #2135, opened 2021-03-02
Fixes Arb.bigdecimal getting hanged for min and max value which other than Double min and max

### LINKED ISSUE #2132: Improve BigDecimal arbs
Currently the BigDecimal arb is quite innefficient when we try to to create BigDecimals between certain values.

For example:

```kotlin
Arb.bigDecimal(BigDecimal.valueOf(-100_000.00), BigDecimal.valueOf(100_000.00)).take(100).forEach {
    println(it)
}
```

This didn't ever print a single number, and the test never completed.

ashishkujoy:
Earlier we were using `Arb.double` under to generate BigDecimal but since `Arb.double` generate only values between `Double.MAX_VALUE` and `Double.MIN_VALUE`, the `Arb.bigDecimal` was getting hanged if the supplied rannge is beyond  `Double.MAX_VALUE` and `Double.MIN_VALUE`.

I have no experience in data driven test, so not sure how the Shrinker will work for BigDecimal.

closes: #2132

### DIFF
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/bigdecimal.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/bigdecimal.kt
@@ -14,7 +14,11 @@ private val bigDecimalEdgecases = listOf(
 
 fun Arb.Companion.bigDecimal(): Arb<BigDecimal> {
    return arbitrary(bigDecimalEdgecases) {
-      BigDecimal(it.random.nextInt() * it.random.nextDouble())
+      if (it.random.nextInt() % 2 == 0) {
+         BigDecimal(it.random.nextLong()) * BigDecimal(it.random.nextDouble())
+      } else {
+         BigDecimal(it.random.nextInt()) * BigDecimal(it.random.nextDouble())
+      }
    }
 }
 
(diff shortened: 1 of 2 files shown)


kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/bigdecimal.kt
```diff
@@ -4,9 +4,22 @@ import io.kotest.property.Arb
...
+   BigDecimal(1.0),
+   BigDecimal(-1.0),
+   BigDecimal("1e-300"),
+   BigDecimal("-1e-300"),
+)
 
-fun Arb.Companion.bigDecimal(scale: Int, roundingMode: RoundingMode) = bigDecimal().map { it.setScale(scale, roundingMode) }
+fun Arb.Companion.bigDecimal(): Arb<BigDecimal> {
+   return arbitrary(bigDecimalEdgecases) {
+      BigDecimal(it.random.nextInt() * it.random.nextDouble())
```
sksamuel: Perhaps we should do BigDecimal(it.random.nextDouble()).multiply(BigDecimal(it.random.nextLong()) to allow for actual big decimals.

ashishkujoy: Doing this way is always produce quite a larger digit BigDecimal `5494701162849313711.73708734` to get smaller BigDecimal can combine both nextInt and nextLong.

```kotlin
fun Arb.Companion.bigDecimal(): Arb<BigDecimal> {
   return arbitrary(bigDecimalEdgecases) {
      if (it.random.nextInt() % 2 == 0) {
         BigDecimal(it.random.nextLong()) * BigDecimal(it.random.nextDouble()) 
      } else {
         BigDecimal(it.random.nextInt()) * BigDecimal(it.random.nextDouble()) 
      }
   }
}
```

sksamuel:
Ok lgtm

Issue #2065, opened 2021-02-06
Expanding config - style discussion


sksamuel:
As part of #2049 #2050 #1370 I want to expand config to parent tests.

Currently, we add config to a test like this:

```
test("this test has config").config(timeout = 412.milliseconds, enabled = true) {
  // test here
}
```

We don't support this syntax on parent tests. So our options are:

1. Expand same style to parents:

```kotlin
  context("this context should support config").config(timeout = 65.milliseconds, tags = setOf(linuxTag)) {
     test("this test inherits config") {
     }
  }
```

The pro is that it works like it does currently, the con is that not all specs can support this (eg word spec) due to limitations in kotlin's DSL support and we can't share config.

2. Introduce config objects that tests can use as a receiver:

```kotlin
      val config = TestConfig(timeout = 412.milliseconds, tags = setOf(linuxTag))
      config.context("this context is defined on the config object") {
         test("this test inherits config") {
         }
         config2.test("this test uses another config receiver that would override the first") { 
         }
      }
```

The big pro here is that config can be defined and then re-used and the spec level config looks the same. The con is that it perhaps looks a bit uglier and harder to read with the config coming first (??)

3.  Similar to two, but define config as lambas:

```kotlin
      config(timeout = 412.milliseconds, tags = setOf(linuxTag)) {
         context("this context gets its config from the surrounding closure") {
            test("this test inherits config") {
            }
            config(invocations = 4) {
              test("this test overrides some config") {
              }            
            }
         }
      }
```

I think this looks nice and is clear when nesting and overriding, but you couldn't define it and reuse it (well you could make a custom function yourself).

4. Your proposal here...

ghost:
Of the three options you mentioned I like option 2 best due to the reusability.

I don't know if it's possible but could what about combining the reusable config with the test declaration.
This might be more confusing if it's even possible but could address your concern about the config tag being unclear that's it's a test.

Could look something like this:
```      
val shortLinuxTest = TestConfig(timeout = 412.milliseconds, tags = setOf(linuxTag))
    shortLinuxTest.context("this context is defined on the config object") {
         test("this test inherits config") {
         }
         LongLinuxTest("this test uses another config receiver that would override the first") { 
         }
      }
```

jschneidereit:
I'm partial to re-usability (see the controversial `EventuallyConfig`) but I think we might be able to get away with the object as well as overriding invoke for the lambda style as well. I can play around with it tonight.

sksamuel:
[...]

PR #2155, opened 2021-03-07
Add container level config for should and fun specs #2065

sksamuel:


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ShouldSpecContextScope.kt
```diff
@@ -43,6 +45,26 @@ class ShouldSpecContextScope(
...
+      ShouldSpecContextConfigBuilder(
+         createTestName(name),
+         description,
+         testContext,
+         defaultConfig.toTestContainerConfig(),
+         lifecycle,
+         false
+      )
+
+   fun xcontext(name: String) =
```
ashishkujoy, line 71: We should add a `suspend` version of `xcontext` here as well, like have added in `FunSpecContextScope`.

sksamuel: Thanks, done.

sksamuel:
Superseded by https://github.com/kotest/kotest/pull/2258

PR #2158, opened 2021-03-08
Allow tags to come from env as well as sysprop #2098

### LINKED ISSUE #2098: kotest.tags should come from properties or environment
Currently it only comes from properties

sksamuel:
Fixes #2098

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/KotestEngineProperties.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/KotestEngineProperties.kt
@@ -1,6 +1,9 @@
 package io.kotest.core.internal
 
-object KotestEngineSystemProperties {
+@Deprecated("Renamed to KotestEngineProperties. Will be replaced in 4.7")
+val KotestEngineSystemProperties = KotestEngineProperties
+
+object KotestEngineProperties {
 
    const val springIgnoreWarning = "kotest.listener.spring.ignore.warning"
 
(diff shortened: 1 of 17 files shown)


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/KotestEngineProperties.kt
```diff
@@ -1,6 +1,9 @@
 package io.kotest.core.internal
 
-object KotestEngineSystemProperties {
+@Deprecated("Renamed to KotestEngineProperties. Will be replaced in 4.7")
+val KotestEngineSystemProperties = KotestEngineProperties
```
LeoColman, line 4: I suppose this could be a type alias

LeoColman (approved):

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/KotestEngineProperties.kt
sksamuel: Yes I guess so.

jschneidereit (approved):

ashishkujoy:
Build gradle file of `kotest-framework-api` need to be fixed. 
Currently it is failing with.
```bash
* What went wrong:
Could not create task ':kotest-framework:kotest-framework-api:buildConfigDocs'.
> Fail to find files -> {[KotestEngineSystemProperties.kt]} in project, found only these files -> {[]}
```

Will need to correct this line, to `KotestEngineProperties`
https://github.com/kotest/kotest/blob/b640d3e2cec257eb318a886ba89442489ff1cb56/kotest-framework/kotest-framework-api/build.gradle.kts#L102

PR #2177, opened 2021-03-22
Limit Spec concurrency properly

hktechn0:
This patch fixes unlimited concurrency of Spec execution.
`DefaultSpecLauncher` should limit concurrency of coroutines execution up to `maxConcurrent` (from `concurrentSpecs` or `parallelism` in configuration).
But concurrency is not limited properly in current implementation because `launch` of coroutines is non-blocking.


sksamuel:
Great shout thanks

PR #2181, opened 2021-03-25
Update Exhaustive function names

seanf:
Related to https://github.com/kotest/kotest/pull/2110

https://kotest.io/docs/proptest/property-test-generators-list.html


seanf:
It seems like the file `doc/generators.md` (from #2110) might not be referenced. It seems to be a duplicate of `docs/proptest/genslist.md`. Maybe the two files need to be merged?

sksamuel:
The old doc/ files can just be updated to point to the micro site in docs/

PR #2183, opened 2021-03-27
Change `beEqualToIgnoringFields` api to avoid runtime crashes

### LINKED ISSUE #2179: Suggestion: change `beEqualToIgnoringFields` signature to avoid runtime crashes
### Context
I wanted to compare object eqaulity reflectively, as the object I wanted to compare [didn't have all properties included in equals call](https://github.com/pinterest/ktlint/commit/270584dc189a0babe6b527b9584cf44d9528e1ca#diff-0b1693fe6cb76775c846585ebcb4c1122eef31ff105a5a409dc1895d7a191061R16). I found Kotest exposes `beEqualToUsingFields` and `beEqualToIgnoringFields`, but looking at their signatures they both require passing a `vararg` arguments. Maybe I initially derped (and obviously didn't read docs), but I thought to avoid passing all properties to default method I should use a negation and use `beEqualToIgnoringFields()` with an empty properties list. It crashed in runtime 💥 
Only then I discovered that `beEqualToUsingFields` behaves differently and does actually accept empty list and is the method I was looking for.

### Suggestion
I'd like to suggest to change `beEqualToIgnoringFields` signature (and ofc all other related matches and functions) to something similar to: 
```
fun <T : Any> beEqualToIgnoringFields(
   other: T,
   ignorePrivateFields: Boolean,
   val property: KProperty<*>,
   vararg other: KProperty<*>
): Matcher<T>
```
to reflect it is required to pass at least one ignored property, which would result in a compilation error instead of runtime one.

mateuszkwiecinski:
Fixes #2179
If I'm not mistaken that break backwards compatibility as I'm changing public api. So if someone used named parameters - it would fail

I also improved related `beEqualToUsingFields` documentation as it is allowed not to pass any `properties`

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/equality/ReflectionKtTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/equality/ReflectionKtTest.kt
@@ -54,12 +54,6 @@ class ReflectionKtTest : FunSpec() {
          }.message shouldBe "Foo(a=sammy, b=13, c=true) should be equal to Foo(a=stef, b=13, c=false) ignoring fields [c]; Failed for [a: sammy != stef]"
       }
 
-      test("shouldBeEqualToIgnoringFields should throw exception when no field is mentioned") {
-         assertThrows<IllegalArgumentException>("At-least one field is required to be mentioned to be ignore for checking the equality") {
-            Foo("sammy", 23, false).shouldBeEqualToIgnoringFields(Foo("danny", 23, false))
-         }
-      }
-
       test("shouldBeEqualToIgnoringFields should compare equality for class having private fields") {
          val car1 = Car("C1", 10000, 430)
          val car2 = Car("C1", 123423, 123)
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/equality/ReflectionKtTest.kt
```diff
@@ -54,12 +54,6 @@ class ReflectionKtTest : FunSpec() {
          }.message shouldBe "Foo(a=sammy, b=13, c=true) should be equal to Foo(a=stef, b=13, c=false) ignoring fields [c]; Failed for [a: sammy != stef]"
       }
 
-      test("shouldBeEqualToIgnoringFields should throw exception when no field is mentioned") {
-         assertThrows<IllegalArgumentException>("At-least one field is required to be mentioned to be ignore for checking the equality") {
-            Foo("sammy", 23, false).shouldBeEqualToIgnoringFields(Foo("danny", 23, false))
-         }
-      }
-
```
mateuszkwiecinski, line 62: I removed the test as it simply doesn't compile anymore, which is an expected behavior 👌

ashishkujoy (approved):
Looks good to me.

sksamuel:
I think the breaking change is small enough that it's ok for 4.5
Can we PR to the changelog to make a note that if you're using named params on this method you will need to change the param name.

ashishkujoy:
> I think the breaking change is small enough that it's ok for 4.5
> Can we PR to the changelog to make a note that if you're using named params on this method you will need to change the param name.

`properties`  was earlier varargs , and since kotlin does not support named param for varargs so i think there won't be any breaking change in user codebase.

ashishkujoy:
We should definitely highlight that in changelog

PR #2186, opened 2021-04-01
Some options for better "or" mechanism in matchers

### LINKED ISSUE #1950: Add better mechanism for "or" in matchers

jschneidereit:
Fixes #1950

I couldn't use `or` since that's a bitwise infix fun in the standard library


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/shouldBeEither.kt
```diff
@@ -0,0 +1,27 @@
...
+   val l = Try { this shouldBe either.first }
+   val r = Try { this shouldBe either.second }
+
+   errorCollector.setCollectionMode(previous)
+
+   if (l.isFailure() && r.isFailure()) {
+      val combined = MultiAssertionError(listOfNotNull(l.errorOrNull(), r.errorOrNull()))
+      errorCollector.collectOrThrow(combined)
+   } else {
+      errorCollector.clear()
```
jschneidereit: There's a bug here, make sure we store the existing errors if they're using assert softly before returning

sksamuel:
This looks good but is limited to only a or b. And it doesn't allow things like.

(a shouldBe 2) or (b shouldBe 3)

Throwing some bike shedding out there, what about something like

```kotlin
     any {
      person.name shouldBe "sam",
      person.age shouldBe 41,
      person.city shouldBe "Chicago"
   }
```

Now any would have to work in the same way as assert softly to capture the errors and only throw at the end.
In other words, any { } is like assertSoftly but only fails if ALL the matchers fail.

jschneidereit:
That sounds good @sksamuel do you want me to do the `(a shouldBe 2) or (b shouldBe 3)` as well as `any`?

sksamuel:
Up to you. I think the any { } thing is a bit clearer, but what do you think ?

sksamuel:
You could even go full hoss and do some { } as well.

jschneidereit:
I love this plan, doing it

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/errorScope.kt
```diff
@@ -0,0 +1,33 @@
+package io.kotest.assertions
+
+import io.kotest.common.ExperimentalKotest
+import io.kotest.fp.Try
+
+internal typealias Failures = List<Throwable>
+internal typealias Assertions = Int
+
+@ExperimentalKotest
+suspend fun <T> errorScope(block: suspend () -> T): Triple<Try<T>, Failures, Assertions> {
```
sksamuel: This is not clear if you were to read it new. I think its worth really commenting on what's going on.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/errorScope.kt
```diff
@@ -0,0 +1,33 @@
...
+   errorCollector.setCollectionMode(originalMode)
+   val resultFailures = errorCollector.getAndReplace(originalFailures)
+   val resultAssertions = assertionCounter.getAndReset()
+
+   repeat(originalAssertions) { assertionCounter.inc() }
+
+   return Triple(result, resultFailures, resultAssertions)
+}
+
+fun ErrorCollector.pushErrorAndMaybeThrow(error: Throwable) {
```
sksamuel: same

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/errorScope.kt
sksamuel: Is error scope the right word? Like assertionsScope ?
It's counting assertions as well, not just errors.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/ErrorCollector.kt
```diff
@@ -78,6 +78,15 @@ open class BasicErrorCollector : ErrorCollector {
    override fun clear() = failures.clear()
 }
 
+internal fun ErrorCollector.getAndReplace(errors: Collection<Throwable>): List<Throwable> {
```
sksamuel, line 81: Can you add comments explaining what this is doing.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/either.kt
```diff
@@ -0,0 +1,49 @@
...
+ * ```
+ *   either {
+ *      "foo" shouldBe "bar"
+ *      "foo" shouldBe "foo
+ *      "foo" shouldBe "baz"
+ *   }
+ * ```
+ */
+@ExperimentalKotest
+suspend fun <T> either(assertions: suspend () -> T): T? {
```
sksamuel: In hindsight this nomenclature might clash with arrow / functional programming.
What about calling this "one"

sksamuel (changes requested):
One rename, one comment, and other that its good to merge!

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/either.kt
sksamuel: I've updated this - let me know on slack if you disagree @jschneidereit

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/one.kt
```diff
@@ -0,0 +1,52 @@
...
+   errorCollector.pushErrors(failures)
+
+   val f = when {
+      assertionsCount == failures.size -> failure("One expected a single assertion to succeed, but none succeeded.")
+      assertionsCount > failures.size + 1 -> failure("One expected a single assertion to succeed, but more than one succeeded.")
+      else -> failure("One expected a single assertion to succeed, but there were more failures than assertions.")
+   }
+
+   errorCollector.pushErrorAndMaybeThrow(f)
+   return null
```
sksamuel, line 39: Would we ever have a situation where one wouldn't either throw or return the value ?

jschneidereit: I think if the assertion mode is soft, but also I think we need it to make the compiler happy.

PR #2191, opened 2021-04-07
fix #2190 by making result of eventually state nullable

### LINKED ISSUE #2190: Eventually only calls listener on successful result
In the case that someone is using matchers within their eventually function we should call their listener when the exception produced by those matchers is thrown. Currently we only call the listener on a [successful result from the producer function](https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/timing/eventually.kt#L102).

jschneidereit:


sksamuel:
lgtm

PR #2200, opened 2021-04-10
Fix issues of passing vararg to another function in containsInOrder

yoonho-sean-lee:
This PR has the aim of fixing issues of containsInOrder extensions. These extensions has passed vararg to another function not using spread operator. Please let me know if any further actions is required.


sksamuel:
nice thanks!

PR #2216, opened 2021-04-21
refactor(datatest): replace deprecated typealias with original

malteesch:


sksamuel:
We need to keep the typealias for people who are using it externally, but internally we can switch over.
Can you update the PR ?

malteesch:
Thanks for the heads up!
I updated the PR accordingly.

sksamuel:
👍🏼

PR #2217, opened 2021-04-21
String Shrinker exceeds min size limit #2213

### LINKED ISSUE #2213: Shrinking of Arbs (String, and maybe others?) exceeds min size limit
**Which version of Kotest are you using** 4.4.3

If I’ve got a String Arb and I set the minimum size to 1, it shrinks to size 0.

In this example, `authKey` shrinks to 0, and so gives an invalid reason for failure.

```kotlin

 test("expect 'actual code is failing', _not_ authKey length error") {
    checkAll(
      Arb.string(1..10, Arb.alphanumeric()),
    ) { authKey ->
      withClue("authKey must have length > 0") {
        authKey shouldHaveMinLength 1
      }
      fail("actual code is failing")
    }
  }
```

I’ve tried using this but then the test loops endlessly

```kotlin
Arb.string(1..10, Arb.alphanumeric()).filter { it.isEmpty() },
```

sksamuel:
@asemy used your code thank you

Fixes #2213

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/strings.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/strings.kt
@@ -96,34 +96,29 @@ class StringShrinkerWithMin(
 
       return when {
          isShortest && isSimplest -> emptyList()
-         !isShortest -> shorterVariants(value)
-         !isSimplest -> simplerVariants(value)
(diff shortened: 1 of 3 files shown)


aSemy:
Looks great! Thanks for doing the grunt work after I did the fun bit :)

LeoColman (approved):

PR #2219, opened 2021-04-23
Add watchosX64 target

ajalt:
This is one of the targets that is part of the default `watchos()`  target


sksamuel:
👍🏼

sksamuel:
PR was green but master is red.
I guess the the workflows are out of sync.
I think watchosX64 doesn't have coroutines yet, don't know if that's an issue here.

sksamuel:
I've reverted this because there are no coroutines for watchosX64 yet, and we need that for the build.

PR #2220, opened 2021-04-24
Upgrade arrow to 0.13.1

myuwono:
updating arrow deps and cleaning up unused ones. note that this is only used in `kotest-property-arrow`.


sksamuel:
Thanks, will also move this to a new repo to allow for faster released independent of kotest.

PR #2225, opened 2021-04-26
Most of the matchers on String really should be matchers on CharSeque…

### LINKED ISSUE #2224: Most of the matchers on String really should be matchers on CharSequence
Matchers on String are really too specific in that they only match on String, but CharSequence is the interface that String implements and most of the matchers don't care about the difference.

sksamuel:
Closes #2224

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/string/case.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/string/case.kt
@@ -6,13 +6,13 @@ import io.kotest.matchers.neverNullMatcher
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldNot
 
-fun String?.shouldBeUpperCase(): String? {
-   this should beUpperCase()
+fun <A : CharSequence> A?.shouldBeUpperCase(): A? {
+   this.toString() should beUpperCase()
    return this
 }
 
-fun String?.shouldNotBeUpperCase(): String? {
-   this shouldNot beUpperCase()
+fun <A : CharSequence> A?.shouldNotBeUpperCase(): A? {
+   this.toString() shouldNot beUpperCase()
    return this
 }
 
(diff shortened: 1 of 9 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/string/empty.kt
```diff
@@ -0,0 +1,44 @@
+package io.kotest.matchers.string
+
+import io.kotest.assertions.show.show
+import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.neverNullMatcher
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldNot
+
+fun <A : CharSequence> A?.shouldBeEmpty(): A? {
+   this.toString() should beEmpty()
```
myuwono, line 10: hmm `null.toString()` will yield the string "null"

sksamuel: Good spot.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/string/case.kt
```diff
@@ -6,13 +6,13 @@ import io.kotest.matchers.neverNullMatcher
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldNot
 
-fun String?.shouldBeUpperCase(): String? {
-   this should beUpperCase()
+fun <A : CharSequence> A?.shouldBeUpperCase(): A? {
+   this.toString() should beUpperCase()
```
myuwono, line 10: `null.toString()`

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/string/ends.kt
```diff
@@ -0,0 +1,14 @@
+package io.kotest.matchers.string
+
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldNot
+
+infix fun <A : CharSequence> A?.shouldEndWith(suffix: String): A? {
+   this.toString() should endWith(suffix)
```
myuwono, line 7: `null.toString()`

i start to wonder if we should change `endWith` and other `String` matchers to work based on `A : CharSequence` now? i.e.

```kotlin

fun <A : CharSequence> beEmpty() = neverNullMatcher<A> { value ->
   MatcherResult(
      value.toString().isEmpty(),
      "${value.toString().show().value} should be empty",
      "${value.toString().show().value} should not be empty"
   )
}

```

sksamuel:
Closing in favour of #2278

PR #2227, opened 2021-04-26
tempdir: Delete temporary directories recursively

sschuberth:
While delete() works on directories, they must be empty for delete() to
succeed. However, temporary directories should be deleted recursively,
including any contained files and / or subdirectories.


sschuberth:
@sksamuel can this still make it into the 4.5.0 release?

sksamuel:
Seems low risk so yes.

sksamuel:
Can we get a test

sschuberth:
> Can we get a test

Sorry, I didn't have time until now, but I've adapted the existing test now accordingly. and also added another improvement commit.

sschuberth:
@sksamuel I don't think the test failure on Ubuntu is related to my changes...

sksamuel:
Yeah it's not, lgtm.

PR #2235, opened 2021-05-04
Add DuplicateTestNameMode #2234

### LINKED ISSUE #2234: Duplicate test name mode
Currently we error when a test has a duplicate name.

However, when using data driven tests, this is not great because if you are using a non-data class, or any class on Javascript, the test names will end up with the same value (the class name).

sksamuel:
Closes #2234 

I've added `DuplicateTestNameMode` which allows us to have an error (current behavior), a warning (outputs a message saying dup name, renames test), or none (no warning, renames test).

The renamed tests are like `foo` => `(1) foo` and so on.

I've set default to warning.

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/test/DuplicateTestNameHandler.kt
+++ b/kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/test/DuplicateTestNameHandler.kt
@@ -1,8 +1,8 @@
 package io.kotest.engine.test
 
+import io.kotest.core.test.DescriptionName
 import io.kotest.core.test.DuplicateTestNameMode
 import io.kotest.core.test.Identifiers
-import io.kotest.core.test.TestCase
 
 /**
  * A handler for test names. Will track all test names used for a particular context.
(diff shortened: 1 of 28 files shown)


jschneidereit (approved):
Looks good to me, I'll look again when the tests are added 👍

LeoColman (approved):

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/test/DuplicateTestNameHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A handler for test names. Will track all test names used for a particular context.
+ */
+class DuplicateTestNameHandler(private val mode: DuplicateTestNameMode) {
+
+   private val names = mutableSetOf<String>()
+
+   private fun message(name: String): String =
+      "Duplicated test name ${name}. To disable this message, set DuplicateTestNameMode to None."
```
ferdoran, line 15: Currently the warning suggests to use "None" as value, however, "None" is not a valid enum value of DuplicateTestNameMethod. Instead of "None", "Silent" should be used in the warning message.

LeoColman: This pr is pretty old.
I created #3069 to address this.

Issue #2149, opened 2021-03-05
Please provide a version of "eventually" that is not reliant on "experimental" features of Kotlin


xenomachina:
`eventually` relies on `@ExperimentalTime`. Apparently [this means that even patch releases of kotlinx-coroutines can break binary backwards compatibility](https://github.com/Kotlin/kotlinx.coroutines/issues/2565).  This, in fact, happened with the update from 1.4.2 to 1.4.3 because the mangled name of `delay(Duration)` changed. You can verify this by getting the two jars, kotlinx-coroutines-core-jvm-1.4.3.jar and kotlinx-coroutines-core-jvm-1.4.2.jar, and using javap on DelayKt.class:

``` sh
$ for i in */kotlinx/coroutines/DelayKt.class; do echo "==== $i ===="; javap $i; done
==== kotlinx-coroutines-core-jvm-1.4.2.jar/kotlinx/coroutines/DelayKt.class ====
Compiled from "Delay.kt"
public final class kotlinx.coroutines.DelayKt {
  public static final java.lang.Object awaitCancellation(kotlin.coroutines.Continuation<?>);
  public static final java.lang.Object delay(long, kotlin.coroutines.Continuation<? super kotlin.Unit>);
  public static final java.lang.Object delay-p9JZ4hM(double, kotlin.coroutines.Continuation<? super kotlin.Unit>);
  public static final kotlinx.coroutines.Delay getDelay(kotlin.coroutines.CoroutineContext);
  public static final long toDelayMillis-LRDsOJo(double);
}
==== kotlinx-coroutines-core-jvm-1.4.3.jar/kotlinx/coroutines/DelayKt.class ====
Compiled from "Delay.kt"
public final class kotlinx.coroutines.DelayKt {
  public static final java.lang.Object awaitCancellation(kotlin.coroutines.Continuation<?>);
  public static final java.lang.Object delay(long, kotlin.coroutines.Continuation<? super kotlin.Unit>);
  public static final java.lang.Object delay-VtjQ1oo(double, kotlin.coroutines.Continuation<? super kotlin.Unit>);
  public static final kotlinx.coroutines.Delay getDelay(kotlin.coroutines.CoroutineContext);
  public static final long toDelayMillis-LRDsOJo(double);
}
```
Note `delay-p9JZ4hM` vs `delay-VtjQ1oo`.

So `eventually` from a build of kotest compiled against 1.4.2 will break at runtime if used with 1.4.3.
 
[...]

PR #2242, opened 2021-05-07
2149 no experimental eventually

jschneidereit:
#2149


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/Interval.kt
```diff
@@ -0,0 +1,113 @@
...
+
+   /**
+    * Returns the next delay as a [Millis].
+    *
+    * @param count The number of times the condition has been polled (evaluated) so far.
+    * Always a positive integer.
+    *
+    * @return The duration of the next poll interval
+    */
+   fun next(count: Int): Millis
```
sksamuel: So this is not a suggestion but a discussion point - if you make next() return a Pair<Interval, Millis> then you have no state to worry about anywhere, as the count can be internalized.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/concurrency.kt
```diff
@@ -0,0 +1,20 @@
+package io.kotest.framework.concurrency
+
+import io.kotest.common.ExperimentalKotest
+
+typealias Millis = Long // TODO: inline class this in 4.6.x (when we have kotlin 1.5)
```
sksamuel: Should we use inline classes? If we rely on 1.5 features we're forcing users to upgrade.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/concurrency.kt
```diff
@@ -0,0 +1,20 @@
...
+
+typealias ConcurrencyProducer<T> = suspend () -> T
+
+/**
+ * Will be replaced with [kotlin.time.Duration] when that API is no longer experimental
+ */
+@ExperimentalKotest
+data class Instant(val timeInMillis: Millis)
```
sksamuel: What is the instant for

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/concurrency.kt
```diff
@@ -0,0 +1,20 @@
...
+
+@ExperimentalKotest
+data class PatienceConfig constructor(val duration: Millis, val interval: Interval) {
```
sksamuel: PatienceConfig is a good name, where did you get that from.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/eventually.kt
```diff
@@ -0,0 +1,138 @@
...
+}
+
+@ExperimentalKotest
+suspend fun <T> eventually(
+   duration: Millis,
+   interval: Interval = 25L.fixed(),
+   retries: Int = Int.MAX_VALUE,
+   exceptions: Set<KClass<out Throwable>> = setOf(),
+   listener: EventuallyListener<T>? = null,
+   f: ConcurrencyProducer<T>
```
sksamuel: We could drop eventually supporting return values entirely, and roll them into until ?

until supports booleans, but I wonder if until could be changed to use T rather than Boolean and it just work as is for booleans?

Then it might all be a bit neater ?

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/Interval.kt
jschneidereit: Ooo I like this, I'll play around with that idea 👍

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/concurrency.kt
jschneidereit: You're right 👍 I'm just used to being tied to kotest's kotlin version cause of duration, the whole point of this PR! 😂😅

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/concurrency.kt
jschneidereit: Stole it right out of the old `until.kt`!

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/eventually.kt
jschneidereit: I'll have to think about this, we'd have to move the exception set to until as well since right now it doesn't try to catch anything iirc

kotest-framework/kotest-framework-api/src/jvmTest/kotlin/io/kotest/framework/concurrency/EventuallySpec.kt
```diff
@@ -0,0 +1,368 @@
...
+      shouldThrow<NullPointerException> {
+         eventually(2.seconds()).allowExceptions(IOException::class) {
+            (null as String?)!!.length
+         }
+      }
+   }
+
+   test("eventually passes tests that throws FileNotFoundException for some time") {
+      val end = System.currentTimeMillis() + 250
+      eventually(5.seconds()).allowExceptions(FileNotFoundException::class, listener = { true }) {
```
sksamuel: what's this listener = true thing ?

kotest-framework/kotest-framework-api/src/jvmTest/kotlin/io/kotest/framework/concurrency/EventuallySpec.kt
```diff
@@ -0,0 +1,368 @@
...
+      val end = System.currentTimeMillis() + 250
+      eventually(5.seconds()) {
+         if (System.currentTimeMillis() < end)
+            assert(false)
+      }
+   }
+
+   test("eventually fails tests that throw unexpected exception type") {
+      shouldThrow<NullPointerException> {
+         eventually(2.seconds()).allowExceptions(IOException::class) {
```
sksamuel: is allowExceptions the right word? Sounds like we want them to happen.
maybe forExceptions, handleExceptions, xxxExceptions ?

kotest-framework/kotest-framework-api/src/jvmTest/kotlin/io/kotest/framework/concurrency/EventuallySpec.kt
```diff
@@ -0,0 +1,368 @@
...
+   }
+
+   test("eventually passes tests that throws FileNotFoundException for some time") {
+      val end = System.currentTimeMillis() + 250
+      eventually(5.seconds()).allowExceptions(FileNotFoundException::class, listener = { true }) {
+         if (System.currentTimeMillis() < end)
+            throw FileNotFoundException("foo")
+      }
+   }
+
```
sksamuel: Do we have a test that an exception we dont want immediately fails the block ?

kotest-framework/kotest-framework-api/src/jvmTest/kotlin/io/kotest/framework/concurrency/EventuallySpec.kt
```diff
@@ -0,0 +1,368 @@
...
+   test("eventually allows suspendable functions") {
+      eventually(100.milliseconds()) {
+         delay(25)
+         System.currentTimeMillis()
+      }
+   }
+
+   test("eventually allows configuring interval delay") {
+      var count = 0
+      eventually(200.milliseconds(), 40.milliseconds().fixed()) {
```
sksamuel: I think we should have an overload that doesn't require the .fixed()

kotest-framework/kotest-framework-api/src/jvmTest/kotlin/io/kotest/framework/concurrency/EventuallySpec.kt
```diff
@@ -0,0 +1,368 @@
...
+      result shouldBe "xxxxxxxxxxx"
+   }
+
+   test("eventually with T predicate, interval, and listener") {
+      var t = ""
+      val latch = CountDownLatch(5)
+      val result = eventually(
+         5.seconds(),
+         250.milliseconds().fixed(),
+         listener = { latch.countDown(); t=="xxxxxxxxxxx" },
```
sksamuel: Let's only have durations in the original function.

sksamuel (commented):
The new stuff should be in a new package `kotest-framework-concurrency`

kotest-framework/kotest-framework-api/src/jvmTest/kotlin/io/kotest/framework/concurrency/EventuallySpec.kt
jschneidereit: suppressExceptions?

kotest-framework/kotest-framework-api/src/jvmTest/kotlin/io/kotest/framework/concurrency/EventuallySpec.kt
jschneidereit:
```
test("eventually fails tests throw unexpected exception type")
```

this one was existing but doesn't check how many attempts are made since it doesn't get to the part that logs the state of the eventually before finishing up

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/concurrency.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import io.kotest.common.ExperimentalKotest
+
+/**
+ * Will be replaced with [kotlin.time.Duration] when that API is no longer experimental
+ */
+typealias Millis = Long
+
+typealias ConcurrencyProducer<T> = suspend () -> T
+typealias ConcurrencyConsumer<T> = suspend (T) -> Boolean
```
sksamuel: what's the producer do again ?

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/continually.kt
```diff
@@ -0,0 +1,57 @@
...
+@ExperimentalKotest
+suspend fun <T> continually(
+   duration: Millis,
+   interval: Interval = 25L.fixed(),
+   listener: ContinuallyListener<T> = ContinuallyListener.default,
+   f: ConcurrencyProducer<T>
+): T? = continually(PatienceConfig(duration, interval), listener, f)
+
+@ExperimentalKotest
+suspend fun <T> continually(
```
sksamuel: should we do the same thing with a continually builder

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/eventually.kt
```diff
@@ -0,0 +1,232 @@
...
+   }
+
+   fun allowExceptions(vararg exceptions: KClass<out Throwable>) = copy(exceptions = this.exceptions + exceptions)
+   suspend fun <T> allowExceptions(
+      vararg exceptions: KClass<out Throwable>, listener: EventuallyListener<T>? = null, shortCircuit: ConcurrencyConsumer<T>? = null, f: ConcurrencyProducer<T>,
+   ): T = copy(exceptions = this.exceptions + exceptions).invoke(listener, shortCircuit, f)
+
+   fun allowExceptionIf(allowExceptionIf: ThrowablePredicate) = copy(allowExceptionIf = allowExceptionIf)
+   suspend fun <T> allowExceptionIf(
+      allowExceptionIf: ThrowablePredicate, listener: EventuallyListener<T>? = null, shortCircuit: ConcurrencyConsumer<T>? = null, f: ConcurrencyProducer<T>,
```
sksamuel: Why do we have all these overloads ?

jschneidereit: they're all gone now 👍

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/concurrency.kt
jschneidereit: removed these

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/framework/concurrency/continually.kt
jschneidereit: done

kotest-framework/kotest-framework-concurrency/src/commonMain/kotlin/io/kotest/framework/concurrency/continually.kt
```diff
@@ -0,0 +1,98 @@
...
+   suspend fun withListener(listener: ContinuallyListener<T>, f: suspend () -> T): T? = withListener(listener).invoke(f)
+
+}
+
+@ExperimentalKotest
+data class ContinuallyState(val start: Long, val end: Long, val times: Int)
+
+@ExperimentalKotest
+fun interface ContinuallyListener<in T> {
+   fun onEval(t: T, state: ContinuallyState)
```
jschneidereit: I kinda want to make the EventuallyListener look like this

PR #2248, opened 2021-05-11
Adds WithDataTestName to configure test name for data test

ashishkujoy:


sksamuel:
Looks good overall. Some thoughts for discussion.  Could the interface just be a marker interface that uses the regular toString? It's then just a hint to the system that it's safe to use toString. What about an annotation as an alternative (keep both?).
Also Triple could be added to safe list.

ashishkujoy:
> Looks good overall. Some thoughts for discussion. Could the interface just be a marker interface that uses the regular toString? It's then just a hint to the system that it's safe to use toString. What about an annotation as an alternative (keep both?).
> Also Triple could be added to safe list.

The only reason for which I had not kept that as marker interface and using `toString` is to allow the possibility to use a different test name compare to `toString` . 

May with the annotation idea we can use `toString` for generating test name.
So that we both type of option available for user.

sksamuel:
What about interface if you want to provide your own method, and annotation if you just want to indicate to use toString ?

ashishkujoy:
> What about interface if you want to provide your own method, and annotation if you just want to indicate to use toString ?

Yes Sam, I did the same. With the latest commit.

sksamuel:
Great, but I just noticed that you've done this in framework. It should be in the new datatest module.

ashishkujoy:
> Great, but I just noticed that you've done this in framework. It should be in the new datatest module.

The Identifiers object which is getting used for getting `stableIdentifier` is in `kotest-framework-api` so if I move this interface and annotation in `kotest-framework-datatest` `Identifiers` won't be able access that.

sksamuel:
Ok I see. Identifiers is there, but WithDataTestName should not go there. I would add a function to datatest module that checks whether it should call isStable or not.

ashishkujoy:
I add a method in data test module `getStableIdentifier` which is checking If the type is `WithDataTestName` or annotated with `IsStableType` if not then it is calling the `Identifiers.stableIdentifier` . Moved `WithDataTestName` and `IsStableType` in data test module.

sksamuel (approved):

sksamuel:
Just had a thought on this. What's the difference between using @IsStableType and just passing in the object? Either way its going to call toString right ?

ashishkujoy:
If we dont use @IsStableType and the type is not stable then we are currently using "t::class.bestName" which is interanally checking if there is any `simpleName` for that class if `simpleName` is not present then only we are using `toString`

So as a result if you have a class Person for which you have overridden `toString` and u have not annotated it with `@IsStableType` then its name will be something like `com.something.Person` instead of what toString is going to return.

sksamuel:
Yes good point.

PR #2273, opened 2021-05-22
add overall project timeout

### LINKED ISSUE #2240: ProjectConfig should support an overall suite timeout
Consider a system that has long timeouts per test but you don't want the entire suite to take longer than some time, this allows a few tests to take an extremely long time but the rest should complete in the time allotted.

```kotlin
    override val invocationTimeout = 4.hours.toLongMilliseconds()
    override val projectTimeout = 12.hours.toLongMilliseconds() // use project in name because ProjectListener
}
```

jschneidereit:
Fixes #2240

### DIFF
--- a/documentation/docs/framework/config_props.md
+++ b/documentation/docs/framework/config_props.md
@@ -60,6 +60,8 @@ object KotestEngineProperties {
 
    const val timeout = "kotest.framework.timeout"
 
+   const val projectTimeout = "kotest.framework.project.timeout"
+
    const val invocationTimeout = "kotest.framework.invocation.timeout"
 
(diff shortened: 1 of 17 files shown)


jschneidereit:
Fixes #2240

jschneidereit:
Here's an example in the [gradle build scan](https://scans.gradle.com/s/kmjrpzwcizwki/tests/:kotest-tests:kotest-tests-timeout:test/UnknownClass/executionError?top-execution=1) of the project timeout in action.

sksamuel (commented):
Should be supported by a system property too.
And you need to add support for merging the configs, adding it to the config detect is not enough.
Finally, we need a test - probably instantiate an engine, set the property and make sure it fails.

PR #2274, opened 2021-05-22
add round project extension

### LINKED ISSUE #2270: Provide machinery to share state before and after project run
Currently, to share state between a `beforeProject` and `afterProject` listener we need to create mutable member variables in the `ProjectListener`'s body. Instead we should provide a way for testers to have immutable state that surrounds the scope of a project run. 

old and busted: 

```kotlin
object Foo : ProjectListener {
  private var stuff: Stuff? = null
  
  override suspend fun beforeProject() {
    stuff = setupStuffForProjectRun()
  }
  
  override suspend fun afterProject() {
    when (stuff) {
      null -> throw...
      is Stuff -> cleanup(stuff as Stuff) // cast cause stuff "is a mutable property that could have been changed by this time"
```

new hotness:

```kotlin
object Bar : ProjectExtension {
  override suspend fun aroundProject(f: suspend () -> ProjectResults) {
    val stuff = setupStuffForProjectRun()
    f()
    cleanup(stuff)
  }
}

### PR DESCRIPTION
**Pull request by jschneidereit**
Fixes #2270

### CODE REVIEW

File: kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/KotestEngine.kt [hunk 1]
```diff
@@ -75,12 +77,14 @@ class KotestEngine(private val config: KotestEngineConfig) {
...
                   return
                }
-
-
             }
          )
 
-      Try { submitAll(plan) }
+      Try {
+         runProjectExtensions(projectExtensions) {
```
jschneidereit: After thinking about this more I'm going to move the try inside so the tests that explode still allow the end of the interceptor to complete

jschneidereit: Or do we put the responsibility to catch and re-throw on the extension implementation

sksamuel: I would gather the errors and then fail the test run at the end.

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/KotestEngine.kt
```diff
@@ -49,6 +50,7 @@ class KotestEngine(private val config: KotestEngineConfig) {
     * Starts execution of the given test plan.
     */
    suspend fun execute(plan: TestPlan) {
+      val projectExtensions = configuration.extensions().filterIsInstance<ProjectExtension>().reversed()
```
sksamuel: We don't need to reverse - we simply won't make any guarantees on the order they run.

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/KotestEngine.kt
```diff
@@ -2,18 +2,14 @@ package io.kotest.engine
...
 import io.kotest.core.spec.afterProject
 import io.kotest.core.spec.beforeProject
 import io.kotest.engine.config.ConfigManager
-import io.kotest.engine.extensions.DumpConfigExtension
-import io.kotest.engine.extensions.EmptyTestSuiteExtension
-import io.kotest.engine.extensions.EngineExtension
-import io.kotest.engine.extensions.KotestPropertiesExtension
-import io.kotest.engine.extensions.SpecifiedTagsTagExtension
-import io.kotest.engine.extensions.TestDslStateExtensions
+import io.kotest.engine.extensions.*
```
jschneidereit, line 10: I wonder if there's editorconfig to control this behavior of IJ

PR #2278, opened 2021-05-22
Migrating string matchers to work with CharSequence

### LINKED ISSUE #2224: Most of the matchers on String really should be matchers on CharSequence

sksamuel:
Closes #2224


JordanLongstaff:
Sorry to necropost, but why is the return value of `should(Not)BeLowerCase` nullable but `should(Not)BeUpperCase` isn't?

PR #2282, opened 2021-05-23
Added Lifecycle methods for property test #2161

### LINKED ISSUE #2161: Lifecycle methods for property test
Allow some way to register beforeProperty / afterProperty callbacks that work for all property tests / all property tests in a scope. Possibly using the coroutine context as a way to make arbitrary data available.

sksamuel:
I'm interested in feedback on this approach (ignore the AbstractContainerContext stuff that's just noise).

The before/after listeners are added to the coroutineContext, so it means we can add them from anywhere. The downside is that they have to be called from a suspend function in order to get access to the coroutineContext, which means you can't just call `beforeProperty` in the spec. Instead, `beforeProperty` is an extension function which calls `beforeSpec` to get access to a suspend continuation. So this means a dependency on Spec, which means I had to create a new module to avoid bringing Kotest into the property test library.

Other options are, we make people do
```
beforeSpec {
   beforeProperty {
   }
 }
 ```

Or we say Kotest prop test 5.0 needs Kotest framework 5.0.

Closes #2161

### DIFF
--- a/documentation/docs/why.md
+++ b/documentation/docs/why.md
@@ -1,0 +1,23 @@
+---
+title: Why Kotest
+sidebar_label: Why Kotest
+slug: why-kotest.html
+---
+
+## vs Junit
+
+* Powerful coroutine integration: Every test is a coroutine, therefore, you can invoke suspension methods without requiring `runBlocking` or other boilerplate
+  * One line configuration to enable coroutine debugging for a test or for all tests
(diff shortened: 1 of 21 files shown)


jschneidereit (approved):
I think this looks good. I also think the use case is advanced enough to be fine as another library if the dependency of kotest prop on kotest framework is too heavy handed.

PR #2299, opened 2021-05-28
Improve Arb.primitive consistency

BartvHelvert:
Fixes a lot of inconsistencies in the Arb API for primitives.

Closes #2297.

PR #2301, opened 2021-05-29
Add primitive arrays to Arb

### LINKED ISSUE #2300: Support Arb primitive arrays
Currently only `Arb.byteArray` is supported.
I suggest that we add support for other primitive arrays as well.

BartvHelvert:
Closes #2300.
Depends on #2299.


sksamuel:
I think will need to be rebased.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/longs.kt
```diff
@@ -34,13 +47,20 @@ fun Arb.Companion.positiveLong(max: Long = Long.MAX_VALUE): Arb<Long> = long(1L,
...
 
-class LongShrinker(private val range: LongRange) : Shrinker<Long> {
-   override fun shrink(value: Long): List<Long> = when (value) {
-      0L -> emptyList()
-      1L, -1L -> listOf(0L).filter { it in range }
+/**
+ * Returns an [Arb] that produces [LongArray]s where [generateArrayLength] produces the length of the arrays and
+ * [generateContents] produces the content of the arrays.
+ */
+fun Arb.Companion.longArray(generateArrayLength: Gen<Int>, generateContents: Arb<Long>): Arb<LongArray> =
```
sksamuel, line 54: We tend to use the name somethingArb or somethingGen when composing arbs/gens.
So perhaps lengthGen and contentArb or something along those lines ?

BartvHelvert: I am not sure how useful it is to repeat the type in the parameter name? I think `length`/`content` would be clear enough?

sksamuel: Ok agreed.

PR #2303, opened 2021-05-31
Support for ignoring unknown keys in JSON asserts

### LINKED ISSUE #2298: Option to ignore unspecified keys when using shouldEqualJson matcher
I'd like an option for `shouldEqualJson` to ignore keys in _actual_ that are unspecified in _expected_. It should probably be included for `shouldNotEqualJson` as well.

I think this could be achieved either by:
* Expanding `CompareMode` with new modes: `CompareMode.StrictIgnoringUnknownKeys` and `CompareMode.LenientIgnoringUnknownKeys`
* Adding a new flag to matcher signatures: `ignoreUnknownKeys: Boolean` 
* New matcher: `shouldEqualSpecifiedJson`? Better names? 😅

Following example should pass:

```kotlin
"""{ "a": 5, "timestamp": "2021-05-28T14:44Z" }""".shouldEqualJson("""{ "a": 5 }""", ignoreUnknownKeys = true) 
"""{ "a": 5, "timestamp": "2021-05-28T14:44Z" }""" shouldEqualSpecifiedJson """{ "a": 5 }"""
```

Kantis:
Closes #2298 

Added new matchers `shouldEqualJsonIgnoringUnknown` and
`shouldNotEqualJsonIgnoringUnknown`. Supports for specifying all
other flags that the previous json matcher supported.

The existing compare method already supported the use-case quite easily,
and had good error messages for possible validation errors.

Unknown keys can only exist in object nodes. The scenarios currently
covered by tests are:

* Actual contains a key not specified in expected - key is ignored
* Actual contains a key specified in expected - key is compared
* Actual is missing a key specified in expected - matcher error occurs

Since comparison happens recursively, I also added tests to make sure
the ignoring is in effect deeper in the structure as well.

## Todo
- [x] add in changelog
- [x] update docs


kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/matchers.kt
```diff
@@ -14,11 +14,22 @@ import io.kotest.matchers.MatcherResult
  * regardless of order.
  *
  */
-fun equalJson(expected: JsonTree, mode: CompareMode, order: CompareOrder) = object : Matcher<JsonTree> {
+fun equalJson(
+   expected: JsonTree,
+   mode: CompareMode,
+   order: CompareOrder,
+   ignoreUnknownKeys: Boolean = false
+) = object : Matcher<JsonTree> {
```
Kantis, line 22: I suppose changing this breaks binary compatibility. Is that allowed for a minor release?

Kantis:
Will add documentation as well.

Do you think it would make sense to put it between `shouldEqualJson` and `shouldContainJsonKey` at https://kotest.io/docs/assertions/json-matchers.html? It feels closely related to `shouldEqualJson`.

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/matchers.kt
sksamuel: No, but this woud go into 5.0 so I'm ok with minor breaking changes (please add to changelog).

sksamuel:
Is this what Lenient mode is meant to do ?

Kantis:
> Is this what Lenient mode is meant to do ?

As a user, I would prefer to have strict types while comparing if possible, but I agree it would feel natural and this change might make the API too verbose? Perhaps it might be worth introducing something like this:

```kotlin
class CompareOptions(
   build: CompareOptions.() -> Unit
) {
   var ignoreKeyOrder: Boolean = false
   var coerceTypes: Boolean = false
   var ignoreUnknownKeys: Boolean = false

   init {
       build()
   }
}

fun String.shouldEqualJson(expected: String, options: CompareOptions.() -> Unit): Unit =
   TODO()

fun x() {
   "actual".shouldEqualJson("expected") { ignoreKeyOrder = true }
}
```

it loses out on the infix fun though, which is a shame..

sksamuel:
Oh that's right, strict vs lenient is about the types.
Maybe we should document that since I had forgotten myself and I wrote it :)

Anyway, I think adding another enum works. FieldComparision or something.

Kantis:
Something like this? And we replace the `ignoreUnknownKeys` boolean with this?
```kotlin

enum class FieldComparison {
   /** Verifies all expected key-values match, and that no extra keys exist in the actual */
   ComparingAll,
   /** Verifies all expected key-values match, ignoring keys in actual that are not defined in expected*/
   ComparingExpected
}
```

sksamuel:
> Something like this? And we replace the `ignoreUnknownKeys` boolean with this?
> 
> ```kotlin
> enum class FieldComparison {
>    /** Verifies all expected key-values match, and that no extra keys exist in the actual */
>    ComparingAll,
>    /** Verifies all expected key-values match, ignoring keys in actual that are not defined in expected*/
>    ComparingExpected
> }
> ```

```
enum class FieldComparison {
   /** Verifies all expected key-values match, and that no extra keys exist in the actual */
   Exact,
   /** Verifies all expected key-values match, ignoring keys in actual that are not defined in expected*/
   IgnoreExtra
}
```

Maybe something like that ?

sksamuel:
Is this ready for final review ?

Kantis:
@sksamuel aye, it should be.

sksamuel:
Sorry is this still waiting on me? If so can we fix the conflicts and I will review.

Kantis:
@sksamuel should be ready now, I think the failing test is flaky.

sksamuel (approved):

PR #2304, opened 2021-06-01
Add Arb.distinct that will terminate #2262

### LINKED ISSUE #2262: Add Arb.distinct that will terminate
The current `distinct` is likely to hang if it cannot generate enough unique values. That is why we deprecated it.
But the functionality is still useful.
I propose we add an alternative, that makes those caveats crystal.

`arb.distinct(maxAttempts = 1000, throwOnLimit = true)`

Those parameters would not have defaults so it would be obvious what you're doing.

@myuwono  ?

sksamuel:
@myuwono 

Closes #2262

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/distinct.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/distinct.kt
@@ -1,9 +1,11 @@
 package io.kotest.property.arbitrary
 
+import io.kotest.common.DelicateKotest
 import io.kotest.property.Arb
 import io.kotest.property.RandomSource
 import io.kotest.property.Sample
 
+@DelicateKotest
 fun <A> Arb<A>.distinct(attempts: Int = 100) = object : Arb<A>() {
 
    private val seen = mutableSetOf<A>()
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/distinct.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/distinct.kt
@@ -1,0 +1,20 @@
+package io.kotest.property.arbitrary
+
+import io.kotest.property.Arb
+import io.kotest.property.RandomSource
+import io.kotest.property.Sample
+
+fun <A> Arb<A>.distinct(attempts: Int = 100) = object : Arb<A>() {
+
+   private val seen = mutableSetOf<A>()
+
+   override fun edgecase(rs: RandomSource): A? = this@distinct.edgecase(rs)
+
+   override fun sample(rs: RandomSource): Sample<A> {
+      var iterations = 0
+      return generateSequence {
+         if (iterations++ < attempts) this@distinct.sample(rs) else null
+      }.filter { seen.add(it.value) }.first()
+   }
+
+}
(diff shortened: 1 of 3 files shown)


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/distinct.kt
```diff
@@ -0,0 +1,20 @@
+package io.kotest.property.arbitrary
+
+import io.kotest.property.Arb
+import io.kotest.property.RandomSource
+import io.kotest.property.Sample
+
+fun <A> Arb<A>.distinct(attempts: Int = 100) = object : Arb<A>() {
```
myuwono, line 9: @sksamuel I wonder if we should be making this function terminal and returns a `Set<A>` instead. At this point this Arb is stateful and unsafe. I would not use it and probably tell our developers to not use it..

consider:

```kotlin
val arbNumbers: Arb<Int> = Arb.of(1, 2, 3).distinct()

checkAll(arbNumbers) { number -> // explode after 3 iterations
  // ...
}
```

secondly, we generally store `Arb<A>` as a val in our codebase. This actually means that the arb which has been distinct-ed will never be reusable, unless declared a function.

```kotlin
val arbNumbers: Arb<Int> = Arb.of(1, 2, 3).distinct() 

// none of the arbs below are stable
val arbOne = arbNumbers.flatMap { ... }
val arbTwo = Arb.bind(arbNumbers, ...) { ... }
```

This is ultimately Kotest's decision if we were to introduce this, but i strongly advocate against it.. i think this creates more problems than it solves.

What might be more useful is perhaps:
```kotlin
checkDistinct(arb) { distinctNonRepeatedValue -> ...
}
```

sksamuel: It's deliberately stateful. And yes it flies in the face of FP best practice, but nevertheless it is useful. So I think omitting it entirely would be a shame (people have asked for it).

So what do you think about adding a new annotation - @KotestStateful or something, which acts like this https://blog.jetbrains.com/kotlin/2021/05/kotlin-coroutines-1-5-0-released/#globalscope

It means you get a compiler warning if you choose to use this method.

myuwono: thinking out loud, how about this?
```kotlin
fun <A> Arb<A>.distinct(): Arb<Set<A>> = ...
fun <A> Arb<List<A>>.distinct(): Arb<List<A>> = ...
```

sksamuel: In the first one, you're going to get passed a set of values, which isn't going to be compatible with other arbs.
So like,

checkAll(Arb.strings().distinct(), Arb.ints()) { a, b -> } // now my a is a set

Did you not like the idea of a compiler warning telling you it's dangerous?

myuwono: Right, maybe the compiler warning is indeed sufficient @sksamuel 

Following up on this, I'm just trying to think in terms of general stateful generation. Would it make sense if we delegate any state outside of the arbs, e.g. on the checkAll level instead? And then all of a sudden we can have many inbuilt constraints including skip test if arb has already tested etc.
https://github.com/kotest/kotest/issues/2262#issuecomment-856582355

sksamuel: We can continue the discussion in that thread yeah. The skip thing is intriguing, but as I commented in the thread, I'm not sure I see the use (yet).

myuwono (approved):

myuwono:
Thanks @sksamuel ! Let's get this in

PR #2310, opened 2021-06-02
Improve Arb function naming

### LINKED ISSUE #2305: Improve Arb naming
I want to propose 2 naming changes:

1.  `Arb.bool` -> `Arb.boolean`, `Arb.boolArray` -> `Arb.booleanArray`.
This makes more sense because it matched the Kotlin standard library naming convention. Requires deprecation of `Arb.bool`.

2. `Arb.uint` -> `Arb.uInt` (and similar for all other unsigned types)
Current naming scheme doesn't follow proper lower camelcase convention which is almost always used in Kotlin. Requires deprecation of `Arb.ushort`.

BartvHelvert:
Improves the naiming of some of the Arb functions to be more idiomatic. The following names are changed:
* bool -> boolean
* ~~edgecase -> edgeCase~~
* ubyte -> uByte, ushort -> uShort etc...

Also fixes some typos in `genslist.md` and fixed deprecation replacements.

Closes #2305.

### DIFF
--- a/documentation/docs/changelog.md
+++ b/documentation/docs/changelog.md
@@ -72,7 +72,7 @@ See the full list of [extension modules](https://kotest.io/docs/extensions/exten
 * In order to ensure the `EventuallyListener` is called in `eventually` when an exception is thrown the `ListenerState` field `result` was changed
   from type `T` to type `T?`. This will allow insight into when the eventually producer function is failing for whatever reason
   instead of appearing as if it is hanging. #2190
-* Property tests now randomly cycle between edgecases and samples, rather than iterating all edgecases first. This allows greater number of edgecases to be used and avoids a combinatoral explosion. If you are implementing custom Arb's by extending the Arb class (instead of using the `arbitrary` builders), then you will need to adjust your edgecases method from `fun edgecases(): List<A>` to `fun edgecase(rs: RandomSource): A?`.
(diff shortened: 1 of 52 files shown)


sksamuel:
I can get behind the types changing, but please add deprecated forwarding methods or type aliases.
Also I don't think we should rename edgecases. We only just went through a round of deprecations and changes on the way edgecases work and I think it's ambiguous whether its one word or two anyway.

BartvHelvert:
> I can get behind the types changing, but please add deprecated forwarding methods or type aliases.

I think all of them have deprecation forward methods? The methods that don't have it didn't exist in 4.6.0 so I don't think it is worth adding deprecation warnings for those methods. If you think I missed some, let me know.

> Also I don't think we should rename edgecases. We only just went through a round of deprecations and changes on the way edgecases work and I think it's ambiguous whether its one word or two anyway.

I can't find any usage of `edgecase` as a single word online so I don't think it is really ambiguous. But if it is to much of a breaking change I will remove it from this PR.

sksamuel:
> > I can get behind the types changing, but please add deprecated forwarding methods or type aliases.
> 
> I think all of them have deprecation forward methods? The methods that don't have it didn't exist in 4.6.0 so I don't think it is worth adding deprecation warnings for those methods. If you think I missed some, let me know.

Ah ok I must have missed them first time around. I see them now.

> > Also I don't think we should rename edgecases. We only just went through a round of deprecations and changes on the way edgecases work and I think it's ambiguous whether its one word or two anyway.
> 
> I can't find any usage of `edgecase` as a single word online so I don't think it is really ambiguous. But if it is to much of a breaking change I will remove it from this PR.

Yeah you're probably right but I feel like it's one of those words in English that works both ways. Either way, I'm reluctant to change it again after we just changed it in 4.4.0. For something that most people won't use directly, I feel the benefit is low.

PR #2311, opened 2021-06-02
Made MapEq use failure(expected, actual, msg)

Kantis:
Assertion failures when comparing maps using `shouldBe` should now show diff using IntelliJ

Prior to this PR:
image

With changes:
image

Diff window:
image

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/MapEq.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/MapEq.kt
@@ -1,12 +1,16 @@
 package io.kotest.assertions.eq
 
+import io.kotest.assertions.Actual
+import io.kotest.assertions.Expected
 import io.kotest.assertions.failure
+import io.kotest.assertions.show.Printed
 import io.kotest.assertions.show.show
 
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/MapEq.kt
```diff
@@ -33,26 +41,28 @@ private fun generateError(actual: Map<*, *>?, expected: Map<*, *>?): Throwable {
...
    val newLine = "\n"
-   return toList()
+   return Printed(toList()
       .joinToString(
          separator = ",$newLine",
          prefix = "{$newLine",
          postfix = "$newLine}",
          limit = 10
       ) { "$indentation${it.first.show().value} = ${it.second.show().value}" }
+   )
```
Kantis, line 66: Would it make more sense to just use `actual.show()` ? I saw there's an implementation of `MapShow` which I tried to use, but I felt this formatting was more readable.

sksamuel: We don't have to use show, if we feel this looks better.

sksamuel:
@ashishkujoy happy with this ?

ashishkujoy:
Looks good to me

PR #2315, opened 2021-06-06
Fix throw handling return type

zmunm:
It seems more appropriate to set it to Unit, as it doesn't use the lambda's return.

I have a simple case where this was a problem.

```kotlin
shouldThrow<IllegalStateException> {
    val a = error("")
}

// and

var a: Int
shouldThrow<IllegalStateException> {
    a = error("")
}
```

But I didn't know if it was appropriate to actually add this to test code.
If you comment on this, I will take it.


sksamuel:
It was the case that you couldn't pass a lambda that returned T to a function that required () -> Unit.
So we created `shouldThrowUnit` for this purpose.
This may have changed in 1.5, if this is now working ?

zmunm:
`shouldThrowUnit` works fine. Thank you!
 It would be nice if it was simply added to the [exception document](https://kotest.io/docs/assertions/exceptions.html).

sksamuel:
I will run some checks to see if kotlin has improved lamba return types in 1.50 and update docs.

Issue #2145, opened 2021-03-05
Why does shouldHaveSingleElement and other Collection "shoulds" have predicate instead of block for assertions?


Zordid:
What I do not really get is: when does your library want predicates (T)->Boolean and when does it require a function block that itself can do multiple assertions using matchers and should constructs?

I.e.:
```
list.shouldHaveSingleElement {
   it.name == "Bob"
}
```

vs. (not the case)
```
list.shouldHaveSingleElement {
  it.name shouldBe "Bob"
}
```

I've stumbled across both ways, but the simple predicate one is used for collections - and I do not like it at all, because the error messages don't reveal any reason for failing the match whereas the latter one could state that something was expected to be "Bob" but is "Alice" for example.

Is there a reason behind this that I cannot see?

sksamuel:
It's probably just because the project is large and we have a smaller number of regular committers so there will be some areas that miss functionality like this. If you want to make a PR that would be great, or if you have a list of matchers you want the block for we'll add them.

Kantis:
@sksamuel maybe we should deprecate these with 5.0 and remove in 6.0? For this particular case, I think the proper method to use would be

```kotlin
list.forOne {
  it.name shouldBe "bob"
}
```

sksamuel:
Yeah the inspectors are what is meant to be used.

We should rename the first one to include the word predicate in it. That
indicates its boolean.

sksamuel:
Let's deprecate any block that is essentially doing the same as the inspectors.
But we should keep blocks that have a predicate, as it's nice for one liners, but maybe converge on the name shouldExist ?
`list.shouldExist { it.age == 90 }`

Kantis:
I was wrong about replacing with `forOne`, sadly 

`forOne` asserts there's one, and only one, element in the collection that matches the block.
`singleElement` checks that the subject is a collection with a single item, and that the single item matches the predicate.

So the usage would not be equivalent. 


On top of that, I found these matchers which accept predicates.
`existInOrder` checks that, given a list of predicates, they are fulfillfed in the correct order.
`exist` checks that any item fulfills the predicate. 


Suggest we proceed with:

- [x] Deprecate `exist`, in favor of `forAny`
- [x] ~Add `singleElement` matcher which accepts a block, and deprecate predicate version~
- [x] Add `forSingle` inspector
- [x] Add `existInOrder` matcher which accepts a list of blocks, and deprecate predicate version

[...]

PR #2321, opened 2021-06-10
Convert inspectors to matchers/asserts

Kantis:
Inspectors could possibly be considered _just another assertion_ and be named using the regular `should...` syntax rather than `for..`.

Related issues: #2145 

#### Improved discoverability
When starting an assertion it feels like common practice to being with `subject should` and look for possible assertions. Inspectors will never show up when looking at the API this way

#### Erases the inspector concept
Inspectors become just another assertion. Less concepts to learn and a beginner will hopefully have an easy time picking up what these assertions do.

#### Changes an established concept
Old users will have to change habits and possibly migrate old code at some point. Aliases for the old inspector names could still be provided until Kotest 6.0 perhaps

### Suggested new names:
`forAll` -> `shouldAll`  -  asserts every element passes 
`forNone` -> `shouldContainNone` -  asserts no element passes (alternative name: `shouldNotContain`?)
`forOne` -> `shouldContainOne` -  asserts only a single element passed
`forAtMostOne` -> `shouldContainAtMostOne` -  asserts that either 0 or 1 elements pass
`forAtLeastOne` -> `shouldContainAtLeastOne` - asserts that 1 or more elements passed
`forAtLeast(k)` -> `shouldContainAtLeast(k)` -  asserts that k or more elements passed
`forAtMost(k)` -> `shouldContainAtMost(k)` -  asserts that k or fewer elements passed
`forAny` which is an alias for `forAtLeastOne`  ->  `shouldContainAtLeastOne`
`forSome` -> `shouldContainSome` - asserts that between 1 and n-1 elements passed. Ie, if NONE pass or ALL pass then we consider that a failure.
`forExactly(k)` -> `shouldContain(k)` - asserts that exactly _k_ elements pass

Examples:
```kotlin
val subjects = listOf(person(), person())
subjects.shouldAll {
  it.age shouldBeInRange 18..65
  it.name shouldBeIn listOf("hans", "greta")
}
```

I also tried making matchers out of the inspectors, to unify them with all other assertions.

Opening this PR to gather feedback on what people would feel about a change.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
@@ -13,7 +13,6 @@ fun <T> Iterable<T>?.shouldContainExactly(vararg expected: T) = this?.toList() s
 fun <T> Array<T>?.shouldContainExactly(vararg expected: T) = this?.asList() should containExactly(*expected)
 
 infix fun <T, C : Collection<T>> C?.shouldContainExactly(expected: C) = this should containExactly(expected)
-fun <T> Collection<T>?.shouldContainExactly(vararg expected: T) = this should containExactly(*expected)
 
 fun <T> containExactly(vararg expected: T): Matcher<Collection<T>?> = containExactly(expected.asList())
 
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/inspectors/asserts.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/inspectors/asserts.kt
@@ -1,0 +1,36 @@
+package io.kotest.matchers.collections.inspectors
+
+import io.kotest.matchers.should
+
+//internal typealias MatcherBlock<T> = (T) -> Unit
+fun interface MatcherBlock<T> {
+   operator fun invoke(element: T): Unit
+}
+
+fun <T> Array<T>.shouldContain(times: Int, block: MatcherBlock<T>) = asList() should containExactly(times, block)
+fun <T> Iterable<T>?.shouldContain(times: Int, block: MatcherBlock<T>) = this should containExactly(times, block)
+
+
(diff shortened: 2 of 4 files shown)


sksamuel:
Personally I like the inspector naming because it just seems to read a lot nicer.

`subjects.forAtMostOne `

vs

`subjects.shouldContainAtMostOne` is pretty ugly to me.

forAll, forNone, forOne roll off the tongue much nicer than shouldAll, shouldContainNone, shouldContainOne.

But I can see the argument they are not as discoverable and probably not that many people use inspectors because of that.

sksamuel:
For nomenclature, what about

`shouldForAll`
`shouldForOne`
`shouldForAtLeastOne`
`shouldForNone`

etc

Kantis:
> For nomenclature, what about
> 
> `shouldForAll`
> `shouldForOne`
> `shouldForAtLeastOne`
> `shouldForNone`
> 
> etc

I think that's a great idea. Makes it a lot easier to migrate usages

nilsschmidthamburg:
I totally agree on the point of discoverability. I my self discover(ed) 90% of the matchers by typing 'should' + autocomplete and look for the best fit. That's why I missed the inspectors like forAll, forAtLeastOne etc. in the first place. 

On the other hand, once I discovered inspectors, I love the way how these read at the moment. Look at this beauty:
`val subjects = listOf(person(), person())
subjects.forAtLeastOne { person ->
  person.age shouldBeInRange 18..65
}
`

In my personal opinion, I wouldn't be very happy with something along the lines of `shouldForAtLeastOne` as it doesn't read that well. I would rather prefer `shouldContainAtLeastOne`. The downside is that we would replace 3 letters (for) with 13 letters (shouldContain) in every single place.

To conclude, I would opt for `shouldContain....` as the new default, and deprecate the old `for...` versions. But I am happy to see more feedback from the community.

Looking at the PR, the only version I am not so sure about is `shouldAll`, but I couldn't come up with a better name so far :-/

nilsschmidthamburg:
Regarding the PR, we should also adapt documentation in [documentation/docs/assertions/inspectors.md](https://github.com/kotest/kotest/blob/master/documentation/docs/assertions/inspectors.md)

sksamuel:
I think our best idea so far is to add aliases as shouldXXX but then have the kotest intellij plugin suggest you rename to forXXX

PR #2326, opened 2021-06-17
Update test_functions.md

niqdev:
Hi,
`shouldHaveLength` is defined as `String?.shouldHaveLength(length: Int): String?` and the example is not compiling.
Thanks for the great lib!


sksamuel:
Thanks

PR #2330, opened 2021-06-20
Fix Arb.long/ulong producing values outside range

### LINKED ISSUE #2327: `Arb.long/Arb.uLong` produce values outside `min/max`
Can be reproduced by `Arb.long(-140_737_488_355_328, 140_737_488_355_327)`, just take a large amount of values and check if it's inside`[-140_737_488_355_328, 140_737_488_355_327]`.  The same goes for `Arb.uLong`.

This issue seems to be caused by [KT-47304](https://youtrack.jetbrains.com/issue/KT-47304).
Should we implement a workaround until it's fixed?

BartvHelvert:
Also fixes some naming mistakes in tests. Closes #2327.


sksamuel:
Nice.

Issue #2312, opened 2021-06-03
Change project config based beforeAll / afterAll to be suspendable


sksamuel:
This would be a minor breaking change for 5.0

PR #2333, opened 2021-06-25
#2312 Makes beforeAll and afterAll suspend functions

ashishkujoy:


sksamuel:
This is a breaking change. What about deprecating the old names, and adding beforeProject / afterProject as suspend equivalents ?

ashishkujoy:
> This is a breaking change. What about deprecating the old names, and adding beforeProject / afterProject as suspend equivalents ?

Sure make sense.

sksamuel (changes requested):
Looks good but we should still call both methods from the created config, otherwise people's listeners will silently start to fail in 5.0.

sksamuel (approved):

PR #2339, opened 2021-07-06
update samples of usage of tags

seepick:
- replacing deprecated `StringTag` with `NamedTag`
- to define tags, passing `vararg` instead of `Set`


sksamuel:
Nice spot thanks.

PR #2356, opened 2021-07-12
Property testing should use bind as default for data class #2355

### LINKED ISSUE #2355: Property testing should use bind as default for data class

sksamuel:
Closes #2355


myuwono:
Whoa @sksamuel is this a wrong rebase?

sksamuel:
Wow this is a mess yeah.
I'll make a new one.
The idea is that for a data class it should be able to populate the fields from the defaults arbs for types.

Issue #2293, opened 2021-05-26
shouldContainXXX fails for list of ByteArrays with confusing message


JacekWicka:
Kotest 4.6.0

```kotlin
   //StringSpec
   "test collection" {
        val sut = listOf<Any>("one".toByteArray(), "two".toByteArray())
        sut shouldContainExactly listOf<Any>("one".toByteArray(), "two".toByteArray())
    }
```
results in:
```
Expecting: [
java.lang.AssertionError: Expecting: [
  [111, 110, 101],
  [116, 119, 111]
] but was: [
  [111, 110, 101],
  [116, 119, 111]
]
Some elements were missing: [
  [111, 110, 101],
  [116, 119, 111]
] and some elements were unexpected: [
  [111, 110, 101],
  [116, 119, 111]
]
```

sksamuel:
^^ what was the status on this @ashishkujoy . I know we discussed it on slack.

ashishkujoy:
Not getting enough time to actively work on this, last time i was to make it work for `shouldContainExactly` but was not able to make it work for `shouldContainExactlyInAnyOrder`

ashishkujoy:
Maybe i should at least push the `shouldContainExactly` fix.

sksamuel:
Fixed in master

PR #2360, opened 2021-07-12
#2293 Fixes ShouldContainExactly for collection containing byte arrays

ashishkujoy:
For ShouldContainExactly to work `ByteArray` , `ShortArray` etc we have to use the `eq` instead of using default `equals` .
One downside of using `eq` is that it has `NumberEq` which try to do type change for comparing number of different type, for example if we are comparing a `Int` with a `Long` it will convert the `Long` to `Int` and then check equality using  default `equals`. This number conversion is not desired for `ShouldContainExactly`. 
https://github.com/kotest/kotest/blob/0ba14983d485b1b1caef0e5a2cd7402ba47e34d2/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/ShouldContainExactlyTest.kt#L76-L95

So i order to make it work without changing the current behaviour of `shouldBe` (internally uses `eq`) and `ShouldContainExactly` I added a flag `strictNumberEq` to control conversion of number type in `NumberEq` . The default value for this is kept `false`.

One downside of this is that all implementation of need `Eq` need to keep forwarding `strictNumberEq` whenever they internally call `eq`.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
@@ -1,5 +1,6 @@
 package io.kotest.matchers.collections
 
+import io.kotest.assertions.eq.eq
 import io.kotest.assertions.show.Printed
 import io.kotest.assertions.show.show
 import io.kotest.matchers.*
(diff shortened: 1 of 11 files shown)


ashishkujoy:
This changes still not sufficient to make it work for `shouldContainExactlyInAnyOrder` as `shouldContainExactlyInAnyOrder` internally creates a map of all elements as key and number of their occurrence as value. Now it uses keys from one map and equate value for that key from both the map.  For one of the map it always get the null value as hash code of ByteArray are not equal.

https://github.com/kotest/kotest/blob/0ba14983d485b1b1caef0e5a2cd7402ba47e34d2/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactlyInAnyOrder.kt#L104-L116

sksamuel (approved):

PR #2365, opened 2021-07-14
Rename resp to response to make code compilable

jvmusin:
Code in the last section was non-compilable because of different names for response


sksamuel:
Nice spot thanks

Issue #2369, opened 2021-07-18
minDate not respected in Arb.localDate


unaryops:
**Which version of Kotest are you using**
Version 4.6

This [code](https://github.com/kotest/kotest/blob/e9c046b2ffd7f60ced7a6aea0b0014cd6f484b7c/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt#L52) doesn’t take into account the date/month portion of the specified minDate . So when you specify minDate that is after the default edge case date, it will (intermittently) fail. Here’s an example `Arb.localDate(minDate = LocalDate.now().plusDays(10))` will generate a `2021-02-28` date

PR #2370, opened 2021-07-19
Fix issue #2369 with Arb.localDate

unaryops:
Arb.localDate doesn’t take into account the date/month portion of the
specified minDatex. So when you specify minDate that is after the
default edge case date, it will (intermittently) fail. Here’s an example
Arb.localDate(minDate = LocalDate.now().plusDays(10)) will generate a
2021-02-28 date.


sksamuel:
👍🏼

PR #2376, opened 2021-07-21
Add shouldThrowMessage matcher

faloi:
I propose to add this here just because `shouldThrow`, `shouldThrowAny` and the rest were here and I didn't find this one. Besides this proposal, I believe it could be clearer to move all throwable assertions to another subtitle, but that's up to you. 😄

BTW, thanks for this great framework! I've been using it in a course of software design and the DSL you created is very natural for my students.


sksamuel:
Thanks for your kind words :)

Issue #2378, opened 2021-07-22
Proposal for a better DoubleShrinker.


sielenk-yara:
The Kotest 4.6.1 `DoubleShrinker` does not produce values I would consider "simpler" than the input:

```
val f = 785040.1538306947
val x = DoubleShrinker.shrink(f)
println(f)
println("------")
x.forEach { println(it) }
```
produces
```
785040.1538306947
------
0.0
1.0
-1.0
785040.1538306947
261680.05127689824
392520.07691534737
785035.1538306947
785036.1538306947
785037.1538306947
785038.1538306947
785039.1538306947
785040.0
```
where the first three values are covered by the normal edge cases and most of the rest I would consider about as complicated as the input. 

I propose something like this
```
fun shrink(value: Double): List<Double> {
  if (!value.isFinite())
    return emptyList()
  val (mantissa, exponent) = "%+.17e".format(Locale.ROOT, value).split('e')
  val trimmedMantissa = mantissa.trimEnd('0')
  val shrunkStrings = (3 until trimmedMantissa.length).reversed()
    .map { "${trimmedMantissa.substring(0, it)}e$exponent" }
  return shrunkStrings.map { it.toDouble() }
}
```
which produces
```
785040.1538306947
------
785040.153830694
785040.15383069
785040.1538306
785040.15383
785040.15383
785040.1538
785040.153
785040.15
785040.1
785040.0
785040.0
785000.0
785000.0
780000.0
700000.0
```
I would find this more helpful when looking for "simpler" values.

The same applies to the `FloatShrinker`.

sielenk-yara:
I just tried to create a pull request for this and found that this only works on the JVM platform!

Maybe I can come up with a non-string based solution for this...

sielenk-yara:
I think I misunderstood the `Shrinker` interface:
`shrink` should not return the whole list of simpler numbers, but only the next one: `shrink(785040.15) shouldBe listOf(785040.1)`.

sksamuel:
Did you get any further with this @sielenk

sielenk:
Sadly no, not yet.
And missed the 5.0 deadline ...

sksamuel:
No 5.0 is still in flight.

sielenk:
[...]

PR #2379, opened 2021-07-22
updated FloatShrinker and DoubleShrinker to regard mantissa bit count…

sielenk-yara:
… as complexity measure

This is patch for #2378.

It does not realize a simplification base 10, though.


sksamuel:
Great stuff, thanks for fixing.

sielenk:
You're welcome (this is also me, sielenk-yara is my company account).

You do realize that this does _not_ what is mentioned in #2378, though?
I am still thinking about getting it done platform-independently in base 10  but this is not yet it.

That said, I do think this is an improvement over the old shrinker which was kind of all over the place - but that is the task of the `Arb`!

sksamuel:
yes I think this is fine for a JVM solution.  I will include this in 4.6.2

sielenk:
This works on all platforms, not just JVM!

sksamuel:
Then I'm just getting confused going over all the outstanding messages lol
:)

PR #2390, opened 2021-07-30
improved geo location generator

sielenk:
This morning I noticed that the `Arb.latlon` generator does not produce proper geo coordinates:
- latitude should be in range -90°...+90° and
- it should not be uniformly distributed in its range for the points on the sphere to be uniformly distributed

While updating I also added a dedicated class for geo locations with descriptive property names and degree and radian values.
Plus backward compatibility code with deprecation annotations with `ReplaceWith` hints.


sksamuel:
This is great, thanks.

PR #2395, opened 2021-08-01
Fix autoClose lazyness

### LINKED ISSUE #2388: autoClose breaks Lazy

1.4.6

[Here](kotest/core/TestConfiguration.kt#L118) Lazy is immediately invoked which initializes the Lazy value, to then pass it to the immediate autoClose function which rewraps in Lazy [here](https://github.com/kotest/kotest/blob/master/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/TestConfiguration.kt#L110).

If delegation happens to other way around, meaning `fun <T : AutoCloseable> autoClose(closeable: T): T` uses `fun <T : AutoCloseable> autoClose(closeable: Lazy<T>): Lazy<T>` than Lazy'ness can be preserved.

When closing the `AutoCloseable` in `invokeAfterSpec` you can check `Lazy#isInitialised` before calling `lazy.value.close()` to avoid initializing values that are not used, and thus don't need to be closed.

nomisRev:
Closes #2388

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/TestConfiguration.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/TestConfiguration.kt
@@ -107,7 +107,7 @@ abstract class TestConfiguration {
     */
    @Suppress("PropertyName")
    fun <T : AutoCloseable> autoClose(closeable: T): T {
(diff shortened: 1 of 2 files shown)


sksamuel:
Should we keep a test that autoclosables not wrapped in lazy by the user are still closed ?

nomisRev:
I'll re-add it so we cover both functions.

sksamuel:
👍🏼

PR #2403, opened 2021-08-05
Add BigDecimal edge case for equals vs compareTo discrepancy

dchenbecker:
BigDecimal is not consistent for equals and compareTo when the
scale differs. This adds an edge case using the String constructor for
BigDecimal to force a different scale.


dchenbecker:
I'm looking for some guidance about how to best unit test this. My expectation is that with a proper seed I could get a 2-ary function to get both edge cases, but I'm not sure how simple that is, or if there's another approach to take here.

sksamuel:
You could just call the edge cases method multiple times and assert you get both of the 0.0 and 0.00 in the output.

sksamuel:
Something like `Arb.bigDecimal().edgecases().contains(BigDecimal("0.00"))` ought to do it.

sksamuel:
👍🏼

PR #2404, opened 2021-08-06
Add Percentage tolerance matchers

### LINKED ISSUE #2232: Add tolerance matcher to Float, Int and Long
We only have tolerance matcher for Double at the moment.
We could probably use the fact that all numbers extend from `Number`

LeoColman:
Closes #2232

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/doubles/Tolerance.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/doubles/Tolerance.kt
@@ -2,6 +2,8 @@ package io.kotest.matchers.doubles
 
 import io.kotest.matchers.Matcher
 import io.kotest.matchers.MatcherResult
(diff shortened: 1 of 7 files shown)


sksamuel (approved):

LouisCAD:
Which version has these?
I'm unable to find `shouldBeWithinPercentageOf` in kotest 4.1.3

LouisCAD:
Wait, did the maven coordinates change again? 🤔 

**EDIT:** Ok, I wasn't seeing newer versions because I was sharing the version with junit4 runner which apparently is not longer receiving updates.

sksamuel:
junit4 is still being released, https://search.maven.org/artifact/io.kotest/kotest-runner-junit4-jvm/5.5.0/jar

`shouldBeWithinPercentageOf` is in the assertions core package which is still at https://search.maven.org/search?q=a:kotest-assertions-core-jvm

The only time we did a major change of coordinates was when 5.0 came out last year, and even then it was only the extensions packages for spring, ktor, etc.

PR #2408, opened 2021-08-09
Added IPAddressV4 arb #2407

sksamuel:


myuwono (approved):

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/network.kt
```diff
@@ -0,0 +1,16 @@
+package io.kotest.property.arbitrary
+
+import io.kotest.property.Arb
+
+/**
+ * Returns an [Arb] where each generated value is an IP Address in V4 format represented as a String.
+ * Each part is a number between 0 and 255 except for the first part which is 1-255.
+ */
+fun Arb.Companion.ipAddressV4(): Arb<String> = arbitrary {
```
myuwono, line 9: We should probably add some simple tests around it. Also there's a class called inetaddress, we have a similarly built arb in house that generates both v4 and v6 with that class as a type. Just a thought.

sksamuel: Probably better yeah, I can update.
Or you could PR your in house class :)

myuwono: Haha can do, I'll raise one. I like your idea to use string as type. Converting that to Inetaddress is trivial. Or do you mean we want to have the arb for Inetaddress type as well?

sksamuel: I picked Strings because that was what I wanted, but I guess the proper type is better. Can always call toString on it.

sksamuel: I think the string is best because Inetaddress in java won't be MPP.

sksamuel: Tests added

Issue #2409, opened 2021-08-09
Hex codepoints arb


sksamuel:
**Please describe the feature you'd like to see including any solutions in mind if you have any**

PR #2410, opened 2021-08-09
Added hex codepoints arb #2409

sksamuel:


myuwono (approved):

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/codepoints.kt
```diff
@@ -105,6 +105,13 @@ fun Codepoint.Companion.ascii(): Arb<Codepoint> =
       .map { Codepoint(it) }
       .withEdgecases(Codepoint('a'.code))
 
+/**
+ * Returns an [Arb] that generates HEX codepoints.
+ */
+fun Codepoint.Companion.hex(): Arb<Codepoint> =
```
myuwono, line 111: We should test this as well, a simple `java.lang.Long.parseLong(hexString, 16)` will work i believe

sksamuel: Oh yeah I meant to come back and add tests.
I'll make another PR with tests.

PR #2413, opened 2021-08-13
#2412 Makes Sequence.containExactly work for single pass sequences

### LINKED ISSUE #2412: Sequence.containExactly should work for single pass sequences

ashishkujoy:
closes: #2412

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
@@ -2,6 +2,8 @@
 
 package io.kotest.matchers.sequences
 
+import io.kotest.assertions.eq.eq
+import io.kotest.assertions.show.show
 import io.kotest.matchers.*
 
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
```diff
@@ -79,17 +81,25 @@ fun <T> containExactly(vararg expected: T): Matcher<Sequence<T>?> = containExact
...
+      if (eq(actualElement.value, expectedElement.value) != null) {
+         failMessage += " (expected ${expectedElement.value.show().value} at ${expectedElement.index} but found ${actualElement.value.show().value})"
          passed = false
-         failMessage += " (expected ${diff.value.second} at ${diff.index} but found ${diff.value.first})"
       }
-   } else {
-      failMessage += " (expected ${expected.count()} elements but found ${value.count()})"
    }
+
+   if (actualIterator.hasNext() || expectedIterator.hasNext()) {
```
sksamuel: We could split this into two if statements and then say "actual had more elements" or "expected had more elements"

ashishkujoy: done

PR #2419, opened 2021-08-18
Add NaN matchers to Float

ajalt:
Currently, there are some assertion matchers that exist for Double but not Float. #2404 added some, and this PR adds some more: namely, the matchers relating to NaN.


ajalt:
Given the discrepancies between Double and Float support, it might be worth replacing the Float matchers entirely with a copy of the Double matchers rather than doing piecemeal changes like this, but that's out of scope for this PR.

PR #2426, opened 2021-08-21
Added classifier idea #2267

### LINKED ISSUE #2267: Automatic classification of values in property tests
https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1621517644064900?thread_ts=1621512425.062800&cid=CT0G9SD7Z

sksamuel:
Closes #2267

@myuwono if you have time, would appreciate some validation on the ideas here.

The idea is that Gens can be extended with a classifier, (provided by default by Kotest for common types), and providable by the user for other types.

The classifier labels each value (eg POSITIVE_INT) and then if we have enabled reporting, will use a `LabelsReporter` to output the results.

This allows us to have automatic labeling for common types instead of the manual only approach now.

Eg output:

```
Label statistics for arg 0 (1000 inputs):
ANY LENGTH LETTER OR DIGITS                                    13 (1%)
SINGLE CHARACTER LETTER                                        13 (1%)
EMPTY STRING                                                   11 (1%)
MAX LENGTH                                                     11 (1%)
OTHER                                                         952 (95%)

Label statistics for arg 1 (1000 inputs):
POSITIVE EVEN                                                 242 (24%)
NEGATIVE ODD                                                  254 (25%)
NEGATIVE EVEN                                                 237 (24%)
MIN                                                             5 (1%)
POSITIVE ODD                                                  254 (25%)
ZERO                                                            6 (1%)
MAX                                                             2 (0%)
```


myuwono:
this is awesome @sksamuel !!

sksamuel:
thanks :)

PR #2434, opened 2021-08-24
Bugfix/2380 arb filter shrinks

### LINKED ISSUE #2380: Problem with shrinking filter
**Which version of Kotest are you using** 4.6.1

I tried to do property based testing but got a wrong shrinking number with filter
I got 1 as a result of the shrinking but it should not be a result according to the filter

```
class ShrinkTest : WordSpec(  {
    "Arb int with filter even should not have 1 with shrink" should {
        checkAll(Arb.int().filter { it % 2 == 0 }) { a ->
            a shouldBe 0
        }
    }
})
```

myuwono:
closes #2380


sksamuel (approved):

sksamuel:
Nice.

Issue #2438, opened 2021-08-25
Property Module: Allow global seed configuration


nicorichard:
**Please describe the feature you'd like to see including any solutions in mind if you have any**
While there are undoubted benefits of running tests with different values every time, in practice our team prefers to favour a more reliable approach.

Such that:
* Issues do not crop up in CI which were not produced locally by the developer
* A developer doesn't need to be intimately familiar with the kotest to reproduce an issue found in CI

**Implementation**
To accomplish this I would suggest the inclusion of a `defaultSeed` parameter to the global PropertyTesting configuration object. This value could be transcribed to the default parameters of PropTestConfig if present.

PR #2439, opened 2021-08-25
Property Module: Allow global seed configuration. Synchronize defaults.

nicorichard:
Implementation for enhancement request #2438 (which I raised myself, and is open to critique)

* Synchronized naming across PropertyTesting parameters which ultimately affect the `PropTestConfig` data class
* Synchronized default values for `PropTest`, `PropTestConfig`, and the global `PropertyTesting` object
* Added defaults configuration for the other `PropTestConfig` values
* Correct spelling for `outputClassifiations`. This is a recent addition so I'm not sure if it amounts to a breaking change yet? **Correct me if I'm wrong please**
* Deprecated configuration values not following the pattern
* Started the documentation process for these values

_Note: I have not added `sysprop` declarations for the new configurations as I am not sure what the exact pattern should be for their namespaces_

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
@@ -12,12 +12,27 @@ import kotlin.native.concurrent.ThreadLocal
 @ThreadLocal
 object PropertyTesting {
    var maxFilterAttempts: Int = 10
-   var shouldPrintGeneratedValues: Boolean = sysprop("kotest.proptest.output.generated-values", "false") == "true"
-   var shouldPrintShrinkSteps: Boolean = sysprop("kotest.proptest.output.shrink-steps", "true") == "true"
+
+   // PropTestConfig
+   var defaultSeed: Long? = null
+   var defaultMinSuccess: Int = Int.MAX_VALUE
+   var defaultMaxFailure: Int = 0
+   var defaultShrinkingMode: ShrinkingMode = ShrinkingMode.Bounded(1000)
    var defaultIterationCount: Int = sysprop("kotest.proptest.default.iteration.count", "1000").toInt()
-   var edgecasesGenerationProbability: Double = sysprop("kotest.proptest.arb.edgecases-generation-probability", "0.02").toDouble()
+   var defaultListeners: List<PropTestListener> = listOf()
+   var defaultEdgecasesGenerationProbability: Double = sysprop("kotest.proptest.arb.edgecases-generation-probability", "0.02").toDouble()
+   @Deprecated("Use defaultEdgecasesGenerationProbability instead. This property will be removed")
+   var edgecasesGenerationProbability: Double
+      get() = defaultEdgecasesGenerationProbability
+      set(value) { defaultEdgecasesGenerationProbability = value }
+   var defaultOutputClassifications: Boolean = sysprop("kotest.proptest.arb.output.classifications", "false") == "true"
+
+   var shouldPrintShrinkSteps: Boolean = sysprop("kotest.proptest.output.shrink-steps", "true") == "true"
+   @Deprecated("This property is no longer used and will be removed")
+   var shouldPrintGeneratedValues: Boolean = sysprop("kotest.proptest.output.generated-values", "false") == "true"
+   @Deprecated("This property is no longer used and will be removed")
    var edgecasesBindDeterminism: Double = sysprop("kotest.proptest.arb.edgecases-bind-determinism", "0.9").toDouble()
-   var outputClassifiations: Boolean = sysprop("kotest.proptest.arb.output.classificaions", "false") == "true"
+
 }
 
 /**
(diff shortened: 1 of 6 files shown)


kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
```diff
@@ -71,15 +84,22 @@ fun PropTest.toPropTestConfig() =
...
+   val seed: Long? = PropertyTesting.defaultSeed,
+   val minSuccess: Int = PropertyTesting.defaultMinSuccess,
+   val maxFailure: Int = PropertyTesting.defaultMaxFailure,
+   val shrinkingMode: ShrinkingMode = PropertyTesting.defaultShrinkingMode,
    val iterations: Int? = null,
-   val listeners: List<PropTestListener> = listOf(),
+   val listeners: List<PropTestListener> = PropertyTesting.defaultListeners,
    val edgeConfig: EdgeConfig = EdgeConfig.default(),
-   val outputLabels: Boolean? = null,
+   val outputClassifications: Boolean = PropertyTesting.defaultOutputClassifications,
```
sksamuel, line 100: If you set this to the default here, how would you override globally, because this field will always have a value.

kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
```diff
@@ -12,12 +12,25 @@ import kotlin.native.concurrent.ThreadLocal
 @ThreadLocal
 object PropertyTesting {
    var maxFilterAttempts: Int = 10
-   var shouldPrintGeneratedValues: Boolean = sysprop("kotest.proptest.output.generated-values", "false") == "true"
    var shouldPrintShrinkSteps: Boolean = sysprop("kotest.proptest.output.shrink-steps", "true") == "true"
-   var defaultIterationCount: Int = sysprop("kotest.proptest.default.iteration.count", "1000").toInt()
-   var edgecasesGenerationProbability: Double = sysprop("kotest.proptest.arb.edgecases-generation-probability", "0.02").toDouble()
+   @Deprecated("This property is no longer used and will be removed")
+   var shouldPrintGeneratedValues: Boolean = sysprop("kotest.proptest.output.generated-values", "false") == "true"
```
sksamuel: Why is this deprecated ?

kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
```diff
@@ -12,12 +12,25 @@ import kotlin.native.concurrent.ThreadLocal
...
+   @Deprecated("This property is no longer used and will be removed")
    var edgecasesBindDeterminism: Double = sysprop("kotest.proptest.arb.edgecases-bind-determinism", "0.9").toDouble()
```
sksamuel: Same question

sksamuel:
Thanks for the PR. I've added some comments.

kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
nicorichard: It is not used anywhere in the source.
It likely can be deleted! I did not want to impose any contact changes on old code.

kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
nicorichard: It was my understanding that the globals were defaults not overrides. Please correct me if I was wrong.

For instance: 

`defaultIterationCount` will never override the value you pass to `forAll<Int>(42)`, nor `forAll<Int>(PropTestConfig(iterations = 42))` but it will be used as default for `forAll<Int>` or `forAll<Int>(PropTestConfig(seed = 42))`

In this case of `defaultOutputClassifications` I've moved from the code from `commonMain/kotlin/io/kotest/property/classifications/output.kt`:

```
val enabled = config.outputLabels ?: PropertyTesting.outputClassifiations
```

directly to the PropTestConfig initialization

```
val outputClassifications: Boolean = PropertyTesting.defaultOutputClassifications
```

folded out:

```
val outputClassifications: Boolean = sysprop("kotest.proptest.arb.output.classifications", "false") == "true"
```

Which should have all the same results as the previous code (besides the name change)

* `PropTestConfig(outputClassifications = VALUE)` will respect `PropTestConfig`'s VALUE
* otherwise check `sysprop` and default to `false`

kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
sksamuel: Oh it should be used. I think leave it and create an issue to use it.

kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
sksamuel: I guess to me there is a default, a global and an instance value.
If you don't set the instance then it goes to the global, which itself uses the default if not set.
I suppose the default can be collapsed inline.

sksamuel:
> Note: I have not added sysprop declarations for the new configurations as I am not sure what the exact pattern should be for their namespaces

I think kotest.proptest.anything.reasonable

PR #2464, opened 2021-09-06
Compare JSON literals in Strict mode

### LINKED ISSUE #2458: shouldEqualJson in Strict mode should compare raw values
4.6.2

The following two tests currently fail using the default comparison mode `CompareMode.Strict`. Imo `CompareMode.Strict` should simply compare the value as a string and presence of quotation.

```
      test("comparing doubles") {
         "3.123456789123457" shouldNotEqualJson "3.123456789123456789123456789"
      }

      test("comparing int and float") {
         "3.2" shouldNotEqualJson "3"
      }
```

Kantis:
Aim is to fix #2458 

It got a bit more complicated than simply checking exact equal string representations in Strict mode. Given that `1.0E2` is the same number as `100.0`, `100` or `+1E+2` I think they should be considered equal. What do you think?

[The JSON standard](https://www.json.org/json-en.html) seems to not make any distinction between integer and floating point numbers. 

TODO:
- [x] ~Changelog?~
- [x] Update relevant docs

Side-note: Do you have a good way of writing tests that targets both JVM and JS platforms? I didn't like copying the `JsonLiteralTest` class to both targets. I tried adding a test factory to `commonTest` and including that from both platforms, but then it was hard to run a specific test within the spec.


jschneidereit (approved):
This is awesome work!

sksamuel:
I think if you activate strict mode, then "100.0" does not equal "1.0E2".
Because if we consider those equal, where do we draw the line? Is 00001 the same as 1 ?
Unless the json spec says that numbers in strings should be treated as special numbers, then I think we just consider them strings.
In lenient mode, I think we can do cleverer things.

kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/JsonLiteralsTest.kt
```diff
@@ -0,0 +1,131 @@
...
+               3
+
+               actual:
+               3.2
+            """.trimIndent()
+         )
+      }
+
+      test("comparing exponent-based float with regular float") {
+         "1E3" shouldEqualJson "1000.0"
```
sksamuel: Should this be for lenient only ?

sksamuel:
Let me know when this is ready again for a review.

Kantis:
> I think if you activate strict mode, then "100.0" does not equal "1.0E2".
> Because if we consider those equal, where do we draw the line? Is 00001 the same as 1 ?
> Unless the json spec says that numbers in strings should be treated as special numbers, then I think we just consider them strings.
> In lenient mode, I think we can do cleverer things.

I think we should draw a line between literals (`100`) and strings (`"100"`). 

When working with literals, _IMO_ we should treat every number that is _same number_ but different format (not type) as the same. 
(`1.0 == 1 != "1" != 01` (invalid number)). 

But it's only my opinion 😄 I can revert to exact literal comparison, e.g. value must be identical (`100.0 != 1.0E2`) if you wish

I added test showing that we don't do type-conversion for quoted numbers in strict mode.

sksamuel:
As discussed on chat, I would update compare mode to use this:

```kotlin
enum class CompareMode {

   /**
    * Types and formats must be identical.
    *
    * For example, "true" and true will not match because one is a string, and the other is a boolean.
    * Similarly, 2.99E9 and 299000000 would not match because despite both being doubles, and the same value,
    * they are not using the same representation.
    */
   Exact,

   /**
    * Types must be identical and compare by value.
    *
    * For example, "true" and true will not match because one is a string, and the other is a boolean.
    * But 2.99E9 and 299000000 are considered equal as they are both doubles, and both represent the same number.
    * "100" and 100 would not match as they are different types.
    */
   Strict,

   /**
    * Compare by value, coercing if possible.
    *
    * For example, "true" and true will match because the string value can be coerced into a valid boolean.
    * Similarly, "100" and 100 will match as the former can be coerced into an int.
    */
   Lenient,
}
```

kotest-assertions/kotest-assertions-json/src/jsTest/kotlin/io.kotest.assertions.json/JsonLiteralsTest.kt
```diff
@@ -63,7 +75,69 @@ class JsonLiteralsTest : FunSpec(
...
+               "10.0" shouldExactlyEqualJson "1e1"
+            }.shouldHaveMessage(
+               """
+               The top level expected 1e1 but was 10.0
+
+               expected:
+               10.0
+
+               actual:
+               10.0
```
Kantis, line 96: This looks/feels a bit odd.

What happens is the json document is parsed to `JsonElement` and then encoded back to string again. (`JsonMatchers.kt:73-77`). I think it would be worthwhile  to roll our own prettification of JSON-document that prevents parsing of the values.. Shall I file an issue for it?

sksamuel:
Please do.

sksamuel:
Let me know when you want a re-review.

Kantis:
@sksamuel I think this is ready for final review. I haven't come up with any better names for `CompareMode.Exact` ☹️

sksamuel:
Nice job :)

PR #2469, opened 2021-09-07
Property Module: Mutation of global PropertyTesting fix for Kotlin/Native

### LINKED ISSUE #2440: Property Module: Mutation of global PropertyTesting configuration throws InvalidMutabilityException in Kotlin/Native
4.6.2

**Steps to reproduce**

Simply attempt to modify the global `PropertyTesting` configuration from a test running in Native (I'm using iosX64)

```
@Test
fun test_changing_property_testing_global_config() {
    PropertyTesting.defaultIterationCount = 100
}
```

```
kotlin.native.concurrent.InvalidMutabilityException: mutation attempt of frozen io.kotest.property.PropertyTesting@6bc211d8
```

**More Information**

We are attempting to upgrade to kotest 4.x property testing into a Kotlin Multiplatform Mobile (android + ios) project. 
Previously we were using a barebones internal framework with a similar generator concept.

With Kotlin/Native we find that tests take 2-3x longer than on JVM with much more expensive CI infrastructure so we took the approach of having different default iteration amounts per platform.

e.g. Android checks more business logic running 1000 iterations and iOS confirms that there are not unexpected differences with a more modest 500 iterations. Saving 1/2 the macOS CI time.

nicorichard:
Fixes #2440 

Allows Kotlin/Native modification of the PropertyTesting configuration object's properties. 

* Introduced an expect/actual abstraction for AtomicReference which allows us to use atomicity in the common source for thread-safe mutations.
* Added an accessory AtomicProperty which allows `get`/`set` delegation by `ReadWriteProperty` to use Atomic properties like common Kotlin variables.


sksamuel:
Native memory models is a bit out of my wheelhouse but I think I basically follow it. Could probably use this for the config for the main framework too (do you use kotest framework or just kotest prop test?)

nicorichard:
We've just been using prop-test so far, but looking at `io.kotest.engine.config.ConfigManager` I expect to see similar issues and most likely even calls to `fun initialize` might fail.

PR #2475, opened 2021-09-08
shouldBeEqualToComparingFields handles arrays and computed properties 

### LINKED ISSUE #2375: Improvement of equality.reflection assertions
Hi,

I've played around with some reflect assertions and I think there should be made some changes.

- shouldBeEqualToComparingFields should have an option to skip computed properties (or make skip default and option to add)
- the actually comparing function checkEqualityOfFields could respect some special cases
  - Arrays
  - Others?

If a pull request is welcomed changing these things, I'm willing to provide some code :-)

Kantis:
Should fix #2375 

I didn't quite get the part regarding `bridge` modifier from the issue discussion. What kind of construct is required to create a `bridge` field?

TODO:
- [ ] update changelog?
- [ ] update relevant docs

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
@@ -369,6 +369,15 @@ fun <T : Any> T.shouldNotBeEqualToComparingFields(
    this shouldNot beEqualComparingFields(other, ignorePrivateFields, emptyList(), ignoreComputedFields)
 }
 
+private typealias PropertyPredicate = (KProperty<*>) -> Boolean
+
+// If no java field exists, it is a computed property which only has a getter
+private val nonComputed: PropertyPredicate = { it.javaField != null }
+private val nonPrivate: PropertyPredicate = { it.visibility != KVisibility.PRIVATE }
+
+private infix fun PropertyPredicate.and(other: PropertyPredicate) =
+   { property: KProperty<*> -> this(property) && other(property) }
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
```diff
@@ -383,8 +410,9 @@ private fun <T> checkEqualityOfFields(fields: List<KProperty<*>>, value: T, othe
    return fields.mapNotNull {
       val actual = it.getter.call(value)
       val expected = it.getter.call(other)
-      if (actual == expected) null else {
-         "${it.name}: ${actual.show().value} != ${expected.show().value}"
-      }
+
+      val isEqual = eq(actual, expected) == null
```
Kantis, line 426: Using `eq` gives us handling of arrays for free.. could it cause other problems? It feels like a nice way to delegate responsibility of equality, IMO

ashishkujoy: Using `eq` will be okay, I don't see any problem using it.

sksamuel:
> Should fix #2375
> 
> I didn't quite get the part regarding `bridge` modifier from the issue discussion. What kind of construct is required to create a `bridge` field?

Bridge  and synthetic methods are added by the compiler. So we should just ignore those (if they even show up).

sksamuel:
Don't need to worry about changelog for bugs - I will compile a full list of fixed bugs when I release the next version.
It's good to add details to the changelog / docs for new features tho.

PR #2486, opened 2021-09-10
Feature/deprecate is stable

### LINKED ISSUE #2482: Data test on JS should use toString

sksamuel:
We originally added isStable so that unstable classes would not break data testing.
Since we now support repeated test names by default, and since isStable does not work on JS or native, we should remove this and just use the toString.

Closes #2482


ashishkujoy:
Seems good to me, but looks like some other changes are got committed as part of this.

sksamuel:
Ah yeah, that's on another branch.

sksamuel:
Moved changes to this PR https://github.com/kotest/kotest/pull/2487

PR #2488, opened 2021-09-10
Deprecate is stable annotation and use toString for data testing

### LINKED ISSUE #2482: Data test on JS should use toString

sksamuel:
Fixes #2482


ashishkujoy (approved):

sksamuel:
third attempt :)

sksamuel:
So this is annoying.
I will break some scenarios.... isolation mode leaf
But JS doesn't support the reflection we need for isStable...
Maybe we just use toString in js

PR #2492, opened 2021-09-10
Introduce constraints for property testing

### LINKED ISSUE #2401: Add configuration option to set minimum execution time in property tests
I'd like to see an option to specify the minimum _amount of time_ (as opposed to iteration count) that is required for execution of each test function for property-based testing. So for example if time is set to 1 second, then the test function gets as many iterations as required to make 1 second.

If there are both the time, and the number of iterations configured, then whichever is _greater_ should be considered for each test function. In other words, if the minimum time runs out _before_ the iteration count is exhausted, then the particular test continues to be executed until all iterations "run out". This guarantees at least _n_ iterations run on slower machines, where _n_ would be a configured number set either via `PropertyTesting.defaultIterationCount` or via the explicit argument of the `checkAll` / `forAll` functions. On the other hand, faster machines would have more iterations, which I think would improve coverage.

Optionally, there could be a possibility to specify whether the "greater" or "lesser" of the two configuration options is chosen when both are specified, but if this feature request is implemented with just the "greater" strategy, that should be fine too.

As an advanced feature (for the future), there could probably be an ability to specify custom (i.e. programmatic) stopping strategies. This is just an idea, but not in scope of this feature request.

sksamuel:
Closes #2401 

Introduces constraints for prop tests to allow for restrictions other than simply iteration count.

Eg,

`checkAll<String, String>(PropTestConfig(Constraits.duration(6000))) { a,b -> a + b shouldBe b + a }`

Will run the property test for 6 seconds, rather than a set number of iterations for example.

Existing syntax is still valid, and specifying nothing is the default 1000 iterations as usual.

One downside is the min iterations would no longer work, but I'm not sure how useful that is anyway.

Interested in your thoughts @myuwono


jschneidereit (approved):
This looks great 👍

myuwono:
this is awesome @sksamuel !!!! this solves so much problems. i can't wait for 5.0 release!

sksamuel:
Do we mind about losing the min iterations check ?
Given how little Exhaustives are used, and since we changed edgecases to be randoms, I'm not sure it adds much value, since for Arbs it has no effect anyway.

myuwono:
I think it makes sense to lose min iterations with the current design.

myuwono (approved):

PR #2494, opened 2021-09-11
Feature/introduce continuation arbs builder

### LINKED ISSUE #2493: [proposal] introduce continuation passing style arbitrary builder for improved ergonomics
### Context

it is currently possible to build dependent arbitraries by nesting `flatMaps`. While that works, the level of nestings makes it a bit awkward, in particular when the arb being built requires particularly complex, potentially interdependent conditions. 

### Proposition
I propose we introduce continuation passing style builders for our arbitrary builder. Drawing inspiration introduced to simplify chaining effects in arrow-fx, I realized that the arbitrary builder ergonomics can be improved using suspend functions using continuation passing style (CPS) on `arb.flatMap`. This not only simplifies the whole building of dependent arbs, it also disambiguates how developers would create, compose, or merge arbs. That is, developers can then be lifted from the burden of deciding between `.map`, `.flatMap`, and `bind`. 

The proposition is as follows:
```kotlin
// notice the `ArbitraryBuilderSyntax` context
fun <A> arbitrary(fn: suspend ArbitraryBuilderSyntax.(RandomSource) -> A): Arb<A> =
   arbitraryBuilder { rs -> fn(rs) }

fun <A> arbitraryBuilder(
   shrinker: Shrinker<A>? = null,
   classifier: Classifier<A>? = null,
   edgecaseFn: EdgecaseFn<A>? = null,
   builderFn: suspend ArbitraryBuilderSyntax.(RandomSource) -> A
): Arb<A> = ... // implementation detail omitted 

@RestrictsSuspension
interface ArbitraryBuilderSyntax {
   // this is a CPS for arb.flatMap
   suspend fun <T> Arb<T>.bind(): T
}
```

what this allows are possibilities as follows:
#### Flatmaps
```kotlin
data class Foo(val first: First, val second: Second, val third: Third)

val arbFirst: Arb<First> = ...
fun arbSecond(first: First): Arb<Second> = ...
fun arbThird(first: First, second: Second): Arb<Third> = ...

// with raw flatMaps
val arbFlatMaps: Arb<Foo> =
   arbFirst.flatMap { first ->
      arbSecond(first).flatMap { second ->
         arbThird(first, second).map { third ->
            Foo(first, second, third)
         }
      }
   }

// with continuation
// notice we also have the random source here to possibly do more complex generations
val arb: Arb<Foo> = arbitrary { rs: RandomSource -> 
   val first = arbFirst.bind()
   val second = arbSecond(first).bind()
   val third = arbThird(first, second).bind()
   Foo(first, second, third)
}
```
#### Dependent Binds 
```kotlin
data class Foo(val first: String, val multiplicationResult: Long)

val arbFirst: Arb<String> = ...
fun arbSecond(first: String): Arb<Int> = ...
val arbThird: Arb<Long> = ...

// with raw binds
val arbBind: Arb<String> = arbFirst.flatMap { first ->
   Arb.bind(arbSecond(first), arbThird) { second, third -> 
      Foo(first, second.toLong() * third)
   }
} 

// with continuation
val arb: Arb<String> = arbitrary { 
   val first = arbFirst.bind()
   val multiplicationResult = arbSecond(first).bind().toLong() * arbThird.bind()
   Foo(first, multiplicationResult)
}
```

myuwono:
This is a proposed implementation for arbitrary builders using continuation passing style. This code is inspired by how monad fx on any `Kind<F, A>` was previously done in arrow.

This allows the following simplification as described in #2493:
```kotlin
data class Foo(val first: String, val multiplicationResult: Long)

val arbFirst: Arb<String> = ...
fun arbSecond(first: String): Arb<Int> = ...
val arbThird: Arb<Long> = ...

// with raw binds
val arbBind: Arb<String> = arbFirst.flatMap { first ->
   Arb.bind(arbSecond, arbThird) { second, third -> 
      Foo(first, second.toLong() * third)
   }
} 
// with continuation
// notice we also have access to the random source
val arb: Arb<String> = arbitrary { rs: RandomSource ->
   val first = arbFirst.bind()
   val multiplicationResult = arbSecond(first).bind().toLong() * arbThird.bind()
   Foo(first, multiplicationResult)
}
```

fixes #2493

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/builders.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/builders.kt
@@ -156,7 +156,6 @@ interface ArbitraryBuilderSyntax {
    suspend fun <T> Arb<T>.bind(): T
 }
 
-@Suppress("UNCHECKED_CAST")
 private class ArbContinuation<A> : Continuation<Arb<A>>, ArbitraryBuilderSyntax {
    override val context: CoroutineContext = EmptyCoroutineContext
    private lateinit var returnedArb: Arb<A>
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/builders.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/builders.kt
@@ -11,9 +11,9 @@ import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
 import kotlin.coroutines.RestrictsSuspension
 import kotlin.coroutines.intrinsics.COROUTINE_SUSPENDED
+import kotlin.coroutines.intrinsics.startCoroutineUninterceptedOrReturn
 import kotlin.coroutines.intrinsics.suspendCoroutineUninterceptedOrReturn
 import kotlin.coroutines.resume
-import kotlin.coroutines.startCoroutine
 
 /**
  * Creates a new [Arb] that performs no shrinking, has no edge cases and
@@ -52,28 +52,43 @@ fun <A> arbitrary(
  * Creates a new [Arb] that performs no shrinking, uses the given edge cases and
  * generates values from the given function.
  */
-fun <A> arbitrary(edgecases: List<A>, fn: (RandomSource) -> A): Arb<A> = object : Arb<A>() {
-   override fun edgecase(rs: RandomSource): A? = if (edgecases.isEmpty()) null else edgecases.random(rs.random)
-   override fun sample(rs: RandomSource): Sample<A> = Sample(fn(rs))
-}
+fun <A> arbitrary(edgecases: List<A>, fn: suspend ArbitraryBuilderSyntax.(RandomSource) -> A): Arb<A> =
+   object : Arb<A>() {
+      override fun edgecase(rs: RandomSource): A? = if (edgecases.isEmpty()) null else edgecases.random(rs.random)
+      override fun sample(rs: RandomSource): Sample<A> = delegate.sample(rs)
+
+      private val delegate = arbitraryBuilder { rs -> fn(rs) }
+   }
 
 /**
  * Creates a new [Arb] that performs shrinking using the supplied [Shrinker], uses the given edge cases and
  * generates values from the given function.
  */
-fun <A> arbitrary(edgecases: List<A>, shrinker: Shrinker<A>, fn: (RandomSource) -> A): Arb<A> = object : Arb<A>() {
+fun <A> arbitrary(
+   edgecases: List<A>,
+   shrinker: Shrinker<A>,
+   fn: suspend ArbitraryBuilderSyntax.(RandomSource) -> A
+): Arb<A> = object : Arb<A>() {
    override fun edgecase(rs: RandomSource): A? = if (edgecases.isEmpty()) null else edgecases.random(rs.random)
-   override fun sample(rs: RandomSource): Sample<A> = sampleOf(fn(rs), shrinker)
+   override fun sample(rs: RandomSource): Sample<A> = delegate.sample(rs)
+
+   private val delegate = arbitraryBuilder(shrinker) { rs -> fn(rs) }
 }
 
 /**
  * Creates a new [Arb] that generates edge cases from the given [edgecaseFn] function
  * and generates samples from the given [sampleFn] function.
  */
-fun <A> arbitrary(edgecaseFn: (RandomSource) -> A?, sampleFn: (RandomSource) -> A): Arb<A> =
+fun <A> arbitrary(
+   edgecaseFn: ArbitraryBuilderSyntax.(RandomSource) -> A?,
+   sampleFn: ArbitraryBuilderSyntax.(RandomSource) -> A
+): Arb<A> =
    object : Arb<A>() {
-      override fun edgecase(rs: RandomSource): A? = edgecaseFn(rs)
-      override fun sample(rs: RandomSource): Sample<A> = Sample(sampleFn(rs))
+      override fun edgecase(rs: RandomSource): A? = edgecaseDelegate.edgecase(rs)
+      override fun sample(rs: RandomSource): Sample<A> = delegate.sample(rs)
+
+      private val edgecaseDelegate: Arb<A?> = arbitraryBuilder { rs -> edgecaseFn(rs) }
+      private val delegate: Arb<A> = arbitraryBuilder { rs -> sampleFn(rs) }
    }
 
 /**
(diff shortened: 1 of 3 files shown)


sksamuel:
You could also do this I think ?

```kotlin
val arb: Arb<String> = arbitrary { rs ->
   val first = Arb.string(5, Codepoint.alphanumeric()).withEdgecases("edge1", "edge2").next(rs)
   val second = Arb.int(1..9).withEdgecases(5).next(rs)
   val third = Arb.int(101..109).withEdgecases(100 + second, 109).next(rs)
   "$first $second $third"
}
```

myuwono:
> You could also do this I think ?
> 
> ```kotlin
> val arb: Arb<String> = arbitrary { rs ->
>    val first = Arb.string(5, Codepoint.alphanumeric()).withEdgecases("edge1", "edge2").next(rs)
>    val second = Arb.int(1..9).withEdgecases(5).next(rs)
>    val third = Arb.int(101..109).withEdgecases(100 + second, 109).next(rs)
>    "$first $second $third"
> }
> ```

You could, the downside is one would have to be prepared to also forego the edgecases of the dependent arbs.

I guess, the general question is, how might we chain arbs such that all the intrinsics are propagated fully? We look at flatmap for instance as a central point of coordination. That solves edgecases compositions without cross-contaminating samples, and potentially many other possibilities such as propagation of shrinkers and classifications.

sksamuel:
Yes fair.

sksamuel:
Can you talk me through the implementation a bit. At least I'm not clear how the returnedArb gets populated in the continuation exactly.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/builders.kt
```diff
@@ -56,12 +90,36 @@ fun <A> arbitrary(
...
+      val continuation = ArbContinuation<A>()
+      val wrapReturn: suspend ArbitraryBuilderSyntax.() -> Arb<A> = {
+         val value: A = builderFn(randomSource.bind())
+         ArbitraryBuilder(
+            { value },
+            classifier,
+            shrinker,
+            { rs ->
+               // use edgecase function if provided, otherwise retain those from flatmap compositions
+               if (edgecaseFn != null) edgecaseFn.invoke(rs) else value
```
myuwono: this should always default to value if null

myuwono: done! added test in https://github.com/kotest/kotest/pull/2494/files#diff-2e6b0f63218b6d89d463623c97b883bb2d959b400807214765432ebf17f615ffR88-R107

myuwono:
> Can you talk me through the implementation a bit. At least I'm not clear how the returnedArb gets populated in the continuation exactly.

awesome. so i think i'd properly quote my reference in https://github.com/arrow-kt/arrow-core/blob/40619f87b7fc1c790c3ab7b7341005a960630a27/arrow-core-data/src/main/kotlin/arrow/typeclasses/Monad.kt

This looks a bit weird at first, but it is actually quite simple. It's in fact very clever!

so let's try this code for instance:
```kotlin
arbitrary {
  val string: String = Arb.string(5).bind()
  val int: Int = Arb.int(1..100).bind()
  "$string $int"
}
```

now the sequence actually happen as follows

#### 1. suspending computation: at line 2, i.e. when bind() is called.

that executes this function
```kotlin
   override suspend fun <B> Arb<B>.bind(): B = suspendCoroutineUninterceptedOrReturn { c ->
      // we call flatMap on the bound arb, and then returning the `returnedArb`, without modification
      returnedArb = this.flatMap { b: B ->
         // we resume the suspension with the value passed inside the flatMap function. 
         // this can be either sample or edgecases. This is important
         // because from the point of view of a user of kotest, when we talk about transformation, 
         // we  care about transforming the generated value of this arb for both sample and edgecases.
         c.resume(b) 
         returnedArb // returned arb here is returned as is
      }
      COROUTINE_SUSPENDED
   }
```

Notice this block returns the special `COROUTINE_SUSPENDED` value, this means the Continuation provided to the block shall be resumed by invoking `Continuation.resumeWith` at some moment in the future when the result becomes available to resume the computation. That resume with is as follows:
```kotlin
   override fun resumeWith(result: Result<Arb<A>>) {
     result.map { theNewArb -> returnedArb = theNewArb }.getOrThrow()
   }
```
if you realize, this `returnedArb` is only going to be modified at the end of the suspend computation, which we defined at the call site where we build the final Arb:
```kotlin
   // continuation is a single shot so this needs to be a function to reinstate value randomization through the arbs
   private fun computeArb(): Arb<A> {
      val continuation = ArbContinuation<A>()
      
      // this is where the point action happen, i.e. when we build the final arb from the returned value from the block.
      // essentially at the end of the suspend we got `A`, and here we need to point it into `Arb<A>`
      val wrapReturn: suspend ArbitraryBuilderSyntax.() -> Arb<A> = {
         // the value after finishing our builder block
         val value: A = builderFn(randomSource.bind())
         // we then create a new arb here with that value.
         ArbitraryBuilder(
            { value },
            classifier,
            shrinker,
            { rs -> edgecaseFn?.invoke(rs) ?: value }
         ).build()
      }
 
      // all the functions are set, we start our coroutine, providing our continuation instance, this will run the suspend block
      wrapReturn.startCoroutine(continuation, continuation)
      
      // once that is finished, the `continuation.resumeWith` will be called, where we assign the `returnedArb` var. 
      // after that, we then return that populated value.
      return continuation.returnedArb()
   }
```

myuwono:
I'll pop those comments in the code as well

sksamuel:
Can we add the generateArbitrary variant which allows suspension in the lamba (inline?)
Could be another pr.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/builders.kt
```diff
@@ -6,62 +6,313 @@ import io.kotest.property.RandomSource
...
+      private val delegate: Arb<A> = arbitraryBuilder(shrinker) { rs -> sampleFn(rs) }
+   }
+
+@Suppress("ClassName")
+object arbitrary {
+   /**
+    * Creates a new [Arb] that performs no shrinking, has no edge cases and
+    * generates values from the given function.
+    */
+   suspend inline fun <A> suspendable(
```
sksamuel: I thought we were going with generateArbitrary to match generateSequence from the std lib.
Not totally bothered but suspendable isn't clear that it's a builder at all to me.

myuwono: generateSequence is not suspend apparently. I was trying to find a better syntax and tried to brainstorm this. Looking at arrow, they have `either` and `either.eager` for instance. I'll ping you in slack for better guidance.
 
```kotlin
// non-suspend
arbitrary { ... } 

// alt 1
arbitrary.suspendable { ... }

// alt 2
suspendArbitrary { ... }

// alt 3
generateArbitrary { ... }
```

myuwono: i'm overthinking it, `generateArbitrary` is a great name 🚀

myuwono:
added `inline suspend fun <A> generateArbitrary(...): Arb<A>` 🎉

PR #2496, opened 2021-09-11
Add property spec

sksamuel:


sksamuel:
Playing around with a prop_spec maybe.

myuwono:
@sksamuel what are we envisioning in this new style? seems like a pretty cool one, somehow reminds me of specs2.

sksamuel:
It's just an idea really. Not sure if it's worth it or not.

PR #2498, opened 2021-09-12
make codepoints arbitraries as snappy as possible

### LINKED ISSUE #2497: [bug] String generation is slower on some codepoints
String generation seems to be slower on some codepoints. These problematic codepoints seem to have filter operations and transformation on each sample.

```kotlin
context("compare snappiness") {
   test("alphanumeric") {
      Arb.string(1000, Codepoint.alphanumeric()).take(100000).last()
   }

   test("greek coptic") {
      Arb.string(1000, Codepoint.greekCoptic()).take(100000).last()
   }
}
```

Screen Shot 2021-09-12 at 9 32 10 pm

myuwono:
fixes #2497


sksamuel:
So my understanding is that this premaps the codepoints rather than maps them on every go.

Issue #2230, opened 2021-04-30
Allow clues to be added to timeouts


sschuberth:
It would be nice if there was a way to also give a clue when a timeout occurs (configured either via `withTimeout()` or the test configuration). A use-case would be tests that call external programs. Such programs might run interactive prompts, blocking the test, which then runs into a timeout. In such cases *stdout* could be shown as a clue, which would help to identify the interactive prompt.

sschuberth:
This has just recently been referred to, so I believe it should stay open.

sksamuel:
It's in the roadmap. I've marked it as pinned.

sksamuel:
How do you see this working?
If I have a test that I set a timeout for 10 seconds and it times out - it will throw a timeout exception so you know why it failed. What extra information would you like in there ?

sschuberth:
> What extra information would you like in there ?

See above (at the example of a CLI program that causes a timeout due to waiting for *stdin*): "In such cases *stdout* could be shown as a clue, which would help to identify the interactive prompt."

sksamuel:
Right, so this is specific to stdin in timeouts.
Kotest would have to capture stdout _in case_ the test times out.

I wonder if another construct makes sense.

withStdOutClue { }

Which captures std out and emits it as a clue if the test fails.

sschuberth:
I wasn't meaning to suggest such a specific change to be done to Kotest. But as some one who writes test with Kotest, I'd like to be able to "manually" pass any clue that I see fit to `withTimeout()`. At the time I filed this issue, I could not see a way to do that (e.g. by nesting `withTimeout()` inside `withClue()`).

sksamuel:
ohhh I see.

Right, that sounds like a really good idea.

sksamuel:
You can now next withTimeout inside a withClue,

```kotlin
  withClue("timey timey") {
     withTimeout(2) {
        delay(1000)
     }
  }
```

PR #2500, opened 2021-09-12
Fix withClue and assertSoftly for coroutines switching threads (#2447)

### LINKED ISSUE #2447: withClue() fails with EmptyStackException if a coroutine switches threads
Version: `kotest-assertions-core:4.6.2`

As described in the [coroutine docs on thread-local data](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html#thread-local-data), a coroutine may switch threads at suspension points.

`withClue()` relies on thread-local data via `ThreadLocalErrorCollector`, without taking coroutine thread-switching into account. It fails if its block contains a coroutine resuming on a different thread.

#### Example
```kotlin
import io.kotest.assertions.withClue
import kotlinx.coroutines.Dispatchers
import kotlinx.coroutines.runBlocking
import org.junit.jupiter.api.Test

class TestCase {
    @Test
    fun `coroutine changing threads`() = runBlocking(Dispatchers.Unconfined) {
        withClue("Hello") {
            Thread.currentThread().run { println("withClue() block begins on $name, id $id") }
            delay(10)  // First suspension makes the Unconfined dispatcher resume on a different thread
            Thread.currentThread().run { println("withClue() block ends   on $name, id $id") }
        }
    }
}
```
##### Build script
```kotlin
import org.jetbrains.kotlin.gradle.tasks.KotlinCompile

plugins {
    kotlin("jvm") version "1.5.30"
    application
}

group = "me.oliver"
version = "1.0-SNAPSHOT"

repositories {
    mavenCentral()
}

dependencies {
    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.1")
    testImplementation(kotlin("test"))
    testImplementation("io.kotest:kotest-assertions-core:4.6.2")
}

tasks.test {
    useJUnitPlatform()
}

tasks.withType<KotlinCompile>() {
    kotlinOptions.jvmTarget = "11"
}

application {
    mainClass.set("MainKt")
}
```

##### Result
```
> Task :test FAILED
withClue() block begins on Test worker @coroutine#1, id 12
withClue() block ends   on kotlinx.coroutines.DefaultExecutor @coroutine#1, id 15

java.util.EmptyStackException
	at java.base/java.util.Stack.peek(Stack.java:102)
	at java.base/java.util.Stack.pop(Stack.java:84)
	at io.kotest.assertions.ThreadLocalErrorCollector.popClue(ErrorCollector.kt:32)
	at TestCase$coroutine changing threads$1.invokeSuspend(TestCase.kt:23)
```

OliverO2:
This should fix #2447 by keeping the error collector synchronized with thread-switching coroutines.

The required context element is automatically installed by the Kotest framework. When using the assertions without the framework, it is currently the user's responsibility to install the required context element, e.g. via
```
runBlocking(errorCollectorContextElement) { ... }
```

As has been discussed in https://github.com/kotest/kotest/issues/2447#issuecomment-917174018, `withClue` could be redesigned to avoid the error collector altogether. In this fix, I did not touch the implementation in order not to mess up message formatting.

Another improvement would be to find a way to automatically install the required `errorCollectorContextElement` for those not using the Kotest framework. Some approaches have been contemplated in https://github.com/kotest/kotest/issues/2447#issuecomment-917691838. Another option might be to ask the Kotlin team to provide an API for installing default context elements.

I'd suggest to go with this rather light-weight approach for the initial 5.0 release and follow up later for further improvements.

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
@@ -2,49 +2,24 @@
 
 package io.kotest.assertions
 
-import java.util.Stack
+import kotlinx.coroutines.asContextElement
 
-actual val errorCollector: ErrorCollector = ThreadLocalErrorCollector
+actual val errorCollector: ErrorCollector get() = ThreadLocalErrorCollector.instance.get()
 
-object ThreadLocalErrorCollector : ErrorCollector {
+/**
+ * A [CoroutineContext.Element] which keeps the error collector synchronized with thread-switching coroutines.
+ *
+ * When using [withClue] or [assertSoftly] on the JVM without the Kotest framework, this context element
+ * should be added to each top-level coroutine context, e.g. via
+ * - runBlocking(errorCollectorContextElement) { ... }
+ * - runBlockingTest(Dispatchers.IO + errorCollectorContextElement) { ... }
+ */
+val errorCollectorContextElement get() = ThreadLocalErrorCollector.instance.asContextElement()
 
-   private val clueContext = object : ThreadLocal<Stack<Clue>>() {
-      override fun initialValue(): Stack<Clue> = Stack()
+class ThreadLocalErrorCollector : BasicErrorCollector() {
+   companion object {
+      val instance = object : ThreadLocal<ErrorCollector>() {
+         override fun initialValue() = ThreadLocalErrorCollector()
+      }
    }
-
-   private val failures = object : ThreadLocal<MutableList<Throwable>>() {
-      override fun initialValue(): MutableList<Throwable> = mutableListOf()
-   }
-
-   private val collectionMode = object : ThreadLocal<ErrorCollectionMode>() {
-      override fun initialValue() = ErrorCollectionMode.Hard
-   }
-
-   override fun setCollectionMode(mode: ErrorCollectionMode) = collectionMode.set(mode)
-
-   override fun getCollectionMode(): ErrorCollectionMode = collectionMode.get()
-
-   override fun pushClue(clue: Clue) {
-      clueContext.get().push(clue)
-   }
-
-   override fun popClue() {
-      clueContext.get().pop()
-   }
-
-   override fun clueContext(): List<Clue> = clueContext.get()
-
-   override fun errors(): List<Throwable> = failures.get().toList()
-
-   /**
-    * Adds the given error to the current context.
-    */
-   override fun pushError(t: Throwable) {
-      failures.get().add(t)
-   }
-
-   /**
-    * Clears all errors from the current context.
-    */
-   override fun clear() = failures.get().clear()
 }
--- a/kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/io/kotest/assertions/CluesTests.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/io/kotest/assertions/CluesTests.kt
@@ -1,6 +1,7 @@
 package io.kotest.assertions
 
 import io.kotest.core.spec.style.FunSpec
+import io.kotest.matchers.collections.shouldHaveSize
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.withContext
@@ -8,11 +9,13 @@ import kotlinx.coroutines.withContext
 class CluesTests : FunSpec({
    test("withClue should not fail on coroutine thread switch") {
       withContext(Dispatchers.Unconfined) {
+         val threadIds = mutableSetOf<Long>()
          withClue("should not fail") {
-            Thread.currentThread().run { println("withClue block begins on $name, id $id") }
+            threadIds.add(Thread.currentThread().id)
             delay(10)
-            Thread.currentThread().run { println("withClue block ends on $name, id $id") }
+            threadIds.add(Thread.currentThread().id)
          }
+         threadIds shouldHaveSize 2
       }
    }
 })
--- a/kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/io/kotest/assertions/CluesTests.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/io/kotest/assertions/CluesTests.kt
@@ -1,0 +1,19 @@
+package io.kotest.assertions
+
+import io.kotest.core.spec.style.FunSpec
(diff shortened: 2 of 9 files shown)


kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
```diff
@@ -2,49 +2,24 @@
...
-
-   override fun setCollectionMode(mode: ErrorCollectionMode) = collectionMode.set(mode)
-
-   override fun getCollectionMode(): ErrorCollectionMode = collectionMode.get()
-
-   override fun pushClue(clue: Clue) {
-      clueContext.get().push(clue)
+import kotlinx.coroutines.asContextElement
+
+actual val errorCollector: ErrorCollector get() = ThreadLocalErrorCollector.instance.get()
```
sksamuel, line 7: Do you think it's a bit unusual to have the val with get() ?
What about just making this a function?

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
```diff
@@ -2,49 +2,24 @@
...
+
+/**
+ * A [CoroutineContext.Element] which keeps the error collector synchronized with thread-switching coroutines.
+ *
+ * When using [withClue] or [assertSoftly] on the JVM without the Kotest framework, this context element
+ * should be added to each top-level coroutine context, e.g. via
+ * - runBlocking(errorCollectorContextElement) { ... }
+ * - runBlockingTest(Dispatchers.IO + errorCollectorContextElement) { ... }
+ */
+val errorCollectorContextElement get() = ThreadLocalErrorCollector.instance.asContextElement()
```
sksamuel, line 17: Does this need the get() if you're returning the same instance ?

kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/io/kotest/assertions/CluesTests.kt
```diff
@@ -0,0 +1,18 @@
...
+import io.kotest.core.spec.style.FunSpec
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.withContext
+
+class CluesTests : FunSpec({
+   test("withClue should not fail on coroutine thread switch") {
+      withContext(Dispatchers.Unconfined) {
+         withClue("should not fail") {
+            Thread.currentThread().run { println("withClue block begins on $name, id $id") }
```
sksamuel: Does the Dispatchers.Unconfined guarantee it'll switch threads after delay? 
I would add the thread name into a set, and after withContext do an assertion that the set has size 2, to ensure it ran on two threads.

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
OliverO2: It's an `actual` for an `expect val`. Conceptually, it behaves as a val. Would you prefer changing the entire `expect` declaration and its target implementations?

sksamuel: Yeah, I think it's clearer if we're using this one as a function to have the expect be a function and on JS / Native return a singleton if needed.

kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/io/kotest/assertions/CluesTests.kt
OliverO2: That's how i understood [the doc](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html#unconfined-vs-confined-dispatcher) where it is used as an example for the exact situation:
> The Dispatchers.Unconfined coroutine dispatcher starts a coroutine in the caller thread, but only until the first suspension point. After suspension it resumes the coroutine in the thread that is fully determined by the suspending function that was invoked.

Could add that check, though.

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
OliverO2: Conceptually, it is the same instance, that's why I made it a `val` here, too. The `get()` ensures that we don't have to worry about the internals of `asContextElement()`, which is invoked.

OliverO2:
Commit bf78dc7 added a fix for `BasicErrorCollector`, which returned clues in an order different from what the tests expect. (Tests for clues and assertSoftly were JVM only, so it seems like BasicErrorCollector did not get tested before.)

kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/io/kotest/assertions/CluesTests.kt
OliverO2: Done via 7b69d9757a2abffa063b7a5e06e6987f52503da9.

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
OliverO2: IntelliJ detects 37 usages within Kotest. Probably better to change this separately from this PR. User-created matchers might also be affected.

sksamuel: Ok fair enough.

sksamuel:
👍🏼

OliverO2:
Cool, thanks! One more note regarding the vals: As I understand it, their use this case is entirely consistent with the [Kotlin coding conventions on "Functions vs properties"](https://kotlinlang.org/docs/coding-conventions.html#functions-vs-properties). I have also seen languages as old as "Eiffel" which do not distinguish at all between zero-parameter functions and properties. :)

sksamuel:
Scala does that too, 

```scala
trait Foo {
   def a:String
}

class FooImpl extends Foo {
  val a = "string"
}
```

sksamuel:
I guess this could be implemented by delegating to withClue as well
https://github.com/kotest/kotest/issues/2230

PR #2503, opened 2021-09-13
Upgrade gradle wrapper 7.1.1 and unify publication

i-walker:


sksamuel:
I had some issues with gradle 7 before but we can try again.

PR #2510, opened 2021-09-15
Fix endless recursion in 2-arity checkAll

dimsuz:
2-arity checkAll with PropTestConfig overload used to call itself,
leading to StackOverflow exception.


sksamuel:
👍🏼

PR #2512, opened 2021-09-16
Added display name format extension point #2507

### LINKED ISSUE #2507: Allow a custom function to name a test
Right now we allow users to append tags to tests names, which is fine - but it would be even better to provide users an extension point to modify the name of the tests based on the test itself and tags (and perhaps other information).

Given a test with tags `Slow` and `NickNack`

status quo:
```
TestClass.nicknack does some stuff[tags = Slow, NickNack]
```

new hotness, in kotlin pseudocode:
```kotlin
val Slow = Tag("Slow")

val testNamer = object : TestNameExtension {
  override fun nameTest(testCase: TestCase, tags: Set<Tag>): String {
    if (tags.contains(Slow)) {
      return "Slow: ${testCase.name} [tags=${(tags - Slow).joinToString()}]"
   }
   return testCase.name
```

would give us a nice test name like 

```
Slow: TestClass.nicknack does some stuff [tags = NickNack]
```

sksamuel:
Closes #2507


jschneidereit:
Does the append tags implementation require an update after this?

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/extensions/DisplayNameFormatterExtension.kt
```diff
@@ -0,0 +1,31 @@
...
+
+   /**
+    * Returns a formatted name for a test.
+    */
+   fun format(testCase: TestCase): String
+
+   /**
+    * Returns a formatted name for a spec class.
+    */
+   fun format(kclass: KClass<*>): String
```
jschneidereit, line 23: Does this work for js and native?

jschneidereit (approved):

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/extensions/DisplayNameFormatterExtension.kt
sksamuel: Yes

sksamuel:
> Does the append tags implementation require an update after this?

The default formatter appends tags in the old way.

Issue #2211, opened 2021-04-19
Add docs for HTML Report


sksamuel:
As added here: https://github.com/kotest/kotest/issues/2011

caramelomartins:
I could maybe help out with this one if you'd like.

sksamuel:
Awesome please do

caramelomartins:
I haven't had any time to work on this but I'm commenting just so that the stale bot doesn't close the issue.

sksamuel:
I've pinned it so it won't set it stale.
If you have time to do this that's cool @caramelomartins it not I will try to get to it.

caramelomartins:
@sksamuel I'll have some time this week to do this. Thanks!

PR #2514, opened 2021-09-16
Add documentation for HTML Reporter

caramelomartins:
As spoken in #2211, I'm opening this PR in order to add some documentation of the HTML Reporter so that it appears on the website. I've kept it simple and mostly used JUnit XML as an inspiration.


sksamuel:
👍🏼

Issue #1234, opened 2020-02-23
After project failure "Class"


sksamuel:
After project / before project errors currently manifest themselves as a "unknown class" as intellij doesn't know what to do with them.

![afterp](https://user-images.githubusercontent.com/743706/75120792-dab83100-5653-11ea-9e09-4bc250532f64.png)

It might be nicer if we create a dummy test parent called "project listener" and attach the failure there.

Issue #2340, opened 2021-07-07
ShouldEqualJson comparing order misbehaving in lists


fabridinapoli:
**Which version of Kotest are you using**
4.6.0

Hello,

I'm using shouldEqualJson https://kotest.io/docs/assertions/json-matchers.html#shouldequaljson and standing to the docs, if I set CompareOrder.Lenient, in theory the order shouldn't matter. As you can see in the following example, I've two lists with the same elements, but in different order:

```
import io.kotest.assertions.json.CompareMode
import io.kotest.assertions.json.CompareOrder
import io.kotest.assertions.json.shouldEqualJson
import org.junit.jupiter.api.Test

class SomeSillyTest {
    @Test
    fun `some silly test`() {
        val currentOne = """
            {
                "some": "thing",
                "listOfStuff": ["a2", "a1"],
                "someList": [
                    {"type": "SOME_TYPE_2"},
                    {"type": "SOME_TYPE_1"}
                ]
            }
        """.trimIndent()
        val expected = """
            {
                "some": "thing",
                "listOfStuff": ["a1", "a2"],
                "someList": [
                    {"type": "SOME_TYPE_1"},
                    {"type": "SOME_TYPE_2"}
                ]
            }
        """.trimIndent()

        currentOne.shouldEqualJson(expected, mode = CompareMode.Lenient, order = CompareOrder.Lenient)
    }
}
```

Result:
```
java.lang.AssertionError: At 'listOfStuff.[0]' expected 'a1' but was 'a2'

expected:
{
  "some": "thing",
  "listOfStuff": [
    "a1",
    "a2"
  ],
  "someList": [
    {
      "type": "SOME_TYPE_1"
    },
    {
      "type": "SOME_TYPE_2"
    }
  ]
}

actual:
{
  "some": "thing",
  "listOfStuff": [
    "a2",
    "a1"
  ],
  "someList": [
    {
      "type": "SOME_TYPE_2"
    },
    {
      "type": "SOME_TYPE_1"
    }
  ]
}
```

sksamuel:
@Kantis do you think this could be fixed easily as part of your other changes ?

Kantis:
Sure, I'll have a look.. Want me to split it into a separate PR?

sksamuel:
Up to you.

Kantis:
[...]

PR #2518, opened 2021-09-18
New CompareOrder for shouldEqualJson to allow out-of-order arrays

Kantis:
Solves #2340 

It's now possible to specify a new `CompareOrder` for `shouldEqualJson` to allow arrays to be shuffled. Example:

```kotlin
"""{ "points": [ {"x": 1.0, "y": 2.0}, {"x": 0.0, "y": 0.0 ] }""".shouldEqualJson(
  """{ "points": [ {"x": 0.0, "y": 0.0 ], {"x": 1.0, "y": 2.0} ] }""", 
  CompareOrder.LenientAll
)
```
Will succeed even though the objects in the array are in a different order.

Current default is kept, which means that array order matters and the above would fail.


sksamuel:
Awesome

sksamuel:
Btw if you write Closes #1234 then it automatically close that issue when the PR is merged.

PR #2523, opened 2021-09-18
CompareJsonOptions

### LINKED ISSUE #2520: Suggestion: Tweaked options API for comparing JSON
The number of parameters to `shouldEqualJson`, along with the possible values of each, is growing with some of my latest work and I feel that perhaps it could be simplified for users by adding the following API.

The primary change would be the creation of a options class, containing all parameters specified. We allow the user to send this options-object and add a helper function for easily creating and configuring it. The user can then either use this as
```kotlin
val jsonOptions = compareJsonOptions {
   typeCoercion = CompareJsonOptions.TypeCoercion.Enabled // Could be statically imported, to allow `TypeCoercion.Enabled`, or `Enabled`
}

test("some test") {
  "{ }".shouldEqualJson(expected, jsonOptions)
}
```

Or define their own methods, example:

```kotlin
infix fun String.myShouldEqualJson(other: String) = shouldEqualJson(other,
   compareJsonOptions {
      fieldComparison = CompareJsonOptions.FieldComparison.Lenient
      typeCoercion = CompareJsonOptions.TypeCoercion.Enabled
   }
)
```

The rest of the API. I think it feels nice that options become more explicit and fine-grained, and it provides plenty of points to add documentation.

```kotlin
fun compareJsonOptions(builder: CompareJsonOptions.() -> Unit): CompareJsonOptions =
   CompareJsonOptions().apply(builder)

class CompareJsonOptions(
   /**
    * Controls whether property order must be identical
    */
   var propertyOrder: PropertyOrder = PropertyOrder.Lenient,

   /**
    * Controls whether array ordering must be identical.
    */
   var arrayOrder: ArrayOrder = ArrayOrder.Strict,

   /**
    * Controls whether the actual document may contain extra fields or not.
    */
   var fieldComparison: FieldComparison = FieldComparison.Strict,

   /**
    * Controls whether number formatting should be taken into consideration. For instance, comparing 1.00 to 1.0, or
    * 1E2 to 100
    */
   var numberFormat: NumberFormat = NumberFormat.Lenient,

   /**
    *  Controls whether types should be coerced when possible. For instance, when strings contain bool or numeric values.
    */
   var typeCoercion: TypeCoercion = TypeCoercion.Enabled
)  {

   enum class PropertyOrder {
      /**
       * Default. Property order in objects does not matter.
       *
       * Example: `"""{ "a": 0, "b": 2 }""".shouldEqualJson("""{ "b": 2, "a": 1 }""", compareJsonOptions { propertyOrder = Lenient })` will pass
       */
      Lenient,

      /**
       * Properties must be in same order. E.g. `{ "a": 0, "b": 2 }` is not considered equal to `{ "b": 2, "a": 1 }`
       */
      Strict
   }

   enum class ArrayOrder {
      /**
       * Default. Arrays must contain the same elements in the same order.
       */
      Strict,

      /**
       * Arrays are allowed to be shuffled, but must still contain same items.
       */
      Lenient,
   }

   enum class FieldComparison {
      /**
       * Default. Objects in [expected] and [actual] must contain the same fields.
       */
      Strict,

      /**
       * Objects in the actual document may contain extraneous fields without causing comparison to fail.
       */
      Lenient,
   }

   enum class NumberFormat {
      /**
       * Default. Numbers will be interpreted before being compared. Meaning we can compare 0E3 to 1000 without fail
       */
      Lenient,

      /**
       * Numbers must also be formatted the same way to be considered equal.
       */
      Strict
   }

   enum class TypeCoercion {
      /**
       * Default. Types will not be converted. Meaning `"true"` and `true` are considered unequal.
       */
      Disabled,

      /**
       * Types may be coerced. Strings containing numbers will be considered equal to their numbers, and booleans in
       * strings will also be compared.
       *
       * For example: `"\"11\"".shouldEqualJson("12", compareJsonOptions { typeCoercion = TypeCoercion.Enabled })` will
       * succeed.
       */
      Enabled
   }
}
```

Kantis:
Closes #2520 

- [x] docs
- [x] changelog

### DIFF
--- a/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
+++ b/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
@@ -2,12 +2,7 @@
 
 package io.kotest.assertions.json
 
-import io.kotest.assertions.json.CompareJsonOptions.ArrayOrder
-import io.kotest.assertions.json.CompareJsonOptions.NumberFormat
-import io.kotest.assertions.json.CompareJsonOptions.PropertyOrder
-import io.kotest.assertions.json.CompareJsonOptions.TypeCoercion
-
-@Deprecated("Json comparison options is now specified with `CompareJsonOptions`")
+@Deprecated("Json comparison options is now specified with `CompareJsonOptions`", ReplaceWith("TypeCoercion"))
 enum class CompareMode {
    /**
     * Types must be identical and compare by value.
@@ -18,7 +13,7 @@ enum class CompareMode {
     */
    @Deprecated(
       "Json comparison options is now specified with `CompareJsonOptions`",
-      ReplaceWith("CompareJsonOptions.TypeCoercion.Disabled")
+      ReplaceWith("TypeCoercion.Disabled")
    )
    Strict,
 
@@ -30,12 +25,12 @@ enum class CompareMode {
     */
    @Deprecated(
       "Json comparison options is now specified with `CompareJsonOptions`",
-      ReplaceWith("CompareJsonOptions.TypeCoercion.Enabled")
+      ReplaceWith("TypeCoercion.Enabled")
    )
    Lenient,
 }
 
-@Deprecated("Json comparison options is now specified with `CompareJsonOptions`")
+@Deprecated("Json comparison options is now specified with `CompareJsonOptions`", ReplaceWith("PropertyOrder"))
 enum class CompareOrder {
    /**
     * All object properties must be in same order as expected.
@@ -44,13 +39,13 @@ enum class CompareOrder {
     */
    @Deprecated(
       "Json comparison options is now specified with `CompareJsonOptions`",
-      ReplaceWith("CompareJsonOptions.PropertyOrder.Strict")
+      ReplaceWith("PropertyOrder.Strict")
    )
    Strict,
 
    @Deprecated(
       "Json comparison options is now specified with `CompareJsonOptions`",
-      ReplaceWith("CompareJsonOptions.PropertyOrder.Lenient")
+      ReplaceWith("PropertyOrder.Lenient")
    )
    Lenient,
 }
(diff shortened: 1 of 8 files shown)


kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/matchers.kt
```diff
@@ -18,59 +19,50 @@ fun equalJson(
...
-   shouldEqualSpecifiedJson(expected, mode, CompareOrder.Lenient)
-
-fun String.shouldNotEqualSpecifiedJson(expected: String, mode: CompareMode) =
-   shouldNotEqualSpecifiedJson(expected, mode, CompareOrder.Lenient)
+   shouldNotEqualJson(expected, legacyOptions(CompareMode.Strict, order))
 
-fun String.shouldEqualSpecifiedJson(expected: String, order: CompareOrder) =
-   shouldEqualSpecifiedJson(expected, CompareMode.Strict, order)
+infix fun String.shouldEqualSpecifiedJson(expected: String) =
+   shouldEqualJson(expected, compareJsonOptions { fieldComparison = FieldComparison.Lenient })
```
Kantis, line 65: Worth saving as a lazy private val, to avoid object creations?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -2,24 +2,24 @@
...
-enum class CompareMode {
-
-   /**
-    * Types and formats must be identical.
-    *
-    * For example, "true" and true will not match because one is a string, and the other is a boolean.
-    * Similarly, 2.99E9 and 299000000 would not match because despite both being doubles, and the same value,
-    * they are not using the same representation.
-    */
-   Exact,
```
Kantis, line 14: Only existed in snapshots

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -28,75 +28,173 @@ enum class CompareMode {
...
+    */
+   var fieldComparison: FieldComparison = FieldComparison.Strict,
+
+   /**
+    * Controls whether number formatting should be taken into consideration. For instance, comparing 1.00 to 1.0, or
+    * 1E2 to 100
+    */
+   var numberFormat: NumberFormat = NumberFormat.Lenient,
 
-enum class FieldComparison {
```
Kantis, line 70: same for this

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -28,75 +28,173 @@ enum class CompareMode {
...
+class CompareJsonOptions(
    /**
-    * Ignore the order of object properties and arrays.
-    *
-    * For example, { "x": 14.2, "y": 13.0 }` and `{ "y": 13.0, "x: 14.2 }` would also be considered equal,
-    * since they have the same properties and values.
-    * However, `[1, 2]` and `[2, 1]` would NOT be considered equal
+    * Controls whether property order must be identical
     */
-   LenientProperties,
```
Kantis, line 58: and this

Kantis:
Updated docs: 
bild
bild

sksamuel (approved):

busches:
Any chance this is released soon? I'm asking as we saw this matcher was available in the [docs](https://kotest.io/docs/assertions/json-matchers.html), attempted to start using it and `CompareOptions` is not a thing, yet is per the documentation. The documentation being so out of sync with the actual project for over a month is not ideal.

PR #2524, opened 2021-09-19
Added engine extensions #2463

### LINKED ISSUE #2463: Make explicitTags accessible to aroundProject or similar and improve Tags to allow interrogation
In #2462 the concept of explicit tags is introduced. Project listeners should be able to interact with the explicit tags as well as modifying the tags specified. Consider a test suite with unit and integration tests. 

If no tags are specified, then I want to set the `Unit` tag. 
If an integration tag `FooIntegration` is specified, then I want to set up the Foo thing for the rest of the test run.

```kotlin
object Unit : Tag()
object FooIntegration : Tag()

suspend fun configureFoo(): Unit = TODO()

object : BeforeProjectExtension {
  override suspend fun aroundProject { config ->
    if (config.explicitTags == null) {
      config.explicitTags = Tags(included = Unit)
    } else {
      if (config.explicitTags.enabled(FooIntegration) {
        configureFoo()
      }
    }
    
    callback()
    
    if (config.explicitTags.enabled(FooIntegration) {
      cleanupFoo()
    }
  }
}
```

We could also have a tags listener to modify the tags (`TagOverrideExtension`) and pass an immutable copy of the enabled/disabled logic to the project extension. There's power in just knowing that a tag is enabled without having to write your own inadequate parser: 

```
val tags = configuration.activeTags().expression
if (tags.contains("Unit") && !tags.contains("!Unit")) { ...
```

sksamuel:
Closes #2463

Introduces EngineExtensions which allow users to write an extension to change:
- tags
- config
- specs

This gives the power needed for #2463

### DIFF
--- a/kotest-common/src/commonMain/kotlin/io/kotest/common/annotations.kt
+++ b/kotest-common/src/commonMain/kotlin/io/kotest/common/annotations.kt
@@ -5,10 +5,22 @@ package io.kotest.common
 @RequiresOptIn(level = RequiresOptIn.Level.WARNING)
 annotation class ExperimentalKotest
 
+/**
+ * Any feature annotated with [DelicateKotest] is an advanced feature.
+ */
 @MustBeDocumented
 @Retention(value = AnnotationRetention.BINARY)
 @RequiresOptIn(level = RequiresOptIn.Level.WARNING)
 annotation class DelicateKotest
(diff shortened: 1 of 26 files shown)


jschneidereit:
This looks great!

PR #2526, opened 2021-09-20
Property testing should use bind as default for data class #2355

### LINKED ISSUE #2355: Property testing should use bind as default for data class

sksamuel:
Closes #2355

### DIFF
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/targetDefaultForClassName.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/targetDefaultForClassName.kt
@@ -9,6 +9,7 @@ import java.time.LocalDate
 import java.time.LocalDateTime
 import java.time.LocalTime
 import java.time.Period
+import kotlin.reflect.KClass
 import kotlin.reflect.full.isSubclassOf
 
 @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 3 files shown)


kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/targetDefaultForClassName.kt
```diff
@@ -48,6 +49,10 @@ actual inline fun <reified A> targetDefaultForClass(): Arb<A>? {
       A::class == LocalTime::class -> Arb.localTime() as Arb<A>
       A::class == Period::class -> Arb.period() as Arb<A>
       A::class == BigDecimal::class -> Arb.bigDecimal() as Arb<A>
+      A::class.isData -> {
```
myuwono, line 52: @sksamuel I wonder would it be useful to provide users the ability to register / inject their own data class arbs?

myuwono (approved):

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/targetDefaultForClassName.kt
sksamuel: Easy enough to do that by just specifying the arb to the checkAll method like `checkAll(1000, Arb.myDataArb())`
This is for when you want to do `checkAll<String, Foo> { ... }`
But with your recent idea who knows how checkAll will advance.

PR #2529, opened 2021-09-20
Feature/checkall

sksamuel:


kotest-property/src/commonMain/kotlin/io/kotest/property/checkall.kt
```diff
@@ -0,0 +1,32 @@
+package io.kotest.property
+
+// todo have these methods delegate to one that uses PropTestConfig
+fun checkAll(iterations: Int, seed: Long, test: PropertyContext.() -> Unit): PropertyContext {
+   val context = PropertyContext(RandomSource.seeded(seed))
+   repeat(iterations) {
+      try {
+         context.test()
+      } catch (t: Throwable) {
+         // shrink here then throw
```
sksamuel, line 10: @myuwono shrinking here

myuwono: this is actually harder than i originally anticipated.. ive been thinking about this but it's so hard to make it reliable for inline created arb.. i'll share some of my findings soon.

in a nutshell if the test has something along these lines, it's somewhat difficult to replay the lambda block with the injected shrunk value unless we can have a reliable mapping between `Arb<*> -> RTree<*>`. We can't do that right now as hashCode of an Arb<*> will always be different all the time.. Neither can we infer that from the value type (see else block on how c was generated)
```kotlin
checkAll {
  val a = Arb.string(...).value() // new arb
  val b = anotherArb(a).value() // dependent arb
  if (somePredicate(b)) {
    val c: C = yetAnotherArb(a, b).value() // another arb inside an if, 
    val expected = ...
    c shouldBe ... // consider the test fails here
  } else {
    val c: C = notTheSameCArbButTheSameType(a, b).value() // another arb that is not the same third arb
    val d: D = arbD(a, c).value() // arbD, a fourth arb that only exists in the else block
    val expected = ...
    d shouldBe ...
  }
}
```

sksamuel:
Perhaps we leave this as a back burner thing that we think about for 5.1

nomisRev:
This feature looks awesome 😍

kotest-property/src/commonMain/kotlin/io/kotest/property/checkall.kt
sksamuel: Yes I see your point. We can't replay the lambda with shrunk values. Even if we hijack .value to return shrinks when in "shrink" mode, the different paths might mean sometimes we need more values than we had originally.

sksamuel:
@myuwono 

```
         checkAll {

            iterations = 100
            seed = 234234234

            beforeProperty {

            }

            afterProperty {

            }

            val a: Arb<Int> = Arb.int()
            val b: Arb<String> = a.map { it.toString() }
            val c: Arb<Boolean> = Arb.boolean()

            test(a, b, c) { a1, b2, c3 ->
               a1.toString() + (b2 + c3) shouldBe (a1.toString() + b2) + c3
            }
         }
```

myuwono:
hmm yeah that's definitely an improvement to the ergonomics! I like the fact that you can specify the iterations and seed directly within the checkAll, but i guess it's not exactly removing the need of pre-committing to arbs upfront. We're still stuck in needing to pass in those arbs to be unpacked and used inside of the lambda.. I understand why we're doing it though, because otherwise shrinking won't quite work.

```kotlin
// currently
test("should test something") {
  checkAll {
    test(arbA, arbB, arbC) { a1, b2, c3 -> // naming can be changed. observe primarily the upfront commitment of arbs
       a1.toString() + (b2 + c3) shouldBe (a1.toString() + b2) + c3
    }
  }
}

// what i sort-of expected to be able to do, but very difficult because of shrinking just won't really work
proptest("should test something") {
  val result = arbA.value().toString() + (arbB.value() + arbC.value())
  result shouldBe ...
}
```

what that also enables is perhaps a seed-propagation via the coroutine context so that developer all of a sudden can do something like this:

```kotlin
proptest("should propagate property-test-context") { // this: PropContext 
  val state = setup(arbParamA.value(), arbParamB.value()) 
  val expected = ...
  sut(state).doSomething() shouldBe expected
}

suspend fun setup(params...): TestState = propContext { // this: PropContext - the random seed and value syntax is available here
  TestState(
    userInStore = Arb.list(arbUser, 1..10).value(),
    somethingElseStateful = arbXyz.value()
  )
} 
```

This is a very powerful feature because all of a sudden devs are able to do more complex aspects of property tests while still keeping those repeatable because of random seed. Currently there's no stopping devs from creating non-repeatable flaky tests by calling `.single()` without random seed everywhere, because they can. :( this feature will be massively beneficial.

I wonder if we should try this out without shrinking support as the first phase? Maybe use an `OptIn:ExpermentalKotest` annotation?

myuwono:
@sksamuel i've opened an issue instead so that we can track that better.

nomisRev:
> what that also enables is perhaps a seed-propagation via the coroutine context so that developer all of a sudden can do something like this:

You can keep any kind of state inside a `suspend` block, but it should be possible to enhance the receiver of the `checkAll` lambda with `fun <A> Arb<A>.bind(): A`.

I.e.

```
interface PropSyntax {
  val ctx: TestContext
  fun <A> Arb<A>.bind(): A
}

fun <A, B> checkAll(arbA: Arb<A>, arbB: Arb<B>, check: suspend PropSyntax.(A, B) -> Unit): Unit =
  TODO()
  
  
// We need to track state here `bind`, `seed`, tec
class PropSynytaxImpl(
  override val ctx: TestContext,
  val seed: Long
): PropSyntax {
  val random: Random = Random(seed)
  
  fun <A> Arb<A>.bind(): A =
     nextValue(random) // ??? 
}
```

It's also possible to put a single SEED for the whole program:

```
data class SeedContext(val value: Long) : AbstractCoroutineContextElement(SeedContext.Key) {
  companion object Key : CoroutineContext.Key<SeedContext>
}

// This is kinda like an implicit which might be there at runtime lookup. O(1) access
suspend fun seed(): Long? = coroutineContext[SeedContext]?.value
```

I'm not familiar enough with the internals to know how to correctly keep track of the state for shrinking etc. Not sure if this helps, these are patterns I've been playing with lately.
Here is a small repo where I'm experimenting with mixing MTL style effects, in a Kotlin idiomatic style. With multiple receivers perhaps it'll become easier to encode this in a less complex way.
https://github.com/nomisRev/KotlinEffectsPlayground

sksamuel:
@nomisRev the issue isn't tracking the values because we can store the result of .value() in some context before returning it.

The issue is that for a given "test" if we don't have an explicit function, say (A,B) -> C we don't know how many values to provide when shrinking. You might use 3 on the first path, but on the shrink case might need four. @myuwono had a good example:

```kotlin
checkAll {
  val a = Arb.string(...).value() // new arb
  val b = anotherArb(a).value() // dependent arb
  if (somePredicate(b)) {
    val c: C = yetAnotherArb(a, b).value() // another arb inside an if, 
    val expected = ...
    c shouldBe ... // consider the test fails here
  } else {
    val c: C = notTheSameCArbButTheSameType(a, b).value() // another arb that is not the same third arb
    val d: D = arbD(a, c).value() // arbD, a fourth arb that only exists in the else block
    val expected = ...
    d shouldBe ...
  }
}
```

I am convinced now that we must always require an explicit function somewhere.

nomisRev:
If `yetAnotherArb` and `notTheSameCArbButTheSameType` are referential transparent, for the same `A`, `B` and `SEED` both `C`s resulting from `(A, B) -> Arb<C>` should be deterministic, no?

nomisRev:
I know `Arb#flatMap` is tricky to implement, but I'm not familiar with the implementation. Not sure how that affects the code here.

myuwono:
@nomisRev right i believe that's a slightly different problem. for flatMaps we've got that covered in https://github.com/kotest/kotest/issues/2493 with `arbitrary` - drawing some inspiration from arrow. This problem is slightly more involved unfortunately. I still can't figure out a way to make shrinking work using the proposed `checkAll` and `.value()` construct. Would love to hear any kind of input from you there..

this is already possible in kotest 5.x with https://github.com/kotest/kotest/pull/2494 merged 
```kotlin
// with continuation
// notice we also have the random source here to possibly do more complex generations
val arb: Arb<Foo> = arbitrary { rs: RandomSource -> 
   val first = arbFirst.bind()
   val second = arbSecond(first).bind()
   val third = arbThird(first, second).bind()
   Foo(first, second, third)
}
```

that syntax already takes into account edgecases as well. As well that arbitrary syntax accepts `Shrinker<T>` as part of the param so that's already a superb ergonomic improvement to what was previously a nested flatmap in kotest 4.x.

This `checkAll` proposal pushes this to the next level.

sksamuel:
What about if .value() on an arb didn't actually bring back just a value,
but instead checked the context for a shrink. If the shrink is null, then
it tries to generate via sample.

That way, if you entered a different code path the next time, all values
would still be available.

PR #2532, opened 2021-09-21
Enhancement: support Arb.bind for more complex data classes

Kantis:
playing around with generating arbs for any data class


sksamuel:
Is this the same as this ?
https://github.com/kotest/kotest/blob/master/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/jvmbind.kt#L29

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/maps.kt
```diff
@@ -71,8 +71,18 @@ class MapShrinker<K, V> : Shrinker<Map<K, V>> {
 
 /**
  * Returns an [Arb] that produces Pairs of K,V using the supplied arbs for K and V.
+ * Edgecases will be derived from [k] and [v].
  */
-fun <K, V> Arb.Companion.pair(k: Arb<K>, v: Arb<V>) = arbitrary { rs ->
+fun <K, V> Arb.Companion.pair(k: Arb<K>, v: Arb<V>) = arbitrary(
+   edgecaseFn = { rs ->
+      when (rs.random.nextInt(100)) {
```
sksamuel: same as random 4 and the first being 1,2 ?

sksamuel (approved):

PR #2559, opened 2021-10-04
Added comparable matcher result and applied to shouldContainExactly

### LINKED ISSUE #1851: shouldContainExactly should show diff in IntelliJ

sksamuel:
Closes #1851

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
@@ -1,5 +1,6 @@
 package io.kotest.matchers.collections
 
+import io.kotest.assertions.AssertionsConfig
 import io.kotest.assertions.eq.eq
 import io.kotest.assertions.show.Printed
 import io.kotest.assertions.show.show
@@ -67,7 +68,7 @@ fun <T, C : Collection<T>> containExactly(expected: C): Matcher<C?> = neverNullM
 
    val negatedFailureMessage = { "Collection should not contain exactly ${expected.printed().value}" }
 
-   if (actual.size <= 20 && expected.size <= 20) {
+   if (actual.size <= AssertionsConfig.maxCollectionEnumerateSize && expected.size <= AssertionsConfig.maxCollectionEnumerateSize) {
       ComparableMatcherResult(
          passed,
          failureMessage,
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
@@ -3,15 +3,35 @@ package io.kotest.matchers.collections
 import io.kotest.assertions.eq.eq
 import io.kotest.assertions.show.Printed
 import io.kotest.assertions.show.show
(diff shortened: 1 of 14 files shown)


ashishkujoy:
Looks Good to me.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
```diff
@@ -3,15 +3,35 @@ package io.kotest.matchers.collections
...
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldNot
 import kotlin.jvm.JvmName
 
+/**
+ * Verifies that this collection contains the sub collections provided in the exact given order.
+ *
+ * So, for example, listOf(1,2,3) contains exactly the sub collections:
+ *
+ * [], [1], [2], [3], [1,2], [2,3] and [1,2,3].
```
Kantis, line 20: Suppose this is out of scope for the PR, but I was a bit surprised when reading this comment. I expected `shouldContainExactly` to match if, and only if, the actual collection contains _exactly_(?) the elements in expected. I.e. no more and no less than the given. 

E.g.

```kotlin 
listOf(1, 2, 3, 3) shouldContainExactly listOf(1, 2, 3) -> false
setOf(1, 2, 3, 3) shouldContainExactly listOf(1, 2, 3) -> true 
setOf(1, 2, 3, 3) shouldContainExactly listOf(2, 3) -> false
```

Perhaps it's you might as well be using `x shouldBe y` in that case, but well. 

Just thought I'd throw it out there.

Kantis: Actually I think this is the actual behaviour, so the comment is lying? 

Given the below test, `actual` contains the subcollection defined in `expected`, but also an additional element causing the test to fail?
bild

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
```diff
@@ -45,10 +65,23 @@ fun <T, C : Collection<T>> containExactly(expected: C): Matcher<C?> = neverNullM
       sb.toString()
    }
 
-   MatcherResult(
-      passed,
-      failureMessage
-   ) { "Collection should not be exactly ${expected.printed().value}" }
+   val negatedFailureMessage = { "Collection should not contain exactly ${expected.printed().value}" }
+
+   if (actual.size <= 20 && expected.size <= 20) {
```
LeoColman: Maybe this 20 could be extracted to a system property, just for the sake of configurability.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
Kantis: I realize I got it the wrong way around. The test that caused failure has an extra in actual, not in expected. Sorry 😞

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactly.kt
sksamuel: Yep good idea

PR #2576, opened 2021-10-17
Refactor `shouldHaveSameStructure`

JavierSegoviaCordoba:
- Rename to `shouldHaveSameStructureAs`
  - It doesn't check files content, only if the tree is the same
- Added `shouldHaveSameStructureAndContentAs`
  - It checks if the tree and the content of the files is the same

For both new assertions, add filters and comparison functions


sksamuel (approved):

sksamuel:
lgtm

PR #2582, opened 2021-10-24
shouldBe returns Element <T>

i-walker:
this would be nice for composition e.g.:

```kotlin
forAll(Arb.int(), Arb.int()) { a, b ->
   val c = (a + b) shouldBe (b + a)
   c - a == b
}
```


sksamuel:
This is nice, but I'm not sure we can change the signature of the most used method in the codebase.
If people have this function as the last statement in a block, then that will cause a compile time error.

i-walker:
If this can be introduced with a different name in the same package that would be nice too, unless we shouldn't have this signature at all?

sksamuel:
two shouldBe's will be very confusing, if they only differ by return value.

i-walker:
Lets wait on this until 1.6 or 1.7 :)

PR #2586, opened 2021-10-27
Update continually.md to remove deprecated function

busches:
`Int.seconds` is deprecated in favor of `Duration.seconds(Int)`.

```kotlin
@SinceKotlin("1.3")
@ExperimentalTime
@Deprecated("Use Duration.seconds() function instead.", ReplaceWith("Duration.seconds(this)", "kotlin.time.Duration"))
public val Int.seconds get() = toDuration(DurationUnit.SECONDS)
```


Kantis:
Thanks! 🙂

PR #2589, opened 2021-10-30
Inspectors should return the collection to allow chaining #2588

### LINKED ISSUE #2588: Inspectors should return the collection to allow chaining
Inspectors currently return Unit. it would be nice if they returned the collection they are invoked on to allow chaining inspector calls

sksamuel:
Closes #2588

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/inspectors/Inspectors.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/inspectors/Inspectors.kt
@@ -1,68 +1,34 @@
 package io.kotest.inspectors
 
-fun <T> Sequence<T>.forAll(fn: (T) -> Unit): Sequence<T> {
-   toList().forAll(fn)
-   return this
-}
-
-fun <T> Array<T>.forAll(fn: (T) -> Unit): Array<T> {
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/inspectors/Inspectors.kt
```diff
@@ -1,84 +1,173 @@
...
-  }
+      val msg = "${passed.size} elements passed but expected $k"
+      buildAssertionError(msg, results)
+   }
+   return this
+}
+
+fun <T> Sequence<T>.forSome(fn: (T) -> Unit): Sequence<T> {
+   toList().forSome(fn)
+   return this
```
dalewking: FYI for more concise style you might think about something like this instead:
```
fun <T> Sequence<T>.forSome(fn: (T) -> Unit) = apply { toList().forSome(fn) }
```

sksamuel: Yes I like that.

sksamuel: Done

PR #2615, opened 2021-11-02
Fixes withEnvironment empty system environment variables on Linux 

### LINKED ISSUE #2614: withEnvironment erases environment variables on Linux
4.6.0

`withEnvironment` stores the originalEnvironment as reference and not as an immutable map on Linux:
grafik

Therefore a call to `envMapOfVariables.clear()` erases the original environment.
grafik

Xaseron:
fixes #2614 

This PR:
- Reorders the test of `withEnvironment` to provoke the failure
- Uses `toMap()` to create an immutable copy of the original system environment


sksamuel:
Nice fix thanks.

PR #2618, opened 2021-11-03
Revert json.md from "CompareJsonOptions (#2523)"

### LINKED ISSUE #2616: JSON Matchers Documentation Incorrect
New options were added with #2523 for JSON Matchers, which look to be a great improvement over what currently exists. The issue is that the documentation, https://kotest.io/docs/assertions/json-matchers.html, makes it look like all of these options are available, but they are not.

As someone who is fairly new to using Kotest, this was super confusing to us, until we found the above PR and realized the documentation is ahead of the released product.

1) Can we get this feature updated so the documentation is accurate
2) How can we avoid this in the future?

busches:
This reverts json.md from commit 3bfeb50c603b2db6e0549e0eb81fbbfccd2e6f0d and resolves #2616.

This PR should be reverted once 5.0.0 is released or rewritten if docs for 4.X and 5.X are to be maintained.

### DIFF
--- a/documentation/docs/assertions/json.md
+++ b/documentation/docs/assertions/json.md
@@ -36,41 +36,85 @@ and
 The inverse of this matcher is `shouldNotEqualJson` which will error if two JSON strings
 _are_ considered equal.
 
-### compareJsonOptions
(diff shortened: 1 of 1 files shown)


Kantis:
Thanks, and sorry for the slip-up.

PR #2625, opened 2021-11-05
[docs] Timeout - Remove deprecated time properties

busches:
`<Int>.<Duration>` are all deprecated in favor of `Duration.<DurationType>(Int)`

This also had a usage of `1.second` which is not a thing.

### DIFF
--- a/documentation/docs/framework/timeout.md
+++ b/documentation/docs/framework/timeout.md
@@ -45,7 +45,7 @@ The most finely grained location for timeouts it on leaf tests directly.
 class TimeoutTest : DescribeSpec({
 
    describe("my test context") {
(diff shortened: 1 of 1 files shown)


Kantis:
There's a few more occurrences. Would you mind grabbing the last few as well? This regex should find most I think `\s\d+\.((milli|micro|nano)?seconds|days|minutes|hours)`

busches:
I did this the lazy way via the browser :) I can replace all occurrences later when I can check this out to a computer.

sksamuel:
I think it's undeprecated in 1.6

Kantis:
@sksamuel seems like it yeah. :) https://youtrack.jetbrains.com/issue/KT-46229

We close this then?

sksamuel:
Yeah they deprecated then undeprecated

Issue #2636, opened 2021-11-09
`shouldNotBeEqualToComparingFieldsExcept` doesn't work as intended + Incorrect Param Type


dongsubkim:
**Which version of Kotest are you using**
Kotest version: 4.6.3 
Issues can be reproduced with 5.0.0.M3 as well

Bugs demonstration.
```kotlin
class Test {

    @Test
    fun test() {
        val john = Person("john", 30)
        val david = john.copy(name = "david")
        val youngerJohn = john.copy(age = 20)
        val johnCopy = john.copy()

        john.shouldBeEqualToComparingFields(johnCopy) // pass
        john.shouldBeEqualToComparingFieldsExcept(johnCopy, Person::name) // pass
        john.shouldBeEqualToComparingFieldsExcept(david, Person::name) // pass
        john.shouldBeEqualToComparingFieldsExcept(youngerJohn, Person::age) // pass
        
        john.shouldNotBeEqualToComparingFieldsExcept(david, Person::age) // fail -> should pass
        john.shouldNotBeEqualToComparingFieldsExcept(youngerJohn, Person::age) // pass -> should fail

        john.shouldBeEqualToComparingFieldsExcept(youngerJohn, john::age) // fail -> should pass
        john.shouldNotBeEqualToComparingFieldsExcept(david, john::age) // fail -> should pass

    }

    data class Person(
        val name: String = "",
        val age: Int = 0
    )
}
```
1. `shouldNotBeEqualToComparingFieldsExcept` does not work as intended.
https://github.com/kotest/kotest/blob/625422441c48a9a3f215df318dc11f5c2530c5bb/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt#L350-L356
it seems it needs to use `shouldNot` instead of `should`

2. `*ComparingFields*` methods accepts `ignoreProperty: KProperty<*>` params, but `beEqualComparingFields` method uses `KProperty1` type. So when the ignoreProperties are provided with `KProperty0`(created with instance like `john::age` instead of class `Person::age`) filtering does not work which causes test failures.
I assume it can be fixed by either change `beEqualComparingFields` method to filter both `KProperty1` and `KProperty0` OR change param type of `*ComparingFields*` methods

PR #2637, opened 2021-11-09
bugfix/2636 Change should to shouldNot on shouldNotBeEqualToComparingFieldsExcept

dongsubkim:
#2636


sksamuel:
Thank you for the fix.

PR #2641, opened 2021-11-11
Renamed Show to Print to align with the output type

sksamuel:


sksamuel:
Going to merge this in so I can continue to work, but please review retroactively and let me know if you find any issues.

PR #2669, opened 2021-11-24
Remove accidentally nested Try

Mahoney:
`submitAll` returns a `Try`. When it was called the returned `Try` was nested
inside another `Try`.

The result is that a failure inside `submitAll` ended up as a
`Success<Failure<Unit>>`, and so the error handling never kicked in and the
failure was swallowed.


sksamuel:
Great thanks

PR #2688, opened 2021-11-28
Compiler plugins should be thread safe #2687

### LINKED ISSUE #2687: Compiler plugins list of specs should be thread safe

sksamuel:
Fixes #2687


sksamuel:
Seems to fix it @JoonasC

PR #2691, opened 2021-11-29
Fix bug where autoClose initialises itself when closing

nomisRev:
When `autoClose` is used with `Lazy` but never invoked it will still `initialise` itself before `AutoCloseable#close`.
This scenario is occurs when not all tests in a `Spec` rely on the `autoClose`, and you run a test form IDEA that doens't rely on `autoClose`. We then see a undeeded `init` + `close`, this can be time-consuming for example with `test-containers`.

This PR levereages `Lazy#isInitialised` before calling `close` so it's possible to use `autoClose` + `lazy` in a completely lazy way.


nomisRev:
@sksamuel any idea why it failed on ubuntu? I can't seem to spot the test-reports again... :/

nomisRev:
Never mind, something broke in the other tests. I will look into it and fix it.

Does this only run on JVM? I saw you had `AutoCloseable` in Kotest common.

sksamuel:
There's a typealias and it maps to the java.lang.AutoCloseable on JVM.

sksamuel:
@nomisRev do you want me to fix up and merge ?

sksamuel (approved):

nomisRev:
@sksamuel I saw this fail with unrelated:

"Shrinks all components to minimum value[jvm]"
```text
java.lang.AssertionError: "Property test failed for inputs

0) MaximumComponents(a=1000, b=1000, c=1000, d=1000, e=1000, f=1000, g=1000, h=1000, i=1000, j=1000, k=1000, l=1000, m=1000, n=1000)

Caused by java.lang.AssertionError: 1000 should be < 100 at
	com.sksamuel.kotest.property.arbitrary.BindShrinkTest$1$2$stdout$1$1$1.invokeSuspend(BindShrinkTest.kt:54)
	com.sksamuel.kotest.property.arbitrary.BindShrinkTest$1$2$stdout$1$1$1.invoke(BindShrinkTest.kt)
	com.sksamuel.kotest.property.arbitrary.BindShrinkTest$1$2$stdout$1$1$1.invoke(BindShrinkTest.kt)
	io.kotest.property.internal.ProptestKt$proptest$3$2.invokeSuspend(proptest.kt:42)

" should include substring "Shrink result (after 45 shrinks) => MaximumComponents(a=0, b=0, c=0, d=0, e=0, f=0, g=0, h=0, i=0, j=0, k=0, l=0, m=100, n=0)"
```

PR #2692, opened 2021-11-29
All collection elements should match a predicate

kkocel:


sksamuel:
The nomenclature doesn't align with the rest of the matchers, as they all begin with should.

See also this ticket: https://github.com/kotest/kotest/issues/2145 where we are trying to standardize and decide what to do with predicate matchers. Your input there is welcome!

kkocel:
@sksamuel I renamed it to `shouldMatchAll`

kkocel:
@sksamuel I saw your answer on SO which is a far better option. Closing this PR

PR #2695, opened 2021-11-29
Collection assertions

### LINKED ISSUE #2145: Why does shouldHaveSingleElement and other Collection "shoulds" have predicate instead of block for assertions?

Kantis:
Highlights:
* Deprecates `shouldExist` in favor of `forAny` 
* Adds `forSingle` inspector, that asserts that a collection has a single element, which matches the assertion
* Adds lambda-based alternatives to predicate-based `existInOrder` matcher
  *  `matchEach` 1:1 match element:assertion
  * `matchInOrder` assert in order, with no gaps
  * `matchInOrderSubset` assert in order, gaps allowed


Kantis:
Any feedback on the naming of `existInAssertedOrder()`? 

I would ideally have liked to use `existInOrder(assertions: List<(T) -> Unit>)` for this one as well, but it seemed to clash with the existing `existInOrder(ps: List<(T) -> Boolean>)`

sksamuel:
I don't like the name `existInAssertedOrder`. 
Perhaps `shouldAssertInOrder` but we need one that works with gaps and one that doesn't.

`shouldAssertSubset` <- gaps ?
`shouldAssert` <- no gaps ?

Kantis:
`shouldAssert` means the matcher would be named `assert`? I think that might be confusing?

Maybe `shouldFullyMatch`, `shouldPartiallyMatch`? 🤔

I agree it would be nice to offer both a complete match opportunity, and a subset match.. Perhaps it would also be nice to be able to specify if they shouold exist in order or not..

sksamuel:
shouldPassEach

shouldApplyInOrder

shouldAssertEach

Kantis:
`shouldMatchInOrder` - Test vararg of assertions against collection, does not require exhaustive test
`shouldMatchInOrderWithoutGaps` - same as above, disallowing gaps
`shouldMatchAll` - same as above, but must be exhaustive for the list. E..g assertion 1 must pass element 1, and one assertion must exist for each element

Matchers would then be `matchAll` and `matchInOrder`, latter taking a parameter to decide whether gaps are allowed. 

Do you prefer another verb?

sksamuel:
I think those names should good, and maybe shouldMatchInOrderWithoutGaps could be shouldMatchInOrderSubset

sksamuel:
@Kantis can you rebase whenever you're able :)

sksamuel:
@Kantis do you have time to finish this up for 5.1 release this week?
If not I can move to 5.2

Kantis:
> @Kantis do you have time to finish this up for 5.1 release this week? If not I can move to 5.2

I'll try ✌🏼 Sorry for not responding for a while, I mentally checked out over the holidays

sksamuel:
So did I. Was wonderful :)

Kantis:
@sksamuel this one's ready ✌🏼

sksamuel:
It's been a while :)
Can you remind me what the matchEach is solving again.

PR #2700, opened 2021-12-01
Remove Arb#long workaround for incorrect randomly generated values

BartvHelvert:
This workaround is redundant because it was fixed in Kotlin 1.6.0.

See [KT-47304](https://youtrack.jetbrains.com/issue/KT-47304).


sksamuel:
lgtm

PR #2705, opened 2021-12-02
Fixes #2620

### LINKED ISSUE #2620: JSON Matchers does not offer "Click to see difference"
When using a JSON matcher and the assertion fails, there is no "Click to see difference" option in IntelliJ to quickly compare them.

Working example:
`401 shouldBe 402`
image

JSON Example:
`"""{"a":1}""" shouldMatchJson """{"a":2}"""`
image

This is very useful when comparing larger JSON objects.

Comparing the output, it looks like the JSON example values need to be wrapped in `<X>`.

Kantis:


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/should.kt
```diff
@@ -78,7 +77,7 @@ fun <T> equalityMatcher(expected: T) = object : Matcher<T> {
          {
             val e = Expected(expected.print())
             val a = Actual(value.print())
-            failure(e, a).message ?: intellijFormatError(e, a)
+            failure(e, a).message ?: "expected:<${e.value.toString()}> but was:<${e.value.toString()}>"
```
sksamuel: Why do it manually and not use intellij format ?

Kantis:
`shouldMatchJson`  edit: Casing is aligned with the 2nd screenshot now..
image

`shouldMatchJsonResource` 
image

`shouldEqualJson` 
image
~This one feels a bit ugly?~ seems it's because of the linebreaks in the json that makes the intellij regexp miss.. looking for a solution

--- 
Update: Regarding the last screenshot - it seems that linebreaks in the `expected:<x> but was:<y>` message always causes message to look garbled. Below is a string comparison with multi-line strings for instance. Not sure what a good solution would be, but I think it's out of scope for this fix.

image

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/should.kt
Kantis: I'm not sure if the intellij format is needed here? 

I wanted to get hide the intellij format. It was only used from `failures.kt`, apart from here. I made the `intellijFormatError` private to `failures.kt` now so no one would use it and think the values would diff (like I did) :)

Kantis: Okay.. seems like it's needed even though values are provided in the `AssertionFailedError`.. I'll revert that bit

Issue #2704, opened 2021-12-02
LocalDateTime arb should accept localdatetimes as min and max


LeoColman:
https://github.com/kotest/kotest/blob/a1399074597304d1bcf7e91a80469772e9cdac9f/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt#L87

The arbs for both `localDate` and `localTime` have an equal type as parameter. LocalDateTime doesn't. As a LocalDateTime is a LocalDate + LocalTime, this should not be an issue.

To fix this:
1. Deprecate `Arb.localDateTime(minYear: Int, maxYear: Int)`
2. Add `Arb.localDateTime(minLocalDateTime: LocalDateTime, maxLocalDateTime: LocalDateTime)`
3. Add tests to the new function

pientaa:
I'd like to take that issue

LeoColman:
Absolutely!
Don't hesitate to ask for help if you need any :)

sksamuel:
We can keep the old one as well as the new one if we want, since it's much simpler to use if you aren't too fussy on the values.

LeoColman:
I suggested deprecating it. Maybe keep it not deprecated?

PR #2711, opened 2021-12-05
LocalDateTime arb should accept localdatetimes as min and max

pientaa:
kotest/kotest#2704


kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
```diff
@@ -84,21 +81,47 @@ fun Arb.Companion.localTime(): Arb<LocalTime> = arbitrary(listOf(LocalTime.of(23
  * @see [localDateTime]
  * @see [localTime]
  */
-fun Arb.Companion.localDateTime(
+fun Arb.Companion.localDateTimeBetweenYears(
```
nomisRev: This seems like a binary breaking change.

Can this be solved with `@JvmName("localDateTimeBetweenYears") and a Kotlin overload instead of 2 different function names?

pientaa: actually the problem is in  `actual inline fun <reified A> targetDefaultForClass(): Arb<A>?`:
```
Overload resolution ambiguity. All these functions match.
public fun Arb.Companion.localDateTime(minLocalDateTime: LocalDateTime = ..., maxLocalDateTime: LocalDateTime = ...): Arb<LocalDateTime> defined in io.kotest.property.arbitrary in file dates.kt
public fun Arb.Companion.localDateTime(minYear: Int = ..., maxYear: Int = ...): Arb<LocalDateTime> defined in io.kotest.property.arbitrary in file dates.kt
```

and `@JvmName` does not fix it.

pientaa: I rolled back function name's change and removed default values in the old function `localDateTime`, to fix it another way

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
```diff
@@ -85,20 +82,46 @@ fun Arb.Companion.localTime(): Arb<LocalTime> = arbitrary(listOf(LocalTime.of(23
  * @see [localTime]
  */
 fun Arb.Companion.localDateTime(
-   minYear: Int = 1970,
-   maxYear: Int = 2030
+   minYear: Int,
+   maxYear: Int
+): Arb<LocalDateTime> {
```
nomisRev, line 123: @sksamuel this is an API breaking change, right?. What should be done here?

Any usage of:
```kotlin
Arb.localDateTime(minYear = 0, maxYear = 3000)
```

sksamuel:
We shouldn't break the existing API at all.
Not looked too closely at this pr yet but anything we do should be just
adding new variants.

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
pientaa: Okay, I see another way of fixing the overload resolution ambiguity I came across earlier. I'll make some changes and request your code review again. Thanks!

sksamuel: The code is great, we just want to avoid breaking changes.
Anyone doing `Arb.localDateTime(minYear = 0)` will now have a compile error, same as `Arb.localDateTime(maxYear = 3000)`

I think simple way around that is two more overloads, each just expecting a single argument.

sksamuel (approved):

PR #2731, opened 2021-12-10
#2646 StringShrinkerWithMin generates samples based on source-Arb

### LINKED ISSUE #2646: Arb.string shrinking simplest character is always 'a' regardless of codepoint
The offending code: https://github.com/kotest/kotest/blob/master/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/strings.kt#L35

but I can't really think of any deterministic way to fix it, because the lowest codepoint isn't known upfront :(

we can collect the sampled codepoints in a mutable set after the fact but it will incur a pretty noticeable runtime cost.

test to replicate this:
```kotlin
test("string shrinking should honor codepoints") {
  val arbNumeric = Arb.of(('0'..'9').map { Codepoint(it.code) })
  val samples = Arb.string(1..5, arbNumeric).generate(RandomSource.seeded(1234L))
  samples.take(1).first().shrinks.children.value.map { it.value() }.forAll { value ->
    value.shouldContainOnlyDigits()
  }
}
```

fails with:
```
The following elements passed:
519
83
1983
5198

The following elements failed:
"a1983" => "a1983" should contain only digits
"5198a" => "5198a" should contain only digits
```

aSemy:
fixes #2646

- `StringShrinkerWithMin` will select a random 'simplest' char, instead of always using `a`, based on the provided codepoints.
- Added KDoc to highlight `StringShrinkerWithMin.simplestChar` 
- `respect min value` test needed updating
- Added a couple of tests to check that samples only contain codepoints of the source Arb


aSemy:
Some cases in `ShouldContainExactlyTest` are failing on my machine. I'm not sure if that's connected to this PR.

https://github.com/kotest/kotest/blob/dae697063533ab01c8c28c1220b045e440b59c43/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/ShouldContainExactlyTest.kt

Update: it looks to be unrelated, I've created #2732

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/strings.kt
```diff
@@ -86,11 +87,30 @@ object StringShrinker : Shrinker<String> {
...
+ * Shrinks a string. Shrunk variants will be shorter (reduced in length, until they are [minLength] in length),
+ * and simplified (characters will be replaced by [simplestChar]).
+ *
+ * Note: by default [simplestChar] is the character "`a`". If shrinking an `Arb<String>` that does not include "`a`",
+ * either provide a suitable alternative, or use the secondary constructor and provide a suitable [Codepoint] [Arb],
+ * and a suitable character will be randomly selected.
+ */
 class StringShrinkerWithMin(
    private val minLength: Int = 0,
    private val simplestChar: Char = 'a',
```
aSemy: I think to be more clear and avoid accidental usage, the primary constructor with the `simplestChar` should be converted to a secondary constructor, and then `@Deprecated`.  The primary, and only, constructor should accept `codepoints: Arb<Codepoint>`, as it is more useful than a static character.

Thoughts?

sksamuel: Do we want shrinks to be random? wouldn't it be better if a shrunk value is always deterministic ?

aSemy: Umm good question. I'm not sure, I hadn't thought about it and I don't have as good overview to answer it to be honest. My instinct would be that it's fine, and allowing for Arbs in shrinkers to make them random is okay. But maybe there's benefit to them being completely deterministic. Or there should be a principle that shrinkers should be deterministic.

If this PR doesn't fit then that's okay :)

sksamuel: I like the idea but I think shrinks should be deterministic.
Like to be, 'a' is the best candidate for a-z shrinking. Perhaps there's a way to always choose the "best" shrink given a list of chars?

aSemy: How about this: The simplest char could be the first character of the pre-shrink String?

```kotlin
   override fun shrink(value: String): List<String> {

      val isShortest = value.length == minLength

      val simplestChar: Char? = value.firstOrNull()
      val isSimplest = simplestChar == null || value.all { it == simplestChar }
      // ...
```

aSemy: I've updated the PR to reflect my suggestion. I hope it's something that works! 

The old behaviour of a static 'simplest char' can still be retained, if necessary. I've written, but commented-out, an example secondary constructor with a `@Deprecated`.

It's not ready for merging as I haven't written tests, just updated the existing ones. In particular there's no testing for when `simplestCharSelector` returns `null`.

sksamuel: Sorry for delay in replying.

I think that the simplest char being the first char in the input string could work.
It would be repeatable for any given input at least.

Another option is to introduce some kind of Codepoints/Alphabet interface which specifies the low point for a given range of chars ?

aSemy: No problem! I hope you had a good time over Christmas and New Years.

I'm not quite sure how an `Alphabet` interface would work, it would help to see an example. 

How should the 'simplest' char be selected? Would users know what this would mean? Would the 'simplest char' be relevant to anything else, or only `StringShrinker`? If 'simplest char' is only relevant to `StringShrinker` then it should only be defined and used there, so the context is clearer.

What happens if an `Alphabet` needs to be filtered or mapped - would the simplest char need to be defined on each operation? 

When defining a 'simplest char', does it have to be contained within the `Alphabet`? What happens if it isn't?

sksamuel:
Yeah OK. I think your idea is the best.

sksamuel:
If we go with your idea, first char, and fix the tests, we can include in 5.1

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/strings.kt
aSemy: @sksamuel  I've added the tests for `simpleCharSelector`. Aside from the deprecation annotations and backwards compatibility, I think it's good to go! Let me know if there's anything more I can do.

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/shrinking/StringShrinkerWithMinTest.kt
```diff
@@ -95,44 +106,184 @@ class StringShrinkerWithMinTest : DescribeSpec({
...
+Attempting to shrink arg "a b c d e f g h i j k"
+Shrink #1: "a b c d e f" fail
+Shrink #2: "a b c ffff" fail
+Shrink #3: "f b c ffff" fail
+Shrink #4: "ffb c ffff" fail
+Shrink #5: "fff c ffff" fail
+Shrink #6: "ffffc ffff" fail
+Shrink #7: "fffff ffff" fail
+Shrink #8: "ffffffffff" fail
+Shrink result (after 8 shrinks) => "ffffffffff"
```
sksamuel, line 286: Why did it end up with f's when a was the first char in the arg ?

sksamuel: never mind I see this test is for a custom char selector.

sksamuel (approved):

sksamuel:
Thanks for all your hard work and patience with the feedback @aSemy.

PR #2734, opened 2021-12-12
Fixes global assert softly causing execution error.

### LINKED ISSUE #2706: Issues when using globalAssertSoftly
5.0.1

I'm seeing multiple issues in Intellij when using the following project config:
```
   override val globalAssertSoftly = true
}
```

Removing the configuration or setting it to `false` causes the functionality below to work.

## Problem 1: Test execution halts on first failure
Example:
```
import io.kotest.property.checkAll
import io.kotest.property.exhaustive.exhaustive

class Kotest : DescribeSpec({
    describe("only one failure is reported") {
        val myList = listOf(1, 2, 3)
        checkAll(myList.exhaustive()) {
            it("Should log for $it") {
                it shouldBe 1
            }
        }
    }
})
```

With `globalAssertSoftly = true`:
image

We see it stops on condition 2 and it never runs condition 3. This is true when run via `gradle test` as well and not just intellij.

With `globalAssertSoftly = false` or unset, we see all test cases are ran and reported.
image

## Problem 2: Test name is suppressed on failure, we only see `executionError` now.

Updating the above test example a bit to have multiple failures, since that's the goal of `globalAssertSoftly` to roll these up per the [documentation](https://kotest.io/docs/framework/project-config.html#global-assert-softly):
```
class Kotest : DescribeSpec({
    describe("only one failure is reported") {
        val myList = listOf(1, 2, 3)
        checkAll(myList.exhaustive()) {
            it("Should log for $it") {
                it shouldBe 2
                it shouldBe "a"
            }
        }
    }
})
```
Now when running, we see the output is no longer the test name, but just `executionError` now.
image

Disabling the condition and we see the output we'd expect again:
image

This problem when combined with the first, makes this combination of features unusable as it's now very difficult to tell what test data input failed. (My contrived example makes it easy, but most tests are not this straight forward.)

Thanks for the help resolving this.

ashishkujoy:
closes: #2706

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/ErrorCollector.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/ErrorCollector.kt
@@ -156,7 +156,9 @@ fun ErrorCollector.collectiveError(): AssertionError? {
    clear()
    return if (failures.isNotEmpty()) {
       if (failures.size == 1) {
-         AssertionError(failures[0])
+         AssertionError(failures[0].message).also {
+            stacktraces.cleanStackTrace(it)
+         }
       } else {
          MultiAssertionError(failures).also {
             stacktraces.cleanStackTrace(it) //cleans the creation of MultiAssertionError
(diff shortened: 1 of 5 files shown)


sksamuel:
Can you check the failing tests.

ashishkujoy:
Done

PR #2736, opened 2021-12-13
Fix issues in `shouldStartWith` and `shouldEndWith`

charleskorn:
This PR addresses two issues:

* If a `shouldEndWith` assertion failed, it would report the first N elements of the list, rather than the last N elements
* If `shouldEndWith` or `shouldStartWith` was used with an actual value that was shorter than the expected slice, an `IndexOutOfBoundsException` would be thrown


sksamuel:
Awesome thanks.

PR #2737, opened 2021-12-13
Fixes data class print recursion #2611

### LINKED ISSUE #2611: shouldBe goes into an infinite loop, when generating diff message for data class with cyclic references
4.6.3

When using shouldBe with a data class that has cyclic references, for example:
```kotlin
data class Bar(val foo: Foo)

data class Foo(val bar: Bar?, val parentFoos: List<Foo>)

val foo1ParentFoos: MutableList<Foo> = mutableListOf()

val foo1 = Foo(null, foo1ParentFoos)
val foo2 = Foo(Bar(foo1), emptyList())

foo1ParentFoos.add(foo2)
```

**Note:** I have already overwritten the equals, hashCode and toString methods from my data classes with custom implementations, to not have a stack overflow error when using them.

The shouldBe method will go into an infinite loop when generating the diff message.
What's weird is that there is no stack overflow.

sksamuel:
Fixes #2611


sksamuel:
@JoonasC

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/BindShrinkTest.kt
```diff
@@ -55,22 +54,7 @@ class BindShrinkTest : StringSpec(
...
-- g: 0
-- h: 0
-- i: 0
-- j: 0
-- k: 0
-- l: 0
-- m: 100
-- n: 0
-)"""
+         stdout shouldContain """MaximumComponents(a=1000, b=1000, c=1000, d=1000, e=1000, f=1000, g=1000, h=1000, i=1000, j=1000, k=1000, l=1000, m=1000, n=1000)"""
```
Kantis: Previously tested that the value was shrunk in all components when using `Arb.bind`. Guessing 1000 for each component would be the initial value, so no longer testing shrinkage?

sksamuel: Good spot thanks, fixed.

Issue #2741, opened 2021-12-19
Not work `js{}` and `js(BOTH){}` - only `js(IR) {}`


BorzdeG:
kotest: 5.0.3
Kotlin: 1.6.10
Gradle: 7.3.2

`Could not resolve io.kotest:kotest-framework-api:5.0.3`

sksamuel:
Yes that's as expected.

From the docs:

![image](https://user-images.githubusercontent.com/743706/146681711-f8ad7cfd-47dc-4b7a-a7d8-a0f0fbe95d0b.png)

BorzdeG:
Why? What is the reason for blocking Legacy in tests? it's just a compilation: https://kotlinlang.org/docs/js-ir-compiler.html#authoring-libraries-for-the-ir-compiler-with-backwards-compatibility

sksamuel:
The compiler plugin uses IR specific generation:
https://github.com/kotest/kotest/blob/master/kotest-framework/kotest-framework-multiplatform-plugin-js/src/jvmMain/kotlin/io/kotest/framework/multiplatform/js/SpecIrGenerationExtension.kt

PR #2742, opened 2021-12-19
#2741 Not work `js{}` and `js(BOTH){}` - only `js(IR) {}`

BorzdeG:
#2741 

* Updated Gradle version
* Updated Kotlin version
* Added support Legacy JS backend

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -35,19 +35,19 @@ jobs:
             run: ./gradlew mingwX64Test --scan
 
          -  name: Run macOS tests
-            if: matrix.os == 'macOS-latest'
+            if: matrix.os == 'macOS-11'
             run: ./gradlew macosX64Test macosArm64TestKlibrary --scan
 
          -  name: Run watchOS tests
-            if: matrix.os == 'macOS-latest'
+            if: matrix.os == 'macOS-11'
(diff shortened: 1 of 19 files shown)


sksamuel:
The framework modules uses a compiler plugin for JS - have you tested that it works for legacy?

BorzdeG:
yes, I local run `./gradlew ﻿﻿jsIrTest jsLegacyTest` in MacOS

GitHub Actions (examples in Ubuntu): [jsIrTest](https://github.com/kotest/kotest/runs/4575649207?check_suite_focus=true#step:4:442) , [jsLegacyTest](https://github.com/kotest/kotest/runs/4575649207?check_suite_focus=true#step:4:712)

I only crash ConcurrentTestsSingleInstanceTest - it takes longer than expected: "2717 should be <1499"
But I don't think it is related to IR - the test is for jvm, not js

sksamuel:
We need to test this against the `kotest-tests-js` module which actually executes javascript tests using the compiler plugin. I'm not sure how this can work given the compiler plugin uses IR classes.

BorzdeG:
Did I understand correctly - now there are no tests for him and it is necessary to supplement the PR with these tests?

sksamuel:
No, we just have to make sure that the tests in `kotest-tests/kotest-tests-js` work and are being detected.
We should build locally using these changes and use that version of the compiler plugin in that project to test.
Like I mentioned in the other issue, I don't see how this can work given the compiler plugin uses IR classes.

PR #2749, opened 2021-12-22
address issue 2746

inquiribus:
details in associated issue

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
@@ -38,6 +38,9 @@ fun <T : Any?> eq(actual: T, expected: T, strictNumberEq: Boolean): Throwable? {
          IterableEq.isValidIterable(actual) && IterableEq.isValidIterable(expected) -> {
             IterableEq.equals(IterableEq.asIterable(actual), IterableEq.asIterable(expected), strictNumberEq)
          }
+         SequenceEq.isValidSequence(actual) && SequenceEq.isValidSequence(expected) -> {
+            SequenceEq.equals(SequenceEq.asSequence(actual), SequenceEq.asSequence(expected), strictNumberEq)
+         }
          shouldShowDataClassDiff(actual, expected) -> DataClassEq.equals(actual as Any, expected as Any, strictNumberEq)
          actual is Throwable && expected is Throwable -> ThrowableEq.equals(actual, expected)
          else -> DefaultEq.equals(actual as Any, expected as Any,strictNumberEq)
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
@@ -38,9 +38,7 @@ fun <T : Any?> eq(actual: T, expected: T, strictNumberEq: Boolean): Throwable? {
          IterableEq.isValidIterable(actual) && IterableEq.isValidIterable(expected) -> {
             IterableEq.equals(IterableEq.asIterable(actual), IterableEq.asIterable(expected), strictNumberEq)
          }
-         SequenceEq.isValidSequence(actual) && SequenceEq.isValidSequence(expected) -> {
-            SequenceEq.equals(SequenceEq.asSequence(actual), SequenceEq.asSequence(expected), strictNumberEq)
-         }
+         actual is Sequence<*> && expected is Sequence<*> -> SequenceEq.equals(actual, expected, strictNumberEq)
          shouldShowDataClassDiff(actual, expected) -> DataClassEq.equals(actual as Any, expected as Any, strictNumberEq)
          actual is Throwable && expected is Throwable -> ThrowableEq.equals(actual, expected)
          else -> DefaultEq.equals(actual as Any, expected as Any,strictNumberEq)
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
@@ -119,9 +119,9 @@ object IterableEq : Eq<Iterable<*>> {
       val iter2 = expected.iterator()
       val elementDifferAtIndex = mutableListOf<Int>()
 
-      fun <T> nestedIterator(item: T): String? = item?.let {
-         if ((it is Iterable<*>) && (it !is Collection<*>) && (it::class.isInstance(actual) || actual::class.isInstance(it))) {
-         """$disallowed $it (${it::class.simpleName ?: "anonymous" }) within $iter1 (${it::class.simpleName ?: "anonymous" }); (use custom test code instead)"""
+      fun <T> nestedIterator(item: T, oracle: Iterable<*>): String? = item?.let {
+         if ((it is Iterable<*>) && (it !is Collection<*>) && (it::class.isInstance(oracle) || oracle::class.isInstance(it))) {
+         """$disallowed $it (${it::class.simpleName ?: "anonymous" }) within $oracle (${oracle::class.simpleName ?: "anonymous" }); (use custom test code instead)"""
       } else null }
 
       var nestedIteratorError: String? = null
@@ -150,8 +150,8 @@ object IterableEq : Eq<Iterable<*>> {
             val b = iter2.next()
             val t: Throwable? = when {
                a?.equals(b) == true -> null
-               nestedIterator(a)?.let { setDisallowedState(it) } == true -> failure(nestedIteratorError!!)
-               nestedIterator(b)?.let { setDisallowedState(it) } == true -> failure(nestedIteratorError!!)
+               nestedIterator(a, actual)?.let { setDisallowedState(it) } == true -> failure(nestedIteratorError!!)
+               nestedIterator(b, expected)?.let { setDisallowedState(it) } == true -> failure(nestedIteratorError!!)
                else -> equalXorDisallowed(eq(a, b, strictNumberEq))
             }
             if (!accrueDetails) break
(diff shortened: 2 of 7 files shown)


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
```diff
@@ -62,45 +65,86 @@ object IterableEq : Eq<Iterable<*>> {
...
+         } else if (accrueDetails) unexpectedElementAtIndex = index else break
          index++
       }
       if (iter2.hasNext()) {
          missingElementAt = index
       }
       val detailErrorMessage = StringBuilder().apply {
          if (elementDifferAtIndex.isNotEmpty()) {
-            append("Element differ at index: ${elementDifferAtIndex.print().value}\n")
+            nestedIteratorError?.let {
```
ashishkujoy: This seems redundant, if there is `nestedIteratorError` then at `line 143` then your returning error of `nestedIteratorError` this `detailErrorMessage` is getting thrown only when `nestedIteratorError` is not present.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/SequenceEq.kt
```diff
@@ -0,0 +1,32 @@
...
+   }
+
+   fun asSequence(it: Any): Sequence<*> {
+      return when (it) {
+         is Sequence<*> -> it
+         else -> error("Cannot convert $it to Sequence<*>")
+      }
+   }
+
+   override fun equals(actual: Sequence<*>, expected: Sequence<*>, strictNumberEq: Boolean): Throwable? =
```
ashishkujoy, line 15: Since anyway we are failing here, not sure then we even need this is eq.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
```diff
@@ -62,45 +65,86 @@ object IterableEq : Eq<Iterable<*>> {
...
+      var accrueDetails = true
       while (iter1.hasNext()) {
          val a = iter1.next()
          if (iter2.hasNext()) {
             val b = iter2.next()
-            val t = eq(a, b, strictNumberEq)
-            if (t != null) {
+            val t: Throwable? = when {
+               a?.equals(b) == true -> null
+               (a is Iterable<*>) && (a !is Collection<*>) && (a::class.isInstance(actual) || actual::class.isInstance(a)) -> {
```
ashishkujoy: Expression seems hard to read may can be converted to function call.

```kotlin
isNestedIterator(b, expected) -> {
      nestedIteratorError = "$unim $b within $iter2 (must use custom code)"
      accrueDetails = false
      failure(nestedIteratorError)
}
```

```kotlin
private fun isNestedIterator(a: Any?, actual: Iterable<*>) =
      (a is Iterable<*>) && (a !is Collection<*>) && (a::class.isInstance(actual) || actual::class.isInstance(a))
```

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
```diff
@@ -38,6 +38,9 @@ fun <T : Any?> eq(actual: T, expected: T, strictNumberEq: Boolean): Throwable? {
          IterableEq.isValidIterable(actual) && IterableEq.isValidIterable(expected) -> {
             IterableEq.equals(IterableEq.asIterable(actual), IterableEq.asIterable(expected), strictNumberEq)
          }
+         SequenceEq.isValidSequence(actual) && SequenceEq.isValidSequence(expected) -> {
```
ashishkujoy: Can be simplified as.
```kotlin
actual is Sequence<*> && expected is Sequence<*> -> SequenceEq.equals(actual, expected, strictNumberEq)
```

ashishkujoy (changes requested):

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/Eq.kt
inquiribus: simplified as recommended

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/SequenceEq.kt
inquiribus: Included for symmetry, and to push an error message that is explicit about the reason for failing. No strong feelings :)  Unconditional failure also is easy to revise if anything changes, but this latter is a weaker justification.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
inquiribus: Yup!  Thanks, removed.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
inquiribus: Converted to function call

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
```diff
@@ -3,86 +3,163 @@ package io.kotest.assertions.eq
...
          is Array<*> -> it.asList()
          is List<*> -> it
          is Set<*> -> it
          is Collection<*> -> it
+         is Iterable<*> -> it
          else -> error("Cannot convert $it to Iterable<*>")
       }
    }
 
+   fun isOrderedSet(item: Iterable<*>): Boolean = item is LinkedHashSet || (item is Set<*> && item.size == 1)
```
ashishkujoy, line 31: I did not understand the reason of checking length is 1.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
```diff
@@ -3,86 +3,163 @@ package io.kotest.assertions.eq
...
+   private const val disallowed = "$trigger nesting iterator"
+
    private fun checkEquality(actual: Iterable<*>, expected: Iterable<*>, strictNumberEq: Boolean): Throwable? {
-      var index = 0
+
       val iter1 = actual.iterator()
       val iter2 = expected.iterator()
       val elementDifferAtIndex = mutableListOf<Int>()
+
+      fun <T> nestedIterator(item: T): String? = item?.let {
```
ashishkujoy: `(it::class.isInstance(actual) || actual::class.isInstance(it)))` part seems wrong, as this method is getting used to check both actual and expected. But in method body instance checking getting perform only with `actual`
This `nestedIterator(a)` need to change to  `nestedIterator(a, actual)`
 `nestedIterator(b)` need to change to  `nestedIterator(a, expected)`

kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/com/sksamuel/kotest/eq/IterableEqTest.kt
```diff
@@ -42,12 +63,126 @@ class IterableEqTest : FunSpec({
       }
    }
 
+   test("should not give error for kotlin ordered set comparison with list") {
+      val error = IterableEq.equals(setOf(1, 2, 3), listOf(1, 2, 3)).shouldBeNull()
```
ashishkujoy, line 67: set and list containing equal elements should not be equal.

inquiribus: I agree, yes.  Full agreement.  At stake here is backward compatibility.  `setOf(something)` returns a `LinkedHashSet` which is an ordered set.  What is wrong and illegal to compare is, down to the metal, an ordered collection vs. a collection that has no ordering guarantee.  An ordered set like `LinkedHashSet` is a creature of convenience (sigh) that is ordered more often than not, and also a set (no duplicates).  The existing code does unordered comparison between `Set and Set`, and ordered comparison between `Set xor Set` so I punted.  Should we break backward compatibility?  That is going to create a very angry nest of wasps among practically minded users.  We can do either, there are good arguments for both.  The issue itself supports backward compatibility ...

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
inquiribus: A Set of one by default is of a unique type different than `LinkedHashSet`; discussion below addresses a larger conundrum in regard to this subject.  One can also argue that a Set of 1 is ordered, but I'm picking nits here.  I'm open to any sensible disposition.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt
inquiribus: modified as requested

kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/com/sksamuel/kotest/eq/IterableEqTest.kt
sksamuel: Backwards compatibility all the way.

sksamuel (approved):

sksamuel:
@inquiribus As long as we're keeping backwards compatibility with all existing use cases, other than adding special support for custom iterables, then I'm happy to merge this as part of 5.1

sksamuel:
I'll merge this in. It would be good to update the docs on what special cases shouldBe handles.

inquiribus:
> I'll merge this in. It would be good to update the docs on what special cases shouldBe handles.

I apologize for tardiness.  Should I write/add a 
```
/**  */
```
header to IterableEq.kt or proceed in some other manner?  If so, would it be OK to submit it as a pull request?

sksamuel:
Pull requests always welcome :)

Issue #2655, opened 2021-11-18
Add mutable test clock


sksamuel:
Would be nice if we had a fixed clock but that could be advanced manually.

pientaa:
I can take that issue, too

sksamuel:
Great!

sksamuel:
its the java.time.Clock that we want to implement.

sksamuel:
@pientaa did you still intend to do this ?

pientaa:
@sksamuel yes, I'm sorry, I was busy recently

sksamuel:
No problem, just wondering :)

pientaa:
@sksamuel after looking at `java.time.Clock` I noticed that there are already implemented such classes: `OffsetClock` and `TickClock`. And they seem to be exactly what you asked for. Or maybe there is something that I missed or misunderstood?

sksamuel:
You can't set an arbitrary value on them after creation. Unless i missed it.

PR #2777, opened 2022-01-13
Feature/2655 add mutable test clock

pientaa:
Implements #2655


kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/time/MutableClock.kt
```diff
@@ -0,0 +1,30 @@
...
+import java.io.Serializable
+import java.time.Clock
+import java.time.Instant
+import java.time.ZoneId
+
+class MutableClock(
+   private var instant: Instant,
+   private var zone: ZoneId,
+) : Clock(), Serializable {
+   fun withInstant(instant: Instant): Clock = MutableClock(instant, zone)
```
sksamuel: This will return a new clock, rather than mutating this one.

pientaa: You're right, gonna change it

PR #2791, opened 2022-01-21
Fix withClue and assertSoftly interference with concurrent coroutines

### LINKED ISSUE #2790: withClue and assertSoftly interfere when concurrent coroutines start on the same thread
Kotest 5.1.0/JVM

When multiple concurrent coroutines start on the same thread, they share an `ErrorCollector`. Whenever they happen to resume on different threads sometime later, they mess up each other's clue stack and/or failure list. This is true regardless of whether the error collector's methods are thread-safe or not.

What is really required is a coroutine-local error collector.

PR upcoming.

OliverO2:
Fixes #2790 by providing a coroutine-local error collector. Test case included.

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
@@ -2,24 +2,59 @@
 
 package io.kotest.assertions
 
-import kotlinx.coroutines.asContextElement
+import kotlinx.coroutines.CopyableThreadContextElement
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlin.coroutines.CoroutineContext
 
-actual val errorCollector: ErrorCollector get() = ThreadLocalErrorCollector.instance.get()
+actual val errorCollector: ErrorCollector get() = threadLocalErrorCollector.get()
 
 /**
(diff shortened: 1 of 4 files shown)


kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
```diff
@@ -2,24 +2,59 @@
...
-      val instance = object : ThreadLocal<ErrorCollector>() {
-         override fun initialValue() = ThreadLocalErrorCollector()
-      }
+
+private val threadLocalErrorCollector = object : ThreadLocal<CoroutineLocalErrorCollector>() {
+   override fun initialValue() = CoroutineLocalErrorCollector()
+}
+
+
+private class CoroutineLocalErrorCollector : BasicErrorCollector() {
```
sksamuel, line 28: Do you think we should copy the parent, or just start a new one ?

OliverO2: Copying the parent collector seems ideal. That way, we keep the parent clues and settings as context, like we do in sequential code, as shown in the test case:
```kotlin
clueContextAsString() shouldBe "$parentClue\n$childClue\n"
```

Seems like good companionship with structured concurrency.

sksamuel: Great

PR #2794, opened 2022-01-22
Fix #2793 by handling the edge cases of lenient json number comparison

### LINKED ISSUE #2793: Incorrect number comparison in lenient json comparisons
master

This is best explained with a failing test so here goes:

```kt

            shouldFail {
               "1.0E-3" shouldEqualJson "0.0001"
            }
            "1.0E-4" shouldEqualJson "0.0001"

```

both of these assertions fail - `1.0E-3` is considered equal to `0.0001` (off by one zero) and `1.0E-4` really is `0.0001` but the assertion doesn't think so.

Digging in to this there are two separate issues:
1. the lenient option is using a regex to trim irrelevant trailing zeroes after the decimal point, but doesn't use a `$` (end of input) sigil so all the intervening (significant) zeroes are stripped
2. the comparison of scientific notation numbers does not deal with the scenario where one of expected or actual is denoted with exponential numbers and the other is not, but parsing as Double does not turn it back to regular notation.

I have a fix, so will raise that as a PR shortly, but raising the issue in case my PR is not accepted and someone else wants to take a crack.

zakhenry:


kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -404,3 +396,17 @@ internal fun compareNull(path: List<String>, b: JsonNode): JsonError? {
       else -> JsonError.ExpectedNull(path, b)
    }
 }
+
+
+fun main() {
```
zakhenry: oops remove this

zakhenry: done

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -377,23 +377,15 @@ private fun compareNumbers(
    }
 }
 
-private val fractionalZeroesRegex =
-   """(\.\d*)0+""".toRegex()
```
zakhenry, line 381: this was moved, but note critically the `$` was added to the end to avoid `0.01` being stripped to `0.1` for example

sksamuel:
awesome thanks

PR #2807, opened 2022-02-01
Add blockingTest to the config options on FreeSpec #2805

### LINKED ISSUE #2805: Add blockingTest to the config options on FreeSpec
It seems at the moment that the only way to have a blocking test is by using `FunSpec`.  Most of our tests are using `FreeSpec`, but we noticed that the `blockingTest` parameter is missing from the config function on `FreeSpec`, so for this particular test class I've had to move them over to `FunSpec`.

Is it possible to make the `blockingTest` parameter available in the other styles of test?

sksamuel:
Fixes #2805


sksamuel:
@kaylanx

kaylanx:
Awesome, thanks for addressing this so quickly :D

sksamuel:
Will be in 5.1.1, expect that to drop by tomorrow.

PR #2812, opened 2022-02-05
Add workaround for issue where no tests are run when using new Kotlin/Native memory model.

charleskorn:
See https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1636921822112900.

In the old memory model, global properties are eagerly initialised, whereas in the new memory model, they are lazily initialised ([source](https://github.com/JetBrains/kotlin/blob/master/kotlin-native/NEW_MM.md)). The impact of this is that a project using the new memory model will not run any tests.

Annotating properties with `@EagerInitialization` changes the behaviour to match the old memory model. 

Caveats:

- I couldn't find an easy way to test this within the existing test setup. If you have any suggestions, happy to add a test for this.

- This is only a workaround: [the docs for `@EagerInitialization`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.native/-eager-initialization/) emphasise that we should not rely on this annotation in the long term.


charleskorn:
I should also say that I have very limited experience with compiler plugins - this works, but there may well be better ways to do this.

sksamuel:
Seems reasonable to me. Assuming it still works on old memory model (I assume it ignores the annotation).

charleskorn:
> Assuming it still works on old memory model (I assume it ignores the annotation).

Yep, that's exactly what it does.

PR #2814, opened 2022-02-05
Fix issue where macOS tests don't run for PRs.

charleskorn:
The `macos-11` runner is used for PR builds on macOS, not `macos-latest`, so the tests were being skipped.


sksamuel:
👍🏼

PR #2823, opened 2022-02-13
Docs: corrected typos in which Spec to extend in test style docs

BjornvdLaan:
The testing style base class used in certain examples did not match the testing style that the docs are talking about. This can be confusing to readers.


sksamuel:
Thank you

PR #2827, opened 2022-02-15
Include the whole list when a `shouldContain` assertion fails.

charleskorn:
Currently, `shouldContain` only prints the first five elements in the list, making understanding why the assertion failed difficult. 

This change instead includes the entire list in the error message when the assertion fails.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/contain.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/contain.kt
@@ -37,7 +37,7 @@ fun <T, C : Collection<T>> contain(t: T, verifier: Equality<T> = Equality.defaul
       value.any { verifier.verify(it, t).areEqual() },
       {
          "Collection should contain element ${t.print().value} based on ${verifier.name()}; " +
-            "listing some elements ${value.take(5)}"
+            "but the collection is $value"
       },
       { "Collection should not contain element ${t.print().value} based on ${verifier.name()}" }
    )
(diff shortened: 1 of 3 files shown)


sksamuel:
This is deliberate. If you have a list of 1000 elements, and they're data classes, you're going to output 10,000 log lines.

charleskorn:
> This is deliberate. If you have a list of 1000 elements, and they're data classes, you're going to output 10,000 log lines.

Makes sense. 

Would you be open to making this configurable? Perhaps the default can be the existing behaviour, and the failure message can include a note to let the reader know that it's possible to change the number of items displayed.

sksamuel:
Yep, configured by a system property which can be included in the output message ?

charleskorn:
> Yep, configured by a system property which can be included in the output message ?

Makes sense for the JVM, what would you suggest for environments like JS and Native? An environment variable?

sksamuel:
There's the assertions object for holding state

charleskorn:
I can see there's [`maxCollectionEnumerateSize`](https://github.com/kotest/kotest/blob/d444794027942d8172d88aad630e284db6470ea3/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/AssertionsConfig.kt#L19) there already - would it make sense to use that, or create a separate configuration option?

I can also see that `ListPrint` already has logic to limit the size of a displayed list, so I'm thinking of modifying that to make respect the configured value, and then modifying `shouldContain` to use `printed()` to print the actual value of the list - does that make sense?

charleskorn:
I've created #2836 as a possible implementation, open to feedback.

Issue #2816, opened 2022-02-06
json.shouldBeObject()


sksamuel:
Asserts that the json element or string is a json object.

Issue #2815, opened 2022-02-06
json.shouldBeArray()


sksamuel:
Asserts that the json element or string is a json array.

PR #2829, opened 2022-02-15
draft for #2815, #2816

Kantis:
@sksamuel is this something in line with what you were thinking? Lots of work remain, mainly curious about what you think of the general idea.

I yoinked the code I was playing around with for making JSON extractions, think it could be useful here..

### DIFF
--- a/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/array.kt
+++ b/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/array.kt
@@ -1,0 +1,95 @@
+import io.kotest.assertions.json.pretty
+import io.kotest.assertions.json.toJsonNode
+import io.kotest.matchers.Matcher
+import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.should
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.decodeFromJsonElement
+
+fun String.shouldBeJsonArray(path: String = "$") = pretty.parseToJsonElement(this) should beJsonArray(path)
+fun JsonElement.shouldBeJsonArray(path: String = "$") = this should beJsonArray(path)
+
+fun String.shouldBeJsonObject(path: String = "$") = pretty.parseToJsonElement(this) should beJsonObject(path)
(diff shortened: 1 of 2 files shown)


sksamuel:
@Kantis this ready for review ?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/array.kt
```diff
@@ -0,0 +1,95 @@
+import io.kotest.assertions.json.pretty
```
LeoColman, line 1: I think this is missing a package statement, isn't it?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/array.kt
```diff
@@ -0,0 +1,95 @@
...
+   while (i < path.length - 1) {
+      when (path[i]) {
+         ']' -> i++
+         '.' -> jsonExtractions.add(ByProperty(path.substring(i + 1, nextToken())))
+         '[' -> jsonExtractions.add(ByIndexing(path.substring(i + 1, nextToken()).toInt()))
+         else -> error("Invalid input $path. (i = $i, actions = $jsonExtractions)")
+      }
+   }
+
+   println("Returning actions: $jsonExtractions")
```
LeoColman, line 92: This seems like a hanging debug println.
Is it supposed to be here?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/array.kt
Kantis: Yup, you're right. Will fix :)

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/array.kt
Kantis: Indeed, will remove.

sksamuel:
@Kantis do you think this is similar to https://github.com/kotest/kotest/pull/2861

Kantis:
> @Kantis do you think this is similar to #2861

Yeah, very. 😁 I could add the path-matching capabilities on top of that PR's work instead, if we're at all interested in it. 

Either way, I will close this.

PR #2831, opened 2022-02-16
Remove inconsistent exceptionClass default values

Pitel:
Why some `eventually` methods has it set to `Throwable` and some doesn't? I suggest to remove it. Or if we want to keep it, put it as defaualt value to `EventuallyConfig`'sa constructor instead of `null`.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/timing/EventuallyTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/timing/EventuallyTest.kt
@@ -46,14 +46,14 @@ class EventuallyTest : WordSpec() {
          }
          "pass tests that completed within the time allowed" {
             val end = System.currentTimeMillis() + 150
-            eventually(1.seconds) {
+            eventually(1.seconds, RuntimeException::class) {
                if (System.currentTimeMillis() < end)
                   throw RuntimeException("foo")
             }
          }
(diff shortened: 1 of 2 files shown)


sksamuel:
Looks like a build failure.

Pitel:
@sksamuel Tests are now fixed. 💪🏻

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/timing/EventuallyTest.kt
```diff
@@ -46,14 +46,14 @@ class EventuallyTest : WordSpec() {
          }
          "pass tests that completed within the time allowed" {
             val end = System.currentTimeMillis() + 150
-            eventually(1.seconds) {
+            eventually(1.seconds, RuntimeException::class) {
```
sksamuel: Would this test fail now if we didn't add the exception ?
Will it break existing users code in other words ?

Pitel: Yes, it will. Other proposed solution might be to add `Throwable` to configuration's default constructor. Then it shouldn't break existing code, so changes to tests might be reverted. Should I do it?

sksamuel:
Yes, we can't break anything.

Pitel:
@sksamuel Ok, now it should be as you want. 👍🏻

Issue #2834, opened 2022-02-17
BigDecimal arb can return edgecases outside min max limits


ciffer:
**Which version of Kotest are you using**
5.0.3

`Arb.bigDecimal(min = .., max = ..)` returns an Arbitrary that has edgecases outside of range min..max.

**Reproduction**:
```kotlin
val edgecases = Arb.bigDecimal(min = BigDecimal.ONE, max = BigDecimal.TEN)
			.edgecases()
		
println(edgecases)
```

**Current output**:
```
[1, 1E-300, -1E-300, -1, 0, 0.00]
```

**Expected output**:
```
[1]
```

I can provide a pull request later today if the report is accepted.

Cheers and keep up the great work!

sksamuel:
Pr always most welcome!

ciffer:
Created PR #2835

ciffer:
I wasn't sure what branch should I target as it isn't a fix for any specific version. Let me know if I need to change it to `release/5.0.x`

myuwono:
This was fixed in #2835

PR #2835, opened 2022-02-18
BigDecimal arb can return edgecases outside min max limits #2834

ciffer:
Edgecases for Arb.bigDecimal(min, max) are now always contrained to be in range [min, max)
  Min is now added as edgecase for Arb.bigDecimal(min, max)

### DIFF
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/bigdecimal.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/bigdecimal.kt
@@ -4,7 +4,7 @@ import io.kotest.property.Arb
 import java.math.BigDecimal
 import java.math.RoundingMode
 
-internal val bigDecimalEdgecases = listOf(
+internal val bigDecimalDefaultEdgecases = listOf(
    BigDecimal(0.0),
    // BigDecimal compareTo and equals are not consistent
    BigDecimal("0.00"),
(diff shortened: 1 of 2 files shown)


kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/bigdecimal.kt
```diff
@@ -28,8 +28,11 @@ fun Arb.Companion.bigDecimal(scale: Int, roundingMode: RoundingMode) =
    bigDecimal().map { it.setScale(scale, roundingMode) }
 
 fun Arb.Companion.bigDecimal(min: BigDecimal, max: BigDecimal): Arb<BigDecimal> {
-   return arbitrary(bigDecimalEdgecases) {
+   val boundedEdgecases = bigDecimalDefaultEdgecases
+      .filter { min <= it && it < max }
+      .plus(min)
+
+   return arbitrary(boundedEdgecases) {
       min.add(BigDecimal(Math.random()).multiply(max.subtract(min)))
```
myuwono: @ciffer thanks heaps! do you mind also fixing this bug? It seems like currently this arb ignores the provided random source
```suggestion
      min.add(BigDecimal(it.random.nextDouble()).multiply(max.subtract(min)))
```

That is technically a different issue but would be great if you can also address that in this PR. Otherwise i'll raise a follow up PR for that. Thanks again!

myuwono (approved):

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/bigdecimal.kt
ciffer: I'll do that later today :+1:

myuwono:
@ciffer you have a compilation failure

```
e: /home/runner/work/kotest/kotest/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/BigDecimalTest.kt: (55, 23): Unresolved reference: shouldNotContain
> Task :kotest-property:compileTestKotlinJvm FAILED
```

sksamuel:
Great thanks!

PR #2836, opened 2022-02-19
Make length of collection snippet included in assertion errors configurable

charleskorn:
Replacement for #2827.

Summary of changes:
* All collection matchers now use `print()` to format collections, which ends up using `ListPrint`
* Added a new `AssertionsConfig.maxCollectionPrintSize` property, which takes its value from the `kotest.assertions.collection.print.size` system property on the JVM, and the `kotest_assertions_collection_print_size` environment variable when running on Kotlin/Native.
* `ListPrint` now respects the value of `AssertionsConfig.maxCollectionPrintSize` when printing lists, and includes a hint on how to configure the limit

~~Still to be done:~~
* ~~Add implementation of `env()` for Kotlin/JS: I have never used Kotlin/JS before and my Googling is not finding an answer for this. Is there an easy way to get an environment variable in Kotlin/JS? `process.env` doesn't seem to be accessible.~~

Future improvements:
* I haven't added any tests for Kotlin/Native as this doesn't seem to be set up and seemed like quite a large task outside the scope of this PR.
* I haven't touched the other values in `AssertionsConfig`, but these could potentially use `ConfigValue` as well.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/contain.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/contain.kt
@@ -37,7 +37,7 @@ fun <T, C : Collection<T>> contain(t: T, verifier: Equality<T> = Equality.defaul
       value.any { verifier.verify(it, t).areEqual() },
       {
          "Collection should contain element ${t.print().value} based on ${verifier.name()}; " +
-            "listing some elements ${value.take(5)}"
+            "but the collection is ${value.print().value}"
       },
       { "Collection should not contain element ${t.print().value} based on ${verifier.name()}" }
    )
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containAll.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containAll.kt
@@ -1,5 +1,6 @@
 package io.kotest.matchers.collections
 
+import io.kotest.assertions.print.print
 import io.kotest.matchers.Matcher
 import io.kotest.matchers.MatcherResult
 import io.kotest.matchers.should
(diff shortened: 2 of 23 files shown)


sksamuel:
Awesome. I'm away atm, but will review on my return.

sksamuel:
There is no env on js yet.

The idea on js and native is you would just use before project to set this on Assertions. I guess that's harder now you have wrapped the values in the Configurable thing. Can you give me a tl;dr on that? Maybe we hide that from the user ?

charleskorn:
> The idea on js and native is you would just use before project to set this on Assertions. 

I think I'm missing something here - is this how it's done for the other options in `AssertionsConfig`? If so, how does that work? All the values on `AssertionsConfig` are immutable.

Or were you thinking we'd break with the existing pattern for this new option?

> Can you give me a tl;dr on that?

The idea was to encapsulate the value of a configuration value and its source, to make it easier to deal with different platforms. For example, Configurable exposes a human-readable description of where the value came from (environment variable or system property), which allows this information to be included in assertion error messages.

charleskorn:
I know you're away at the moment, ignore this until you're back @sksamuel...

Not quite sure what's happened, but it looks like the CI build is failing for reasons unrelated to these changes. It looks like a flaky test - could you try re-running it? (I don't have access to do this myself unfortunately.)

sksamuel:
Brilliant thank you @charleskorn

jmfayard:
Hello @charleskorn @sksamuel , 

After getting the message `Configure the system property kotest.assertions.collection.print.size` I tried to do so,
but it tourns out it's not trivial to do in Gradle, see below

Maybe it should be documented?

```diff
# gradle.properties
+systemProp.kotest.assertions.collection.print.size=200

# build.gradle.kts
tasks.withType<Test> {
    useJUnitPlatform() // Use Junit 5 for tests
+    setSystemProperties(System.getProperties() as Map<String, Any>)
}
```

charleskorn:
> Maybe it should be documented?

Seems sensible to me. I'd suggest just passing through only the Kotest-related properties though, not all system properties. 

It might also be a good idea for the Kotest Gradle plugin to automatically set this up for all test tasks. What do you think @sksamuel?

jmfayard:
So actually what I proposed above made the CI crash https://scans.gradle.com/s/7b4dezdum77ue/console-log?anchor=69

The correct way to do it:

```diff
// build.gradle.kts
tasks.withType<Test> {
    useJUnitPlatform() // Use Junit 5 for tests
+    systemProperty("kotest.assertions.collection.print.size", "200")
}
```

PR #2837, opened 2022-02-20
Use configuration avoidance in quick start snippet

LouisCAD:


sksamuel:
Hi Louis. What does the change do ?

LouisCAD:
Hello, it avoids configuring the test tasks when not needed (e.g. when not running tests), which makes each build and Gradle invocation slightly faster.

sksamuel:
Ahh cool.

PR #2850, opened 2022-02-27
Fix compilation error in example

kdnakt:
https://kotest.io/docs/framework/extensions/extension-examples.html

The code listed above didn't compile with `io.kotest:kotest-framework-api-jvm:5.1.0`.
So I fixed the documents.

I think this pull request introduced the relevant change.
https://github.com/kotest/kotest/pull/2509/files#diff-22ba5bc340d94549bb752f5c91e426910e50d63f4f816490635c98937a0af482


sksamuel:
Good spot thanks.

PR #2851, opened 2022-02-27
Basic documentation for shrinking property-based tests

### LINKED ISSUE #2822: Additional documentation for shrinking.
Hello! First off, thank you for this wonderful library. I just switched over from JUnit-Quickcheck for one of my projects, and it's been working great for me so far.

However, one thing I found that could be improved is that there is seemingly little documentation about shrinking on the kotest website.   The documentation [does mention](https://kotest.io/docs/proptest/property-test-global-config.html#printing-shrink-steps) shrinking in the configuration section -- but this was all I was able to find. 

Some questions I as a new user have about shrinking:

  1. What is it? (I don't have this question, but someone else might)
  2. How does it work in kotest?
  3. Can I define a custom shrinking function for my custom generators? If not, is shrinking still somehow able to be applied to them, or does it only work with built-in types?
  
I think it would be very helpful if a section was added to the docs introducing shrinking and answering these questions.

BjornvdLaan:
Fixes #2822 

Shrinking is a very interesting feature of property-based testing frameworks. Kotest has this functionality, but it is not mentioned in the documentation. Here is a page that explains the basics of shrinking and how it works for built-in and custom generators.


sksamuel:
This is awesome thank you.

PR #2853, opened 2022-02-28
smart cast shouldBeSuccess and shouldBeFailure

### LINKED ISSUE #2852: shouldBeSuccess should return value for chaining
Result<Foo>(foo).shouldBeSuccess().fooname

i-walker:
resolves #2852


sksamuel:
Tests failing, can you check ?

PR #2862, opened 2022-03-11
ignore uniqueid selectors for better interop with other engines

### LINKED ISSUE #2806: Kotest should respect engine name in junit platform launcher
_I have a test suite that uses multiple junit engines, and when i run tests by uniqueId kotest also runs tests even when the uniqueId clearly states a different engine than kotest. [engine:failgood]/[class:...]/[class:...]
I think kotest should just run no tests when the uniqueid selector is used with a different engine._

https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1643716624167609

christophsturm:
fixes #2806


sksamuel:
Should we only ignore unique id if the engine part of the unique id is != kotest ?

christophsturm:
does kotest support executing tests by uniqueid? or should it just run all tests for a uniqueid that has the kotest engine descriptor?

sksamuel:
I would just throw an error if engine is kotest and this selector is found.

sksamuel (approved):

sksamuel:
Maybe throw an error if a kotest unique ID is detected

Issue #2845, opened 2022-02-23
Matcher compose


sksamuel:
```kotlin
Matcher.compose<Person>(
  nameMatcher, { it.name },
  addressMatcher, { it.address }
)
```

(Or another design if you prefer).

pientaa:
What do you mean by `nameMatcher` and `addressMatcher`? And what return type should we expect from this function? `Matcher`?

PR #2864, opened 2022-03-11
#2845 Matcher compose

pientaa:
Implements #2845 

Before I finish PR, I'd like to make sure if that's what you meant @sksamuel in that issue


sksamuel:
Yes you're on the right lines. I don't think you need the fold though, could just do .all { matcher -> ... }

kierans:
Just wondering if there's any documentation that should be added so that new people can learn of this feature?

sksamuel:
I will rework the custom matchers documentation today since it is lacking in so many places anyway.

sksamuel:
Ok much better custom matcher docs are now up.

PR #2881, opened 2022-03-16
Made isOrderedSet platform-specific, to allow TreeSet eq. Fixes #2879

### LINKED ISSUE #2879: shouldContainExactly should work with TreeSet
If you call `treeSet shouldContainExactly listOf("a", "b")`, that should work, but currently blows up[1] because Kotest doesn't treat TreeSets as ordered sets ([source](https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt#L63)). They probably should be -- they have a consistent iteration order.

Side note, this issue was tricky to debug because the toString representation of the expected and actual values were identical, forcing me to step through the execution with a debugger.

Also, I believe this was previously working in Kotest 4.x.

5.1.0.

[1] Exception is thrown here: https://github.com/kotest/kotest/blob/45df11acd0310877af75eb5229b79bcaba2922ce/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt#L71

Kantis:


sksamuel:
Should we expose the inner error too like in #2879 ?

sksamuel (approved):

Kantis:
Thanks for the review! Yeah - I think so.. I added it as well 👍🏼

PR #2888, opened 2022-03-21
modify documentation for merging

### LINKED ISSUE #2865: Combine multiple generators with an even distribution
Consider something like this:
```
val foobar = listOf(
    Arb.negativeLong(),
    ....,
).reduce { acc, gen ->  
    acc.merge(gen)
}
```
Because of the way that the `merge` function is written earlier generators in the list will be exponentially less likely to be called. I've dealt with this in my own project with the following function:
```
fun <T> List<Gen<T>>.mergeAllGen(): Arb<T> = object : Arb<T>() {
    private val arbs = this@mergeAllGen.filterIsInstance<Arb<T>>()

    override fun edgecase(rs: RandomSource): T? =
        arbs.random(rs.random).edgecase(rs)

    override fun sample(rs: RandomSource): Sample<T> =
        when (val gen = this@mergeAllGen.random(rs.random)) {
            is Arb -> gen.sample(rs)
            is Exhaustive -> gen.toArb().sample(rs)
        }
}
```

If you run this sample you can see the impact is quite dramatic:
```
val sample = listOf(
    Arb.double(),
    Arb.constant(0L),
    Arb.constant(0L),
    Arb.constant(0L),
    Arb.constant(0L),
    Arb.constant(0L),
    Arb.constant("hello"),
)
println(sample.reduce { acc, gen -> acc.merge(gen) }.edgecases())
println(sample.mergeAllGen().edgecases())
```

I'm happy to add this + update the docs if there is a template PR I can go off of.

myuwono:
### In this PR
- update documentation for merge as appropriate

fixes #2865

### DIFF
--- a/documentation/docs/proptest/genops.md
+++ b/documentation/docs/proptest/genops.md
@@ -54,23 +54,27 @@ val dependentArbs: Arb<String> = Arb.of("foo", "bar").flatMap { prefix ->
 
 ## Merging
 
-Two generators can be merged together, so that elements 0, 2, 4, ... are taken from the first generator, and elements 1, 3, 5, ... are taken from the second generator.
+Two generators can be merged together, so that the resulting elements are equally sampled from both generators.
 
 ```kotlin
 val merged = arbA.merge(arbB)
 ```
 
-So with the following example:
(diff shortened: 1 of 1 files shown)


sksamuel:
Nice

PR #2893, opened 2022-03-23
remove incorrect comments

coffee-and-tea:
Comment and function does not match


sksamuel:
What's wrong with the comments ?

coffee-and-tea:
> What's wrong with the comments ?

comments seem like for function `shouldContainInOrder`,  not for `shouldContainExactly`, if I'm reading those correctly.

sksamuel:
👍🏼

PR #2895, opened 2022-03-23
haveElementAt Matcher throw ArrayIndexOutOfBoundsException

coffee-and-tea:
ArrayIndexOutOfBoundsException when testing with following:
```
listOf("a", "b", null) shouldNot haveElementAt(10, null)
```
Expect to have a positive result, since index 10 does not exist


sksamuel:
Your PR is changing docs ?

coffee-and-tea:
> Your PR is changing docs ?

Nop, pushed the code change commit

ashishkujoy:
Thanks @coffee-and-tea for the fix.

PR #2897, opened 2022-03-24
Match Json schema

### LINKED ISSUE #2817: json.shouldMatchSchema(schema)
Asserts that the json string or node matches the given schema.
Need a simple schema builder. Doesn't need to be full json schema compatible.

Kantis:
Closes #2817

Adds schema builder. Example: 
```kotlin
val addressSchema = jsonSchema {
  obj {   // object is reserved, obj was chosen over jsonObject for brevity but could be changed ofc, or jsonObject could be added as alternative.
    withProperty("street", required = true) { string() }
    withProperty("zipCode", required = true) { 
      integer { 
        beEven() and beInRange(10000..99999)   // supports constructing a matcher that will be used to test values
      } 
    }
    additionalProperties = false   // triggers failure if other properties are defined in actual
  }
}

val personSchema = jsonSchema {
  obj {
    withProperty("name", required = true) { string() }
    withProperty("address") { addressSchema() }
  }
}
```

Adds schema validation for `String` and `kotlinx.serialization.JsonElement`, which can be tested for compliance with a `JsonSchema`.

```kotlin
"{}" shouldMatchSchema personSchema 


// fails with:
$.name => Expected string, but was undefined
```

Adds partial parsing of [json schema](https://json-schema.org/specification.html)
```kotlin
val parsedSchema = parseSchema(
  """
  {
  "$id": "https://example.com/geographical-location.schema.json",  // will  be ignored
  "$schema": "https://json-schema.org/draft/2020-12/schema",       // will be ignored
  "title": "Longitude and Latitude Values",                                            // will be ignored
  "description": "A geographical coordinate.",                                     // will be ignored
  "required": [ "latitude", "longitude" ],
  "type": "object",
  "properties": {
    "latitude": {
      "type": "number",
      "minimum": -90,
      "maximum": 90
    },
    "longitude": {
      "type": "number",
      "minimum": -180,
      "maximum": 180
    }
  }
}
  """
)
```

Missing support for: 
* `$defs` and `$refs`
* Recursive schemas
* Parsing of schema composition
* `string.format` 
* `array.prefixItems`,
* `array.contains`, 
* `array.items = false`
* `array.maxContains`
* `array.minContains`
* `array.uniqueItems`
* `enum`


kotest-assertions/kotest-assertions-json/build.gradle.kts
```diff
@@ -22,13 +22,13 @@ kotlin {
          dependencies {
             implementation(Libs.Serialization.json)
             implementation(project(Projects.Assertions.Shared))
+            implementation(project(Projects.Assertions.Core))
```
Kantis, line 26: Used for some `Matcher`s, see `parse.kt`

Kantis:
Would it make sense to mark the entire feature as `@ExperimentalKotest` and let it incubate for a while?

documentation/sidebars.js
```diff
@@ -1,203 +1,216 @@
 module.exports = {
-   docs: [
-      'quickstart',
-      'changelog',
-      'blogs',
+   "docs": [
+      "quickstart",
+      "changelog",
+      "blogs",
```
Kantis, line 5: Changed all quotes to make it easier to cut docs

sksamuel:
You happy with this?
I'm going to release 5.3.0 tonight/tomorrow so it could go into that/

Kantis:
Yeah, would love to get it out 🙂

documentation/docs/assertions/json/schema.md
```diff
@@ -0,0 +1,92 @@
...
+  """
+)
+```
+
+or using Kotest's built-in DSL:
+
+```kotlin
+val addressSchema = jsonSchema {
+  obj {   // object is reserved, obj was chosen over jsonObject for brevity but could be changed ofc, or jsonObject could be added as alternative.
+    withProperty("street", required = true) { string() }
```
sksamuel, line 47: What do you think about allowing `string("name", true)` as well, for simplicity ?

Kantis: Yeah, that would be neat :)

Kantis: Everything is marked as experimental, so should be fine to make further additions/changes I think? Would you be ok with dropping a 5.3.1 with some DSL-additions or wait a day or two for me to finish that?

sksamuel: 5.3.1 is fine, or later even.
Gives us chance to gain more feedback.

PR #2903, opened 2022-03-26
Adjust PIT Gradle plugin configuration

szpak:
Starting with PIT 1.6.7 it is just enough to have kotest-extensions-pitest on (test) classpath

Over more that parameter is already deprecated and "marked for deletion". See:
https://github.com/hcoles/pitest/pull/900
https://github.com/szpak/gradle-pitest-plugin/issues/277

### DIFF
--- a/documentation/docs/extensions/pitest.md
+++ b/documentation/docs/extensions/pitest.md
@@ -22,7 +22,7 @@ After doing that, we need to inform Pitest that we're going to use `Kotest` as a
 ```kotlin
 // Assuming that you have already configured the Gradle/Maven extension
 configure<PitestPluginExtension> {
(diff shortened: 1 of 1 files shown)


sksamuel:
Did something change in PIT to automatically register Kotest? I looked at the changelog but couldn't find anything.

szpak:
Please see the first [linked issue](https://github.com/hcoles/pitest/pull/900):
> As a result of this change the 'testPlugin' parameter is now redundant, but has been retained for this release.

sksamuel:
Ok so it must be looking for instances of org.pitest.testapi.TestPluginFactory on the classpath.

szpak:
I guess so. Thanks to that you can also have multiple plugins enabled at the same time.

PR #2907, opened 2022-03-30
print generated input parameters and perform shrinks on any throwable

### LINKED ISSUE #2906: Property Test does not shrink or report failing inputs when function being tested throws
5.2.2

The test below does not print out the generated value when the function being tested throws. 
```kotlin
test("should print out failing inputs and shrinks when a function throws") {
    suspend fun functionThatUnexpectedlyThrows() { throw RuntimeException("aaaa") }
    checkAll(Arb.bigInt(4, 4)) { value ->
       functionThatUnexpectedlyThrows()
    }
}
   
// Property failed after 1 attempts
// java.lang.AssertionError: Property failed after 1 attempts
//
// Repeat this test by using seed 320358835705021305
// 
// Caused by RuntimeException: aaaa
```

I realized that we don’t report the inputs or attempt any shrinking to the users when the exception is not kotest’s assertion errs. i.e. if that function was wrapped around `shouldNotThrowAny { .. }` then that’ll print out the inputs and properly perform shrinking. 

Is this by design? or should we fix that?

myuwono:
### In this PR
- print generated input parameters and perform shrinks on any throwable

potentially fixes #2906 

However, it seems like the explicit throwable type matches were introduced for a reason. I would need @sksamuel to triage whether that is the case or whether this code change is legitimate.

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/internal/test.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/internal/test.kt
@@ -41,18 +41,9 @@ internal suspend fun test(
       fn()
       context.markSuccess()
       coroutineContext[AfterPropertyContextElement]?.after?.invoke()
-   } catch (e: AssertionError) { // we track assertion errors and try to shrink them
+   } catch (e: Throwable) {  // we track any throwables and try to shrink them
       context.markFailure()
       handleException(context, shrinkfn, inputs, seed, e, config)
-   } catch (e: Exception) {
-      context.markFailure()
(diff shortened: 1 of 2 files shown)


sksamuel (approved):

sksamuel:
The only reason it's assertion errors not throwables is to avoid shrinking on a non-assertion problem, but I think this is a valid change.

myuwono:
@sksamuel makes sense. I'll go ahead and merge this within the next 8 hours if you don't have any concerns.

sksamuel:
Sure merge away.

Issue #2892, opened 2022-03-23
shouldStartWith and shouldEndWith should support Regex


nkiesel:
Kotlin stdlib overloads methods for strings and regex (e.g. `replace`).  The same should be done for kotest assertions. I should be able to write
```kotlin
"I love KoTest assertions!" shouldStartWith "I love"
"I love KoTest assertions!" shouldStartWith "I (love|liebe)".toRegex()
```
Right now I instead have to use
```kotlin
"I love KoTest assertions!" shouldMatch "^I (love|liebe).*".toRegex()
```

sksamuel:
Good idea.

nor-ek:
You can assign me.

sksamuel:
Done

nor-ek:
PR -> https://github.com/kotest/kotest/pull/2909

PR #2909, opened 2022-03-31
Matcher support startsWith regex

nor-ek:
Solution is based on **matches** matcher since I don't see **startsWith** method in Kotlin stdlib. If this is not acceptable then I'll find another way.
I've also prepared tests and updated documentation.

I also noticed that there would be need to write similar method in **endsWith** matcher. Are you able to prepare task for that?

Closes feature/gh-2892


nkiesel:
Looks good to me. I should have mentioned this earlier: could you please also add a `shouldEndWith` with regex parameter (given that this exists for string parameters)? (Update: I just noticed that this was already mentioned above.)

nkiesel:
for `startsWith`, we could use `matchesAt(regex, 0)` to avoid wrapping in `^` and `.*`.

nor-ek:
@nkiesel Of course I will add `shouldEndWith`. However I see `matchesAt` is annotated with  `ExperimentalStdlibApi`. Is it fine? I've noticed that `@OptIn` is used so you can check changes on PR.

sksamuel:
👍🏼

PR #2919, opened 2022-04-11
docs: mention Native platforms on the framework main page

vitorhugods:
Even though https://kotest.io/ mentions "multi-platform", it limits itself to `JVM or Javascript`.

This can be a bit deceiving and lead people to think that native targets are not supported before digging deeper into docs and understanding that this isn't the case.

This PR adds a mention to "native platforms".

How it currently looks on the website, with "Jvm or Javascript" highlighted:
grafik


sksamuel (approved):

sksamuel:
Nice thanks

PR #2921, opened 2022-04-13
Remove deprecated listeners override

CucumisSativus:
`listeners` are deprecated and will be removed. I suggest we update the HTML generator documentation


sksamuel:
Thanks

PR #2931, opened 2022-04-21
Version selection for docs

Kantis:
Adds a version selector:
image

Release process of new minor version would include cutting docs by following the steps:

1. Create `versioned_docs/version-<VersionToRelease>`
2. Copy current `docs` content to the dir created in step 1
3. Copy current sidebar to `versioned_sidebars/version-<VersionToRelease>-sidebars.json`

I think it could be worthwhile having this approach for the docs. 
- 🟢 Old docs will still be available for people stuck on older versions.
- 🟢 Contributors can update current `docs` folder documentation with their changes for next version (while adding the regular change, instead of separating documentation into a later PR). 
- 🔻 Tedium added to release process.
- 🔻 More complex documentation structure.

Latest version is the one shown by default (e.g. latest cut version of the docs). Going to the current docs shows a warning  that the docs are for an unreleased version:
image


sksamuel:
146 files to review is a lot, so what's the tl'dr
Each time we do a minor release bump we copy all the docs from /5.2 to /5.3 and add a new selector ?

Kantis:
Support the TLDR is: suffer some copy-pasting required when releasing versions to gain versioned docs + ability to work on next version docs as part of regular PRs. I think it would be worthwhile.

sksamuel:
But I guess I was asking what that copy n paste would be. Is it just a folder of all the docs with a prefix ?

Kantis:
Yeah, just make a copy of the current docs dir to freeze it. Same thing for the sidebars.

143 of the files changed are just the docs freeze for 5.2.

sksamuel (approved):

sksamuel:
Ok cool

PR #2935, opened 2022-04-22
YearMonth Arbitrary implementation

### LINKED ISSUE #2928: Add YearMonth Arbitrary
We have Arbitraries for LocalDate, LocalDateTime, LocalTime and more of these cute `java.time` guys.

I propose we also add a `YearMonth` generator, which would behave similarly (and can even be derived from the others if they contain a date `localDateTime.yearMonth`)

# YearMonthArb

I propose this function:

```
fun Arb.Companion.yearMonth(
minimumYearMonth: YearMonth = YearMonth.of(1970, 1),
maximumYearMonth: YearMonth = YearMonth.now().plusYears(10)
)
```

## Edge cases
I think there are some good edgecases that we can add:
- `YearMonth.now()` if it's in the range
- Minimum of the range
- Max of the range
- All leap years/leap februaries in the range

### Kotest Property Arb Project

This could be added to https://github.com/kotest/kotest-property-arbs instead of the main repository. I think it could be part of the main Arbs in the JVM, but I also agree that it might be a little bit of extra-bloat.

Where should we stop? There are many possible `Temporal`s. We did have a generic solution to mock the `now` function with `CurrentInstantListener`, but I don't think this can be as generic.

nor-ek:
YearMonth arbitrary implementation and tests.

Edge cases of arbitrary:
- Minimum of the range
- Max of the range
- All leap february in the range

Closes #2928

### DIFF
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
@@ -7,6 +7,7 @@ import java.time.LocalDateTime
 import java.time.LocalTime
 import java.time.Period
 import java.time.Year.isLeap
+import java.time.YearMonth
 import java.time.temporal.ChronoUnit
 import java.time.temporal.TemporalQueries.localDate
 import java.time.temporal.TemporalQueries.localTime
@@ -160,6 +161,28 @@ fun Arb.Companion.localDateTime(
          }.first { !it.isBefore(minLocalDateTime) && !it.isAfter(maxLocalDateTime) }
       }
    )
+}
+
+/**
+ * Arberates a stream of random YearMonth
+ *
+ * If any of the years in the range contain a leap year, the date [02/YEAR] will always be a constant value of this
+ * generator.
+ *
+ * This generator creates randomly generated YearMonth, in the range [[minYearMonth, maxYearMonth]].
+ *
+ * @see [yearMonth]
+ */
+fun Arb.Companion.yearMonth(
+   minYearMonth: YearMonth = YearMonth.of(1970, 1),
+   maxYearMonth: YearMonth = YearMonth.of(2030, 12)
+): Arb<YearMonth> {
+   val leapYears = (minYearMonth.year..maxYearMonth.year).filter { isLeap(it.toLong()) }
+   val february = leapYears.map { YearMonth.of(it, 2) }
+
+   return arbitrary(february + minYearMonth + maxYearMonth) {
+      minYearMonth.plusMonths(it.random.nextLong(ChronoUnit.MONTHS.between(minYearMonth, maxYearMonth)))
+   }.filter { it in minYearMonth..maxYearMonth }
 }
 
 typealias InstantRange = ClosedRange<Instant>
--- a/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/DateTest.kt
+++ b/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/DateTest.kt
@@ -1,5 +1,6 @@
 package com.sksamuel.kotest.property.arbitrary
 
+import io.kotest.assertions.throwables.shouldNotThrowAny
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.inspectors.forAll
 import io.kotest.matchers.collections.shouldContain
@@ -14,6 +15,7 @@ import io.kotest.property.arbitrary.localDateTime
 import io.kotest.property.arbitrary.localTime
 import io.kotest.property.arbitrary.period
 import io.kotest.property.arbitrary.take
+import io.kotest.property.arbitrary.yearMonth
 import io.kotest.property.checkAll
 import io.kotest.property.forAll
 import java.time.LocalDate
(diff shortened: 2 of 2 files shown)


kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/DateTest.kt
```diff
@@ -212,4 +214,46 @@ class DateTest : WordSpec({
          checkAll(10) { _: Period -> /* No use. Won't reach here if unsupported */ }
       }
    }
+
+   "Arb.yearMonth(minYearMonth, maxYe)" should {
+      "generate valid YearMonths (no exceptions)" {
+         Arb.yearMonth().take(10_000).toList()
+      }
```
LeoColman:
```suggestion
      "generate valid YearMonths" {
         shouldNotThrowAny {
             Arb.yearMonth().take(10_000).toList()
         }
      }
```

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/DateTest.kt
```diff
@@ -212,4 +214,46 @@ class DateTest : WordSpec({
...
+         val years = mutableSetOf<Int>()
+         val months = mutableSetOf<Int>()
+
+         checkAll(10_000, Arb.yearMonth(YearMonth.of(1998, 1), YearMonth.of(1999, 12))) {
+            years += it.year
+            months += it.monthValue
+         }
+
+         years shouldBe setOf(1998, 1999)
+         months shouldBe (1..12).toSet()
```
LeoColman: I think this test needs improvement.

We're not guaranteeing here that we're not generating YearMonths from outside the range specified

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/DateTest.kt
```diff
@@ -212,4 +214,46 @@ class DateTest : WordSpec({
...
+      }
+
+      "Generate YearMonth always in the range" {
+         repeat(1_000) {
+            val min = YearMonth.of(1970, 1).plusMonths(it.toLong())
+            val max = min.plusYears(20)
+            Arb.yearMonth(min, max).forAll { it >= min && it <= max }
+         }
+      }
```
LeoColman: This adds to my previous comment about improving the tests.
I think having some glanularity for months should be validated in the tests as well

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/DateTest.kt
```diff
@@ -212,4 +214,46 @@ class DateTest : WordSpec({
...
+
+      "Contain Feb if leap year" {
+         val leapYear = 2016
+         Arb.yearMonth(YearMonth.of(leapYear, 1), YearMonth.of(leapYear, 12)).edgecases() shouldContain YearMonth.of(
+            2016,
+            2
+         )
+      }
+
+      "Contain now if leap year" {
```
LeoColman: I'd say that maybe `now` should always be a constant if it's within the range

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
```diff
@@ -162,6 +163,31 @@ fun Arb.Companion.localDateTime(
...
+ */
+fun Arb.Companion.yearMonth(
+   minYearMonth: YearMonth = YearMonth.of(1970, 1),
+   maxYearMonth: YearMonth = YearMonth.of(2030, 12)
+): Arb<YearMonth> {
+   val leapYears = (minYearMonth.year..maxYearMonth.year).filter { isLeap(it.toLong()) }
+   val leapMin = YearMonth.of(leapYears.first(), minYearMonth.month)
+   val leapMax = YearMonth.of(leapYears.last(), maxYearMonth.month)
+   val leapNowIfBetween = YearMonth.now().takeIf { it in leapMin..leapMax }
+   val february28s = leapYears.map { YearMonth.of(it, 2) }
```
LeoColman: It's possible to simplify this logic a bit:

```
val yearMonthRange = (minYearMonth..maxYearMonth) // Maybe this method doesn't exist, but something similar might
val leapYearMonths = yearMonthRange.filter { isLeap(it.year) }

val edgeCases = leapYearMonths
// ...

```

LeoColman (changes requested):

LeoColman:
Nice pull request, @nor-ek ! 😄

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
no-u42: It's impossible to filter ClosedRange :/  However I've simplified it a bit by removing now from leap years.

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/DateTest.kt
no-u42: I've also improved `generate valid LocalDates (no exceptions)` by this assertion.

nor-ek:
@LeoColman thanks for review. I've improved it a bit. Not sure about https://github.com/kotest/kotest/pull/2935#discussion_r856376041 how would you do this?

sksamuel:
> @LeoColman thanks for review. I've improved it a bit. Not sure about [#2935 (comment)](https://github.com/kotest/kotest/pull/2935#discussion_r856376041) how would you do this?

Maybe just a single test with a specific year range ?

nor-ek:
Okay so let's say it's done. I removed this test in case of test `generate YearMonths between minYearMonth and maxYearMonth` is checking specific yearMonthRange.

sksamuel:
👍🏼

PR #2938, opened 2022-04-26
Update junit_xml.md

obecker:
Fix: `JunitXmlReporter` is a `Listener`

### DIFF
--- a/documentation/versioned_docs/version-5.2/extensions/junit_xml.md
+++ b/documentation/versioned_docs/version-5.2/extensions/junit_xml.md
@@ -26,7 +26,7 @@ To configure in your project, you need to add the `JunitXmlReporter` using [proj
 
 ```kotlin
 class MyConfig : AbstractProjectConfig() {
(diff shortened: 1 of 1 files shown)


sksamuel:
Listeners are subsumed into extensions in 5.x

Are you running 4.x?

obecker:
I'm using `io.kotest:kotest-extensions-htmlreporter:5.2.3`.
In my IDE I'm seeing `HtmlReporter` extends `ProjectListener`, which extends `BeforeProjectListener` and `AfterProjectListener`, and both extend `Listener`.
I got a compile error because `HtmlReporter` was not an `Extension` ...

obecker:
Ah, I got it. I have `io.kotest:kotest-framework-api-jvm:4.6.4` in my classpath. Sorry, my fault.

sksamuel:
You should be using 5.2.3 framework, and then the docs are current.
Does that not work for you ?

sksamuel:
Ok glad you got it sorted :)

PR #2939, opened 2022-04-26
Update html_reporter.md

obecker:
Fix: both reporters are of type `Listener`


obecker:
I have used a wrong version.

PR #2944, opened 2022-04-29
SortedBy matcher

### LINKED ISSUE #2940: Add shouldBeSortedBy matcher
Maybe I missed it, but I can't find a quick way to check if a list of things is sorted according to one of the fields of the objects in the list

```kotlin
class DataPoint(timestamp: Instant, count: Long)
val data = listOf(DataPoint(Instant.now(), 10), DataPoint(Instant.now().plusSeconds(1)))

data shouldBeSortedBy { it.timestamp } // <- shouldBeSortedBy doesn't exist
```

nlochschmidt:
Adding a `sortedBy` turned out to be surprisingly easy 👏 

The old `sorted()` simply forwards to the new `sortedBy` with the identity function `{ it }`.

Closes #2940


sksamuel:
👍🏼

Issue #2913, opened 2022-04-05
Pitest fails to run Junit5/Kotest tests because it assigns unique id selectors


ffeltrinelli:
**Which version of Kotest are you using**
I'm upgrading kotest from 5.1.0 (where it works fine) to 5.2.2 (where I have the error described below).

**Description**

In #2862, kotest was updated to throw an exception if tests running with `kotest` engine have a unique id selector.

Unfortunately, this is exactly what [Pitest](http://pitest.org/) is doing:

```
PIT >> INFO : MINION : org.pitest.util.PitError: TestEngine with ID 'kotest' failed to discover tests
PIT >> FINE : MINION :      at org.pitest.util.Unchecked.translateCheckedException(Unchecked.java:20)
PIT >> FINE : MINION :      at org.pitest.coverage.execute.CoverageWorker.run(CoverageWorker.java:51)
PIT >> FINE : MINION :      at org.pitest.coverage.execute.CoverageMinion.main(CoverageMinion.java:88)
PIT >> FINE : MINION : Caused by: org.junit.platform.commons.JUnitException: TestEngine with ID 'kotest' failed to discover tests
PIT >> FINE : MINION :      at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discoverEngineRoot(EngineDiscoveryOrchestrator.java:160)
PIT >> FINE : MINION :      at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discoverSafely(EngineDiscoveryOrchestrator.java:134)
PIT >> FINE : MINION :      at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discover(EngineDiscoveryOrchestrator.java:108)
PIT >> FINE : MINION :      at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discover(EngineDiscoveryOrchestrator.java:80)
PIT >> FINE : MINION :      at org.junit.platform.launcher.core.DefaultLauncher.discover(DefaultLauncher.java:110)
PIT >> FINE : MINION :      at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:86)
PIT >> FINE : MINION :      at org.junit.platform.launcher.core.DefaultLauncherSession$DelegatingLauncher.execute(DefaultLauncherSession.java:86)
PIT >> FINE : MINION :      at org.junit.platform.launcher.core.SessionPerRequestLauncher.execute(SessionPerRequestLauncher.java:53)
[...]

PR #2955, opened 2022-05-03
Updated docs on Pitest extension for Maven users

ffeltrinelli:
Following the conversation in #2913 , I'm updating the Pitest extension docs to get more clarity on how it should be configured for Maven users.

### DIFF
--- a/documentation/docs/extensions/pitest.md
+++ b/documentation/docs/extensions/pitest.md
@@ -8,6 +8,8 @@ slug: pitest.html
 
 The Mutation Testing tool [Pitest](https://pitest.org/) is integrated with Kotest via an extension module.
 
+## Gradle configuration
+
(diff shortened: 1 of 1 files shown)


Kantis:
I think it would be great to add the same change in the version 5.3 docs (otherwise they'll only show for the unreleased 5.4 which most users will probably not be looking at for a while)

https://github.com/kotest/kotest/blob/master/documentation/versioned_docs/version-5.3/extensions/pitest.md

sksamuel:
Oops merged without reading @Kantis' comment. Can we get another PR for the 5.3 docs ?

ffeltrinelli:
@Kantis @sksamuel sorry for the late answer, yes of course I raised #2982 to target also kotest 5.3 docs!

PR #2957, opened 2022-05-04
allow nested js tests

benkuly:


benkuly:
Update:

It runs the tests without problems. I just have two issues with this solution and don't really know, how to fix them:
* `beforeSpec` and `afterSpec` seems to get ignored.
* The fail output doesn't make clear,  which subtest has failed.

benkuly:
Please keep open for the case that someone could solve the two remaining issues.

benkuly:
> Please keep open for the case that someone could solve the two remaining issues.

benkuly:
Would be cool to find a solution, because many advantages of kotest disappear without support of nested test in all targets.

aSemy:
Hi @benkuly, I've checked out your branch locally and updated it with the latest commits from the master branch, and I think it's working now. Could you please update your branch and check? If you can't, no worries, just let me know and I'll make a new PR.

benkuly:
@aSemy feel free to open a new MR and close this. I just tried to update to upstream, but there were too many changes and it has been to long since I implemented it...

benkuly:
see https://github.com/kotest/kotest/pull/3913

PR #2963, opened 2022-05-09
[docs] Fix description of `Arb.harryPotterCharacter`

AndreyBozhko:
Looks like the same description is used for both `Arb.wines` and `Arb.harryPotterCharacter` - this PR fixes the doc for the latter.


sksamuel:
👍🏼

PR #2968, opened 2022-05-14
Added assumptions for property testing

### LINKED ISSUE #2966: Property test assumptions
Inspired by https://jqwik.net/docs/current/user-guide.html#assumptions

sksamuel:
Closes #2966

Details of this feature provided in the docs in the PR.
Feedback on syntax and requirements welcome.


Kantis:
Looks awesome! :) 

Perhaps we could allow defining assumptions using matchers as well? 

```kotlin

fun interface Assume {
  operator fun invoke()
}

checkAll<String, String> { a, b ->
  withAssumption(Assume { a shouldNotBe b }) {
    levenshtein(a, b) shouldBeGreaterThan 0
  }
}
```

Perhaps we could also have `fun withAssumptions(vararg assumptions: Assume, test: suspend () -> Unit)`  for defining multiple assumptions, to avoid deep nesting.. disclaimer: Haven't used assumptions before so I'm not sure how common it is to apply more than 1-2 assumptions. 

```kotlin
checkAll<String, String> { a, b ->
  withAssumptions(
    Assume { a shouldNotStartWith b[0] },
    Assume { a shouldHaveMinLength(b.length) } 
  ) {
    levenshtein(a, b) shouldBeGreaterThan 0
  }
}
```

We could ditch the interface if it just clutters the API.

```kotlin
checkAll<String, String> { a, b ->
  withAssumption({a shouldNotBe b}) {
    levenshtein(a, b) shouldBeGreaterThan 0
  }
}

checkAll<String, String> { a, b ->
  withAssumptions(
    { a shouldNotStartWith b[0] },
    { a shouldHaveMinLength(b.length) } 
  ) {
    levenshtein(a, b) shouldBeGreaterThan 0
  }
}
```


Jqwik seems to be a bit more forgiving with the discard ratio (80% vs 10%), but I'm not sure if it's good to allow that much discard... 😄

sksamuel:
Oh I read jqwik's ratio as 20% not 80%, 80% seems really high.
Either way, we could increase this.

I like your Assume suggestion.

sksamuel:
Updated based on feedback @Kantis

enolive:
coming here from your reply to my twitter post https://twitter.com/_sksamuel/status/1526210446176526338

looks promising so far!

I noticed however 2 possible problems. correct me if I'm wrong:

- the discard counter seems to increase on any `AssertionError`, even on the execution of the test() method (for instance when using checkAll instead of forAll). Wouldn't it be better to only count real assumption failures that occur while invoking the assumptions? 
- as `Arb.filter` is often used for a simple assumption alternative (you even mention it in your docs), would it be possible/smart to also consider filtered out values as a discard? I remember the only time I ruined the performance running properties while experimenting with QuickCheck was a really awful `suchThat` filter ;-)

enolive:
coming here from your reply to my twitter post https://twitter.com/_sksamuel/status/1526210446176526338

looks promising so far!

I noticed however 2 possible problems. correct me if I'm wrong:

- the discard counter seems to increase on any `AssertionError`, even on the execution of the test() method (for instance when using checkAll instead of forAll). Wouldn't it be better to only count real assumption failures that occur while invoking the assumptions? 
- as `Arb.filter` is often used for a simple assumption alternative (you even mention it in your docs), would it be possible/smart to also consider filtered out values as a discard? I remember the only time I ruined the performance running properties while experimenting with QuickCheck was a really awful `suchThat` filter ;-)

sksamuel:
> 

Not sure I understand your second point @enolive but your first point I understand and I will clarify the statistics.

enolive:
here some explanation:

a filter statement can be used interchangeably to an assumption (for instance, with Arb.pair(...).filter). It has the same risks as using assumptions with too many values being discarded.

consider the following arbitrary that will run endlessly:

```kotlin
checkAll(Arb.int().filter { false }) {
  ...
}
```

With checking the discard attempts we could abort after x tries, just like with assumptions. I am not sure if this is even possible to implement (as the Arb would need some context knowledge from the running propTest).

sksamuel:
> here some explanation:
> 
> a filter statement can be used interchangeably to an assumption (for instance, with Arb.pair(...).filter). It has the same risks as using assumptions with too many values being discarded.
> 
> consider the following arbitrary that will run endlessly:
> 
> ```kotlin
> checkAll(Arb.int().filter { false }) {
>   ...
> }
> ```
> 
> With checking the discard attempts we could abort after x tries, just like with assumptions. I am not sure if this is even possible to implement (as the Arb would need some context knowledge from the running propTest).

Ahh I see.
I think with assumptions, you don't get "extra" values, you just lose some iterations. So if you filter 50%, you just get 50% less values.

Perhaps that should be changed, so assumptions are skipped but it doesn't reduce the iteration count like how filters work.

@myuwono any opinions ?

kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
```diff
@@ -101,6 +101,7 @@ data class PropTestConfig(
    val outputClassifications: Boolean = PropertyTesting.defaultOutputClassifications,
    val labelsReporter: LabelsReporter = StandardLabelsReporter,
    val constraints: Constraints? = null,
+   val maxDiscardPercentage: Int = 20,
```
LeoColman, line 117: Docs say 10%
Here it's 20%

documentation/docs/proptest/assumptions.md
```diff
@@ -0,0 +1,97 @@
...
+Kotest expands on basic boolean assumptions by allowing you to specify any assertion in the assumption predicate, in
+addition to supporting an arbitrary number of assumptions.
+
+For example, building on the previous example:
+
+```kotlin
+checkAll(Arb.string(3..4, Codepoint.az()), Arb.string(3..4, Codepoint.az())) { a, b ->
+  withAssumptions(
+     { a shouldNotBe b },
+     { a shouldHaveLength (b.length) },
```
LeoColman, line 62: I supposed these assertions shouldn't be counted in the AssertionsCounter, mainly for the purposes of not falsely counting assertions if no test is actually made

LeoColman:
Syntax Proposal:
```
checkAll<String, String> { a, b ->
    assuming(a != b) 
    assuming { a != b }
    assuming { a shouldNotBe b }

    assume(a != b)
    assume { a != b }
    assume { a shouldNotBe b} 

   levenshtein(a, b)
} 

```

This brings the syntax closer to what we have with
```
require(boolean)
assert(boolean)
check(boolean)
```
from the STDLib

It also brings a bit the concept of contracts, although we're not directly adding a contract here. It will be an assumption from 'now on'

sksamuel:
> Syntax Proposal:
> 
> ```
> checkAll<String, String> { a, b ->
>     assuming(a != b) 
>     assuming { a != b }
>     assuming { a shouldNotBe b }
> 
>     assume(a != b)
>     assume { a != b }
>     assume { a shouldNotBe b} 
> 
>    levenshtein(a, b)
> } 
> ```
> 
> This brings the syntax closer to what we have with
> 
> ```
> require(boolean)
> assert(boolean)
> check(boolean)
> ```
> 
> from the STDLib
> 
> It also brings a bit the concept of contracts, although we're not directly adding a contract here. It will be an assumption from 'now on'

I like this, as it aligns with the stdlib.
Although it's going to need to use exceptions for control flow and breaks the functional approach I prefer.

Kantis:
I think the previous proposal also aligns with the stdlib. withAssumptions could be considered a scoping function, such as with and withContext. Everything in the scope is tested under the given assumption. 

I think the main benefit of @LeoColman's proposal would be getting rid of an additional level of nesting. Personally I think it's a bit clearer with the nesting but just my 2c.

sksamuel:
I like withAssumption as it is more FP to apply a function to a scope, rather than throwing an exception as control flow.
But the nesting approach can be ugly.

Should we support both ?

LeoColman:
> Should we support both ?

image


I like supporting both.

- The implementation will probably very similar (and shared) for both versions
- I don't see it as being confusing. @Kantis pointed out how the STDlib also likes some direct FP. `with(x)` is used everywhere
- Testing/maintaining both don't seem like a big hassle

myuwono (approved):

myuwono:
I think we can definitely go ahead and introduce this as an experimental feature and keep iterating on it. It's definitely a great feature to have! This reminds me of ScalaCheck's conditional property, i.e. the `==>` function which also works similarly with scopes.

Around discards. I agree I think we should treat it similarly to filter. We can give user the config how many discards are "acceptable" with a sane default e.g. `maxAssumptionAttempts`. For instance the test below should run exactly 10 times with the input set that satisfies all the assumptions, instead of failing the entire suite. Shrinking will be a little bit interesting..
```kotlin
var n: Int = 0
checkAll(iterations = 10, arbUser, arbAuthenticationPolicy) { user, policy ->
  assume(user.isActive) 
  assume(policy.enforcements in listOf(PolicyEnforcement.Password, PolicyEnforcement.Session))

  // do the rest of the test in here ...
  n++
}
// after this i expect n == 10
``` 
The rationale behind that is often times in integration tests (which is often slower), one may opt to have relatively lower number of iterations to strike a balance between exhaustiveness and duration. As an illustration, if user picks 10 iterations, and puts in some assumptions, most likely in these situations what users really meant was - "I want to get all inputs such that <assumption(s)> hold, and then run the test exactly 10 times".

myuwono:
I've done some experiment to see how ScalaCheck conditional property works. Specifying conditions doesn't consume iterations. This can be a data point.
```scala
class DummySpec extends Specification with ScalaCheck {
  var outer = 0
  var inner = 0
  "dummy conditional property" in prop { (first: Int, second: Int) =>
    outer += 1
    println(s"$outer - $inner")
    (first > 0) ==> {
      (second < 0) ==> {
        inner += 1
        first must beGreaterThan(second)
      }
    }
  }
}

// outputs:
// 1 - 0
// 2 - 0
// 3 - 1
// ...
// 595 - 99
// 597 - 99
```

sksamuel:
In think you're right, assumptions should just be like filter. I'll change
the pr.

LeoColman:
We still need to address assertion counter. Is it easy to say

"Hey, disable the assertion counter for this block"?

Jun. 25, 2022 12:05:26 Sam ***@***.***>:

sksamuel:
Should be.

sksamuel:
`assume(boolean)` and `assume { }` variants added.
Changed logic so that discards assumptions do not count towards the iteration total.

PR #2970, opened 2022-05-15
Support persisting and reusing seed for property tests #2967

### LINKED ISSUE #2967: Property testing - re use failed seed automatically
Inspired by https://jqwik.net/docs/current/user-guide.html#rerunning-falsified-properties

sksamuel:
Closes #2967


documentation/docs/proptest/seed.md
```diff
@@ -0,0 +1,60 @@
...
+sidebar_label: Seeds
+---
+
+
+When a property test is executed, the values are generated using a random source that is created from a seed value. By
+default this seed value is itself randomly chosen (using the default `kotlin.random.Random` instance). However, there
+are times when this value needs to be fixed or repeated.
+
+You can change the default used by all tests, unless overriden through the options listed below, by changing the
+configuration value `PropertyTesting.defaultSeed`.
```
Kantis, line 14: An example here would be nice 😊  I'm not sure what the most appropriate place to put that config would be.

kotest-property/src/jvmMain/kotlin/io/kotest/property/seed/seedio.kt
```diff
@@ -0,0 +1,30 @@
...
+fun seedDirectory(): Path = Paths.get(System.getProperty("user.home")).resolve(".kotest").resolve("seeds")
+
+fun seedPath(path: TestPath): Path {
+   return seedDirectory().resolve(path.value)
+}
+
+actual fun writeSeed(path: TestPath, seed: Long) {
+   val f = seedPath(path)
+   f.parent.toFile().mkdirs()
+   Files.write(f, seed.toString().encodeToByteArray())
```
Kantis: Perhaps should be wrapped in a try/catch. In case this fails (for instance due to lacking permissions) it would be nice if it didn't break the test run. Or would that be overly defensive? Perhaps mostly relevant on CI/CD pipelines..

kotest-property/src/desktopMain/kotlin/io/kotest/property/seed/loadSeedFromFile.kt
```diff
@@ -0,0 +1,7 @@
+package io.kotest.property.seed
+
+import io.kotest.common.TestPath
+
+actual fun readSeed(path: TestPath): Long? = null
+actual fun writeSeed(path: TestPath, seed: Long) {}
```
Kantis, line 6: warn/error log that saving seeds is not implemented on JS/Desktop?

kotest-property/src/commonMain/kotlin/io/kotest/property/internal/checks.kt
```diff
@@ -2,8 +2,14 @@ package io.kotest.property.internal
 
 import io.kotest.property.PropTestConfig
 import io.kotest.property.PropertyContext
+import io.kotest.property.PropertyTesting
 import kotlin.math.min
 
+fun PropTestConfig.checkFailOnSeed() {
```
Kantis:
```suggestion
internal fun PropTestConfig.checkFailOnSeed() {
```

Kantis (commented):
added my 2c 😃

kotest-property/src/desktopMain/kotlin/io/kotest/property/seed/loadSeedFromFile.kt
sksamuel: Good idea.

myuwono (approved):

kotest-property/src/jvmMain/kotlin/io/kotest/property/seed/seedio.kt
sksamuel: exception handling added

kotest-property/src/desktopMain/kotlin/io/kotest/property/seed/loadSeedFromFile.kt
sksamuel: Hmm now I think about it, it's not optional per platform. It would just be loads of warnings everytime.

PR #2972, opened 2022-05-15
Added stateful testing #2654

### LINKED ISSUE #2654: Stateful testing
https://jqwik.net/docs/current/user-guide.html#stateful-testing
https://hypothesis.readthedocs.io/en/latest/stateful.html

sksamuel:
Closes #2654

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
@@ -1,0 +1,54 @@
+package io.kotest.property.state
+
+import io.kotest.property.RandomSource
+
+/**
+ * A function that is applied against the current state to mutate and/or test.
+ */
+interface Action<T> {
+
+   /**
+    * A predicate which decides if this action should be applied.
+    *
+    * For example, if you were writing a stateful test for a stack, you would want
+    * to skip 'pop' operations if the stack was empty.
+    *
+    * @return true if this action should be applied or false if the action should be skipped.
+    */
+   suspend fun precondition(state: T): Boolean
+
+   /**
+    * Invokes the next transition to the state.
+    *
+    * This function should contain your test assertions.
+    */
+   suspend fun run(state: T, rs: RandomSource)
+
+   companion object {
+
+      /**
+       * Creates a new [Action] that is always applied (precondition is set to constant true).
+       */
+      operator fun <T> invoke(f: (T, RandomSource) -> Unit): Action<T> {
+         return object : Action<T> {
+            override suspend fun precondition(state: T): Boolean = true
+            override suspend fun run(state: T, rs: RandomSource) = f(state, rs)
+         }
+      }
+   }
+}
+
+suspend fun <T> checkState(input: T, vararg actions: Action<T>) {
+   checkState(input, 1000, *actions)
+}
+
+suspend fun <T> checkState(input: T, iterations: Int, vararg actions: Action<T>) {
+   checkState(input, iterations, RandomSource.default(), *actions)
+}
+
+suspend fun <T> checkState(input: T, iterations: Int, rs: RandomSource, vararg actions: Action<T>) {
+   repeat(iterations) {
+      val action = actions.random()
+      if (action.precondition(input)) action.run(input, rs)
+   }
+}
--- a/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/state.kt
+++ b/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/state.kt
@@ -1,0 +1,64 @@
+package com.sksamuel.kotest.property
+
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.matchers.shouldBe
+import io.kotest.property.Arb
+import io.kotest.property.arbitrary.long
+import io.kotest.property.arbitrary.next
+import io.kotest.property.state.Action
+import io.kotest.property.state.checkState
+
+/**
+ * A cache that tracks how often a user's profile has been viewed.
+ *
+ * Each time [add] is invoked, that user is added to the list of viewers, and when
+ * [remove] is invoked, that user is removed.
+ * [get] returns the viewers for a user.
+ */
+class ViewerCache {
+
+   val map = mutableMapOf<Long, MutableList<Long>>()
+
+   override fun toString(): String {
+      return map.toString()
+   }
+
+   fun add(viewee: Long, viewer: Long) {
+      val viewers = map.getOrPut(viewee) { mutableListOf() }
+      viewers.add(viewer)
+   }
+
+   fun remove(viewee: Long, viewer: Long) {
+      val viewers = map.getOrPut(viewee) { mutableListOf() }
+      viewers.remove(viewer)
+   }
+
+   fun get(viewee: Long): List<Long> {
+      return map.getOrElse(viewee) { emptyList() }
+   }
+}
+
+class ViewerCacheTest : FunSpec() {
+   init {
+
+      val add = Action<ViewerCache> { state, rs ->
+         val userId = Arb.long(1L..100L).next(rs)
+         val viewerId = Arb.long(1L..100L).next(rs)
+         state.add(userId, viewerId)
+         state.get(userId).contains(viewerId) shouldBe true
+      }
+
+      val remove = Action<ViewerCache> { state, rs ->
+         val userId = Arb.long(1L..100L).next(rs)
+         val viewerId = Arb.long(1L..100L).next(rs)
+         if (state.get(userId).contains(viewerId)) {
+            state.remove(userId, viewerId)
+            state.get(userId).contains(viewerId) shouldBe false
+         }
+      }
+
+      test("viewer cache state") {
+         checkState(ViewerCache(), add, remove)
+      }
+   }
+}


Kantis (approved):
Neat! Never used anything like it :)

kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
```diff
@@ -0,0 +1,54 @@
...
+   checkState(input, 1000, *actions)
+}
+
+suspend fun <T> checkState(input: T, iterations: Int, vararg actions: Action<T>) {
+   checkState(input, iterations, RandomSource.default(), *actions)
+}
+
+suspend fun <T> checkState(input: T, iterations: Int, rs: RandomSource, vararg actions: Action<T>) {
+   repeat(iterations) {
+      val action = actions.random()
```
myuwono, line 51: Sam, is this intended? as a user this will randomize my action order.. which I would consider a bug..?

kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
```diff
@@ -0,0 +1,54 @@
...
+    * This function should contain your test assertions.
+    */
+   suspend fun run(state: T, rs: RandomSource)
+
+   companion object {
+
+      /**
+       * Creates a new [Action] that is always applied (precondition is set to constant true).
+       */
+      operator fun <T> invoke(f: (T, RandomSource) -> Unit): Action<T> {
```
myuwono, line 32: I wonder if we can instead use a `GenerateArbitraryBuilderContext` on this so we actually discourage users from calling `.next(rs)`.. To my experience that habit can be detrimental to the project health, especially with a team of many.. That is because the temptation of just calling `.next()` without random seed is real, and can easily creep. I'm currently thinking of writing a ktlint rule to ban the use of `.next()` and `.single()` throughout our (in my company, not kotest) codebase.

my suggestion would be:
```kotlin
interface Action<T> {
   // etc...
   companion object {
      operator fun <T> invoke(f: suspend ActionBuilderContext.(T) -> Unit): Action<T> {
         return object : Action<T> {
            override suspend fun precondition(state: T): Boolean = true
            override suspend fun run(state: T, rs: RandomSource) {
               generateArbitrary { 
                  with(DefaultActionBuilderContext(this)) {
                     f(state)
                  }
               }.generate(rs).first()
            }
         }
      }
   }
}


interface ActionBuilderContext {
   suspend fun <T> Arb<T>.bind(): T
   suspend fun randomSource(): RandomSource
}

private class DefaultActionBuilderContext(
   private val generateArbitraryBuilderContext: GenerateArbitraryBuilderContext
) : ActionBuilderContext {
   override suspend fun <T> Arb<T>.bind(): T = with(generateArbitraryBuilderContext) {
      this@bind.bind()
   }

   override suspend fun randomSource(): RandomSource = with(generateArbitraryBuilderContext) {
      arbitrary { it }.bind()
   }
}

```

in call site that will look like:
```kotlin
val add: Action<ViewerCache> = Action { state ->
   val userId = Arb.long(1L..100L).bind()
   val viewerId = Arb.long(1L..100L).bind()
   state.add(userId, viewerId)
   state.get(userId).contains(viewerId) shouldBe true
}
```

kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
sksamuel: You just provide the actions, and Kotest randomly makes a sequence of them. That's the point here, you might get ABCDBDBCBCAAABCBABADBDABDADA so it's like property testing but with actions/state transitions.

kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
sksamuel: I looked at using an ArbitraryBuilderContext but won't it capture the continuation so we can't use suspend functions inside the action? So it means we can't have an action to delay say, useful for testing cache expiry.

myuwono: There's another one Sam, `generateArbitrary { }` the suspend version

sksamuel: Can I use delay etc inside that one ?

sksamuel: Can I use delay etc inside that one ?

myuwono: `generateArbitrary` is not using restricted continuation and is designed to be used in conjunction with suspend functions in mind. we have several of those which setup dynamodb state etc in our codebase with arbitrary data.

myuwono: and yes of course, you can use delay 💯

kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
myuwono: right, say we have `checkState(in, ..., action1, action2, action3)` what do I expect at the end of each test? As a user, do I:
- have only one of the action executed, or
- have these actions executed in sequence, or
- have these actions executed in parallel, or
- have these actions executed in random sequence

kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
sksamuel: Oh awesome, then that's the one to use, I'll update.

kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
sksamuel: executed in a random sequence of `iterations`.

myuwono: right, ok - yeah that makes sense. I guess in future improvements, we can have some sort of setting which allow user to select the behaviours, maybe using an enumerable, perhaps something like:
```kotlin
enum class ActionExecutionMode { Random, Sequential, Parallel }
```

We do have a couple of tests for race-conditions to simulate a parallel processing of many out of order events.

myuwono: Having said that, shouldn't this be `actions.random(rs.random)`?

kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
```diff
@@ -0,0 +1,54 @@
...
+}
+
+suspend fun <T> checkState(input: T, iterations: Int, vararg actions: Action<T>) {
+   checkState(input, iterations, RandomSource.default(), *actions)
+}
+
+suspend fun <T> checkState(input: T, iterations: Int, rs: RandomSource, vararg actions: Action<T>) {
+   repeat(iterations) {
+      val action = actions.random()
+      if (action.precondition(input)) action.run(input, rs)
```
myuwono, line 52: maybe we need to handle failures in here i.e.  printing out the inputs, initial state of the system, the name/alias of the action being performed (currently an action doesn't have a name, but maybe it's useful to have one), as well as the random seed?

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/state.kt
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * A cache that tracks how often a user's profile has been viewed.
+ *
+ * Each time [add] is invoked, that user is added to the list of viewers, and when
+ * [remove] is invoked, that user is removed.
+ * [get] returns the viewers for a user.
+ */
+class ViewerCache {
+
+   val map = mutableMapOf<Long, MutableList<Long>>()
```
myuwono, line 20: hmm - i was looking at the type, should this be `MutableMap<Long, MutableSet<Long>>`. Lists allow multiple entry with the same ids. Unless this model is intended, this will make `remove` action test flakey. Maybe I'm missing something

sksamuel: Yeah this was just an example use case for demoing, and in my real world example that inspired it, we allow duplicates.

kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
sksamuel: Good ideas.

kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
sksamuel: I like the enum idea, and yes we should be using rs.random.

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/state.kt
myuwono: or maybe Sam wanted to preserve the order? I guess that makes sense, but maybe we need a data class describing the consistent ordering id, e.g. timestamp if that's the case? e.g. `data class ViewData(val viewedAt: Instant, val userId: Long)`?

kotest-property/src/commonMain/kotlin/io/kotest/property/state/state.kt
myuwono: @sksamuel  i think i've just discovered a pretty bad bug in generateArbitraries 😭 , I'll raise an issue for it. don't use that Sam. We can instead bake it in the ActionContext
```kotlin
interface ActionBuilderContext {
   suspend fun <T> Arb<T>.bind(): T
   suspend fun randomSource(): RandomSource
}

private class DefaultActionBuilderContext(private val rs: RandomSource, private val edgeConfig: EdgeConfig) : ActionBuilderContext {
   override suspend fun <T> Arb<T>.bind(): T = this.generate(rs, edgeConfig).first().value
   override suspend fun randomSource(): RandomSource = rs
}
```

the builder for action can be:
```kotlin
operator fun <T> invoke(f: suspend ActionBuilderContext.(T) -> Unit): Action<T> {
   return object : Action<T> {
      override suspend fun precondition(state: T): Boolean = true
      override suspend fun run(state: T, rs: RandomSource, edgeConfig: EdgeConfig) {
         with(DefaultActionBuilderContext(rs, edgeConfig)) {
            f(state)
         }
      }
   }
}
```

fedefernandez:
@myuwono @sksamuel this would a great feature to have it, are there any intentions to continue? How can I help?

PR #2974, opened 2022-05-16
Array size validation

nor-ek:
JsonArray size validation and tests.


nor-ek:
Done:
- https://github.com/kotest/kotest/issues/2979
- https://github.com/kotest/kotest/issues/2978
- https://github.com/kotest/kotest/issues/2977

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/matchSchema.kt
```diff
@@ -77,13 +77,17 @@ private fun validate(
    fun violation(message: String) =
       listOf(SchemaViolation(currentPath, message))
 
+   fun violation(conditionResult: Boolean, message: String) = if(conditionResult) violation(message) else emptyList()
```
Kantis: It was not immediately clear to me that this was a different `violation` function than the other invocations. I think we should either rename this one, or just inline the conditional
```suggestion
fun violationIf(condition: Boolean, messageFn: () -> String) = if(conditionResult) violation(message) else emptyList()
   
val sizeViolation = violationIf(tree.elements.size < expected.minItems || tree.elements.size > expected.maxItems) {
   "Expected items between ${expected.minItems} and ${expected.maxItems}, but was ${tree.elements.size}"
}
```

or:

```suggestion
val sizeViolation = if (tree.elements.size < expected.minItems || tree.elements.size > expected.maxItems) violation("Expected items between ${expected.minItems} and ${expected.maxItems}, but was ${tree.elements.size}") else emptyList()
```

Kantis (commented):
Great work!! 😄 Could you please update the docs as well?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/matchSchema.kt
no-u42: @Kantis reason why I did it is because when we deal with more violations (in future) then we will have huge if block `if ... else emptyList` so I tried to do something with it. Another thought was to implement array violations in separate file. 

But your idea to rename this is totally fine !

Kantis: Ah.. I understand and agree that it would be a good idea to break it down a bit somewhere down the line :)

nor-ek:
@Kantis yes of course. Could you tell me where it fit best? I don't see any array validation section in `schema.md` (remember there will be more in future). Should this be an table with external links?

Kantis:
Ah crap.. I thought I added a per-schema-element overview of what you could build into the schema... 🤡 

I think we can skip the external links if we focus on the parts that we have support for. 

I think we can add it here: https://kotest.io/docs/assertions/json/json-schema-matchers.html#defining-schemas
Perhaps splitting "Defining schemas" into "Parsing schemas" and "Building schemas" headers, and under "Building schemas" we can start documenting the different elements and what features we have support for (Array, Object, String, etc..)

Adding KDoc is also a good idea, it will show up when people check out the APIs through IntelliJ for instance

nor-ek:
@Kantis Okay, check out if it's fine after fixes.

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/builder.kt
```diff
@@ -208,10 +212,14 @@ fun JsonSchema.Builder.obj(dsl: JsonSchema.JsonObjectBuilder.() -> Unit = {}) =
 
 /**
  * Defines a [JsonSchema.JsonArray] node where the elements are of the type provided by [typeBuilder].
+iI * The length of the array can be specified using the [minItems] and [maxItems] keywords.
+ *
+ * @param minItems - minimum array length, default value is 0
+ * @param maxItems - maximum array length, default value is [Int.MAX_VALUE]
```
Kantis:
```suggestion
 * The length of the array can be specified using the [minItems] and [maxItems] keywords.
 *
 * [minItems] - minimum array length, default value is 0
 * [maxItems] - maximum array length, default value is [Int.MAX_VALUE]
```

Kantis: I think the `@param` does nothing for KDoc (please correct me if I'm wrong) :)

Kantis (commented):
Just a small thing, I think this is looking good! 😊

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/builder.kt
no-u42: To be honest I've seen @param usage in `spring-security` project. It's also described here https://kotlinlang.org/docs/kotlin-doc.html#param-name. I also noticed that it's used among a project.

Kantis: Ahh, okay.. sorry :) Would've been fine to keep it then.. Anyway, I'll approve and if you want you can change it back later if you wish

Kantis (approved):

Kantis (approved):

nor-ek:
@Kantis by the way couldn't we use matcher as you suggested in another validation https://github.com/kotest/kotest/issues/2979#issuecomment-1129201285 ? What is a driver for it's usage ?

Kantis:
@nor-ek IMO there's primarily two factors:
1. Parsing of JSON schema (textual style, which I just noticed we are lacking a test for in this PR)
2. Clean implementation. For more complex validations we get a nice separation of concerns by delegating the validation to the matcher.. Size might also be a good candidate. It feels important though that we don't lose validations due to matchers perhaps only testing until first violation.
e.g. testing: 

```kotlin
"[1, 1]" shouldMatchSchema jsonSchema { array(beUnique = true, minItems = 3) { number() } }

// Expect BOTH violations to be reported (actual msg made up, other msg would be fine)
$ => Sequence should be unique, but wasn't
$ => Expected minimum length 3, but was 2
```

nor-ek:
@Kantis I'm not sure what test case is in your mind according to `Parsing of JSON schema (textual style, which I just noticed we are lacking a test for in this PR)`. 

Do you mean we should also test strings, not only numbers? 

I've checked below test case, should add it?
```kotlin
    test("Array size larger than maxItems and wrong type") {
       val array = "[1,2,\"tom\"]"
       val sizeBoundedArray = jsonSchema {
          array(minItems = 0, maxItems = 1) { number() }
       }
       array shouldNotMatchSchema sizeBoundedArray
       shouldFail { array shouldMatchSchema sizeBoundedArray }.message shouldBe """
          $ => Expected items between 0 and 1, but was 3
          $[2] => Expected number, but was string
       """.trimIndent()
    }
```

Kantis:
No, I mean where we try to parse and use a schema containing a minItems validation. Something like:

```kotlin
val sizedSchema = parseSchema("""
{
  "type": "array",
  "minItems": 2
}""")

shouldFail {
  "[1]" shouldMatchSchema sizedSchema
}
```

nor-ek:
@Kantis Done.

nor-ek:
@Kantis Could you tell me about merge policy? This PR is blocking rest of json array validation tasks. However may you know what's wrong with failing github checks?

Kantis:
> @Kantis Could you tell me about merge policy? This PR is blocking rest of json array validation tasks. However may you know what's wrong with failing github checks?

I think if it's approved and builds are  passing then it sohuld be fine to merge :) 

We have some issues with builds atm - same on master.. sorry 😬

nor-ek:
@Kantis Ahh fine I thought it's something wrong with this PR. However will it merge automatically?

Kantis:
@nor-ek please pull latest from master, should fix the build failures

nor-ek:
@Kantis how is it merging? automatically?

Kantis:
Merges now, thanks @nor-ek 🙌

PR #2975, opened 2022-05-17
Make orNull() provide a shrink to null

agile-jordi:
This PR makes orNull() add a shrink to null to whatever shrinks the base Arb already generates. It includes a test as well.

Please let me know if there is anything else I should or could do.

Thank you for your attention!


myuwono (approved):
Thanks @agile-jordi !

PR #2982, opened 2022-05-18
Updated docs on Pitest extension for Maven users

ffeltrinelli:
Following the conversation in #2913 , I'm updating the Pitest extension docs to get more clarity on how it should be configured for Maven users.

This PR is the same as #2955 , to target kotest 5.3 docs.


Kantis:
Thanks! 💯

PR #2987, opened 2022-05-20
JsonArray uniqueItems support

nor-ek:
I was forced to annotate `JsonNode` with `@Serializable`. I'm not sure if I should write custom serializer please check this @Kantis. 

I also wonder about creating custom violation mathod that would support all matcherss. Maybe in another task. 

By the way this PR covers:
- documentation
- tests
- code implementation


nor-ek:
@Kantis this PR is ready too.

Kantis:
Please add a test with `parseSchema` as well :)

nor-ek:
Ahh. But there is an problem since I've added `matcher` to **JsonArray** so I need to test `parseSchema` method for `matcher` instead of `uniqueItems` which is in **JsonSchema.Builder.array.** Is that right? However have you any idea how to pass matcher in json?

Maybe I should write serializer?

Kantis:
Actual JSON scheme should be parsed, so I think you will need to write a (de)serializer for JsonArray to be used instead of the derived serializer. I believe a few other JsonElements have explicit serializers

nor-ek:
@Kantis done

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/nodes.kt
```diff
@@ -1,5 +1,8 @@
 package io.kotest.assertions.json
 
+import kotlinx.serialization.Serializable
+
+@Serializable
```
Kantis, line 5: I believe this shouldn't be required? You can try removing it in next PR perhaps :)

Kantis (approved):

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/nodes.kt
no-u42: Nice catch. I thought I've deleted it

Issue #2993, opened 2022-05-22
unrestricted `suspendArbitraryBuilder` throws when there were suspend functions within the builder lambda


myuwono:
**Which version of Kotest are you using**
5.3.0

## Context
A while back we introduced support to non-restricted suspend function to the arbitrary builders for building arbitraries. Unfortunately, the suspend arbitrary builder throws when there were suspend functions within the builder lambda. For instance this function fails:
```kotlin
context("suspend arbitrary builder with unrestricted continuation") {
   test("should build arb on the parent coroutine context") {
      val arb = withContext(Foo("hello")) {
         generateArbitrary {
            val hello = coroutineContext[Foo]?.value
            val world = arbitrary { "world" }.bind()
            val first = Arb.int(1..10).bind()
            val second = Arb.int(11..20).bind()
            delay(10.milliseconds) 
            withContext(Dispatchers.IO) {
               listOf(hello, world, first, second).joinToString(" ")
            }
         }
      }

      arb.generate(RandomSource.seeded(1234L)).take(4).toList().map { it.value } shouldContainExactly listOf(
         "hello world 2 20",
         "hello world 6 12",
         "hello world 7 19",
         "hello world 9 13"
      )
   }
}
```
This is because the suspensions were called from a blocking context, since both `Arb<A>.sample` and `Arb<A>.edgecase` is non-suspend. 

## Propositions
I observed we have several options to solve this:
1. option 1: make `generateArbitrary` return a new `SuspendArb<A>` which has both sample and edgecase functions as suspend. 
2. option 2: keep the status quo - make `generateArbitrary` return the old `Arb<A>`. Here we need to block the calling thread whenever sample and edgecase were called.
3. option 3: change sample and edgecase functions in `Arb<A>` to suspend - this will need to be done in a non-breaking way / through deprecation cycle.
4. option 4: we remove `generateArbitrary` in favour of the new actions as that natively support suspend.

[...]

PR #2995, opened 2022-05-23
make unrestricted suspend arbitrary builder works with the current Arb model

myuwono:
## In this PR
- use runBlocking to ensure that all suspend functions in `suspendArbitraryBuilder` are ran in a blocking manner to ensure the unrestricted suspend arbitrary builder keep working with the current Arb model.

## Context
This PR implements a stopgap while we are figuring out a way forward for #2993 .. 

The reason why it is necessary to block the thread is because `Arb<A>` has this shape:
```kotlin
abstract class Arb<out A> : Gen<A>() {
   abstract fun edgecase(rs: RandomSource): A?
   abstract fun sample(rs: RandomSource): Sample<A>
}
```
while our builder function requires us to dispatch various suspend functions, as in:
```kotlin
suspend fun suspendArb(): Arb<String> = generateArbitrary {
  suspendFn1()
  suspendFn2()
  suspendFn3()
  "hello world"
}
```
notice since we have many suspend functions, which may hop dispatchers, suspend the coroutines, etc. we have to execute `runBlocking` in order to satisfy the Arb's construct. In a nutshell it's like trying to create an arb from suspend functions. This obviously requires us to block the execution of sampleFn.
```kotlin
class TheArb(
   val sampleFn: suspend () -> Sample<A>, 
   val edgecaseFn: suspend () -> A?
): Arb<String> {
   abstract fun edgecase(rs: RandomSource): A? = runBlocking { edgecaseFn() }
   abstract fun sample(rs: RandomSource): Sample<A> = runBlocking { sampleFn() }
}
```

Obviously this is not great. 

Ultimately I want to have deeper discussions with @sksamuel and the kotest team what's the future of `Arb<T>`. I understand that `edgecase` and `sample` are currently not suspend. `checkAll` is suspend. `forAll` is suspend. Is there anything stopping us to make `Arb<T>`'s sample and edgecase suspend? I have my inclination on making them suspend as it solves many issues, including this and ones that I've highlighted in #2993. I've poured what I've been thinking on why I believe there's merit in redesigning `Arb<T>` and make both of these functions suspend but I'd like to run it past Sam and the rest of the kotest team as that'll need to be quite a major shift.


myuwono:
@sksamuel have you had any chance looking into this? I'm keen to merge this to at least make sure this is fixed for end users, albeit as a stopgap.

sksamuel:
There's nothing stopping us making edgecases and sample suspend other than it being a large breaking change for everyone who has implemented custom arbs without using the builders.

PR #3017, opened 2022-05-25
Reverted MDX version to 1.6.22 and added a build pipeline for docs ch…

Kantis:
…anges


Kantis:
Not sure why all PR-test builds were triggered.. But docs are building again so will proceed with merge

PR #3021, opened 2022-05-28
Enable building native targets for kotest-assertions-json.

charleskorn:
No implementation or documentation changes were required - this was already supported by the current implementation, and the documentation suggested Kotlin/Native was already supported.

Only major change was moving the existing tests into `commonTest` so that they run against all targets.


kotest-assertions/kotest-assertions-json/src/commonTest/kotlin/io.kotest.assertions.json/schema/ArraySchemaTest.kt
```diff
@@ -3,11 +3,10 @@ package io.kotest.assertions.json.schema
 import io.kotest.assertions.shouldFail
 import io.kotest.core.spec.style.FunSpec
 import io.kotest.matchers.shouldBe
-import org.intellij.lang.annotations.Language
 
 class ArraySchemaTest : FunSpec(
    {
-      fun json(@Language("JSON") raw: String) = raw
+      fun json(raw: String) = raw
```
Kantis: This method and it's usage should be completely removed. It's only point was to provide the language injection 😄

kotest-assertions/kotest-assertions-json/src/commonTest/kotlin/io.kotest.assertions.json/schema/ObjectSchemaTest.kt
```diff
@@ -3,12 +3,11 @@ package io.kotest.assertions.json.schema
 import io.kotest.assertions.shouldFail
 import io.kotest.core.spec.style.FunSpec
 import io.kotest.matchers.shouldBe
-import org.intellij.lang.annotations.Language
 
 
 class ObjectSchemaTest : FunSpec(
    {
-      fun json(@Language("JSON") raw: String) = raw
+      fun json(raw: String) = raw
```
Kantis: likewise

kotest-assertions/kotest-assertions-json/src/commonTest/kotlin/io.kotest.assertions.json/schema/ArraySchemaTest.kt
charleskorn: Good catch - I've fixed this in https://github.com/kotest/kotest/pull/3021/commits/c1dbc548621d9f2dc7cfac13fece6275fd08ae0d.

kotest-assertions/kotest-assertions-json/src/commonTest/kotlin/io.kotest.assertions.json/schema/ObjectSchemaTest.kt
charleskorn: Good catch - I've fixed this in https://github.com/kotest/kotest/pull/3021/commits/c1dbc548621d9f2dc7cfac13fece6275fd08ae0d.

PR #3024, opened 2022-05-28
Added gradle convention plugins

Kantis:


buildSrc/src/main/kotlin/kotest-js-conventions.gradle.kts
```diff
@@ -0,0 +1,12 @@
+plugins {
+   id("kotlin-conventions")
+}
+
+kotlin {
+   targets {
+      js(BOTH) {
```
Kantis: We sometimes used to target `js(IR)` and sometimes `js(BOTH)`. Are we ready to go to IR only?

buildSrc/build.gradle.kts
```diff
@@ -1,9 +1,12 @@
...
 repositories {
-   mavenCentral()
+   gradlePluginPortal()
 }
 
-plugins {
-   `kotlin-dsl`
+dependencies {
+   implementation("com.adarshr:gradle-test-logger-plugin:3.2.0")
+   implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.21")
```
Kantis: TODO: Should pull kotlin version from version catalog..

buildSrc/src/main/kotlin/kotest-multiplatform-conventions.gradle.kts
```diff
@@ -0,0 +1,30 @@
...
+      watchosArm32()
+      watchosArm64()
+      watchosX86()
+      watchosX64()
+      watchosSimulatorArm64()
+
+      iosX64()
+      iosArm64()
+      iosArm32()
+      iosSimulatorArm64()
```
Kantis: TODO: Option to configure targets, all multiplatform targets should still be default.

Kantis: Having a hard time figuring this one out.. perhaps hard to solve with buildSrc plugins, might be more suited for a proper internal kotest gradle plugin..

buildSrc/src/main/kotlin/kotest-js-conventions.gradle.kts
sksamuel: I guess once 1.7 comes out we could ditch BOTH ?

sksamuel (approved):

Issue #2977, opened 2022-05-17
Add JsonSchema array support for contains


nor-ek:
According to task comment https://github.com/kotest/kotest/issues/2961#issuecomment-1128122006 it should be possible to do
```
jsonSchema {
  array(contains=String)) { string() } 
}
```
- https://json-schema.org/understanding-json-schema/reference/array.html#contains

nor-ek:
@Kantis How it should behave in such example? 
```kotlin
         val array = "[\"life\", \"universe\", \"everything\", 42]\n"
         val containsStringArray = jsonSchema {
            array(contains = containsSpec { number() }) { string() }
         }
```

Reason why I'm asking is because now we have two `elementTypes` one from containSpec and second one from `expected/ JsonArray`. 
I see two possibilities:
1. Respect `contains.elementType` rules if `ContainsSpec` its not null and don't look on `JsonArray.elementType` 
2. Make `JsonArray.elementType` nullable and allow only one elementType 

I would choose option 2 since it's more like https://json-schema.org/understanding-json-schema/reference/array.html#contains where we can provide contains or type
```json
{
   "type": "array",
   "contains": {
     "type": "number"
   }
}
```

However I see that it's not possible to construct such json with current validations:
```json
["life", "universe", "everything", 42]
```

nor-ek:
@Kantis PR: https://github.com/kotest/kotest/pull/3026. By the way you can assign me.

PR #3026, opened 2022-05-31
Json array contains validation support

nor-ek:
In this PR I've implemented basic json arary contains validation. It's described here https://json-schema.org/understanding-json-schema/reference/array.html#contains.

I also set `elementType` to null since it's not necessary when `contains` is used. 
However I think about adding validation that will check if any `value` was passed. 

Github task: https://github.com/kotest/kotest/issues/2977


Kantis:
@nor-ek I think we should consider that: `array()` is valid as long and will match any array element.

Also, contains and items won't always be primitives. Simply comparing types is not sufficient. Example: https://www.jsonschemavalidator.net/s/5mTF2EY9

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/specs.kt
```diff
@@ -0,0 +1,5 @@
+import io.kotest.assertions.json.schema.JsonSchemaElement
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class ContainsSpec(val schema: JsonSchemaElement)
```
Kantis: Please add a package directive

nor-ek:
@Kantis Please check out if its fine. I've done:
- contains serialzier
- improved matchSchema - now it validates objects inside contains spec instead only types
- increased tests quantity

sksamuel:
@Kantis feel free to merge once you're happy with this.

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/matchSchema.kt
```diff
@@ -84,6 +85,18 @@ private fun validate(
       violationIf(!matcherResult.passed(), matcherResult.failureMessage())
    } ?: emptyList()
 
+   fun ContainsSpec.violation(tree: JsonNode.ArrayNode): List<SchemaViolation> {
+      val foundElements = tree.elements.mapIndexed { i, node ->
+         validate("$currentPath.contains[$i]", node, schema).isEmpty()
+      }.count { it }
+      return violationIf(foundElements == 0, "Expected any item of type ${schema.typeName()}")
```
Kantis: If the `contains` specifies an object or array, there might be items in the array which are of the correct type but fails to validate due to failing other constraints. Is there any way we can get a better message here? Perhaps from the recursive validation?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/matchSchema.kt
```diff
@@ -92,9 +105,10 @@ private fun validate(
                "Expected items between ${expected.minItems} and ${expected.maxItems}, but was ${tree.elements.size}"
             )
             val matcherViolation = violation(expected.matcher, tree.elements.asSequence())
-            matcherViolation + sizeViolation + tree.elements.flatMapIndexed { i, node ->
-               validate("$currentPath[$i]", node, expected.elementType)
-            }
+            matcherViolation + sizeViolation +
+               (expected.elementType?.violation(tree)
+                  ?: expected.contains?.violation(tree)
```
Kantis: I think JSON schema definition  allows both elementType and contains to be specified in conjunction? For instance, the elementType might be object, and contains might be object with a certain property. Both can be true at the same time I think :)

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/parse.kt
```diff
@@ -73,26 +70,28 @@ internal object JsonSchemaArraySerializer : KSerializer<JsonSchema.JsonArray> {
    override fun deserialize(decoder: Decoder): JsonSchema.JsonArray =
       decoder.decodeStructure(descriptor) {
          var matcher: Matcher<Sequence<JsonNode>>? = null
-         var minItems = 1
-         var maxItems = Int.MAX_VALUE
-         val elementType = decodeSerializableElement(descriptor, 4, SchemaDeserializer, null)
+         val minItems = kotlin.runCatching { decodeIntElement(descriptor, 1) }.getOrDefault(1)
```
Kantis:
```suggestion
         val minItems = runCatching { decodeIntElement(descriptor, 1) }.getOrDefault(1)
```
kotlin prefix shouldnt be needed

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/specs.kt
```diff
@@ -0,0 +1,33 @@
+import io.kotest.assertions.json.schema.JsonSchemaElement
```
Kantis, line 3: Please specify the package for the file

kotest-assertions/kotest-assertions-json/src/commonTest/kotlin/io.kotest.assertions.json/schema/ArraySchemaTest.kt
```diff
@@ -100,5 +100,54 @@ class ArraySchemaTest : FunSpec(
             $ => Sequence should be Unique
          """.trimIndent()
       }
+
```
Kantis, line 105: I would like to add a test with a more complex structure. Preferably with an array with objects with a contains-spec that fails all objects in the array, so we can see that the error message provided is useful

nor-ek:
@Kantis you can check it now

nor-ek:
@Kantis

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/matchSchema.kt
```diff
@@ -84,6 +85,24 @@ private fun validate(
...
+   fun ContainsSpec.violation(tree: JsonNode.ArrayNode): List<SchemaViolation> {
+      val schemaViolations = tree.elements.mapIndexed { i, node ->
+         validate("$currentPath.contains[$i]", node, schema)
+      }
+      val foundElements = schemaViolations.count { it.isEmpty() }
+      return when {
+         foundElements != 0 -> emptyList()
+         foundElements == 0 -> violation("Expected any item of type ${schema.typeName()}") + schemaViolations.flatten()
+         schemaViolations.isNotEmpty() -> schemaViolations.flatten()
+         else -> emptyList()
```
Kantis: These two branches should be impossible to reach, since `foundElements` must either be `== 0` or `!= 0`? Can we remove the branches?

kotest-assertions/kotest-assertions-json/src/commonTest/kotlin/io.kotest.assertions.json/schema/ArraySchemaTest.kt
```diff
@@ -102,5 +102,81 @@ class ArraySchemaTest : FunSpec(
...
+      test("Array not contains string") {
+         val array = "[1,1]"
+         val containsStringArray = jsonSchema {
+            array(contains = containsSpec { string() })
+         }
+         "[\"bob\"]" shouldMatchSchema  containsStringArray
+         shouldFail { array shouldMatchSchema containsStringArray }.message shouldBe """
+            $ => Expected any item of type string
+            $.contains[0] => Expected string, but was number
+            $.contains[1] => Expected string, but was number
```
Kantis:
```suggestion
            $ => Expected some item to match contains-specification:
                $[0] => Expected string, but was number
                $[1] => Expected string, but was number
```

Perhaps, to make it a bit more distinguished in case we get it in an object with other properties as well. What do you think?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/specs.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlinx.serialization.encoding.decodeStructure
+
+@ExperimentalKotest
+class ContainsSpec(val schema: JsonSchemaElement)
+
+@ExperimentalKotest
+object ContainsSpecSerializer : KSerializer<ContainsSpec> {
+   override val descriptor: SerialDescriptor = buildClassSerialDescriptor("ContainsSpec") {
+      element<String>("type")
+      element<String>("elementType")
```
Kantis: I don't think there should be an element called `elementType` in a Contains spec. See docs [here](https://json-schema.org/understanding-json-schema/reference/array.html#contains) 

I think we could remove it and just use `type`?

Kantis (changes requested):
Sorry that it's  taken a while @nor-ek. Feel free to join the kotlin slack and poke me there, I'm typically more responsive there 🙂

I think we're getting close here. I had some small comments that I think we should address before merging.

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/specs.kt
no-u42: Yeah, I'm facing an issue with parsing this when there is only "type" will try to fix that.

nor-ek:
@Kantis done, you can checkout changes

nor-ek:
btw @Kantis can you send me an invitation to this slack?

Kantis (approved):

Issue #3028, opened 2022-05-31
Nesting `When` under `When` in `BehaviorSpec` results in silently ignored tests


andrey-bolduzev:
As illustrated in the [sample project](https://github.com/andrey-bolduzev/kotest-when-when/blob/main/src/test/kotlin/demo/DemoTest.kt), doing something along the lines of:

```kotlin
class DemoTest : BehaviorSpec({

    Given("given") {
        When("top when") {
            When("bottom when") {
                Then("assertion") {
                    1 shouldBe 2
                }
            }
        }
    }
})
```

and the isolation mode set to `IsolationMode.InstancePerLeaf` results in all tests passing without execution.

If the isolation mode is left at the default, the outer `When` is ignored as not containing any tests/leafs, so also not great.

If I understand correctly, `When`s are not supposed to be nested under `When`s but it's something that is very easy to miss when writing or reviewing code, and you end up with silently broken test code.

Is it possible to mitigate this in any way?

**Which version of Kotest are you using**

5.3.0

Kantis:
I think we can solve it by adding a `@DslMarker`. I can have a look :)

Kantis:
So all spec scopes already have a `@DslMarker` in the form of [@KotestTestScope](https://github.com/kotest/kotest/blob/9fcf94066f462e9337db1e8153f39106edde277e/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/KotestTestScope.kt), see [BehaviorSpecGivenContainerScope](https://github.com/kotest/kotest/blob/master/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/BehaviorSpecGivenContainerScope.kt#L30) for instance.

I checked out your repro (thanks a lot for providing!) and can confirm the problem exists outside the Kotest repo. I wonder if the retention type is insufficient to propagate to consumer sites? Trying to do a local build with altered retention type to see if it makes a difference..

andrey-bolduzev:
Interesting. Just looking around other dependencies in a project, any marker annotation annotated with `@DslMarker` I can see uses the default retention (`RUNTIME`).

Kantis:
I tried it out with every possible retention type, and still no luck.. Maybe something to do with how our stuff is packaged? Ktor also uses `@DslMarker` in a Kotlin multiplatform project and it's working there.. 🤔

sksamuel:
[...]

PR #3038, opened 2022-06-09
Set runtime retention on dsl marker

sksamuel:
Fix for #3028 hopefully.
@Kantis


Kantis:
@sam I tried this out by building locally and it didn't matter what retention was set. I'm starting to think that maybe something is wrong in our packaging? We could probably try to compare our setup with Ktor and try to spot differences. I noticed Ktor's DSL markers are propagating properly.

sksamuel:
It definitely had some effect as a test was failing because it was incorrectly nested.
Just have to figure out how to propogate it now.

sksamuel:
Ktor has the same as us, so maybe we're not packaging it properly. I'll do some tests on api vs implementation.

PR #3043, opened 2022-06-10
Fix publish of kotest-bom

Kantis:


kotest-bom/build.gradle.kts
```diff
@@ -4,6 +4,7 @@ plugins {
    signing
 }
 
+group = "io.kotest"
```
Kantis, line 7: I think this was the issue. the rest is just slight refactoring.

sksamuel (approved):

Issue #3047, opened 2022-06-12
Writing failed property test seed uses invalid file-name parameters, causes exception `Illegal char <<> at index`


aSemy:
**Which version of Kotest are you using** 

`5.3.0.1010-SNAPSHOT`

When this test fails, it causes an exception

```none
Illegal char <<> at index 76: FailedSeedTest/I've got illegal file characters <> -- test
```

```kotlin
import io.kotest.core.spec.style.FunSpec
import io.kotest.matchers.ints.shouldBeExactly
import io.kotest.property.Arb
import io.kotest.property.arbitrary.int
import io.kotest.property.checkAll

class FailedSeedTest : FunSpec({
  context("I've got illegal file characters <>") {
    test("test") {
      checkAll(Arb.int(0..1)) { i ->
        i shouldBeExactly 1
      }
    }
  }
})
```

I don't think the [new failed-seed-writer](https://github.com/kotest/kotest/issues/2967) is escaping the names.

https://github.com/kotest/kotest/blob/7bf2a995d186549c4c955f7111055115c9b36a46/kotest-property/src/jvmMain/kotlin/io/kotest/property/seed/seedio.kt#L27-L38

<details>
  <summary>Full stacktrace</summary>
  
```stacktrace

Illegal char <<> at index 76: FailedSeedTest/I've got illegal file characters <> -- test
java.nio.file.InvalidPathException: Illegal char <<> at index 76: FailedSeedTest/I've got illegal file characters <> -- test
	at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
	at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
	at sun.nio.fs.AbstractPath.resolve(AbstractPath.java:53)
	at io.kotest.property.seed.SeedioKt.seedPath(seedio.kt:25)
	at io.kotest.property.seed.SeedioKt.readSeed(seedio.kt:11)
	at io.kotest.property.seed.SeedKt.getFailedSeed(seed.kt:24)
	at io.kotest.property.seed.SeedKt.getFailedSeedIfEnabled(seed.kt:18)
[...]

PR #3049, opened 2022-06-12
Property test seed persisting: catch invalid seedPath exceptions

aSemy:
Prevent errors with invalid paths crashing the tests - Put the `seedPath(path)` call in the try/catch.

This will help, but not resolve, https://github.com/kotest/kotest/issues/3047


sksamuel:
Thank you for this, but I've added another PR that escapes them properly so should work for all.

PR #3050, opened 2022-06-12
Use XDG_CACHE_HOME for seed home

### LINKED ISSUE #3048: Use XDG base directories for writing failed seeds

`5.3.0.1010-SNAPSHOT`

Presently the seeds are written in the subdirectory of the home directory.

https://github.com/kotest/kotest/blob/7bf2a995d186549c4c955f7111055115c9b36a46/kotest-property/src/jvmMain/kotlin/io/kotest/property/seed/seedio.kt#L21-L23

It's a huge bugbear of mine, and others, when the home directory is littered with dot files. [XDG Base directories](https://wiki.archlinux.org/title/XDG_Base_Directory) is a way of avoiding it, but it the libraries have to implement it.

While it can be fixed later, especially because the seeds aren't important, but if the `.kotest` directory does start having important config, fixing it later can be quite ugly. It's nicer to get it right first.

I couldn't find a Kotlin Multiplatform library, but there's [lots of examples in Github](https://github.com/search?q=xdg+language%3Akotlin&type=code), including this snippet from Jetbrains

https://github.com/JetBrains/ideavim/blob/da9b6ea76271a82c4042d30af6f1d42231efdde2/vim-engine/src/main/kotlin/com/maddyhome/idea/vim/vimscript/services/VimRcService.kt#L52-L62

So, in short

* Instead of using `$USER_HOME/.kotest`...
  * if its defined, use `$XDG_CACHE_HOME`, else use `$USER_HOME/.cache`
  * then plus `kotest/seeds`

There are also platform libraries
* JVM https://github.com/harawata/appdirs
* JVM https://github.com/dirs-dev/directories-jvm

sksamuel:
Fix #3048

### DIFF
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/seed/seedio.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/seed/seedio.kt
@@ -19,9 +19,9 @@ actual fun readSeed(path: TestPath): Long? {
    }
 }
 
-fun xdgConfigHomeOrNull(): String? = System.getenv("XDG_CONFIG_HOME")?.takeIf { it.isNotBlank() }
+fun xdgCacheHomeOrNull(): String? = System.getenv("XDG_CACHE_HOME")?.takeIf { it.isNotBlank() }
 fun userHome(): String = System.getProperty("user.home")
-fun configDirectory(): String = xdgConfigHomeOrNull() ?: userHome()
+fun configDirectory(): String = xdgCacheHomeOrNull() ?: userHome()
 fun seedDirectory(): Path = Paths.get(configDirectory()).resolve(".kotest").resolve("seeds")
 
 fun seedPath(path: TestPath): Path {
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/seed/seedio.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/seed/seedio.kt
@@ -19,10 +19,10 @@ actual fun readSeed(path: TestPath): Long? {
    }
 }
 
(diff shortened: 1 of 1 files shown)


kotest-property/src/jvmMain/kotlin/io/kotest/property/seed/seedio.kt
```diff
@@ -19,7 +19,10 @@ actual fun readSeed(path: TestPath): Long? {
    }
 }
 
-fun seedDirectory(): Path = Paths.get(System.getProperty("user.home")).resolve(".kotest").resolve("seeds")
+fun xdgCacheHomeOrNull(): String? = System.getenv("XDG_CACHE_HOME")?.takeIf { it.isNotBlank() }
+fun userHome(): String = System.getProperty("user.home")
+fun configDirectory(): String = xdgCacheHomeOrNull() ?: userHome()
+fun seedDirectory(): Path = Paths.get(configDirectory()).resolve(".kotest").resolve("seeds")
```
aSemy:
```suggestion
internal fun xdgCacheHomeOrNull(): String? = System.getenv("XDG_CACHE_HOME")?.takeIf { it.isNotBlank() }
internal fun userHome(): String = System.getProperty("user.home")
internal fun configDirectory(): String = xdgCacheHomeOrNull() ?: userHome()
internal fun seedDirectory(): Path = Paths.get(configDirectory()).resolve(".kotest").resolve("seeds")
```

One thought: It might be prudent to not expose these publicly - though I haven't checked out the project locally so don't know if 'internal' is the right visibility.

aSemy (approved):
looks good, thanks! 👍

Issue #2960, opened 2022-05-06
Incompatibility with Kotlin-1.7.0-Beta


WToon:
Using the latest version of Kotest (5.3.0) and the Kotlin 1.7.0-Beta, I am encountering the following error:

```
java.lang.IncompatibleClassChangeError: Found interface kotlin.time.TimeMark, but class was expected
	at io.kotest.engine.spec.interceptor.SpecFinishedInterceptor.intercept-0E7RQCE(SpecFinishedInterceptor.kt:37)
	at io.kotest.engine.spec.interceptor.SpecFinishedInterceptor$intercept$1.invokeSuspend(SpecFinishedInterceptor.kt)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
```

The full stacktrace is attached.

[stacktrace.txt](https://github.com/kotest/kotest/files/8641823/stacktrace.txt)

sksamuel:
Sounds like they've changed the time stuff again.

sksamuel:
We'll prob need to wait for 1.7 final and then cut a 5.4 release that targets that.

jandryml:
I had the same issue with Kotest (5.3.0), but with Kotlin version 1.6.21 also I was using kotlinx-datetime on version 0.3.3. Downgrade of kotlinx-datetime to 0.3.2 solved it for me.

WToon:
> I had the same issue with Kotest (5.3.0), but with Kotlin version 1.6.21 also I was using kotlinx-datetime on version 0.3.3. Downgrade of kotlinx-datetime to 0.3.2 solved it for me.

I tried the following to force version 0.3.2:
`
    implementation("org.jetbrains.kotlinx:kotlinx-datetime:0.3.2"){
        force = true
    }
`
but I get the same error.

sksamuel:
Once 1.7 final is out we will release. 
See https://github.com/kotest/kotest/issues/2990

jesperancinha:
Hi, I just came across this same issue @sksamuel . Can you give an estimate as to when this issue might be solved? Thanks!

sksamuel:
We'll do a 5.4 build once 1.7 is fully out.

ismarslomic:
I'm also facing this issue with kotest `v5.3.0` and kotlin `v1.7.0` (which is available as we speak)

sksamuel:
1.7 just got released a few hours ago but we'll wait for the official blog
post.
[...]

PR #3052, opened 2022-06-12
Kotlin 1.7 compat test - replace `measureTime` with manual duration measure

aSemy:
Just a quick stab at fixing the issue here: https://github.com/kotest/kotest/issues/2960

This might help compatibility with 1.6 and 1.7

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -68,4 +68,3 @@ junit-jupiter = ["junit-jupiter-api", "junit-jupiter-engine"]
 gradle-plugin-publish = { id = "com.gradle.plugin-publish", version.ref = "plugin-publish" }
 shadowjar = { id = "com.github.johnrengelman.shadow", version = "7.1.2" }
 kotlin-serialization = { id = "org.jetbrains.kotlin.plugin.serialization", version.ref = "kotlin" }
-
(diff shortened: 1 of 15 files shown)


jschneidereit:
Wow did they get rid of that function?

aSemy:
> Wow did they get rid of that function?

No, they improved it: https://kotlinlang.org/docs/whatsnew17.html#time-marks-based-on-inline-classes-for-default-time-source

The issue is that internally it uses [`ValueTimeMark`](https://github.com/JetBrains/kotlin/blob/bcd8a28d4c77d1652ddcaaf0fdbe3524df2a2f05/libraries/stdlib/src/kotlin/time/TimeSource.kt#L54), which used to be a regular class - now it's a value class.

sksamuel:
We need to do this for all places.
We use monotonic in continually, eventually, test case executor, constraits, until and logger.

sksamuel:
@aSemy You can prob leave it in the tests because the tests will run with the same version of kotlin as we use in kotest.

aSemy:
I tried re-implementing TimeMark. I got it working in Kotlin 1.7, but it's tricky because in Kotlin 1.6 it was an abstract class (now it's an interface)

* 1.6 https://github.com/JetBrains/kotlin/blob/ed3967eb585b63acdc7339726908bc0f2e6d8101/libraries/stdlib/src/kotlin/time/TimeSource.kt#L49
* 1.7 https://github.com/JetBrains/kotlin/blob/e4b2a8121b6a8bf6d2a5a3405053945b1ac9db68/libraries/stdlib/src/kotlin/time/TimeSource.kt#L78

I got up to `logger.kt` and then got a bit stuck, I'm not sure how to handle `writeLog(...)`

https://github.com/kotest/kotest/blob/33735bcd7571a41e4c6f328cac619c1e3563bdf2/kotest-common/src/commonMain/kotlin/io/kotest/mpp/logger.kt#L41

I'll commit what I have so far. If someone wants to take what I have and improve on it - please feel free!

sksamuel:
That log thing is really just for kotest contributors to track bugs. Feel free to change it in any way you want. You can remove the timestamp entirely if you want.

sksamuel:
Will be easier to merge this and continue in kotest repo.

PR #3053, opened 2022-06-12
Added new statistics for property testing

### LINKED ISSUE #2997: Collecting statistics of property tests just like in jqwik or quickcheck
As of 5.3, Kotest has some nice features for analysing the test case distribution of the used arbitraries, namely

- Auto Classifiers for many of its arbitraries that report the statistics for their input values (when `PropTestConfig::outputClassifications` is enabled)
- `checkCoverage` combined with `classify` to test the value distribution of an arbitrary
- anything else I might be missing? Currently the docs are lacking here a bit :wink: 

But I still miss some features that would help exploring the actual distribution (without necessarily writing a coverage test for it). This can be great to get a better confidence in the quality of custom arbitraries

1. percentage of labels put into classify should be reported (regardless of the usage of a checkCoverage closure)

```kotlin
checkAll<Int> { n ->
  classify(n % 2 == 0, "even", "odd")
}
```

would report something like

```
1000 attempts (51% even, 49% odd)
```

The only way I can see to achieve this right now is by printing the classifications by myself:

```kotlin
val cs = checkAll { ... }.classifications()
println(cs) // or some more fancy custom reporter
```

2. report collected values with/without labels, not only directly coming from the input, but also from computed values. jqwik and quickcheck are calling these commands `collect`

```kotlin
checkAll<List<Int>> { xs ->
  collect("list size", xs.size)
  val result = isUnique(xs)
  collect("result", result)
}
```

would report something like (sorted by its percentage descending)

```
list size:
  7% 7
  6% 3
  5% 4
  4% 6
  ...

result:
  94% false
    5% true
```

to some extent I can simulate this by using (or abusing?) `classify(Int, String)` which seems to be used internally by the Auto Classifiers, but

- it is limited to the number of my arbitrary inputs
- does not sort its values by the percentage
- it ruins the default output of the Auto Classifiers

Resources:

- [Statistics in jqwik](https://jqwik.net/docs/current/user-guide.html#collecting-and-reporting-statistics)
- [Analysing the Test Case Distribution in QuickCheck](https://hackage.haskell.org/package/QuickCheck-2.14.2/docs/Test-QuickCheck.html#g:21)

sksamuel:
Closes #2997

### DIFF
--- a/kotest-common/src/commonMain/kotlin/io/kotest/common/testPath.kt
+++ b/kotest-common/src/commonMain/kotlin/io/kotest/common/testPath.kt
@@ -11,3 +11,10 @@ class TestPathContextElement(
 ) : AbstractCoroutineContextElement(Key) {
    companion object Key : CoroutineContext.Key<TestPathContextElement>
 }
+
+@ExperimentalKotest
+class TestNameContextElement(
+   val testName: String,
+) : AbstractCoroutineContextElement(Key) {
+   companion object Key : CoroutineContext.Key<TestNameContextElement>
+}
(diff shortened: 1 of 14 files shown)


sksamuel:
@enolive Adds new `collect` functions, which are then output by default, depending on the statistcs report mode setting in `PropertyTesting`.

myuwono (approved):

myuwono:
This is really cool @sksamuel !!

PR #3063, opened 2022-06-19
Adds deprecation section in changelog

ashishkujoy:


documentation/docs/changelog.md
```diff
@@ -4,7 +4,7 @@ sidebar_label: Changelog
 slug: changelog.html
 ---
 
-## 5.4.0
+## 5.4.0 (Unreleased)
```
Kantis: The 5.4 docs are only available by selecting that WIP-version using the version selector. All documentation for that version has a notice like this:

image

So I would argue that adding _Unreleased_ is a bit redundant, although it probably doesn't hurt either, as long as we remember to remove it when doing the release 🙂

Kantis (approved):

documentation/docs/changelog.md
ashishkujoy: Sure I will remove it, I was not aware of this.

PR #3071, opened 2022-06-22
Fix problem with isolation mode when duplicate names occurr

### LINKED ISSUE #2915: Unexpected execution order and failure with duplicated test names
I tried with Kotest 5.0 and later updates

The issue seems to be related to be triggered by two tests _not on the same level_ having the same name. The following tests should work: each `load()` is properly followed by a `complete()`, and the tests pass on Kotest 4.x. However, on 5.x they fail with two exceptions: one somewhere in JUnit code, the other in the `FakeService`. Looks like either `InstancePerLeaf` is not respected, or the execution order is just messed up because of the duplicated names.

Renaming one of the duplicated test names causes the tests to pass, also removing some `it`s (I think).

This behavior is unexpected, because the "duplicated" test names appear on different levels. My understanding was that the names are considered duplicate on the same tests level only.

Repro:

```kotlin
package com.example
​
import io.kotest.core.TestConfiguration
import io.kotest.core.spec.IsolationMode
import kotlinx.coroutines.CompletableDeferred
import kotlinx.coroutines.Job
import kotlinx.coroutines.cancel
import kotlinx.coroutines.launch
import kotlinx.coroutines.test.TestCoroutineScope
​
fun TestConfiguration.customTestScope() = TestCoroutineScope(Job()).also { testScope ->
    afterSpec {
        testScope.cancel()
        testScope.cleanupTestCoroutines()
    }
}
​
internal class Repro : DescribeSpec() {
​
    override fun isolationMode() = IsolationMode.InstancePerLeaf
​
    private val service = FakeService()
​
    init {
        describe("test") {
            val load: suspend () -> Unit = { runCatching { service.await() } }
​
            describe("foo") {
                customTestScope().launch { load() }
​
                it("a") {}
​
                it("b") {}
​
                describe("THIS TEST NAME IS DUPLICATED") {
                    service.complete()
​
                    it("a") {}
​
                    it("b") {}
                }
​
                describe("bar") {
                    service.complete()
​
                    it("a") {}
​
                    it("b") {}
​
                    describe("baz") {
                        customTestScope().launch { load() }
​
                        it("a") {}
​
                        it("b") {}
​
                        describe("THIS TEST NAME IS DUPLICATED") {
                            service.complete()
​
                            it("a") {}
​
                            it("b") {}
                        }
                    }
                }
            }
        }
    }
}
​
class FakeService {
​
    var deferred: CompletableDeferred<Unit>? = null
​
    suspend fun await() {
        if (deferred != null) error("Can't await without completing first")
​
        val cd = CompletableDeferred<Unit>()
        deferred = cd
​
        val value = runCatching { cd.await() }
        deferred = null
        value.getOrThrow()
    }
​
    fun complete() {
        deferred!!.complete(Unit)
    }
}
```

Kantis:
Fixes #2915

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/descriptors/descriptor.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/descriptors/descriptor.kt
@@ -111,7 +111,7 @@ sealed interface Descriptor {
     */
    fun isAncestorOf(descriptor: Descriptor): Boolean = when (descriptor) {
       is SpecDescriptor -> false // nothing can be an ancestor of a spec
-      is TestDescriptor -> this.id == descriptor.parent.id || isAncestorOf(descriptor.parent)
+      is TestDescriptor -> descriptor.path().value.startsWith(this.path().value)
    }
 
    /**
@@ -141,21 +141,18 @@ sealed interface Descriptor {
    }
 }
 
-fun descriptorId(name: String, parentDescriptor: Descriptor? = null) =
-   DescriptorId(name + parentDescriptor?.hashCode())
-
 data class DescriptorId(
    val value: String,
 )
 
 fun SpecDescriptor.append(name: TestName): TestDescriptor =
-   TestDescriptor(this, descriptorId(name.testName, this))
+   TestDescriptor(this, DescriptorId(name.testName))
 
 fun TestDescriptor.append(name: TestName): TestDescriptor =
    this.append(name.testName)
 
 fun Descriptor.append(name: String): TestDescriptor =
-   TestDescriptor(this, descriptorId(name, this))
+   TestDescriptor(this, DescriptorId(name))
 
 /**
  * Returns the [TestDescriptor] that is the root for this [TestDescriptor].
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/descriptors/descriptor.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/descriptors/descriptor.kt
@@ -141,16 +141,21 @@ sealed interface Descriptor {
    }
 }
 
-data class DescriptorId(val value: String)
+fun descriptorId(name: String, parentDescriptor: Descriptor? = null) =
+   DescriptorId(name + parentDescriptor?.hashCode())
+
+data class DescriptorId(
+   val value: String,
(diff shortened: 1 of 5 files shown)


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/descriptors/descriptor.kt
```diff
@@ -103,15 +103,15 @@ sealed interface Descriptor {
...
+      is TestDescriptor -> this == descriptor.parent
    }
 
    /**
     * Returns true if this descriptor is ancestor (1..nth-parent) of the given [descriptor].
     */
    fun isAncestorOf(descriptor: Descriptor): Boolean = when (descriptor) {
       is SpecDescriptor -> false // nothing can be an ancestor of a spec
-      is TestDescriptor -> this.id == descriptor.parent.id || isAncestorOf(descriptor.parent)
+      is TestDescriptor -> this == descriptor.parent || isAncestorOf(descriptor.parent)
```
sksamuel: could replace this with isParent || isAncestor

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/descriptors/descriptor.kt
```diff
@@ -103,15 +103,15 @@ sealed interface Descriptor {
     */
    fun isParentOf(descriptor: Descriptor): Boolean = when (descriptor) {
       is SpecDescriptor -> false // nothing can be the parent of a spec
-      is TestDescriptor -> this.id == descriptor.parent.id
+      is TestDescriptor -> this == descriptor.parent
```
sksamuel, line 106: This fails if two things on the same path have a parent name that is the same ?
Can we add an explicit test for this to the descriptor tests ?

Kantis: [Tests added](https://github.com/kotest/kotest/pull/3071/commits/b0230f0c43f29169b77bd33535b081e56104ed53)

sksamuel:
Tests failed but otherwise lgtm.

Could add to release branch too for a 5.3.3 at some point

Kantis:
@sksamuel needs an approval for merging :)

What's the plan for 5.4? Anything major we want to get done before pushing it?

sksamuel (approved):

Issue #2969, opened 2022-05-14
Support automatic binding of value classes


Kantis:
See #2937

Should be possible to automatically derive an Arb for a value class if it is possible to generate the wrapped value.

PR #3072, opened 2022-06-22
Allow Arb.bind<T>() to directly bind to all supported types

tbvh:
Fix for https://github.com/kotest/kotest/issues/2969

Note: this change also allows to bind to normal classes, not just data classes.


kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/targetDefaultForClassName.kt
```diff
@@ -51,11 +50,10 @@ fun targetDefaultForType(providedArbs: Map<KClass<*>, Arb<*>> = emptyMap(), type
       type == typeOf<Period>() -> Arb.period()
       type == typeOf<BigDecimal>() -> Arb.bigDecimal()
       type == typeOf<BigInteger>() -> Arb.bigInt(maxNumBits = 256)
-      clazz?.isData == true -> {
```
tbvh, line 54: If type matches are split off and handled first, maybe clazz can be handled with a single null check?

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/merge.kt
```diff
@@ -38,3 +38,23 @@ fun <A, B : A> Arb<A>.merge(other: Gen<B>): Arb<A> = trampoline { sampleA ->
...
+
+/**
+ * Returns a new [Arb] which merges the provided arbs.
+ *
+ * Each sample is generated by randomly selecting one of the provided arbs and returning a sample from it.
+ *
+ * @param arbs the args to merge
+ * @return the merged arg.
+ */
+fun <A : Any> Arb.Companion.merge(arbs: List<Arb<A>>): Arb<A> {
```
myuwono: @tbvh I'm not used to the type signature of this merge function. Normally merge would take exactly one covariant type of the receiver. For it to take a list and without receiver, this may need to be renamed like mergeAll or combineAll. You can also put that together in the same file as Arb.choose and mention that it is an alias of choose. Having the two functions that have the same purpose risks the two diverging without us noticing. Instead, I highly encourage checking out `Arb.choice` or `Arb.choose`. that does this.

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/ReflectiveBindTest.kt
```diff
@@ -123,6 +129,55 @@ class ReflectiveBindTest : StringSpec(
...
+      "Can bind to no-arg constructor classes" {
+         val noArgArb = Arb.bind<NoArgConstructor>()
+         noArgArb.next().shouldBeInstanceOf<NoArgConstructor>()
+      }
+
+      "Can bind to sealed classes" {
+         val shape3dArb = Arb.bind<Shape3d>()
+         shape3dArb.next().shouldBeInstanceOf<Shape3d>()
+
+         val shapes3d = shape3dArb.take(100).toList()
```
myuwono: Do you want to specify a random seed to ensure this test and all others which uses this pattern not flakey?

myuwono:
Looking great! Asking reviews from @Kantis as well

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/merge.kt
tbvh: Thanks, I wasn't aware of that function, I'll refactor to choice. I figured that Arb<T>.merge does the same thing for two arbs, so Arb.merge would be appropriate for a list of Arbs. Seeing that the functionality is already there I will not add unnecessary duplication to the api.

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/jvmbind.kt
```diff
@@ -2,86 +2,110 @@ package io.kotest.property.arbitrary
...
+private fun arbForParameter(providedArbs: Map<KClass<*>, Arb<*>>, className: String?, param: KParameter): Arb<*> {
+   val errorMsg = "Could not locate generator for parameter $className.${param.name}, consider providing an Arb for it."
+   val arb =
+      try {
+         Arb.forType(providedArbs, param.type)
+      } catch (e: IllegalStateException) {
+         throw IllegalStateException(errorMsg, e)
+      }
+   return arb ?: error(errorMsg)
+}
```
Kantis: Could be reduced to this I believe? Unless I miss something in `Arb.forType` that might throw `IllegalStateException`?

```suggestion
private fun arbForParameter(providedArbs: Map<KClass<*>, Arb<*>>, className: String?, param: KParameter): Arb<*> =
   checkNotNull(Arb.forType(providedArbs, param.type)) { "Could not locate generator for parameter $className.${param.name}, consider providing an Arb for it." }
```

Added benefit: Error message is created lazily 🙂

Kantis (approved):
I think this is great, nice work! 😄  added a small comment, please have a look but no major feedback

Btw, could you please update docs in `/documentation/docs/proptest/reflective_arbs.md` also?

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/jvmbind.kt
tbvh: For parameters that are themselves classes `targetDefaultForType` will call `forClassUsingConstructor` recursively. An issue with the primary constructor then throws IllegalsStateException. But since the parameter name is lost in that context I figured it worthwhile to rethrow the exception.

See `ArbDefaultsTest` "throw on complex data class".

I agree this is not pretty though. I would like to see this improved too.

tbvh: I've added a test for a nested parameter exception, and changed the exception message what. 
I hope it makes more sense now.

I suppose even better would be to have one exception with the path to the failing parameter, but that requires to pass along the 'parameter stack' and change the signature of targetDefaultForType etc. I'm not sure if that change is better.

Kantis: Thanks for clarifying!

documentation/docs/proptest/reflective_arbs.md
```diff
@@ -25,16 +28,24 @@ context("All currencies converts to EUR") { // In some spec
     converted.currency shouldBe EUR
   }
 }
+
+context("Converting to a currency and back yields the same amount") { // In some spec
```
sksamuel, line 32: Can we leave in a test for a data class specifically, just in case in the future the way the reflection calls diverges.

sksamuel (approved):

sksamuel:
lgtm

documentation/docs/proptest/reflective_arbs.md
tbvh: Dataclasses are still part of the tests. This is just the example from the documentation.

I added a second example here to illustrate that check<SomeClass>() and check(Arb.bind<SomeClass>()) are both valid.
I pushed a small change to illustrate the different use cases better.

sksamuel:
Great work

PR #3086, opened 2022-07-04
Add suspendable mountable extensions

### LINKED ISSUE #3081: Add suspendable installable extension
https://github.com/kotest/kotest-extensions-arrow/pull/189

sksamuel:
@nomisRev 

This is a suspendable version of mountables. There are some caveats:

You can't invoke from the init block, because that that is not suspendable.
So you are limited to tests or callbacks.

Secondly, I am not happy with the nomenclature because it clashes with existing mountables, so we might want to think of another name.

closes #3081


nomisRev:
@sksamuel this actually doesn't cover the use-case we have in Kotest Arrow.
https://github.com/kotest/kotest-extensions-arrow/pull/189

There it also allows installing for a `Spec`, the initialisation in that case is lazy but it also requires a `suspend fun get` or `operator suspend fun invoke`.

sksamuel:
Yeah but there's no way to use suspend functions inside the spec body, because that's just a class constructor which as you know isn't suspendable.

We could add a lazy install, that runs the function later when there's a coroutine scope available, as part of the spec setup process, but the user would need to be careful not to try using the resource from inside the spec body. I think this is basically what you suggested in one of the comments.

```kotlin
class Foo : FunSpec() {
  init {
     val resource = lazyinstall(r) // wrapped in some kind of lazy container
     // I can't stop you doing resource.get here but an would throw an exception at least
     test("foo) {
        resource.get // now can you get the value out as you know it's been initialized.
     }
  }
}
```

Another alternative is to introduce a function that contains all your tests and that could be suspendable, replacing the init block entirely.

```kotlin
class Foo : FunSpec() {
  override suspend fun tests() {
     val resource = install(r)
     test("foo) {
        resource.dosomething
     }
  }
}
```

But that's a shift from our standard DSL and we'd have to be certain of the benefits before adding it.

If we're going down that route then we can get as clever as we want and add non-class based support.

```kotlin
val propertyTests = describeSpec {
   val resource = install(resource)
   describe("a") {
      it("b") {
        r // use r here
      }
  }  
}
```

sksamuel:
Might be able to go all the way and use kotlin scripts for tests, then this would be all you need for an entire test.

```kotlin
describeSpec {
   val resource = install(r)
   describe("foo") {
      it("bar") {
      }
   }
}
```

nomisRev:
Wow that looks very interesting! :O 
Getting rid of classes, and just exposing top-level (and perhaps compose-able) Specs sounds interesting to me.

On the other hand it also makes me wonder why an `suspend` init block cannot be provided in that case 🤔
If Kotest can support this would it also not be possible to expose a constructor for `StringSpec` that takes a `suspend` lambda, and executes it immediately after calling the regular constructor?

nomisRev:
> Yeah but there's no way to use suspend functions inside the spec body, because that's just a class constructor which as you know isn't suspendable.

What I meant in my other comment is that I think this can be overcome. Some pseudo code since I'm not so familiar with the Kotest codebase (yet).

```kotlin
// Update constructor in non-binary breaking way

class StringSpec(
  body: StringSpec.() -> Unit = {},
  private val init: suspend StringSpec.() -> Unit
) : DslDrivenSpec(), StringSpecRootScope {

  constructor(body: StringSpec.() -> Unit) : this(body, { })
    
   init {
      body()
   }
}
```

```kotlin
// My imaginary test runner
val foundSpecs = scanForSpecs()
foundSpecs.forEach { spec ->
  spec.init()
  spec.beforeSpec()
  spec.runAllTests()
  spec.afterSpec()
}
```

Or am I wildly incorrect here? Reason I bring it up is I do like the `override suspend fun tests()` example you showed above, but I'd prefer not to have the additional nesting 😅
Or loose the support to write, `class GithubEventProcessorSpec : StringSpec({`.

I am also a fan of the `val propertyTests = describeSpec {` but this would be a completely new DSL.
It would be great if we can support people leveraging new stuff in older codebases as well. There are already quite some big test suites out there build with the current DSL style that I would love to also support with this new Resource functionality in Kotest Arrow.

sksamuel:
In your example, you have two lambdas

```kotlin
class StringSpec(
  body: StringSpec.() -> Unit = {},
  private val init: suspend StringSpec.() -> Unit
) : DslDrivenSpec(), StringSpecRootScope {
```

So you'd be providing two?

Also if you just did:

```kotlin
class MyClass : FunSpec() {
  init {
    //
   }
}
```

Then you're still stuck with a non suspendable constructor.

nomisRev:
> So you'd be providing two?

No, to not break binary-compat you'd need to maintain the current `non-suspend` constructor, right?
Since currently `body: StringSpec.() -> Unit = {}` is called from within `init { }` I didn't want to change that because I am not sure what impact that would have if you'd call it in a slightly later step. If that is not necessary, or the impact is non-existent, you could make the primary constructor `suspend` and maintain a non-suspend constructor for binary compat. Perhaps even as `@Deprecated(level = DeprecationLevel.HIDDEN)` so everyone would automatically switch to the `suspend` version without a break in the binary so old code keeps working as well.

Having a constructor with both allows you to choose. I would only be using the `suspend` lambda, and hopefully in a way that I could write:

```kotlin
class MySpec : StringSpec({
   ...
})
```

Personally I never use the syntax below because I dislike the additional indentation it causes, so I always go for the syntax shown above.

```kotlin
class MySpec2 : FunSpec() {
  init {
    //
   }
}
```

sksamuel:
If the inline variant has it's init lambda be Suspendable, it can't be invoked from the constructor init{} block anymore. So it needs to move to the instantiation logic inside Kotest, which may work, but it might be breaking things as it would be invoked at a different point now. Also, if we have two as in your example, you would need to specify a name in the constructor which might be ugly. But let me play about with it, if I can get it work I'll come back.

nomisRev:
Ah right, you need to name the lambda in that case. I was thinking in the case of a trailing lambda, but that doesn't apply here.

> But let me play about with it, if I can get it work I'll come back.

Awesome, if there is anything I can help with to investigate let me know!

sksamuel:
What we can do is make the body function suspendable and run it in a run blocking context.

```kotlin
abstract class FunSpec(body: suspend FunSpec.() -> Unit = {}) : DslDrivenSpec(), FunSpecRootScope {
   init {
      runBlocking {
         body()
      }
   }
}
```

Then this works:

```kotlin
class BarTest : FunSpec({
   delay(1000)
})
```

It means however that we now have functional differences between the above style and:

```kotlin
class FooTest : FunSpec() {
   init {
      delay(1000) // doesn't work.
   }
}
```

We'd have to also be sure that running the constructor in a run blocking block is ok. It feels like it wouldn't break anything because it was blocking before anyway, since it wasn't suspendable.

nomisRev:
@sksamuel is the `Spec` constructor called from a place where we're in `suspend`?

Otherwise it seems that "So it needs to move to the instantiation logic inside Kotest, which may work, but it might be breaking things as it would be invoked at a different point now." would not actually be a breaking change, right?

Well... maybe for this scenario it is breaking 🤔

```kotlin
class MySpec : StringSpec({
  println("Hello StringSpec")
}) {
  init {
    println("Hello MySpec")
  }
}
```

Now this prints:

```text
Hello StringSpec
Hello MySpec
```

but if you delay `body()` being called after construction then that would switch around 🤔
So yes, `runBlocking` would be the only option. Damn :( `runBlocking` always makes me a bit sad :D

sksamuel:
>

sksamuel:
After slack discussion with @nomisRev we're going to add a kind of Lazy install instead. Will do this in a separate PR.

PR #3096, opened 2022-07-06
Fix tolerance matchers for negative values.

### LINKED ISSUE #3095: Tolerance-based matchers don't work with negative values
Hi!

I believe I've run into a bug using the tolerance matchers today when writing a test that had a negative expected value. After some digging, I think I've figured out the cause.

Negative numbers will never match the `x plusOrMinus y.percent` and `x.shouldBeWithinPercentageOf(y, z)` forms.

This is true for each typed implementation: `Int`, `Float`, `Long`, `Double`  

Trivial example (notice how the "lowest" and "highest" values are reversed in the message)

```kotlin
-1.1 shouldBe (-1.0 plusOrMinus 90.percent)
// -1.1 should be equal to -1.0 within tolerance of -0.9 (lowest acceptable value is -0.09999999999999998; highest acceptable value is -1.9)
```

There are tests covering these methods, (see: [ToleranceMatcherTest]), but they appear to be set up incorrectly, as `Arb.sample()` is not called.

Using the latest master commit (56573a7)

[ToleranceMatcherTest]: https://github.com/kotest/kotest/blob/56573a7c300608dd46368167dac31cfa66548011/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/ints/ToleranceMatcherTest.kt#L30-L33

erik:
This PR fixes tolerance matchers for negative values.

Previously, these constructions would fail for any negative `y`, and any numeric `x`:

- `x shouldBe (y plusOrMinus z)`
- `x.shouldBeWithinPercentageOf(y, z)`

There were existing tests for this, but they appear not to have been running. When `Arb.sample()` was called for unmodified existing tests, they consistently failed.

`Long.shouldBeWithinPercentageOf(...)` also was untested, so this PR adds some tests for it.

Fixes: #3095

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/doubles/Tolerance.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/doubles/Tolerance.kt
@@ -5,6 +5,7 @@ import io.kotest.matchers.MatcherResult
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldNot
 import kotlin.math.abs
+import kotlin.math.absoluteValue
 
 /**
  * Creates a matcher for the interval [[this] - [tolerance] , [this] + [tolerance]]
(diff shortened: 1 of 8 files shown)


sksamuel:
awesome thank you. Looks like one test is failing FloatPercentageTest

erik:
Must have gotten lucky when testing locally! 

Changing all the `Arb.bind(...)` calls to `checkAll(...)`, there are actually a few more issues with the matchers and tests around edge case inputs (overflows, NaN, Infinity, etc. don't seem to be handled correctly). I'll fix the cases related to negative numbers at least and open a new ticket for other issues I find.

LeoColman:
Nice work!

PR #3075, opened 2022-06-26
Gradle plugin improvements

charleskorn:
This PR includes a number of related changes:

* it introduces a test for the Gradle plugin that verifies that tests can run successfully for Kotlin/JVM, Kotlin/JS and Kotlin/Native with both Kotlin 1.6.21 and 1.7.0
* it fixes the issue described in #3060 for Kotlin/JS targets
* it adds additional warnings from the Gradle plugin to warn when the project is not configured correctly (these were really helpful when I was trying to get the composite build configured correctly for the test project)
* it changes the Gradle plugin to apply the generic compiler plugin artifacts rather than the JVM-specific ones (this allows Gradle's composite builds feature to work correctly, making debugging the compiler plugins much easier, and has no impact on normal builds)
* it fixes the issue described in #3059, which allows Kotest to be used to test Gradle plugins

However, the Kotlin 1.7.0 test is failing for the JS browser target. I've never used Kotlin/JS, so I'm not quite sure what is going on here - can anyone help me with this? (Maintainers should be able to edit this branch, so please feel free to push up a fix directly)

```
> Task :jsBrowserTest FAILED
Cannot load "sourcemap", it is not registered!
  Perhaps you are missing some plugin?
Server start failed on port 9876: Error: No provider for "framework:mocha"! (Resolving: framework:mocha)
```

PR #3107, opened 2022-07-12
Add test for Gradle plugin and fix issues running tests for native targets on Kotlin 1.7+

charleskorn:
This PR adds the Gradle plugin test that was removed from #3075, which turned into a much bigger ordeal than I expected...

## Native targets on Kotlin 1.7+
Adding the test revealed that tests weren't being run for native targets due to the unification of the compiler plugin ABI in Kotlin 1.7: https://kotlinlang.org/docs/whatsnew17.html#unified-compiler-plugin-abi-with-jvm-and-js-ir-backends.

I've resolved this by bringing the native compiler plugin into the formerly JS-only compiler plugin, and then reduced the duplication this introduced. (It might be easiest to review this by stepping through each commit one-by-one rather than trying to absorb all of the changes in one hit.)

## JS browser target on Kotlin 1.7
The JS browser target test was failing for Kotlin 1.7 since the beginning (ie. since I added it in #3075). Given how many issues this test is uncovering for native targets, I think there's value in adding the test now, but disabling the JS browser target test and coming back to sort out what the issue is for the JS browser target separately.

For reference, this is what the failure for the JS browser target on Kotlin 1.7 looks like:

```
> Task :jsBrowserTest FAILED
Cannot load "sourcemap", it is not registered!
  Perhaps you are missing some plugin?
Server start failed on port 9876: Error: No provider for "framework:mocha"! (Resolving: framework:mocha)
```

## Running the tests on GitHub Actions
I've had to add the Gradle plugin tests as a separate CI job, as it seems as though the existing Linux PR test job was hitting some resource limits and failing unpredictably. Separating the jobs has the additional benefit of bringing the PR workflow more in line with the `master` branch build workflow. (These two workflows could probably be combined into one, but that's a task for another PR.)


sksamuel:
Thanks for putting all this effort in.

I'm not following the changes though. You've replaced with the js/native compiler plugins with an embeddable / legacy plugins right. What's the reason for that?

sksamuel:
And is the transformer just intended to unify some of the duplication between the plugins?

charleskorn:
> I'm not following the changes though. You've replaced with the js/native compiler plugins with an embeddable / legacy plugins right. What's the reason for that?

In Kotlin 1.6 and earlier, the Kotlin compiler for JVM and JS targets used the embeddable compiler and therefore plugins for JVM and JS targets needed to reference the embeddable compiler too. The Kotlin compiler for native targets required plugins to reference a different, non-embeddable compiler. This is why Kotest previously had a JS compiler plugin (that was really a plugin for the embeddable compiler, but only handled JS, not JVM), and a native compiler plugin.

In Kotlin 1.7, JetBrains has unified the compiler plugin ABI to use only the embeddable compiler for all targets (JVM, JS and native): https://kotlinlang.org/docs/whatsnew17.html#unified-compiler-plugin-abi-with-jvm-and-js-ir-backends. This means that in order to continue supporting native targets on Kotlin 1.7 and above, Kotest needs to have a compiler plugin for the embeddable compiler that supports native targets.

The existing JS compiler plugin is a embeddable compiler plugin. So instead of creating another embeddable compiler plugin for native targets, I've renamed the JS compiler plugin and added support for native targets based on what was in the old native plugin. 

However, we need to continue supporting the old native compiler plugin in order to support users of Kotest that are using Kotlin 1.6 or earlier.

This might make things clearer:

**Before this PR:**

| Target | Kotlin 1.6           |                                                | Kotlin 1.7           |                                                                                              |
|--------|----------------------|------------------------------------------------|----------------------|----------------------------------------------------------------------------------------------|
|        | **Kotlin compiler type** | **Kotest plugin name**                             | **Kotlin compiler type** | **Kotest plugin name**                                                                           |
| **JVM**    | Embeddable           | n/a                                            | Embeddable           | n/a                                                                                          |  
| **JS**     | Embeddable           | `kotest-framework-multiplatform-plugin-js`     | Embeddable           | `kotest-framework-multiplatform-plugin-js`                                                   |
| **Native** | Non-embeddable       | `kotest-framework-multiplatform-plugin-native` | Embeddable           | `kotest-framework-multiplatform-plugin-js` (:warning: which doesn't work for native targets) |

**After this PR:**

| Target | Kotlin 1.6           |                                                             | Kotlin 1.7           |                                                             |
|--------|----------------------|-------------------------------------------------------------|----------------------|-------------------------------------------------------------|
|        | **Kotlin compiler type** | **Kotest plugin name**                                          | **Kotlin compiler type** | **Kotest plugin name**                                          |
| **JVM**    | Embeddable           | n/a                                                         | Embeddable           | n/a                                                         |  
| **JS**     | Embeddable           | `kotest-framework-multiplatform-plugin-embeddable-compiler` | Embeddable           | `kotest-framework-multiplatform-plugin-embeddable-compiler` |
| **Native** | Non-embeddable       | `kotest-framework-multiplatform-plugin-legacy-native`       | Embeddable           | `kotest-framework-multiplatform-plugin-embeddable-compiler` |


> And is the transformer just intended to unify some of the duplication between the plugins?

Yep, exactly - almost all of the logic can be shared between both JS and native targets now. There might be an opportunity to reduce some of the remaining differences in the future as well.

sksamuel:
Ok that makes sense thank you.

sksamuel (approved):

charleskorn:
Any chance you could publish a new release with this change @sksamuel?

sksamuel:
Yes I can release 5.4.2.
Is the gradle plugin smart enough to know if you're on 1.6 vs 1.7 and configure accordingly?

charleskorn:
> Yes I can release 5.4.2.

Thanks!

> Is the gradle plugin smart enough to know if you're on 1.6 vs 1.7 and configure accordingly?

This is handled by the Kotlin compiler infrastructure automatically - it will call the appropriate method on the Gradle plugin to get the correct artifact for the type of compiler in use. So the Kotest Gradle plugin doesn't need to know about the differences in behaviour with different Kotlin versions.

PR #3114, opened 2022-07-21
Replace testCoroutineDispatcher with coroutineTestScope in the docs

yongjoon-km:
[This code](https://github.com/kotest/kotest/blob/4790f479720173ce5e9e95d47df3615021d32064/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/UnresolvedTestConfig.kt#L72-L73) said `testCoroutineDispatcher` is deprecated to test coroutines from 5.3. 
So the document seems to be updated to `coroutineTestScope`.


sksamuel:
👍🏼

PR #3115, opened 2022-07-22
Fix issue where compiler plugins aren't published.

charleskorn:
This seems to have been accidentally removed in #3024.

`kotest-framework-standalone` is also not being published - should it be published?


charleskorn:
Looks like the build failed for one target because downloading something from GitHub Actions' cache failed 😢 I'm not able to retry it - could someone with access please restart it?

Kantis:
Thanks for fixing this @charleskorn

PR #3119, opened 2022-07-24
♻️ Refactors Success Matchers to its own file

### LINKED ISSUE #3111: Result matchers should include the failure reason
https://github.com/kotest/kotest/blob/b47f92247e12bdb43617c762950ed1f392e0afe1/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/result/matchers.kt#L62

Currently, a test that tests for

```
a shouldBeSuccess b
```
will fail if a is `Failure` (as expected).

What it won't do is print what was the failure, so there's no way to debug other than, well, debugging.

LeoColman:
Fixes #3111


LeoColman:
Still pending refactor on the Failure side.
Wanting some early opinions on this

sksamuel (approved):
Looks like a nice clean up

LeoColman:
I think this is everything
@sksamuel check by commit, it will be easier to review

sksamuel (approved):
This looks great.

PR #3128, opened 2022-07-28
fix: [kotest-assertions-json] `shouldContainJsonKey` should be true for keys with `null` values

msewell:
We noticed that for a JSON object which contains property with `null` values like this:

```json
{
  "nullable": null
}
```

the `shouldContainJsonKey("nullable")` matcher would throw even though the key was present, which was against our expectations.

The changes in this PR will make `shouldContainJsonKey(:)` not throw for keys with `null` values. **_This is a breaking change ⚠️,_** as tests relying on the previous behavior may break.

### DIFF
--- a/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/keys.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/keys.kt
@@ -25,8 +25,9 @@ fun containJsonKey(path: String) = object : Matcher<String?> {
          else -> if (value.length < 50) value.trim() else value.substring(0, 50).trim() + "..."
       }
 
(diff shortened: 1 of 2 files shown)


sksamuel:
So the question is, should `shouldContainJsonKey` consider a null value as present or not.

msewell:
> So the question is, should `shouldContainJsonKey` consider a null value as present or not.

I think that's the wrong way to think about it, even if it does address the implementation issue. The question is (IMO): should `someJsonString.shouldContainJsonKey(keyName)` _ever_ throw  if `keyName` is present as a key in `someJsonString`? I don't think it should, regardless of the _value_ belonging to that key; it is only the presence of the key itself that should be considered.

sksamuel:
I think that is the right way to think about this. The name being jsonKey
not jsonValue.

Issue #3137, opened 2022-08-04
java.lang.NoSuchMethodError: 'long kotlin.time.Duration$Companion.milliseconds-UwyO8pc(long)'


minzoovv:
I'm encontering the error  while running my kotest

My kotlin version is 1.7.10
I cannot figure out how cause this runtime error

```
'long kotlin.time.Duration$Companion.milliseconds-UwyO8pc(long)'
java.lang.NoSuchMethodError: 'long kotlin.time.Duration$Companion.milliseconds-UwyO8pc(long)'
	at io.kotest.engine.spec.interceptor.SpecFinishedInterceptor.intercept-0E7RQCE(SpecFinishedInterceptor.kt:25)
	at io.kotest.engine.spec.interceptor.SpecFinishedInterceptor$intercept$1.invokeSuspend(SpecFinishedInterceptor.kt)
```

using Kotest Version : 5.4.1
and all 5.X.X version gives this error when I run the kotest test 

Is there any solution to resolve this error?

SonYoonSeok:
I think it's because function Duration.milliseconds is deprecated. I'll modify it, and send PR Thanks.

charleskorn:
Could you please try with Kotest 5.4.2? This issue should be fixed in that version.

Issue #3059, opened 2022-06-14
NoSuchMethodError: kotlin.time.DurationKt.toDuration with version 5.3.1 when testing Gradle plugins


DanySK:
Hi,
I tried to update Kotest to 5.3.1, but I could not in projects testing Gradle plugins, for instance https://github.com/DanySK/Template-for-Gradle-Plugins, on the branch [`renovate/patch-kotest`](https://github.com/DanySK/Template-for-Gradle-Plugins/tree/renovate/patch-kotest).

Here is the build scan of the failing build: https://scans.gradle.com/s/fyvjl7xhjrazm/failure?anchor=e30&expanded-stacktrace=WyIwIl0&focused-exception-line=0-0#1

I guess that there is some feature of a recent version of Kotlin being used by Kotest 5.3.1 and not present in the version of Kotlin that Gradle embeds.

Kantis:
Seems to be caused by 
```

Exception
java.lang.NoSuchMethodError: kotlin.time.DurationKt.toDuration(JLkotlin/time/DurationUnit;)J	
at io.kotest.engine.spec.interceptor.SpecFinishedInterceptor.intercept-0E7RQCE(SpecFinishedInterceptor.kt:22)	
•••
at io.kotest.runner.junit.platform.KotestJunitPlatformTestEngine.execute(KotestJunitPlatformTestEngine.kt:82)	
at io.kotest.runner.junit.platform.KotestJunitPlatformTestEngine.execute(KotestJunitPlatformTestEngine.kt:46)	
•••
```

I guess Gradle 7.4.2 still embeds Kotlin 1.5.31?

DanySK:
This is very much possible indeed

Kantis:
I guess this wasn't working on 5.x.x ? Kotest requires Kotlin 1.6 since 5.0.0 (see https://kotest.io/docs/changelog.html#kotlin-16-is-now-the-minimum-supported-version)

sksamuel:
This is a duplicate of https://github.com/kotest/kotest/issues/2804
It's because gradle embed a kotlin version. It causes lots of issues.

sksamuel:
https://github.com/gradle/gradle/issues/16345

DanySK:
@sksamuel It can't be an exact duplicate, as it works with Kotest 5.3.0: https://github.com/DanySK/Template-for-Gradle-Plugins/blob/master/gradle/libs.versions.toml

charleskorn:
[...]

PR #3138, opened 2022-08-05
Fix issue deprecated method error

SonYoonSeok:


SonYoonSeok:
# Summary
You can see this issue #3137.

There is a deprecated method in `SpecFinishedInterceptor` and it doesn't work.

## Before
```Kotlin
val duration = Duration.milliseconds(timeInMillis() - start)
```

## After
```Kotlin
val duration = (timeInMillis() - start).milliseconds 
```

sksamuel (approved):

Kantis:
I think this was a workaround for supporting Kotlin 1.5. With this change it won't be possible to write and test gradle plugins for version 7.4 and earlier. I think it should be fine since 7.5 is stable now, but it would perhaps be good to add a compatibility matrix to the readme or something

SonYoonSeok:
Thanks for comment @Kantis !!, I know from #3116 the Gradle version has been updated to 7.5. So I think, It will be fine, right?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/interceptor/SpecFinishedInterceptor.kt
```diff
@@ -22,7 +23,7 @@ internal class SpecFinishedInterceptor(private val listener: TestEngineListener)
 
       // We deliberately use the deprecated method here to maintain backwards compatibility with Kotlin 1.5.
       // See https://github.com/kotest/kotest/issues/3059.
```
Kantis, line 25: Comment is no longer relevant and should be removed

Kantis (commented):
The problem wasn't the Gradle version used by Kotest itself.  See https://github.com/kotest/kotest/issues/3059

When writing Gradle plugins you're forced to use Gradle's embedded kotlin version which for < v7.5 was Kotlin 1.5.  By using the deprecated method, we could enable those use cases still. But if this is now blocking users from migrating to Kotlin 1.7.10 I think we should go ahead with this change.

Users writing Gradle plugins could perhaps upgrade to Gradle 7.5 which moves the embedded Kotlin to 1.6 which should work with this change. What do you think @charleskorn ?

charleskorn:
I think #3137 was fixed in 5.4.2 (the issue says they were using 5.4.1). Could you please confirm that you're seeing the issue with 5.4.2?

SonYoonSeok:
The codes related to the error has not been updated in the master branch. I think it will still throw a error @charleskorn

charleskorn:
I'm using Kotest 5.4.2 with Kotlin 1.7.10 with no issues, and I was previously running into this problem. There's also a test in the CI suite that runs Kotest with Kotlin 1.7.

The deprecated method [has not been removed in Kotlin 1.7.10](https://github.com/JetBrains/kotlin/blob/v1.7.10/libraries/stdlib/src/kotlin/time/Duration.kt#L268), but simply marked as deprecated, so there _shouldn't_ be any issues... but I may be wrong...

For this reason, I suspect there's another issue at play in #3137 - are you able to share a sample project that demonstrates the issue?

Of course, if we want to stop supporting Kotlin 1.5 and remove the deprecated method, go ahead, but I'm not sure this is going to fix #3137.

sksamuel:
Kotest 5 is already 1.6 only, https://github.com/kotest/kotest/blob/master/gradle/libs.versions.toml#L12.
But I agree with you @charleskorn that changing this shouldn't fix anything as the method still exists in 1.7

Issue #3146, opened 2022-08-10
Show unsigned numbers with `u` suffix 


Kantis:
Example:

```kotlin
class UnsignedTest: StringSpec({
      @Serializable
      data class Wham(val id: UInt)

      "Bam" {
         val dam = json.decodeFromString<Wham>("{ \"id\": 1 }")
         dam.id shouldBe 1
      }
})
```
fails with:
```
expected:<1> but was:<1>
Expected :1
Actual   :1
```
the u would make it a bit clearer why this fails:

```
expected:<1> but was:<1u>
Expected :1
Actual   :1u
```

PR #3149, opened 2022-08-12
Support printing unsigned integers clearly

yongjoon-km:
Solves #3146 

Unsigned Integers cause confusing error messages when assertion failed.
There are four unsigned types in Kotlin [kotlin docs](https://kotlinlang.org/docs/unsigned-integer-types.html) (UByte, UShort, UInt, ULong).
Since there is no clear representation among four unsigned types, I put `(UnsignedType)` at the end of a printed number.


LeoColman (approved):
Nice one!
Merging when green

PR #3156, opened 2022-08-13
Gradle Multiplatform plugin: set Kotest Compiler dependency version as a constant, configurable in the extension

### LINKED ISSUE #3153: Kotest is not enabled in Gradle Kotlin Multiplatform project: "no explicit dependency version has been provided"
* Kotest 7.4.2
* Kotlin Multiplatform 1.7.10
* Gradle 7.5.1

I'm getting a warning from Kotest in my Kotlin Multiplatform project about the Gradle config

```
...
> Configure project :modules:events-library
Warning: Kotest plugin has been added to project project ':modules:events-library', and the project does contain a Kotest engine dependency, but no explicit dependency version has been provided. Kotest will not be enabled.
Warning: Kotest plugin has been added to project project ':modules:events-library', and the project does contain a Kotest engine dependency, but no explicit dependency version has been provided. Kotest will not be enabled.
Warning: Kotest plugin has been added to project project ':modules:events-library', and the project does contain a Kotest engine dependency, but no explicit dependency version has been provided. Kotest will not be enabled.
Warning: Kotest plugin has been added to project project ':modules:events-library', and the project does contain a Kotest engine dependency, but no explicit dependency version has been provided. Kotest will not be enabled.
Warning: Kotest plugin has been added to project project ':modules:events-library', and the project does contain a Kotest engine dependency, but no explicit dependency version has been provided. Kotest will not be enabled.
Warning: Kotest plugin has been added to project project ':modules:events-library', and the project does contain a Kotest engine dependency, but no explicit dependency version has been provided. Kotest will not be enabled.
...
```

I'm not sure what this means. 

1. I've defined Kotest dependencies in `build.gradle.kts` `commonTest`, but the version is provided by a [Java Platform](https://docs.gradle.org/current/userguide/java_platform_plugin.html) subproject.
2. the Kotest plugin version is applied, but the version is defined in `buildSrc/build.gradle.kts` - I can't set the version again in a subproject
3. the message repeats the word "project": "has been added to _project_ _project_"

```kotlin
// ./modules/events-library/build.gradle.kts
plugins {
  buildsrc.convention.lang.`kotlin-multiplatform`
  id("io.kotest.multiplatform") // version set in buildSrc/build.gradle.kts
  kotlin("plugin.serialization")
  distribution
}

kotlin {
  sourceSets {
    val commonTest by getting {
      dependencies {
        implementation(dependencies.platform(projects.modules.versionsPlatform)) // applies the Kotest BOM

        implementation(kotlin("test"))

        implementation(libs.kotest.core)
        implementation(libs.kotest.datatest)
        implementation(libs.kotest.frameworkEngine)
        implementation(libs.kotest.json)
        implementation(libs.kotest.prop)
      }
    }
  }
}
```

aSemy:
Update: I'm breaking this PR up into smaller PRs

* #3167 
* #3168
* #3163

---

Updates the Multiplatform Gradle plugin to have a `const val` for the compiler version. This is automatically set by a Gradle task. The version can be manually overridden in the Kotest extension.

> Manual version override example: 
> 
> ```kotlin
> //build.gradle.kts
> plugins {
>    id("io.kotest.multiplatform") version "5.4.0"
> }
> 
> kotest {
>   kotestCompilerVersion.set("5.4.2")
> }
> ```

I also added additional tests using Gradle TestKit for this.

This change is required because the method of determining the version from the project dependencies was not robust. It could break when the version wasn't explicitly set (e.g. it's determined by a BOM, or it's a project dependency or a direct jar), and the order of the Gradle plugins could make a difference (e.g. if the Kotest plugin is applied before the Kotlin Mutliplatform plugin, then no dependencies will be caught).

resolves #3153 

I tried really hard to find a way to automatically determine the version from the dependencies, or from the Gradle buildscript plugins classpath, but this was far too difficult, and almost impossible to do in a way that covered all edgecases. Some sort of fallback `const val` would be necessary, and if there's a fallback then we might as well not bother with trying to figure it out in a complicated way.

#### Summary

- move KotestPluginExtension into different file
- move test-project out of resource directory (so IntelliJ doesn't try to resolve the compiled files in test-project/build
- test-project adjustments to improve manually opening it in a different IntelliJ instance
- apply kotlin-dsl plugin (DSL matches the .kts files better, and aligns Gradle/Kotlin versions)
- replace test-prep in Gradle script with running `build -x check`
- enable caching in test project
- fix gradlew script resolution on Windows
- fix test.xml line comparison on Windows


aSemy:
Hi @charleskorn, are you able to take a look? I'd like your thoughts since you recently refactored the plugins.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
```diff
@@ -23,47 +25,28 @@ repositories {
...
-   }
-
-   dependsOn("jar")
-   dependsOn(":kotlinNpmInstall")
-
+tasks.withType<Test>().configureEach {
    useJUnitPlatform()
 
    systemProperty("kotestVersion", Ci.publishVersion)
+   val gradleWrapper = if ("windows" in System.getProperty("os.name").toLowerCase()) {
```
charleskorn, line 57: I would strongly recommend against invoking the Gradle wrapper from the tests, and instead encourage retaining the task dependencies. 

If these tasks are run from within the tests, Gradle may end up running the same tasks at the same time (once from the original Gradle invocation that is running the tests at the same time, and again from the Gradle invocation from inside the tests). This can lead to conflicts, weird errors and all other kinds of pain.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
```diff
@@ -90,3 +73,39 @@ tasks {
...
+            |// Do not edit manually. This file was created by ${this.path}
+            |
+            |package io.kotest.framework.multiplatform.gradle
+            |
+            |const val KOTEST_COMPILER_PLUGIN_VERSION: String = "${Ci.gradleVersion}"
+            |
+         """.trimMargin()
+   inputs.property("kotestConstantsFileContent", kotestConstantsFileContent)
+
+   val kotestConstantsOutputFile = project.layout.projectDirectory.file(
```
charleskorn: Rather than writing this to the main source code directory, it's probably best to write this to a file under the build directory and attach the file to the `main` source set. 

If the `kotestPluginConstants.kt` file is declared as an output file from this task, you could even attach the task's outputs to the source set, which will then ensure that any other task that requires source files will correctly depend on this task, removing the need for the `tasks.withType<KotlinCompile>().configureEach { dependsOn(updateKotestPluginConstants) }` block below.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
```diff
@@ -90,3 +73,39 @@ tasks {
...
+      "src/main/kotlin/io/kotest/framework/multiplatform/gradle/kotestPluginConstants.kt"
+   )
+   outputs.file(kotestConstantsOutputFile)
+
+   doLast {
+      logger.lifecycle("Updating Kotest Gradle plugin constants\n${kotestConstantsFileContent.indexOf("  > ")}\n")
```
charleskorn: Is this logging still required?

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/kotestPluginConstants.kt
```diff
@@ -0,0 +1,5 @@
+// Do not edit manually. This file was created by :kotest-framework:kotest-framework-multiplatform-plugin-gradle:updateKotestPluginConstants
```
charleskorn: This file shouldn't be checked in - instead, we should let Gradle generate it when it's needed.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePluginSpec.kt
```diff
@@ -1,60 +1,74 @@
...
       "1.6.21",
-      "1.7.0"
-   ).forEach { kotlinVersion ->
-      context("when the project targets Kotlin version $kotlinVersion") {
-         val testProjectPath = Paths.get("src", "test", "resources", "test-project").toAbsolutePath()
-         val testReportsDirectory = testProjectPath.resolve("build").resolve("test-results")
+      "1.7.0",
+   )
+
+   beforeAny {
```
charleskorn, line 26: See comment above - I wouldn't recommend doing this.

If we do do this, setting the `kotlinVersion`, `kotestVersion` and `useNewNativeMemoryModel` properties is not needed - these are only needed for the test project.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePluginSpec.kt
```diff
@@ -1,60 +1,74 @@
...
+            val testReportContents: String = Files.readAllBytes(testReportFile).decodeToString()
 
-            testReportContents shouldStartWith """
-               <?xml version="1.0" encoding="UTF-8"?>
-               <testsuite name="TestSpec" tests="2" skipped="0" failures="0" errors="0"
-            """.trimIndent()
+            testReportContents.shouldContainInOrder(
+               """  <?xml version="1.0" encoding="UTF-8"?>                                    """.trim(),
+               """  <testsuite name="TestSpec" tests="2" skipped="0" failures="0" errors="0"  """.trim(),
+            )
```
charleskorn, line 60: Is there a reason for this change?

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestPluginExtensionSpec.kt
```diff
@@ -0,0 +1,198 @@
...
+  }
+}
+""".trimIndent()
+         }
+         `verify Gradle can configure the project`(gradleTest)
+         `verify Kotest compiler version is set`(gradleTest, "1.2.3")
+         `verify Kotest plugin warnings`(gradleTest)
+      }
+
+      context("groovy") {
```
charleskorn, line 58: Do we need to test both Kotlin and Groovy scripts? I wouldn't expect to find any issues with the Groovy test that wouldn't already be uncovered by the Kotlin test.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/test-project/.gitignore
```diff
@@ -0,0 +1,9 @@
+
+# lock files are not important in a test project
+yarn.lock
```
charleskorn, line 3: Is there a reason why we wouldn't keep the lock file? This helps ensure everyone is running the same versions of each dependency.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
```diff
@@ -1,105 +1,93 @@
...
+      )
 
    // This will soon be deprecated and removed, see https://youtrack.jetbrains.com/issue/KT-51301.
    override fun getPluginArtifactForNative(): SubpluginArtifact =
-      SubpluginArtifact(KotestGroupId, KotestNativeArtifactId, version)
+      SubpluginArtifact(KotestGroupId, KotestNativeArtifactId, kotestExtension?.kotestCompilerPluginVersion?.orNull)
 
    override fun isApplicable(kotlinCompilation: KotlinCompilation<*>): Boolean {
+      val project = kotlinCompilation.target.project
+      val kotestExtension = project.extensions.findByType<KotestPluginExtension>()
```
charleskorn: Should we just use the existing `kotestExtension` value rather than retrieving it again here?

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
```diff
@@ -1,105 +1,93 @@
...
+
       return when {
-         // if we can't find a version to use then we won't apply to this module
-         engineDeps().isEmpty() -> {
-            println("Warning: Kotest plugin has been added to project $target, but the project does not contain a Kotest engine dependency. Kotest will not be enabled.")
+         kotestExtension == null                                -> {
+            // this shouldn't happen
```
charleskorn: If this shouldn't happen, should we just throw an exception and give up?

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
```diff
@@ -1,105 +1,93 @@
...
+            logger.warn("Warning: could not find Kotest extension in $project. Kotest will not be enabled.")
             false
          }
-         version == null -> {
-            println("Warning: Kotest plugin has been added to project $target, and the project does contain a Kotest engine dependency, but no explicit dependency version has been provided. Kotest will not be enabled.")
+
+         !kotestExtension.kotestCompilerPluginVersion.isPresent -> {
+            logger.warn("Warning: Kotest plugin has been added to $project, but could not determine Kotest engine version. Kotest will not be enabled.")
```
charleskorn: Should this ever happen? Shouldn't this always be set to `KOTEST_COMPILER_PLUGIN_VERSION` by default?

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
```diff
@@ -1,105 +1,93 @@
...
+   private val logger: Logger = Logging.getLogger(this::class.java)
 
    companion object {
+      const val kotestPluginExtensionName = "kotest"
       const val compilerPluginId = "io.kotest.multiplatform"
       const val KotestGroupId = "io.kotest"
       const val KotestEmbeddableCompilerArtifactId = "kotest-framework-multiplatform-plugin-embeddable-compiler"
       const val KotestNativeArtifactId = "kotest-framework-multiplatform-plugin-legacy-native"
-      const val missingProjectValError = "Project is not initialized"
       const val engineDepPrefix = "kotest-framework-engine"
```
charleskorn: `engineDepPrefix` isn't being used any more, so can be removed.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
```diff
@@ -1,105 +1,93 @@
...
 import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
 import org.jetbrains.kotlin.gradle.plugin.KotlinCompilerPluginSupportPlugin
 import org.jetbrains.kotlin.gradle.plugin.SubpluginArtifact
 import org.jetbrains.kotlin.gradle.plugin.SubpluginOption
 import org.jetbrains.kotlin.gradle.plugin.mpp.AbstractKotlinNativeCompilation
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinJsCompilation
 
-class KotestMultiplatformCompilerGradlePlugin : KotlinCompilerPluginSupportPlugin {
+
+@Suppress("unused")
```
charleskorn: What's unused? Can we remove it?

charleskorn (commented):
Thanks for looking at this @aSemy. Comments below...

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
aSemy: I agree that manually invoking the Gradle wrapper twice is not ideal (TestKit would be better), but I don't see how the PR is different to the current code, could you explain? The Gradle wrapper script is still invoked from within the test the same way. But It's better now though because the test code doesn't need to know about the project structure to discover the Gradle wrapper, and it works on all OSes.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
aSemy: Good points, I'll set up a generated source set 👍

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
aSemy: I think some logging is useful. It will only run when the input/output changes, so it make it obvious when something's happened, and logging the full file makes it easy to search the whole project to find what's going on.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePluginSpec.kt
aSemy: The previous assertion failed on Windows because of line endings differences

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestPluginExtensionSpec.kt
aSemy: Better safe than sorry 🙂  Plus the test setup is there, in case a Groovy test is needed in the future - it will be easier to copy-paste and adjust the existing test. But you're right, I also don't expect any issues.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/test-project/.gitignore
aSemy: Each test run changed the Kotlin version, which changed the lock file, so it was constantly changing. Normally yes, it should be committed.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
aSemy: `kotestCompilerPluginVersion` has a convention, but a user can override that to be null.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
aSemy: The `@Suppress("unused")` was already there, but at the top of the file. I moved it so it was easier to see.

The plugin class itself `KotestMultiplatformCompilerGradlePlugin` is unused. I might write a simple test so this can go...

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
aSemy: Possibly... I'm not sure what is best practice when it comes to Kotlin compiler plugins. I made the assumption that it's better to try and use the Gradle API as much as possible.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
aSemy: If there's a bug (perhaps the Kotlin compiler plugin is not created in the correct order, or the Kotest plugin isn't applied correctly), then personally I'd rather have a warning message so I can continue without Kotest tests than an exception that completely blocks all development.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
aSemy: I've set up a separate source directory. I'm not sure how to attach the task output to the source set, could you explain or show some code?

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePluginSpec.kt
aSemy: The properties are required because test-project has no default values for them, so Gradle will be unable to configure the project.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
aSemy: I just realised I had put `indexOf` instead of `prependIndex` - it's now fixed

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
charleskorn: Ah, I think I attached my comment to the wrong part of the diff sorry. Passing through the path to the Gradle wrapper is fine - my comment was about the removal of all of the task dependencies for the test tasks and then running `./gradlew build -x check` in the tests.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
charleskorn: I can't find a good example unfortunately, but I believe the way to do it is to call [`srcDir`](https://docs.gradle.org/current/javadoc/org/gradle/api/file/SourceDirectorySet.html#srcDir-java.lang.Object-) on the `main` source set, passing the `Task` (or `Provider<Task>`).

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePluginSpec.kt
charleskorn: Ah, I misunderstood which project this was referring to, so, yes, you're correct, the properties are required. 

I would still not recommend doing this however (see comment on `build.gradle.kts` above).

kotest-framework/kotest-framework-multiplatform-plugin-gradle/test-project/.gitignore
charleskorn: Ah, I see, makes sense.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
charleskorn: I'm not sure I agree - if there's something wrong setting up Kotest, I'd want it to fail early and loud, otherwise there's a risk Gradle is silently not running the tests and the build is passing when in fact there are failing tests.

sksamuel: Yeah I think we'd want to fail, the user can remove the kotest plugin if there's no workaround for them.

sksamuel:
Before I review this myself, is the version now required to be set ?

```
kotest {
  kotestCompilerVersion.set("5.4.2")
}
```

Or is this just an optional way of specifying it ?

sksamuel:
I also wouldn't call the field `kotestCompilerVersion` since that's not clear to me what it's setting. Is it the compiler plugin version?

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
aSemy: Well, depending on the output of other projects by referencing them isn't best practice either :)

> Don’t reference other project tasks directly
> https://docs.gradle.org/current/userguide/cross_project_publications.html

I've worked on a way to make the Gradle TestKit work well, I think this would be the best way:  https://github.com/kotest/kotest/pull/3163

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
aSemy: Good point, done!

aSemy:
> Before I review this myself, is the version now required to be set ?
> 
> ```
> kotest {
>   kotestCompilerVersion.set("5.4.2")
> }
> ```
> 
> Or is this just an optional way of specifying it ?

No it's not required, it's an override for the default.

A default value is set when `KotestPluginExtension` is created, and that default value is `KOTEST_COMPILER_PLUGIN_VERSION`

https://github.com/aSemy/kotest/blob/26b3cb4df79d04ff69943ed1e3175d0224d24bbd/kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt#L46-L50


However a user can override this convention by manually specifying a version in `build.gradle.kts`

```kotlin
//build.gradle.kts
plugins {
   id("org.jetbrains.kotlin.multiplatform") version "5.4.2" // this contains a default value of "5.4.2"
}

kotest {
  kotestCompilerVersion.set("5.4.0") // the user overrides it to be "5.4.0"
}
```

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestPluginExtension.kt
```diff
@@ -0,0 +1,13 @@
...
+
+
+abstract class KotestPluginExtension {
+   /**
+    * The version to use for the Kotest compiler plugins.
+    *
+    * Defaults to [KOTEST_COMPILER_PLUGIN_VERSION].
+    */
+   abstract val kotestCompilerPluginVersion: Property<String>
+}
```
aSemy: Continuing @sksamuel's [comment...](https://github.com/kotest/kotest/pull/3156#issuecomment-1216493957)

> I also wouldn't call the field `kotestCompilerVersion` since that's not clear to me what it's setting. Is it the compiler plugin version?

Yeah, I'm happy to rename it. It's the version of either 
*  [`io.kotest:kotest-framework-multiplatform-plugin-embeddable-compiler`](https://search.maven.org/artifact/io.kotest/kotest-framework-multiplatform-plugin-embeddable-compiler)
* [`io.kotest:kotest-framework-multiplatform-plugin-legacy-native`](https://search.maven.org/artifact/io.kotest/kotest-framework-multiplatform-plugin-legacy-native)

Perhaps it makes sense to make these both separate properties. The legacy one can be marked as deprecated?

```suggestion
abstract class KotestPluginExtension {
  /** <insert docs here>, defaults to [KOTEST_COMPILER_PLUGIN_VERSION]. */ 
  abstract val kotestMultiplatformPluginEmbeddableCompilerVersion Property<String>
  /** <insert docs here>, defaults to [KOTEST_COMPILER_PLUGIN_VERSION]. */ 
  abstract val kotestMultiplatformPluginLegacyNativeVersion Property<String>
}
```

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
aSemy: Alright, I think I've fixed this!

This was the best example I found: https://docs.gradle.org/current/userguide/declaring_dependencies_between_subprojects.html#sec:depending_on_output_of_another_project

I also *finally* found a good use for the `resources.text.fromString(...)` Gradle util 🎉

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
aSemy: Okay yeah the logging got old quickly (plus it happens more regularly now the file isn't checked into source). I changed it to be a debug.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePluginSpec.kt
aSemy: 👍 Sure thing. I'll resolve this thread and we can discuss how to run the tests in the other thread.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestPluginExtension.kt
aSemy: I renamed it to `kotestEmbeddableCompilerVersion` to match the artifact name, `kotest-framework-multiplatform-plugin-embeddable-compiler`. 

I didn't add an additional property for the legacy compiler. It would be nice, but since it's slated for removal it's probably best to keep things slimmer. I'm pasting it here as a demo/in case it is useful.


```kotlin
   /**
    * The version of the **legacy** Kotest compiler plugin. This is only relevant for projects
    * that use Kotlin 1.6 or earlier.
    *
    * Defaults to [KOTEST_EMBEDDABLE_COMPILER_VERSION].
    */
   @Deprecated("Scheduled for removal https://youtrack.jetbrains.com/issue/KT-51301, update to Kotlin 1.7+")
   abstract val kotestLegacyEmbeddableCompilerVersion: Property<String>
```

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
charleskorn: > Well, depending on the output of other projects by referencing them isn't best practice either :)

Totally :) This was the best way I could find to ensure that Gradle wasn't building Kotest in two processes at once (once in the Gradle instance running the plugin tests, and another time as a result of the included build into the dummy Gradle project used by the tests). If there's a better and more reliable way of doing this, then go for it.

> I've worked on a way to make the Gradle TestKit work well, I think this would be the best way: https://github.com/kotest/kotest/pull/3163

Awesome - using TestKit wasn't an option when I added the test originally due to the issue described in #3059 (Gradle embedded an old version of Kotlin, which caused problems running the tests).

Can this PR (#3156) be merged independently of #3163? Is there some way to break these changes up to make them smaller and easier to review? Both of these PRs are quite large.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestPluginExtension.kt
sksamuel: SInce we're moving to one, I think kotestCompilerPluginVersion is the best name.

aSemy: Done 👍

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
aSemy: I've broken down this PR into smaller ones - see the main comment. 

> Awesome - using TestKit wasn't an option when I added the test originally due to the issue described in https://github.com/kotest/kotest/issues/3059 (Gradle embedded an old version of Kotlin, which caused problems running the tests).

Yeah, the embedded Kotlin version is a pain to work around. It's frustrating that the Gradle team is so slow to update it. Although [7.5 does use Kotlin 1.6.21,](https://docs.gradle.org/current/userguide/compatibility.html#kotlin) so that should match Kotest....

Hmm actually I suspect what was wrong is that currently on master branch the compiler plugin sets the Kotlin Kotlin language level to 1.6. Gradle basically requires the language level is 1.4. I added the `kotlin-dsl` plugin, which sets the language level to 1.4, thus resolving the issues.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
aSemy: Resolved in https://github.com/kotest/kotest/pull/3168

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
aSemy: This was committed in https://github.com/kotest/kotest/pull/3168

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
aSemy: Resolved in https://github.com/kotest/kotest/pull/3168

sksamuel:
Should this be re-opened @aSemy

aSemy:
@sksamuel No it's okay, this can be closed. I'll pick up the remaining bit of work (creating automated tests) in #3163 if/when I have time/motivation

sksamuel:
Ok

PR #3159, opened 2022-08-15
Update test-coroutine-dispatcher docs

yongjoon-km:
testCoroutineDispatcher is deprecated from 5.3. [Code](https://github.com/kotest/kotest/blob/a7195a4d1c4a0d9e20526bc336338d394e028003/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/UnresolvedTestConfig.kt#L66-L75)

However, `AbstractProjectConfig` does not have `coroutineTestScope`. So I didn't update that part on the docs.


Kantis:
Since the deprecation is since 5.3, perhaps we should update the 5.3 docs as well? Other than that, looks good to me.

yongjoon-km:
> Since the deprecation is since 5.3, perhaps we should update the 5.3 docs as well? Other than that, looks good to me.

I previously updated 5.3 docs as well. Thank you :)

Kantis (approved):

PR #3160, opened 2022-08-15
Fix invalid example on test_coroutine_dispatcher docs

yongjoon-km:
AbstractProjectConfig does not have coroutineTestScope.


Kantis:
Thanks! Looks like it was correct in 5.2, 5.4 and current docs, but not 5.3. Suggested fix is in place for other versions.

Kantis (approved):

PR #3162, opened 2022-08-15
minor tidy up of buildSrc convention plugins

aSemy:
Some minor config updates

* use Kotlin DSL accessors, like `tasks.javadoc`
* use [Gradle config avoidance API](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html) (replace `forEach {}` with `configureEach {}`
* merge related code blocks (`publications.withType<MavenPublication>().configureEach {}` was in there twice)

### DIFF
--- a/buildSrc/src/main/kotlin/kotest-publishing-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotest-publishing-conventions.gradle.kts
@@ -1,30 +1,18 @@
 plugins {
    signing
-   `java-library`
    `maven-publish`
 }
 
 val javadocJar by tasks.registering(Jar::class) {
    group = JavaBasePlugin.DOCUMENTATION_GROUP
-   description = "Assembles java doc to jar"
+   description = "Empty Javadoc Jar (required by Maven Central)"
    archiveClassifier.set("javadoc")
-   from(tasks.javadoc)
 }
 
 val ossrhUsername: String by project
 val ossrhPassword: String by project
 val signingKey: String? by project
 val signingPassword: String? by project
-
-signing {
-   useGpgCmd()
-   if (signingKey != null && signingPassword != null) {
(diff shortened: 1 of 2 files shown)


Kantis:
I actually think we could remove the javadoc stuff.. It's not used, right @sksamuel ?

aSemy:
Javadoc is required for Maven Central publishing, that might be why?

Kantis:
Ah, okay.. I had a [discussion with Jetbrains](https://kotlinlang.slack.com/archives/C19FD9681/p1656541828249869) about some issues with included builds that they hinted might be caused by the usage of java-library/javadoc plugins. I haven't gotten around to trying out removing them, but this could be another problem we run into then. 😄

aSemy:
Ahh yeah, so it would be best to remove the `java-library` plugin and instead create an empty Javadoc jar. 

https://github.com/adamko-dev/kotlinx-serialization-typescript-generator/blob/2914dd6256600671f37583c5371759cc8ec27393/buildSrc/src/main/kotlin/buildsrc/convention/maven-publish.gradle.kts#L127-L149

While you could add Dokka HTML/Javadoc in the Javadoc jar, I don't really see the point. It really slows down the build, and how many people are going to unzip it and look at the contents, vs looking at the Kotest website? Especially since the source-jar is included, so that includes the documentation.

aSemy:
I removed the `java-library` plugin, updated the JavadocJar task with an explanation, and moved the signing config after the publishing config (to try and avoid this issue https://github.com/gradle/gradle/issues/16961)

sksamuel:
> I actually think we could remove the javadoc stuff.. It's not used, right @sksamuel ?

We need to generate the jar at least for publication.

sksamuel:
> 

As long as we create an empty one it'll work - as maven central insists on one.

sksamuel (approved):

PR #3164, opened 2022-08-17
enable Gradle configuration cache

aSemy:
[Gradle configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html) improves development speed, especially when switching branches.

The cache is local per-project. This feature is incubating, but can be set to only emit warnings if problems are detected, which is what I've done.

The Kotest project doesn't use many Gradle plugins or unusual Gradle config, so I expect that this should be stable. In case there are issues, then remove the `$projectDir/.gradle/configuration-cache` directory.

I've also enabled the recommended preview flag `STABLE_CONFIGURATION_CACHE` https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:stable


Kantis:
Nice! 💯 I have tried out configuration caching in some other projects and it helps a lot. I haven't run into any problems with it for a while now so I think we should merge this 👍🏼

sksamuel:
This is cool. Nice PR.

Kantis (approved):

PR #3165, opened 2022-08-17
Use Gradle type-safe project accessors

aSemy:
This PR enables [Gradle typesafe project accessors](https://docs.gradle.org/7.5.1/userguide/declaring_dependencies.html#sec:type-safe-project-accessors), replacing the manual definitions here 
https://github.com/kotest/kotest/blob/master/buildSrc/src/main/kotlin/Projects.kt

The benefits are that these accessors are automatically created and updated by Gradle. The downsides are that these accessors are incubating, and might not be as stable as the manual definitions - although I've never experienced a problem.

I also did some tidying up of the root `settings.gradle.kts` to make it a little more legible.

(I branched from the wrong branch so this PR includes https://github.com/kotest/kotest/pull/3164, but the changes here don't necessarily depend on it)


buildSrc/src/main/kotlin/kotest-jvm-conventions.gradle.kts
```diff
@@ -20,7 +20,7 @@ kotlin {
    sourceSets {
       val jvmTest by getting {
          dependencies {
-            implementation(project(Projects.JunitRunner))
+            implementation(project(":kotest-runner:kotest-runner-junit5"))
```
aSemy, line 23: The type-safe project accessors aren't available in buildSrc, so the project has to be referenced manually here.

settings.gradle.kts
```diff
@@ -5,95 +5,94 @@ pluginManagement {
    }
 }
 
-val ideaPlatformPrefix: String? = System.getProperty("idea.platform.prefix")
-val isIntelliJ = ideaPlatformPrefix.equals("idea", ignoreCase = true)
```
aSemy, line 9: I don't think these are used? So I removed them

buildSrc/src/main/kotlin/kotest-jvm-conventions.gradle.kts
Kantis: Makes sense. Would it be possible to solve this by moving buildSrc -> build-logic and using it as an included build instead? From what I can tell, that's the recommended way to handle build logic moving forward?

aSemy: Hmm no I don't think so, the generated project accessors are only available within a single Gradle build (i.e. there's a root `settings.gradle.kts`), not across included builds. It might be possible in other ways (moving `:kotest-runner-junit5` to an included build, and doing some dependency substitution, maybe?), but nothing that will be as simple as what's in this PR :)

On buildSrc... Normally I think buildSrc is fine. There's a drawback (every change in buildSrc requires a full rebuild), but that's not really such a big deal. The config in buildSrc doesn't change that much, and even when it does, it's not a heavy hit to re-build.

However, I think setting up an included build for the build plugins would help Kotest. It's a much larger project than most, and there are a lot more modules. Included builds would also help with #3171.

As a replacement for buildSrc, I'd probably set up two included builds. One for 'settings plugins'. These would be for plugins like version catalog, shared repository definitions, version alignment. These plugins would be used in all other projects, including the other included build, which would be more of a direct replacement for buildSrc, where you'd define convention plugins (for Kotlin, publishing, verification plugins, etc). 

And after that, once the ball is rolling, you could even include the Kotest extension projects as an included build. That would help with alignment, and sharing resources - but I'm just throwing out ideas.

sksamuel:
I think this looks good @Kantis

Kantis (approved):

sksamuel:
Just needs a fix for the gradle plugin module then I think we're set.

aSemy:
Done! Just a small typo https://github.com/kotest/kotest/pull/3165/commits/e0085770d89174ae7e1998b99b9732899282fde4

PR #3167, opened 2022-08-17
Kotest Gradle plugin [part 1]: move test-project out of resources directory

aSemy:
Breaking down #3156 into smaller PRs

This change moves the Gradle plugin functional test project out of the test/resources directory. This is important, as including the project in the resources dir means the `build/` and `.gradle/` directories are included, and IntelliJ tries to resolve the compiled files and gets very confused, and will refuse to build the project.

The tests still don't work for me because they require some specific set up that I haven't done (publishing the latest Kotest libs to Maven Local, setting up Kotlin native building (the `jvmOnly` flag isn't taken into account in these tests), but the tests do at least run and fail!


charleskorn:
> The tests still don't work for me because they require some specific set up that I haven't done (publishing the latest Kotest libs to Maven Local, setting up Kotlin native building (the jvmOnly flag isn't taken into account in these tests), but the tests do at least run and fail!

Could you share the command you're running and the output you get? The tests should work fine locally both before and after this change. 

There should be no need to publish anything to a local Maven repo - Gradle should pick everything up through the composite build.

charleskorn (approved):

aSemy:
> Could you share the command you're running and the output you get? The tests should work fine locally both before and after this change.

I'm running `./gradlew :kotest-framework:kotest-framework-multiplatform-plugin-gradle:test -PjvmOnly=true`, so the test setup fails because it depends on tasks that `jvmOnly` disabled 

https://github.com/kotest/kotest/blob/2ba0cf84d0eace0f652258b4b0cae29e1492f125/kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts#L44-L52

I can't build the project without the `jvmOnly` flag as I don't have xcode setup on my Mac, or some other build tool on my PC.

If I disable the specific tasks, then the tests do run, but they fail with Gradle errors saying it can't find the correct versions of dependencies, and it searches in Maven Local.

I'm going to focus on getting TestKit working (#3163) rather than trying to debug these errors though :)

PR #3168, opened 2022-08-17
Kotest Gradle plugin [part 2]: use configurable property for Kotest compiler plugin version

### LINKED ISSUE #3153: Kotest is not enabled in Gradle Kotlin Multiplatform project: "no explicit dependency version has been provided"

aSemy:
(Breaking down https://github.com/kotest/kotest/pull/3156 into smaller PRs)

Depends on https://github.com/kotest/kotest/pull/3167

Add configurable `kotestCompilerPluginVersion` to Kotest extension.

Resolves #3153

More thorough tests are included in #3163

- adjust code to use `kotestCompilerPluginVersion` property
- generate default value for k`otestCompilerPluginVersion` - `KOTEST_COMPILER_PLUGIN_VERSION`
- minor Gradle formatting, configuration avoidance updates
- update dependencies (Gradle dependencies are automatically added by `kotlin-dsl`)
- add basic test (mostly to get rid of the `@Suppress("unused")`)
- move `KotestPluginExtension` to new separate file
- add `kotlin-dsl` plugin (better compatibility with Gradle embedded version, adds Gradle Kotlin DSL library to better match how `gradle.kts` works)

### DIFF
--- a/kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
+++ b/kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
@@ -72,7 +72,7 @@ abstract class KotestMultiplatformCompilerGradlePlugin @Inject constructor(
 
       return when {
          !kotestExtension.kotestCompilerPluginVersion.isPresent -> {
-            logger.warn("Warning: Kotest plugin has been added to $project, but could not determine Kotest engine version. Kotest will not be enabled.")
+            logger.warn("Warning: the Kotest plugin has been added to $project, but kotestCompilerPluginVersion has been set to null. Kotest will not be enabled.")
             false
          }
 
--- a/kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
+++ b/kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
@@ -1,11 +1,13 @@
-@file:Suppress("unused")
-
 package io.kotest.framework.multiplatform.gradle
 
+import javax.inject.Inject
 import org.gradle.api.Project
-import org.gradle.api.artifacts.Dependency
-import org.gradle.api.provider.Property
+import org.gradle.api.logging.Logger
+import org.gradle.api.logging.Logging
 import org.gradle.api.provider.Provider
+import org.gradle.api.provider.ProviderFactory
+import org.gradle.kotlin.dsl.create
+import org.gradle.kotlin.dsl.findByType
 import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
 import org.jetbrains.kotlin.gradle.plugin.KotlinCompilerPluginSupportPlugin
 import org.jetbrains.kotlin.gradle.plugin.SubpluginArtifact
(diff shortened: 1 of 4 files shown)


kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/main/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePlugin.kt
```diff
@@ -1,105 +1,87 @@
...
+      val project = kotlinCompilation.target.project
+      val kotestExtension = project.extensions.findByType<KotestPluginExtension>()
+         ?: error("Could not find Kotest extension in $project")
+
       return when {
-         // if we can't find a version to use then we won't apply to this module
-         engineDeps().isEmpty() -> {
-            println("Warning: Kotest plugin has been added to project $target, but the project does not contain a Kotest engine dependency. Kotest will not be enabled.")
+         !kotestExtension.kotestCompilerPluginVersion.isPresent -> {
+            logger.warn("Warning: Kotest plugin has been added to $project, but could not determine Kotest engine version. Kotest will not be enabled.")
```
charleskorn: For this case, should we explain that the version on the extension has been set to `null`, and that a possible way to fix the problem is to set an explicit version on the `kotest` extension?

aSemy: This error should only happen if the property is explicitly set to something nullable in a `build.gradle.kts`, and the stacktrace would lead here, so I think the wording isn't very important. So if you want to suggest something then I'll update it :)

charleskorn: How about `Warning: the Kotest plugin has been added to $project, but kotestCompilerPluginVersion on the kotest extension has been set to null. Kotest will not be enabled for this project.`?

aSemy:
```suggestion
            logger.warn("Warning: the Kotest plugin has been added to $project, but kotestCompilerPluginVersion has been set to null. Kotest will not be enabled.")
```

charleskorn (approved):

PR #3171, opened 2022-08-18
Publish Gradle version catalog

Kantis:
As @LeoColman suggested. 

_WIP: I haven't gotten the below to work due to a gradle error when importing the catalog_

You can try it out by using `./gradlew publishKotestVersionCatalogPublicationToMavenLocal` to publish the version catalog locally and then adding the following in `/settings.gradle.kts` in the project where you want to use the definitions.

```
dependencyResolutionManagement {
   repositories { mavenLocal() }
   versionCatalogs {
      create("kotest") {
         from("io.kotest:kotest-version-catalog:5.5.0-LOCAL")
      }
   }
}
```

After importing like above, you should be able to add dependencies like: `testImplementation(kotest.runner.junit5)`


aSemy:
Looks cool! Is the intention to publish it to Maven Central, or just keep it locally? If it's locally, then setting it up as an [included build](https://docs.gradle.org/current/userguide/composite_builds.html) might be better, as Gradle will then manage building and sharing new versions.

Kantis:
> Looks cool! Is the intention to publish it to Maven Central, or just keep it locally? If it's locally, then setting it up as an [included build](https://docs.gradle.org/current/userguide/composite_builds.html) might be better, as Gradle will then manage building and sharing new versions.

Thanks! It's intended to be shared on Maven Central :)

build.gradle.kts
```diff
@@ -7,3 +15,63 @@ repositories {
...
+         "runner-junit5",
+         "extensions-htmlreporter", // Extension built and released with main kotest project
+         "extensions-junitxml" // Extension built and released with main kotest project
+      ).forEach { alias ->
+         library(alias, groupId, "kotest-$alias").version(publishVersion)
+      }
+
+      // Extensions
+      listOf(
+         "spring" to "1.1.2",
```
LeoColman: How are we going to keep these versions up to date?

Kantis: Manually, I think. Up for suggestions though :)

If we publish with latest available version every time, then at least we should offer compatible versions. 

Given that extensions can release faster than Kotest itself, users might want to override the versions we specify here using [Gradle's mechanism for that
](https://docs.gradle.org/current/userguide/platforms.html#sec:overwriting-catalog-versions)

LeoColman: I think no way we'll remember to do this manually.
Is there maybe a way to publish the version catalog outside the main pipeline? Otherwise as you said dependencies might get updated before we are able to update the version catalog

Kantis: Sure, it would be possible to externalize the version catalog pipeline. We could have a completely separate repo for the version catalog even. 

When would the version catalog be published, and how would the version of the catalog map to kotest version?

Kantis: @sksamuel what do you think?

LeoColman: Is it possible to maybe update a catalog?

"Hey, for version 5.4.0 of the catalog, now take extension v1.2.1 instead of v1.2.0"

Then we could keep 'updating' the main catalog with the side versions.
No idea on the complexity of this, but knowing if this is possible at all would be nice

Kantis: I think that's forbidden, but I'm having trouble finding docs to confirm it. :(

sksamuel: You can't override a release once it's on maven central

sksamuel:
This is really cool.
I'm not sure about the extensions though. Having them in the main repo will mean they go out of date pretty quickly.

So it either needs to be totally obvious how to override them, or we have another versioning scheme for the catalog which updates more frequently,  or we don't include them.

Kantis:
> This is really cool. I'm not sure about the extensions though. Having them in the main repo will mean they go out of date pretty quickly.

In practice, I get the feeling that most extensions move along slower than the main repo.. I'm not really sure how to quantify it, but it would be interesting to see how often this would've occurred.

One thing that would be hard to workaround though is when extension releases depend on main repo releases.. Like Spring extensions going to 1.1.0 as a result of Kotest going to 5.0.0. That would require synchronizing the release to keep the version catalog in sync for 5.0.0

sksamuel:
Why don't we start without extensions. We can always add them later?

Kantis:
@aSemy @charleskorn anything you'd like to see changed?

build.gradle.kts
```diff
@@ -7,3 +15,42 @@ repositories {
...
+
+catalog {
+   val groupId = project.group.toString()
+   val extensionsGroupId = "$groupId.extensions"
+
+   println("using groupId: $groupId and version: $publishVersion for version catalog")
+
+   versionCatalog {
+      listOf(
+         "assertions-api",
```
charleskorn, line 27: Is there any way to generate this list automatically? From inspecting subprojects perhaps?

build.gradle.kts
LeoColman: But are version catalogues published to the maven central?

LeoColman: I thought that they might be published to some gradle repository

LeoColman:
> Why don't we start without extensions. We can always add them later?
I like this idea a lot.
Seems easier to understand things

LeoColman:
As changes were already made in this direction, I think this PR is good enough to get us started!
LGTM?

LeoColman (approved):

PR #3181, opened 2022-08-27
Add new matchers (shouldCloseTo)

jaehyunup:
Hello bro,

issue - [#3183 ](https://github.com/kotest/kotest/issues/3183)

i request the addition of the "shouldCloseTo" function, inspired by "closeTo" provided by various matcher libraries such as "hamcrest" and "assertj".

The newly added Instant::shouldCloseTo function is to check that value and otherInstant are as close as the developer expected. As well as me, other developers will often use it in situations where they cannot accurately capture the stubbed time.

Although it is a function that is frequently used, I know that kotest does not provide it, so I until used extensions function in my code, but I raise this PR because I think it has sufficient usage and versatility.

In order to reconsider the need for this feature, Instant has implemented it first, but if this PR is Approved and proves necessary, they want to add it to all Datetime forms.

Thank you.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/date/instant.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/date/instant.kt
@@ -44,8 +44,8 @@ fun closeTo(anotherInstant: Instant, point: Long, unit: ChronoUnit) = object : M
       val diff = kotlin.math.abs(value.toEpochMilli() - anotherInstant.toEpochMilli())
       return MatcherResult(
          diff <= Duration.of(point, unit).toMillis(),
-         { "Expected $value to be close to $point${unit.name}, but it's not." },
-         { "$value is not close as $point${unit.name} to $anotherInstant." }
+         { "Expected $value to be close to $anotherInstant in range by $point${unit.name}, but it's not." },
+         { "$value is not close to $anotherInstant in range by $point${unit.name} " }
       )
    }
 }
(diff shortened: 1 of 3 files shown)


kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/date/instant.kt
```diff
@@ -37,6 +39,17 @@ fun between(fromInstant: Instant, toInstant: Instant) = object : Matcher<Instant
    }
 }
 
+fun closeTo(anotherInstant: Instant, point: Long, unit: ChronoUnit) = object : Matcher<Instant> {
```
sksamuel: What do you think about using a duration here?
so,

closeTo(instant, 2.seconds)

jaehyunup: I understand the intention of your review.
If i modify the function based on your example, it will be as follows.

**closeTo(instant, 2L.toDuration(DurationUnit.SECONDS))** 
_cause **"2L.seconds"** style has been Deprecated_

```kotlin
fun closeTo(anotherInstant: Instant, duration: Duration) = object : Matcher<Instant> {
   override fun test(value: Instant): MatcherResult {
      val range = duration.inWholeNanoseconds
      return MatcherResult(
         abs(value.toEpochMilli() - anotherInstant.toEpochMilli()) <= range,
         { "Expected $value to be close to $anotherInstant in range by $range, but it's not." },
         { "$value is not close to $anotherInstant in range by $range " }
      )
   }
}
```

sksamuel: The 5.seconds style was un-deprecated in Kotlin 1.6 or 1.7, so that's the best way to go I think.
I'd be ok having both styles.

jaehyunup: I was mistaken., it's un-deprecated 
good 👍🏻

I modified the code and fixed the bug. in this [commit](https://github.com/kotest/kotest/commit/6368a19c9a940df8f8943620ed57b50f9fc4cb4c)

sksamuel (approved):

PR #3182, opened 2022-08-27
Fix Instant::shouldNotBeBetween comment was written incorrectly

jaehyunup:
I modified the comment of the function "Instnat::shouldNotBeBetween" because it seemed wrong.
(The comment for the current "shouldNotBeBetween" function appears to be the same as "shouldBeBetween".)


sksamuel:
👍🏼

sksamuel (approved):

PR #3185, opened 2022-08-28
Allow registering custom arbs through GlobalArbResolver.

### LINKED ISSUE #3136: Allow registering custom default Arb for your own types in checkAll/forAll
Currently `checkAll` and `forAll` use `defaultForClass` and `targetDefaultForClass` to get a hardcoded list of global default Arb for those classes.

I have some more types (like `Vector3d` from JOML) that I would like to register default generators for, but it doesn't seem possible.

----

Maybe the definition of  `checkAll` could be amended like this:

```kt
suspend inline fun <reified A> checkAll(
   config: PropTestConfig,
   noinline property: suspend PropertyContext.(A) -> Unit
): PropertyContext = proptest(
   Arb.default<A>(config),
   config,
   property
)
```

sksamuel:
Closes #3136

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/resolution/ArbResolver.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/resolution/ArbResolver.kt
@@ -1,0 +1,19 @@
+package io.kotest.property.resolution
+
+import io.kotest.property.Arb
+import kotlin.reflect.KType
+
+/**
+ * Looks up default [Arb] instances for types.
+ */
+interface ArbResolver {
+
+   /**
+    * Returns a default [Arb] for the given [KType] or null if one is not available.
+    */
+   fun resolve(type: KType): Arb<*>?
+}
+
+object EmptyArbResolver : ArbResolver {
+   override fun resolve(type: KType): Arb<*>? = null
+}
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/resolution/resolve.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/resolution/resolve.kt
@@ -1,0 +1,37 @@
+package io.kotest.property.resolution
+
+import io.kotest.property.Arb
+import kotlin.reflect.KType
+import kotlin.reflect.typeOf
+
+/**
+ * Resolves a default [Arb] for reified type [A].
+ * See [resolve].
+ */
+inline fun <reified A> Arb.Companion.default(): Arb<A> = resolve(typeOf<A>()) as Arb<A>
+
+/**
+ * Resolves a default [Arb] for a given [type] through the following
+ * [ArbResolver] strategies:
+ *
+ * 1. Checks [CommonTypeArbResolver] for Kotlin common type mappings.
+ * 2. Checks [platformArbResolver] for platform specific mappings.
+ * 3. Checks user defined mappings in [GlobalArbResolver].
+ *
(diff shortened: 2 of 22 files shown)


kotest-property/src/commonMain/kotlin/io/kotest/property/resolution/ArbResolver.kt
```diff
@@ -0,0 +1,19 @@
+package io.kotest.property.resolution
+
+import io.kotest.property.Arb
+import kotlin.reflect.KType
+
+/**
+ * Looks up default [Arb] instances for types.
+ */
+interface ArbResolver {
```
Kantis, line 9:
```suggestion
internal interface ArbResolver {
```

kotest-property/src/commonMain/kotlin/io/kotest/property/resolution/ArbResolver.kt
```diff
@@ -0,0 +1,19 @@
...
+
+   /**
+    * Returns a default [Arb] for the given [KType] or null if one is not available.
+    */
+   fun resolve(type: KType): Arb<*>?
+}
+
+object EmptyArbResolver : ArbResolver {
```
Kantis, line 17:
```suggestion
internal object EmptyArbResolver : ArbResolver {
```

kotest-property/src/commonMain/kotlin/io/kotest/property/resolution/resolve.kt
```diff
@@ -0,0 +1,38 @@
...
+import io.kotest.property.Arb
+import kotlin.reflect.KType
+import kotlin.reflect.typeOf
+
+/**
+ * Resolves a default [Arb] for reified type [A].
+ * See [resolve].
+ */
+@Suppress("UNCHECKED_CAST")
+inline fun <reified A> Arb.Companion.default(): Arb<A> = resolve(typeOf<A>()) as Arb<A>
```
Kantis, line 12: I'm wondering if there's some way to use an `ArbResolver` as a context receiver here, to allow users to override Arb resolution.

```suggestion
context(ArbResolver)
inline fun <reified A> Arb.Companion.default(): Arb<A> = resolve(typeOf<A>()) as Arb<A> // should use ArbResolver#resolve
```

Ideally the method below (which could be considered a composite `ArbResolver` as well) would be added to the context by default.

Kantis (approved):

sksamuel:
Oops auto merge merged this. I will apply your suggestions manually @Kantis

kotest-property/src/commonMain/kotlin/io/kotest/property/resolution/resolve.kt
sksamuel: I don't want to introduce context receivers to Kotest as that will move us to 1.8 as the minimum version (if context receivers are stable by then). We've been bitten before by experimental stuff so not yet.

kotest-property/src/commonMain/kotlin/io/kotest/property/resolution/ArbResolver.kt
sksamuel: done

kotest-property/src/commonMain/kotlin/io/kotest/property/resolution/ArbResolver.kt
sksamuel: done

PR #3186, opened 2022-08-28
shouldBe/shouldNotBe chain

### LINKED ISSUE #3170: shouldBe() / shouldNotBe() don't chain
kotest 5.4.2

I'm testing some mapping and transformation code, and sometimes there can be some benefits to asserting against external test data in stead of hardcoded inline constants. But this carries the risk that the test data may be incomplete, outdated or corrupted, and you end up with a test that looks god but actually compares null to null, which might pass even if the relevant mapping code is faulty or missing entirely. 

To guard against this, i add a null check for the nullable feilds, and chaining is a clear and succinct way to do it:  
```
with(outboundEvent.data) {
    ...
    purpose.shouldNotBeNull() shouldBe src.purpose
```
(i use infix syntax for consistency with the rest of my test, and function syntax for the null check because I have to) 
This works fine, but my first attempt did not:
```
    purpouse.shouldNotBe(null) shouldBe src.purpouse
```
Apparently shouldtNotBe (and shouldBe) returns Unit, not T as all the chainable asserts.
Why this inconsistency?

sksamuel:
Closes #3170


LeoColman (changes requested):
Missing some tests on the return

LeoColman (approved):

LeoColman (approved):

myuwono (approved):

philBrown:
This represents a breaking change to the API and should never have been released as a minor update

PR #3187, opened 2022-08-28
Added lazy mountable extensions

sksamuel:
Would something as simple as this suffice for your needs @nomisRev


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/extensions/MountableExtension.kt
```diff
@@ -27,11 +27,54 @@ interface MountableExtension<CONFIG, MATERIALIZED> : Extension {
...
+   fun mount(configure: CONFIG.() -> Unit): LazyMaterialized<MATERIALIZED>
+}
+
+data class LazyMaterialized<MATERIALIZED>(private var value: MATERIALIZED) {
+
+   fun set(value: MATERIALIZED) {
+      this.value = value
+   }
+
+   fun get() = value
```
nomisRev: Should this not be protected from incorrect access during initialisation? It can only be safely accessed from `TestScope`? That actually depends on how the underlying `Extension` initialises it, right?

This would work for my use-case, but the API is rather unsafe.

In the PR with the original use-case [here]() there is `LazyMaterialized` with `suspend fun get()` so it can only be accessed from `suspend TestScope.() -> Unit`. If it's access before it needs to be somehow in a `suspend` context and will get loaded lazily then. If it's accessed **after** `clear/close` is called (on `afterSpec` or similar) then it'll throw an `IllegalStateException`.

Should we have a similarly "safer" API? I.e. the example here `Kafka` also involves some heavy I/O, blocking code both on `start` and `close`.

LeoColman (approved):

nomisRev (commented):
I'm so sorry @sksamuel, my comment has been pending on this PR for 2 + weeks.. 🤦‍♂️

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/extensions/MountableExtension.kt
nomisRev: So, short-answer: yes this works.
Long-answer: should we try to make the API safer?

sksamuel: I agree the API isn't great, but I don't know how we can limit it to test scope, or even if we care about that, as long as it supports suspension.
Like you might want to initialize in a before spec listener. That's a suspendable context.
Let me think more on this.

sksamuel: Ok I have followed your guidelines and it is much clearer now.

A `LazyMountableExtension` returns a `LazyMaterialized` which has a SAM.

```kotlin
interface LazyMaterialized<MATERIALIZED> {
   suspend fun get(): MATERIALIZED
}
```

Once get is invoked, the implementation can initialize if required.

Here is the example from the tests:

```kotlin
object : LazyMaterialized<String> {
  var state: String? = null
  override suspend fun get(): String {
    delay(1)
    if (state == null) state = "ready"
    return state ?: error("Must be initialized")
  }
}
```

nomisRev:
Thank you for looking into this @sksamuel 🙏
This is going to be great to offer even better support for Arrow in Kotest auto-of-the-box ❤️

PR #3192, opened 2022-08-29
[DO NOT MERGE] Test with Kotlin 1.7.20-Beta

sksamuel:
This PR is to test Kotest with 1.7.20


LeoColman:
@sksamuel I think you've fixed Kotlin version in main, right?
For that reason I'll close this one to clear things up :)

Issue #3097, opened 2022-07-06
Error message when using shouldContainExactly with HashSet is hard to understand


JonathanLennox:
**Which version of Kotest are you using**
kotest 5.3.0, JVM

When I apply the `shouldContainExactly` to a `HashSet`, the resulting error message is very hard to understand:

```
Disallowed: Set can be compared only to Set
java.lang.AssertionError: Disallowed: Set can be compared only to Set
May not compare HashSet with ArrayList
expected:<*> but was:<*>
```

As I eventually figured out, `shouldContainExactly` requires a collection with a fixed iteration order, which `HashSet` does not satisfy, but that is not at all clear from that error.

The error message should say something about the set's iteration order.

PR #3194, opened 2022-08-29
Attempt to improve the error message for `containExactly` when used with non-stable sets

Kantis:
See #3097


sksamuel (approved):

Kantis:
Having a look at the failures. I only ran tests for `containExactly` 🙈

PR #3197, opened 2022-08-30
Read/write table data to a file

### LINKED ISSUE #3179: Read/write Table data to a file format maybe CSV
Table testing is nice

```kotlin
val table: Table3<String, String, String> = table(
    headers("mavenDependency", "kotlinName", "versionName"),
    row("io.kotest:kotest-core", "Testing.kotest.core", "version.kotest"),
    row("io.kotest:kotest-assertions-arrow", "Testing.kotest.assertions.arrow", "version.kotest"),
)
table.forAll { mavenDependency, kotlinName, versionName ->
    listOf(mavenDependency, kotlinName, versionName).all { it.contains("kotest") } shouldBe true
}
```

The issue is my table contains 1000+ rows currently so I'm reinventing the wheel on how to read it from a text file / write it to a text file https://github.com/jmfayard/refreshVersions/blob/main/plugins/dependencies/src/test/resources/dependencies-mapping-validated.txt

It would be really useful to have a simple way to read that data from a (CSV?) file

`fun File.readTable3<A,B,C>(headers: Headers3, map: Sometype -> Row3<A, B, C>) : Table<A,B,C>`

I can volunteer to implement that if you think the feature makes sense.

**CsvDataSource**

While researching for writing this ticket, I found out that there used to be a `CsvDataSource`class for this https://github.com/kotest/kotest/search?q=CsvDataSource&type=code

```kotlin
val source = CsvDataSource(javaClass.getResourceAsStream("/user_data.csv"), CsvFormat())

val table = source.createTable<Long, String, String>(
     { it: Record -> Row3(it.getLong("id"), it.getString("name"), it.getString("location")) },
     { it: Array<String> -> Headers3(it[0], it[1], it[2]) }
)
```

But it was removed in an apparently unrelated commit? https://github.com/kotest/kotest/pull/1457

jmfayard:
Resolves #3179

See the discussion in the README

I'm starting with readme-oriented programming, documenting how the end result should look like to make sure it makes sense.

**Update 1st september** needs refactoring but most features are there I think.

Documentation needs updates:

- [ ] it's now `file.writeTable(headers, rows)` and only works for `RowN<String, String, ..., String>`
- [ ] use `table.mapRows { }` if you are using `RowN` with types other than `String`
- [ ] `map.toTable()` is a quick way to produce a table from a map
- [ ] `table(headers, list.map { row(...) })` is a quick way to produce a table from a list


documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
```diff
@@ -0,0 +1,90 @@
...
+Now that your table data lives in a file, it's time to read it!
+
+## Read table from files and execute your asserts
+
+Here is how you read your `.table` file:
+
+```kotlin
+val tableFromFile = table(
+    source = testResources.resolve("users.table"),
+    headers = headers("id", "username", "fullName"),
```
Kantis, line 72: Why do we need to define the headers again?

documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
```diff
@@ -0,0 +1,90 @@
...
+
+Now that your table data lives in a file, it's time to read it!
+
+## Read table from files and execute your asserts
+
+Here is how you read your `.table` file:
+
+```kotlin
+val tableFromFile = table(
+    source = testResources.resolve("users.table"),
```
Kantis: Could source just be a string? That way you can define, load and potentially transform the data any way you'd like before turning it into a table 🤔

documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
```diff
@@ -0,0 +1,90 @@
...
+
+```csv
+id | username | fullName
+4  | jmfayard | Jean-Michel Fayard
+6  | louis    | Louis Caugnault
+```
+
+Curious why it's not just a .csv file?
+
+Well CSV is not a well defined format. Everyone has its flavor and we have too. The `.table` has its rules: it always uses `|` as separator, it must have an header, it can have comments, it can't have newlines inside the columns. Basically it's optimized for putting table data in a `.table` file.
```
Kantis: I like this. Pragmatic way forward. I think the choice of `|` makes the table read nicely, plus it should be seldomly used for other things.

documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
```diff
@@ -0,0 +1,90 @@
...
+
+Here is how you read your `.table` file:
+
+```kotlin
+val tableFromFile = table(
+    source = testResources.resolve("users.table"),
+    headers = headers("id", "username", "fullName"),
+    transform = { a: String, b: String, c: String ->
+        row(a.toInt(), b, c)
+    }
```
Kantis, line 76: Nice, makes it possible to convert any way we'd like!

documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
```diff
@@ -0,0 +1,90 @@
...
+val tableFile = testResources.resolve("users.table")
+table.writeTo(tableFile)
+```
+
+The `users.table` file looks like this:
+
+```csv
+id | username | fullName
+4  | jmfayard | Jean-Michel Fayard
+6  | louis    | Louis Caugnault
```
Kantis, line 48: What should happen if the user tries to load this table?
```csv
id | username | fullName
4  | jmfayard 
6  | louis    | Louis Caugnault
```

documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
jmfayard: As a sanity check
Being explicit that the file must contain an header `id | username | fullName`
The alternative being to assume that whatever the first line in the header is implicitly always correct.

documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
jmfayard: yes there can be an overload

documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
jmfayard: and then the only escaping we have to do is  `\|` for an actual `|`

documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
jmfayard:
```
AssertionError when parsing File("users.table") with headers("id | username | fullName")
Expected: 3 columns for all rows
Actual: 2 or less columns at lines [1, 42]
```

jmfayard: I think I would allow and ignore empty lines or lines that start with `#` as a comment
https://stackoverflow.com/questions/1961006/can-a-csv-file-have-a-comment

documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
jmfayard: actually we need the headers only when the source is a file, not when it's passed as a raw string

jmfayard:
@Kantis probably need some refactoring and handle variants with 20 generic types, but I think I cover the main use cases now. ouch was lots of work :)

documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
jmfayard: done

documentation/versioned_docs/version-5.4/assertions/table_driven_testing.md
```diff
@@ -0,0 +1,89 @@
...
+
+Defining a table of data in code is convenient... until you start to have too much rows.
+
+## Export a table to a text file
+
+You can export your data to a text file with the `.table` extension.
+
+```kotlin
+val tableFile = testResources.resolve("users.table")
+table.writeTo(tableFile)
```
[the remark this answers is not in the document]
jmfayard: TODO: it would be nice if it was possible to programmatically add comments and new lines, maybe something like this:

```kotlin
val comments = mapOf(1 to "comment at first line", 5 to "", 6 to "other comment", rows.lastIndex to "That's all I've got")
file.writeTable(headers, rows, comments)
```

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/data/TableFile.kt
```diff
@@ -0,0 +1,71 @@
...
+internal fun File.readStringTable(headers: List<String>): StringTable {
+   if (exists().not()) throw AssertionError("Can't read table file")
+   if (extension != "table") throw AssertionError("Table file must have a .table extension")
+   val lines = readLines()
+   if (lines.isEmpty()) throw AssertionError("Table file must have a header")
+   StringTable.parseRow(lines.first()) shouldBe headers
+   return StringTable(headers, lines, skipFirstLine = true)
+}
+
+fun File.writeTable(headers: List<String>, cells: List<List<String>>): String {
```
jmfayard, line 38: TODO: it would be nice if it was possible to programmatically add comments and new lines, maybe something like this:

```kotlin
val comments = mapOf(1 to "comment at first line", 5 to "", 6 to "other comment", rows.lastIndex to "That's all I've got")
file.writeTable(headers, rows, comments)
```

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/data/StringTableTest.kt
```diff
@@ -0,0 +1,213 @@
...
+            fileMissingTableExtension.writeTable(usersTable.headers, emptyList())
+         }
+      }
+
+      test("Cells con't contain new lines") {
+         val tableWithNewLines =
+            mapOf("1" to "one\n", "two" to "two", "three" to "three\nthree").toTable()
+         shouldThrowMessage(testCase.name.testName) {
+            tempfile(suffix = ".table").writeTable(tableWithNewLines.headers, tableWithNewLines.rows)
+         }
```
Kantis, line 221: It's a bit unclear to me what the expected failure message is here

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/data/StringTableTest.kt
```diff
@@ -0,0 +1,213 @@
...
+
+      test("| should be escaped") {
+         val table = mapOf("greeting" to "Hello || world").toTable()
+         val file = tempfile(suffix = ".table")
+         file.writeTable(table.headers, table.rows) shouldBe """
+            key      | value
+            greeting | Hello \|\| world
+         """.trimIndent()
+      }
+   }
```
Kantis, line 206: I think we also need to add escaping of backslash itself. E.g.: 

```
Hello \|\| World ->  (Hello || World)  // 1 item
Hello \\|| World -> (Hello \,, World)   // 3 items
Hello \\\|\\\| World -> (Hello \|\| World) // 1 item
```

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/data/StringTableTest.kt
```diff
@@ -0,0 +1,213 @@
...
+
+   test("All rows must have the right number of columns") {
+      val invalidRows = """
+      4  | jmfayard | Jean-Michel Fayard
+      5  | victor | Victor Hugo | victor.hugo@guernesey.co.k
+      6  | louis    | Louis Caugnault
+      7  | edgar
+   """.trimIndent()
+      shouldThrowMessage("Expected all rows to have size 3, but got rows at lines [1, 3]") {
```
Kantis: Perhaps we can build an even more elaborate failure message. Could help if/when there's rows with comments etc as well. Plus it removes ambiguity regarding 0 vs 1-indexing. Just an idea, feel free to skip if it adds too much scope.

```suggestion
      shouldThrowMessage("""
        Expected all rows to have 3 columns but some rows differed:
          - Row 1 has 4 columns ("5  | victor [...]")
          - Row 3 has 2 columns ("7  | edgar")
      """ ) {
```

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/data/StringTable.kt
```diff
@@ -0,0 +1,43 @@
...
+   }
+
+   companion object {
+      internal fun parseRow(line: String): List<String> {
+         val notAPipeSeparator = "🫓"
+         return line
+            .replace("\\|", notAPipeSeparator)
+            .split("|")
+            .map { it.trim().replace(notAPipeSeparator, "|") }
+      }
```
Kantis: Could avoid the `notAPipeSeparator` back and forth with a regex 

```suggestion
         return line.split("[^\\]\|").map(String::trim)
      }
```

Kantis: Also I think we should add a note in the docs that whitespace will be trimmed from the cell content 🙂

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/data/TableFile.kt
```diff
@@ -0,0 +1,71 @@
...
+// TODO
+
+private fun Row.strings(): List<String> = values().map { it.toString() }
+
+
+internal fun File.readStringTable(headers: List<String>): StringTable {
+   if (exists().not()) throw AssertionError("Can't read table file")
+   if (extension != "table") throw AssertionError("Table file must have a .table extension")
+   val lines = readLines()
+   if (lines.isEmpty()) throw AssertionError("Table file must have a header")
```
Kantis, line 33: Would it make sense to throw `IllegalArgumentException`s instead? No assertion has been made at this point, I think?

Kantis (commented):
Nice work! 🤘 Added various suggestions. Would like a third pair of eyes on this once it's ready 👀

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/data/StringTableTest.kt
jmfayard: so I started with something like

```
context("file.writeTable - validation") {
   test("check extension") {
      shouldThrowMessage("Table file must have a .table extension") {
         val fileMissingTableExtension = tempfile()
         fileMissingTableExtension.writeTable(usersTable.headers, emptyList())
      }
   }
}
```

Then I was like let's give a better name to the test... like the expected error message

```
context("file.writeTable - validation") {
   test("Table file must have a .table extension") {
      shouldThrowMessage("Table file must have a .table extension") {
         val fileMissingTableExtension = tempfile()
         fileMissingTableExtension.writeTable(usersTable.headers, emptyList())
      }
   }
}
```

But then I wondered if there was a way to not duplicate the string?

And sure enough it's easy to do with kotest

```
context("file.writeTable - validation") {
   test("Table file must have a .table extension") {
      shouldThrowMessage(testCase.name.testName) {
         val fileMissingTableExtension = tempfile()
         fileMissingTableExtension.writeTable(usersTable.headers, emptyList())
      }
   }
}
```

I can revert to the first version if you prefer

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/data/StringTableTest.kt
jmfayard: @Kantis my issue is more that there could be a large number of rows inside the file.
Should we take the first 5 with errors or something?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/data/StringTableTest.kt
jmfayard: done

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/data/StringTable.kt
jmfayard: > Also I think we should add a note in the docs that whitespace will be trimmed from the cell content 🙂

Done

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/data/TableFile.kt
jmfayard: you can decide. Personally I don't like exceptions since they don't work well with `assertSoftly` for example.

```kotlin
   test("invalid file") {
      val file = resourcesDir.resolve("invalid.table") 
      val table = table(headers, file, transform) // implicit assertion that the file exists
      table shouldBe expectedTable
   }
```

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/data/StringTableTest.kt
jmfayard: @Kantis done

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/data/StringTable.kt
jmfayard: > @Kantis Could avoid the notAPipeSeparator back and forth with a regex

that was my first idea too but it removes the character before the separator

```kotlin
test("regex? now you have two problems") {
      val input = "hello|world"
      val regex = Regex("(?:[^\\\\])\\|")
      input.split(regex) shouldBe listOf("hello", "world")
}
Element differ at index: [0]
Expected :["hello", "world"]
Actual   :["hell", "world"]
```

I had to do a lame imperative for loop in StringTable.parseRow() to manage all cases.
If you have a better solution I am open

jmfayard: found the right regex

jmfayard:
> @Kantis Would like a third pair of eyes on this once it's ready 👀

It's ready-ish now.
I just added the versions with 1 to 9 generic arguments. Fun :)
Who can provide a third pair of eyes?

Also 

- The CI is still in mode **2 workflows awaiting approval**
- locally `./gradlew allTests` throws an error [`org.gradle.api.GradleException: Could not evaluate onlyIf predicate for task ':kotest-common:jsIrGenerateExternalsIntegrated'.`](https://scans.gradle.com/s/ay64yqnzii54g/failure?expanded-stacktrace=WyIwIiwiMC0yIl0#1) which doesn't seems related to anything I did

sksamuel:
@Kantis I'm happy if you are.

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/data/TableFile.kt
Kantis: I think there's a difference between an assertion (E.g. validating the output of production code) and a bug in test code. IMO this `AssertionError` would indicate a fault in the test code, not the main code and I think the distinction is clearer when output as an exception. What do you think?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/data/StringTableTest.kt
Kantis: No big deal, lets keep it as is. :)

Kantis (approved):

Kantis:
There's a chance we'll try to get https://github.com/kotest/kotest/pull/3264 out tonight, so disabled auto-merge for now.

Kantis:
Thanks @jmfayard 💯

jmfayard:
amazing, thanks!

polarene:
This is a very cool feature, why isn't it documented? I stumbled upon it randomly while digging through the source code.

Kantis:
@polarene It was documented but only in the 5.4 version of the docs. Fixed in https://github.com/kotest/kotest/commit/7b29815cc4b67b14118460b99203b6cbe81e6d9c

polarene:
Thank you! I found out about it because I was trying to come up with something similar, either using Markdown tables or with a DSL (I did a nice hack but it doesn't work if the first value is a number, due to standard lib operators).

But is the page released? Still can't find it in the docs... maybe the index needs to be updated as well?

polarene:
Mhh yes, it seems to be missing from these sidebar files:
https://github.com/kotest/kotest/blob/7b29815cc4b67b14118460b99203b6cbe81e6d9c/documentation/versioned_sidebars/version-5.8-sidebars.json#L48

PR #3199, opened 2022-08-31
Add ability to toggle the inheritance of @Tags

Kantis:


sksamuel:
We probably need to hide behind a setting

LeoColman (approved):

documentation/docs/framework/tags.md
```diff
@@ -167,6 +167,10 @@ class MyTestClass : FunSpec()
 Note that when you use these annotations you pass the tag string name, not the tag itself. This is due to Kotlin annotations only allow "primitive" arguments
 :::
 
+### Inheriting tags
+
+By default, the `@Tags` annotation will only be considered on the immediate Spec which it was applied to. However, a Spec can also inherit tags from superclasses and superinterfaces. To enable this, toggle `tagInheritance = true` in your [project config](./project-config.html)
```
LeoColman, line 172: Is this the documentation that goes to production?
If so, we should release that only when the next version is ready

Kantis: It will be in the "next" version until we perform the release and cut the docs, at which point it'll show in the stable docs. 

Next version shows like this:

image

Issue #3161, opened 2022-08-15
Add option to ignore differences in line endings during string comparison, so tests are not OS dependent


aSemy:
When writing tests that verify multi-line strings, for example file contents, it's common to use Kotest string assertions. However these tests are highly sensitive to line endings.

This assertion will pass on Unix, and fail on Windows:

https://github.com/kotest/kotest/blob/a7195a4d1c4a0d9e20526bc336338d394e028003/kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePluginSpec.kt#L38-L43

This PR also fixed line-ending tests failing https://github.com/kotest/kotest/pull/2743

I have some thoughts on how to implement this.

### Additional arg on matchers

One option is to add an arg, like `ignoreLineEndings`, to *all* string comparison matchers which defaults to true, but this would mean they could no longer be `infix fun`s. Adding an overload with a `ignoreLineEndings` arg could work around that, but that's a lot of work, code, maintenance, and testing.

> Example: new overloads are verbose and repetitive
> 
> ```kotlin
> // existing function
> infix fun <A : CharSequence?> A.shouldStartWith(prefix: CharSequence): A {
>    this should startWith(prefix)
>    return this
> }
> 
> // new overload
> fun <A : CharSequence?> A.shouldStartWith(prefix: CharSequence, ignoreLineEndings: Boolean = false): A {
>    this shouldNot startWith(prefix, ignoreLineEndings)
>    return this
> }
> 
> // existing function
> infix fun <A : CharSequence?> A.shouldNotStartWith(prefix: CharSequence): A {
>    this shouldNot startWith(prefix)
>    return this
> }
> 
> // new overload
> fun <A : CharSequence?> A.shouldNotStartWith(prefix: CharSequence, ignoreLineEndings: Boolean = false): A {
>    this shouldNot startWith(prefix, ignoreLineEndings)
>    return this
> }
> ```

### Boolean vs Sealed class flag

Even if there is an additional arg, there are drawbacks to using a boolean flag, as it is restricted in allowing for different options. Instead a sealed class/enum parameter would allow for more customisation, for example controlling exactly how the line endings are handled. Perhaps someone wants line endings in the 'expected' string to be converted from `/r/n` to `/n`, but on the 'actual' string left alone. Or vice versa, or both. 

### Project config option

[...]

PR #3207, opened 2022-09-10
Adjust Kotest test assertions to be runnable on Windows

aSemy:
When I run the tests on my Windows machine, I get several failing tests because the test assertions are Unix only.

I also adjusted the Gradle tests to only run when the required tasks are present. This is a bit of a hack, and should be fixed by implementing Gradle TestKit tests #3163

related #3161 

I've enabled the `jvmOnly` flag so I might have missed some tests

### DIFF
--- a/kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
+++ b/kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
@@ -103,19 +103,19 @@ gradlePlugin {
 }
 
 
-val kotestPluginConstantsFileContents = resources.text.fromString(
-   """
-      |// Generated file, do not edit manually
-      |@file:org.gradle.api.Generated
-      |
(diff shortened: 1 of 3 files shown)


kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
```diff
@@ -99,20 +103,20 @@ gradlePlugin {
...
+      """
+         |// Generated file, do not edit manually
+         |@file:org.gradle.api.Generated
+         |
+         |package io.kotest.framework.multiplatform.gradle
+         |
+         |const val KOTEST_COMPILER_PLUGIN_VERSION: String = "${Ci.gradleVersion}"
+         |
+      """.trimMargin()
+   )
```
aSemy, line 118: I moved this to avoid an issue with the Gradle configuration cache, which likes having variables defined within the task.

Issue #3204, opened 2022-09-06
Config scanning cannot be disabled in kotest.properties


PrinceOfDubai24:
I have the following `kotest.properties` file
```
kotest.framework.config.fqn=com.acme.kotest.KotestConfig
kotest.framework.classpath.scanning.config.disable=true
kotest.framework.classpath.scanning.autoscan.disable=true # not relevent for the issue
```
**Config** scanning still takes place instead of instantiating the one from my file because `detectAbstractProjectConfigs` and `loadProjectConfigFromClassname` are run **before** `applyConfigFromSystemProperties` (within the `initialize` method) has a chance to populate system props from the file.

https://github.com/kotest/kotest/blob/e47889cc3129e4be3ee432d78ae3da12114b05d3/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestEngineLauncher.kt#L169-L172

https://github.com/kotest/kotest/blob/e47889cc3129e4be3ee432d78ae3da12114b05d3/kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/config/detectAbstractProjectConfigs.kt#L7-L18

https://github.com/kotest/kotest/blob/e47889cc3129e4be3ee432d78ae3da12114b05d3/kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/config/loadProjectConfigFromClassname.kt#L6-L9

PR #3208, opened 2022-09-11
Detect Abstract config after applying config from system properties

pt2121:
Addressed #3204
I delayed scanning the abstract project configs by wrapping the call in lambda.
Please let know if you want more refactoring and make the execution order more explicit.


sksamuel (approved):

sksamuel:
Nice :)

sksamuel (approved):

PR #3216, opened 2022-09-25
feat(3213): add should contain only assertion

### LINKED ISSUE #3213: Add shouldContainOnly
It would be useful to have an assertion that is exactly equivalent to AssertJ's `assertThat(list).containsOnly(these,elements,in,any,number,and,any,order,and,nothing,else)`.

For example:

```
listOf(1, 1, 1).shouldContainOnly(1)   // expected to pass
listOf(1, 1, 2, 2).shouldContainOnly(1, 2)  // expected to pass
listOf(1, 1, 2).shouldContainOnly(1)   // expected to fail because it shouldn't contain 2
listOf(1, 1, 2).shouldContainOnly(1, 2, 3)  // expected to fail because it doesn't contain 3
```

p1729:
Closes #3213

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containOnly.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containOnly.kt
@@ -12,6 +12,34 @@ import io.kotest.assertions.AssertionsConfig.maxCollectionPrintSize as printSize
 infix fun <T> Iterable<T>?.shouldContainOnly(expected: Iterable<T>) =
     this?.toList() should containOnly(expected.toList())
 
+/**
+ * Assert that an array contains only the given elements.
+ */
+@JvmName("shouldContainOnly_array")
+infix fun <T> Array<T>?.shouldContainOnly(expected: Array<T>) =
+    this?.asList() should containOnly(*expected)
+
+/**
+ * Assert that a collection contains only the given elements.
+ */
+fun <T> Iterable<T>?.shouldContainOnly(vararg expected: T) =
+    this?.toList() should containOnly(*expected)
+
+/**
+ * Assert that an array contains only the given elements.
+ */
+fun <T> Array<T>?.shouldContainOnly(vararg expected: T) =
+    this?.asList() should containOnly(*expected)
+
+/**
+ * Assert that a collection contains only the given elements.
+ */
+infix fun <T, C : Collection<T>> C?.shouldContainOnly(expected: C) = this should containOnly(expected)
+
+/**
+ * Assert that a collection contains only the given elements.
+ */
+fun <T> Collection<T>?.shouldContainOnly(vararg expected: T) = this should containOnly(*expected)
 
 /**
  * Assert that a collection contains only the given elements
@@ -46,3 +74,16 @@ fun <T, C : Collection<T>> containOnly(expectedCollection: C): Matcher<C?> = nev
 
     MatcherResult(passed, failureMessageSupplier, negatedFailureMessageSupplier)
 }
+
+@JvmName("shouldNotContainOnly_iterable")
+infix fun <T> Iterable<T>?.shouldNotContainOnly(expected: Iterable<T>) =
+    this?.toList() shouldNot containOnly(expected.toList())
+
+@JvmName("shouldNotContainOnly_array")
+infix fun <T> Array<T>?.shouldNotContainOnly(expected: Array<T>) = this?.asList() shouldNot containOnly(*expected)
+
+fun <T> Iterable<T>?.shouldNotContainOnly(vararg expected: T) = this?.toList() shouldNot containOnly(*expected)
+fun <T> Array<T>?.shouldNotContainOnly(vararg expected: T) = this?.asList() shouldNot containOnly(*expected)
+
+infix fun <T, C : Collection<T>> C?.shouldNotContainOnly(expected: C) = this shouldNot containOnly(expected)
+fun <T> Collection<T>?.shouldNotContainOnly(vararg expected: T) = this shouldNot containOnly(*expected)
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containOnly.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containOnly.kt
@@ -1,0 +1,48 @@
+package io.kotest.matchers.collections
+
+import io.kotest.assertions.print.print
+import io.kotest.matchers.*
+import kotlin.jvm.JvmName
+import io.kotest.assertions.AssertionsConfig.maxCollectionPrintSize as printSize
+
+/**
+ * Assert that a collection contains only the given elements.
+ */
+@JvmName("shouldContainOnly_iterable")
+infix fun <T> Iterable<T>?.shouldContainOnly(expected: Iterable<T>) =
+    this?.toList() should containOnly(expected.toList())
+
+
+/**
+ * Assert that a collection contains only the given elements
+ */
+fun <T> containOnly(vararg expected: T): Matcher<Collection<T>?> = containOnly(expected.asList())
+
+/**
+ * Assert that a collection contains only the given elements.
+ */
+fun <T, C : Collection<T>> containOnly(expectedCollection: C): Matcher<C?> = neverNullMatcher { actualCollection ->
+    val actualSet = actualCollection.toSet()
+    val expectedSet = expectedCollection.toSet()
+    val notExpectedSet = actualSet.minus(expectedSet)
+    val missingSet = expectedSet.minus(actualSet)
+
+    val passed = notExpectedSet.isEmpty() && missingSet.isEmpty()
+    val negatedFailureMessageSupplier = { "Collection should not have just ${expectedCollection.print().value}" }
+    val failureMessageSupplier = {
+        buildString {
+            if (missingSet.isNotEmpty()) {
+                append("Some elements were missing: ${missingSet.take(printSize.value).print().value}")
+            }
+            if (missingSet.isNotEmpty() && notExpectedSet.isNotEmpty()) {
+                append(" and some elements were unexpected: ${notExpectedSet.take(printSize.value).print().value}")
+            }
+            if (missingSet.isEmpty() && notExpectedSet.isNotEmpty()) {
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/ShouldContainOnlyTest.kt
```diff
@@ -0,0 +1,43 @@
...
+                }
+                shouldThrow<AssertionError> {
+                    actual shouldContainOnly arrayOf(3, 2)
+                }
+                shouldThrow<AssertionError> {
+                    actual.shouldNotContainOnly(1, 2, 3)
+                }
+                shouldThrow<AssertionError> {
+                    actual shouldNotContainOnly arrayOf(1, 2, 3)
+                }
```
charleskorn: Would be good to assert that the assertion error messages are as expected for these test cases

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/ShouldContainOnlyTest.kt
```diff
@@ -0,0 +1,43 @@
...
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.collections.shouldContainOnly
+import io.kotest.matchers.collections.shouldNotContainOnly
+import io.kotest.matchers.throwable.shouldHaveMessage
+
+class ShouldContainOnlyTest : WordSpec() {
+
+    init {
+        "containOnly" should {
+            "test that an array contains given elements only" {
```
charleskorn, line 16: Would it be possible to break this up into smaller tests, rather than one big test?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/ShouldContainOnlyTest.kt
cb-pankajk: sure, I will make the changes

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/ShouldContainOnlyTest.kt
cb-pankajk: sure, I will make the changes

p1729: sure, I will make the changes

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/ShouldContainOnlyTest.kt
p1729: sure, I will make the changes

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containOnly.kt
```diff
@@ -0,0 +1,89 @@
...
+ */
+fun <T> Collection<T>?.shouldContainOnly(vararg expected: T) = this should containOnly(*expected)
+
+/**
+ * Assert that a collection contains only the given elements
+ */
+fun <T> containOnly(vararg expected: T): Matcher<Collection<T>?> = containOnly(expected.asList())
+
+/**
+ * Assert that a collection contains only the given elements.
```
LeoColman, line 86: I'd like this documentation to be more explicit on how the method will do this. We can do that with a couple of code examples if you think it will be more instructive, like:

listOf(1, 1, 1).shouldContainOnly(1)   // Assertion passes
listOf(1, 2, 1).shouldContainOnly(1) // Assertion fails

This way we have a couple of samples on why we should use this method.

This applies for all the other documentations. What do you think?

LeoColman (changes requested):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containOnly.kt
p1729: yes, that would be helpful as people may get confused and would want to know how it is different than `containExactly`. Shall I add the examples to all flavours? e.g array, iterable, vararg

Also, for this PR I picked a few things from `containExactly` like error messages pattern, etc which are repeated here in `containOnly` as well. Do let me know if you think something needs refactoring. // cc @sksamuel

LeoColman: > . Shall I add the examples to all flavours? e.g array, iterable, vararg

Yes.
Although honestly, I'd just copy paste it a couple of times and be done with it 😅

p1729: @LeoColman I have some more queries on the assertions. What should be the expectation in these 
```
listOf(1, 2) shouldContainOnly setOf(1, 2) // Assertion passes ?
listOf(1, 2) shouldNotContainOnly listOf(3) // Assertion passes ? 
```

LeoColman: That's a good question.
When things don't make much sense to be used in that way I usually care less about its documentation.

So feel free to not include these cases in the documentation. However, we should define a explicit behavior, yes. As the proposer of the feature, what do you think should be the behavior?

p1729: For, `listOf(1, 2) shouldContainOnly setOf(1, 2)` I think assertion should pass in this case since the equality is on the items and not really on what type you are passing as long as both are iterable. And, for `listOf(1, 2) shouldNotContainOnly listOf(3)` assertion should fail since the actual collection doesn't even have `3` (not able to think of any use-case though were this will be required). 

Also, I think since no one has asked for these flavours as of now. So, we can be more restrictive in what we offer on public APIs and add these later once requirements come in. Let me your thoughts especially on being restrictive upfront and just disclosing the below 2.

```
fun <T> Iterable<T>?.shouldContainOnly(vararg expected: T)
fun <T> Array<T>?.shouldContainOnly(vararg expected: T)
```

@klitoskyriacou Let us know your thoughts as well

klitoskyriacou: I would treat `shouldContainOnly` so as to be consistent with `shouldContainExactly` with regards to their signatures and overloads. The main difference between the two is when the actual collection contains duplicates: `listOf(1, 1, 2).shouldContainOnly(1, 2)` would pass, whereas `shouldContainExactly` fails. So, since `listOf(1, 2) shouldContainExactly setOf(1, 2)` is well defined and it passes, I would agree that we should also support `listOf(1, 2) shouldContainOnly setOf(1, 2)` and that it should also pass.

As for `listOf(1, 2) shouldNotContainOnly listOf(3)`, I think that should pass, because:

1. `shouldNotContainOnly` passes whenever `shouldContainOnly` fails, and vice versa;
2. `listOf(1, 2) shouldContainOnly listOf(3)` would obviously fail
3. since that fails, its opposite (`listOf(1, 2) shouldNotContainOnly listOf(3)`) should pass.

p1729: thanks, @klitoskyriacou. 

I'm done with all the suggested changes @LeoColman, @charleskorn. Please re-review

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containOnly.kt
```diff
@@ -0,0 +1,181 @@
...
+import io.kotest.assertions.AssertionsConfig.maxCollectionPrintSize as printSize
+
+/**
+ * Assert that a collection contains only the given elements.
+ * For example,
+ *  listOf(1, 1, 2) shouldContainOnly listOf(1, 2)    // Assertion passes
+ *  listOf(1, 2, 1) shouldContainOnly listOf(1)       // Assertion fails
+ *  listOf(1) shouldContainOnly listOf(2, 2)          // Assertion fails
+ *
+ *  Note: Comparison is via the standard Java equals and hash code methods.
```
LeoColman, line 15: We are in common kotlin land, so I think these statements are innacurate

LeoColman (changes requested):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containOnly.kt
p1729: @LeoColman I didn't follow. I just wanted to highlight that clients can change the comparison behavior based on their implementations of hashing contracts per [docs](https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#hashCode--) since I am using maps internally. Also, I picked up this existing note from [here](https://github.com/kotest/kotest/blob/cfb48c74ddcabed017d6baffb936b04ac3b304b3/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactlyInAnyOrder.kt#L21)

sksamuel: I think he's saying that the equals comes from the Kotlin common module not the JVM object superclass, since this isn't a JVM module.

But I don't think it matters too much we all know what it means.

sksamuel:
There's been a lot of review on this 2 file change. I think it's fine to merge now.

sksamuel:
@LeoColman you ok to approve unless you think there's a blocker somewhere.

LeoColman (approved):

LeoColman:
Alright, I'll drop the nitpicking.
Approved. Auto-merged when green.

PR #3217, opened 2022-09-26
Fix json compareArrays

jostrzol:
In the original `expected` was in place of `actual` and vice versa.

It rendered `fieldComparison = Lenient` useless if there was an array in the json, because after flipping, it expected additional fields in `actual` to be in `expected`.

### DIFF
--- a/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
+++ b/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
@@ -268,7 +268,12 @@ internal fun compareArrays(
          fun findMatchingIndex(element: JsonNode): Int? {
             for (i in availableIndexes()) {
                // Comparison with no error -> matching element
-               val isMatch = compare(path + "[$i]", expected.elements[i], element, options) == null
+               val isMatch = compare(
+                  path = path + "[$i]",
+                  expected = expected.elements[i],
+                  actual = element,
+                  options = options
+               ) == null
 
                if (isMatch) {
                   return i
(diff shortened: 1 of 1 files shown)


kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -268,7 +268,7 @@ internal fun compareArrays(
          fun findMatchingIndex(element: JsonNode): Int? {
             for (i in availableIndexes()) {
                // Comparison with no error -> matching element
-               val isMatch = compare(path + "[$i]", element, expected.elements[i], options) == null
+               val isMatch = compare(path + "[$i]", expected.elements[i], element, options) == null
```
piotr-proszowski: Maybe named parameters would be useful here?

```
               val isMatch = compare(
                  path = path + "[$i]",
                  expected = expected.elements[i],
                  actual = element,
                  options = options
               ) == null
```

jostrzol: Done

sksamuel (approved):

piotr-proszowski (approved):

sksamuel:
Do we need to apply this to master branch as well ?

jostrzol:
Sorry for late reply, missed the notification.

> Do we need to apply this to master branch as well ?

Yeah, it would be best.

I misunderstood your contribution guidelines - I thought that all bug fixes should be applied to a release branch. 

Should I make another PR to master?

sksamuel:
The bug fix to the release branch will only apply to 5.4.3, so we'd need it in master if we want it on 5.5 and onward.
I can do it quickly.

jostrzol:
Ok, then go ahead.

PR #3229, opened 2022-10-08
Fix typo from asset to assert

jama5262:
Update a typo from `asset` to `assert` in https://kotest.io/docs/assertions/inspectors.html


sksamuel:
Nice catch

Issue #3232, opened 2022-10-11
checkAll with >5 exhaustives should run for each cross product


zakhenry:
**Which version of Kotest are you using**
5.5.1

In the internal definition of `suspend fun <A, B, C, ...etc> proptest`, there is a special case generated for the scenario where every single element is exhaustive, however this only goes up to the `<A, B, C, D, E>` case, however there is up to `<A, B, C, D, E, F, G, H, I, J, K, L>` cases supported.

This means that while code will compile and run, it will not produce the expected result.

## Demonstration of the behaviour:

```kotlin
suspend fun main() {

    var count = 0
    checkAll(
        listOf('1', '0').exhaustive(),
        listOf('1', '0').exhaustive(),
        listOf('1', '0').exhaustive(),
        listOf('1', '0').exhaustive(),
        listOf('1', '0').exhaustive(),
    ) { a, b, c, d, e ->

        println("$a$b$c$d$e")

        count++
    }

    println("Count: $count")

    var count2 = 0

    checkAll(
        listOf('1', '0').exhaustive(),
        listOf('1', '0').exhaustive(),
        listOf('1', '0').exhaustive(),
        listOf('1', '0').exhaustive(),
        listOf('1', '0').exhaustive(),
        listOf('1', '0').exhaustive(),
    ) { a, b, c, d, e, f ->

        println("$a$b$c$d$e$f")

        count2++
    }

    println("Count2: $count2")
}

```

The first prints all 32 binary digits up to 2^5-1 and count is 32, whereas the second case prints alternating `111111` and `000000`, for 1000 iterations (expected 64).

## Expected behaviour:

Either of (in order of preference):

1. have the same behaviour as the first 5 declarations
2. not compile
3. throw exception that varargs length > 5 where all are exhaustive is not supported

PR #3233, opened 2022-10-11
test(proptest): add (failing) arity tests for checkAll and forAll with arity 6-12

zakhenry:
tests: #3232


sksamuel (approved):
not actually failing it seems

zakhenry:
Yea sorry I wrote the failing test first then raised the PR and then added the fix, forgot to update the title

PR #3234, opened 2022-10-11
feat(Exhaustive.cartesian): support up to 12 args to Exhaustive.cartesian for arity parity with checkAll & forAll

zakhenry:
`Exhaustive.cartesian` is a more explicit alternative to generating the full matrix of possible combinations when all generators are Exhaustive, and so it should support the same number of arguments as checkAll/forAll

also relates to #3233 as this function could be used to simplify the all-exhaustive cartesion branches there, however that would require either changing `Exhaustive.Companion.cartesian` to be suspending, or make NTuple args then forEach on the result.


zakhenry:
relates to #3233 as this function could be used to simplify the all-exhaustive cartesion branches there, however that would require either changing `Exhaustive.Companion.cartesian` to be suspending, or make NTuple args then forEach on the result

sksamuel (approved):

Issue #3099, opened 2022-07-07
Tolerance matchers fail on edge-case numeric inputs


erik:
While working on #3096, the property-based checker found some edge cases for tolerance matchers which don't appear to work correctly (or at least have somewhat strange behavior).

Cases where we should probably throw an `IllegalArgumentException`: 

```kotlin
(Float.NaN).shouldBeWithinPercentageOf(Float.NaN, 1.0)         // => NaN should be in NaN..NaN
(Double.NaN).shouldBeWithinPercentageOf(Double.NaN, 1.0)       // => NaN should be in NaN..NaN
0.shouldBeWithinPercentageOf(0, Double.POSITIVE_INFINITY)      // => 0 should be in NaN..NaN
0L.shouldBeWithinPercentageOf(0L, Double.POSITIVE_INFINITY)    // => 0 should be in NaN..NaN
0f.shouldBeWithinPercentageOf(0f, Double.POSITIVE_INFINITY)    // => 0.0 should be in NaN..NaN
0.0.shouldBeWithinPercentageOf(0.0, Double.POSITIVE_INFINITY)  // => 0.0 should be in NaN..NaN
```

Cases where behavior differs from `x shouldBe (y plusOrMinus z.percent)`:

```kotlin
// Applies to both Double and Float

NEGATIVE_INFINITY shouldBe (NEGATIVE_INFINITY plusOrMinus 1.percent) // OK
POSITIVE_INFINITY shouldBe (POSITIVE_INFINITY plusOrMinus 1.percent) // OK

NEGATIVE_INFINITY.shouldBeWithinPercentageOf(NEGATIVE_INFINITY, 1.0) // -Infinity should be in -Infinity..NaN
POSITIVE_INFINITY.shouldBeWithinPercentageOf(POSITIVE_INFINITY, 1.0) // Infinity should be in NaN..Infinity
```

**Which version of Kotest are you using**
#3096 - branched from 56573a7

LeoColman:
Nice catch!

sksamuel:
Was fixed in 5.6.0

PR #3240, opened 2022-10-14
Feature/3099 nan tolerance

piotrb5e3:
Hi,

This is a proposed solution to #3099.

I unified the code of all tolerance matchers and improved handling around NaNs.

It has one issue: since the new code operates all on doubles, the old "Match close enough numbers" test fails, as the ranges of acceptable values are slightly different with double rounding. Not really sure what to do about it.

I'd also happily remove `FloatToleranceMatcher`, but I'm not sure if it's not part of the API.


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/numeric/ToleranceMatcher.kt
```diff
@@ -0,0 +1,33 @@
+package io.kotest.matchers.numeric
+
+import io.kotest.matchers.Matcher
+import io.kotest.matchers.MatcherResult
+import kotlin.math.abs
+
+open class ToleranceMatcher<T : Number>(private val expected: T, private val tolerance: Double) : Matcher<T> {
```
LeoColman, line 7: Interesting solution by using `Number`!

sksamuel (approved):

sksamuel:
Just need to fix the test @piotrb5e3 and I think we're good!

Kantis:
In my opinion we should revert the unification of ToleranceMatchers. It'll lead to some duplicate code, but it will let us be as correct as possible which I think is more important.

sksamuel:
I agree

LeoColman:
@piotrb5e3 Could you give maintainers access to your branch so we can fix the git history?

piotrb5e3:
@LeoColman Should I add specific people to the access list for this branch? _Allow edits by maintainers_ is already enabled.

Issue #2945, opened 2022-04-29
largeString shouldMatchFile "workflow.yaml"


jmfayard:
**Please describe the feature you'd like to see including any solutions in mind if you have any**

To compare very large strings, it's easier to write the expected content in a separate file.

https://github.com/krzema12/github-actions-kotlin-dsl/blob/6fc0b3f279ee7372e084545245a08d9e51f36317/library/src/test/kotlin/it/krzeminski/githubactions/IntegrationTest.kt#L25-L39

```kotlin
class IntegrationTest : FunSpec({
    setupShouldMatchFile(
        baseDirectory = File("src/test/resources/integration"),
        normalize = { it.replace("\r\n", "\n") }, // test shouldn't fail on CI running on windows
    )

    test("workflowTest : toYaml()") {
        workflowTest.toYaml() shouldMatchFile "workflowTest.yaml"
    }

}
```

When the test fails, the error message is:

```
Files differ in /Users/jmfayard/IdeaProjects/github-actions-kotlin-dsl/library/src/test/resources/integration
Expected: expected/workflowTest.yaml
Actual:     actual/workflowTest.yaml 
``` 

That tells me exactly what I need to use IntelliJ Diff tool, the perfect tool to check what happens and fix it.

<img width="1018" alt="github-actions-kotlin-dsl_–_workflowTest_yaml__et_New_Issue_·_kotest_kotest" src="https://user-images.githubusercontent.com/459464/166074129-14596fc5-934f-40f8-93ff-5b10ecc629ea.png">

The alternative to put the very long string inside the test is ugly

https://github.com/krzema12/github-actions-kotlin-dsl/blob/main/library/src/test/kotlin/it/krzeminski/githubactions/IntegrationTest.kt#L77-L150

Especially modifying the large expected string is hard.

My implementation is done at https://github.com/krzema12/github-actions-kotlin-dsl/blob/6fc0b3f279ee7372e084545245a08d9e51f36317/library/src/test/kotlin/it/krzeminski/githubactions/kotest/ShouldMatchFile.kt

I fall back to use `this shouldBe file.readText()` on the CI (for me GitHub Actions) because there I can't easily compare the content of the actual and the expected file.

sksamuel:
Hi. I'm not sure what you're asking for :)
Are you suggesting we add a "string" shouldBe file matcher?

jmfayard:
Yes that was the suggestion :)

nor-ek:
Can assign me

[...]

PR #3241, opened 2022-10-14
Added shouldMatchResource, shouldNotMatchResource matchers (#2945)

bzareba:


kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/resource/matchers.kt
```diff
@@ -0,0 +1,68 @@
...
+import io.kotest.matchers.shouldNot
+import java.io.File
+import java.net.URL
+import java.nio.file.Path
+import kotlin.contracts.contract
+import kotlin.io.path.createDirectories
+import kotlin.io.path.writeText
+
+infix fun String?.shouldMatchResource(path: String) {
+   contract {
```
LeoColman: fancy!

LeoColman (changes requested):
What should happen when resources are binaries or any format other than text?
Can we add some test cases to cover this?

bzareba:
@LeoColman Hi, I think we should let users compare string to any file. I've added test case to cover this. 

When it comes to binary files I think something like `ByteArray shouldMatchResource "/resource.bin"` would be a better fit.

Kantis:
Can we let the user provide a matcher to compare actual with resource content with? For instance, what if the resource contains a json blob and you want to compare specific fields of the json.

Current implementation could also see issues due to line endings

bzareba:
@Kantis I've added possibility to add custom matcher in form of matcher provider, so the usage is:
`givenValue.shouldMatchResource("/testResource.txt", ::startWith)`
or like this, enabling resource value modification before compare:
~`givenValue.shouldMatchResource("/testResource.txt") { s -> startWith(s.lowercase()) }`~

edit: I've added optional `ignoreLineSeparators` parameter so the syntax with lambda is:
`givenValue.shouldMatchResource("/testResource.txt", { s -> startWith(s.lowercase()) })`

bzareba:
I also added possibility to ignore line separators differences. I made it default behavior.

sksamuel:
@bzareba looks great. I think @LeoColman wanted a test on non-text resource and then this is good to merge.

bzareba:
Hi @sksamuel I've added the test, but I think `ByteArray` matchers would be better with binary files, so I just added them :D I've pushed them here, please let me know if they should go in the separate PR

bzareba:
I see there are some failing tests. That is probably due to autocrlf handling in git. I'll find a way to handle this, I'm thinking of creating the resources programmatically

bzareba:
> I'm thinking of creating the resources programmatically

I forgot that `.gitattributes` can be used to keep original line-ending for these files. It should work now. :)

LeoColman (approved):

PR #3248, opened 2022-10-17
include extraction of regex patterns defined within build gradle

### LINKED ISSUE #3247: Kotest should respect test `filter { }` defined in gradle.build.kts
Kotest does not include filters defined within build files

i.e. something like this doesn't get properly picked up.
```kotlin
tasks {
  useJUnitPlatform()
  withType<Test> {
    when(name) {
      "configIT" -> {
        filter {
          includeTestsMatching("*ConfigurationIT")
        }
      }
    }
  }
}
```

I put a debug point in there and realized that Gradle's `ClassMethodNameFilter.matcher.commandLineIncludePatterns` returned an empty list, whereas `ClassMethodNameFilter.matcher.buildScriptIncludePatterns`  returned the supplied pattern.

image

5.5.1

myuwono:
fixes #3247 

However, I'm not too sure how I could test this.. as everything in gradle internal is private. asking for @sksamuel for inputs


myuwono:
Superseded by #3257

PR #3252, opened 2022-10-19
Add missing brackets to example core matcher calls

xoliver:
Looking at the docs I noticed the brackets were missing here, so here's a fix. Thanks for your great work!


sksamuel:
Good spot thanks.

PR #3269, opened 2022-10-28
Use the original failure when ErrorCollector contains single failure

### LINKED ISSUE #3236: throw AssertionFailedError when shouldBe is the only failure in the assertSoftly block
When just using `shouldBe`:

```kotlin
val longStringA = "sjdigjiosgoidfguiohdfispdjgpiodjgiosdjgiosdjoijsdiogjsdiogjsdiojgiosdjgiosdjgiojsdoigjsdoigjisodjgoisdjgiosdjgoisdjoijgoisjg"
val longStringB = "sjdigjiosgoidfguiohdfispdjgpiodjgiosdjgiosdjoijsdiogjsdiogjsdiojgiosdjgxosdjgiojsdoigjsdoigjisodjgoisdjgiosdjgoisdjoijgoisjg"

longStringA shouldBe longStringB
```

kotest throws AssertionFailedError, which generates a nice "Click to see difference" button in IntelliJ:

image

However, when that same assert is wrapped with `assertSoftly`:

```kotlin
val longStringA = "sjdigjiosgoidfguiohdfispdjgpiodjgiosdjgiosdjoijsdiogjsdiogjsdiojgiosdjgiosdjgiojsdoigjsdoigjisodjgoisdjgiosdjgoisdjoijgoisjg"
val longStringB = "sjdigjiosgoidfguiohdfispdjgpiodjgiosdjgiosdjoijsdiogjsdiogjsdiojgiosdjgxosdjgiojsdoigjsdoigjisodjgoisdjgiosdjgoisdjoijgoisjg"
                
assertSoftly {
    // Assume more assertions here that pass

    longStringA shouldBe longStringB
}
```

kotest throws just regular plain AssertionError, which does not generate "Click to see difference":

image

I realize that this is normal when there are multiple exceptions, because AssertionFailedError can only cover one comparison. But if assertSoftly's only failure is `shouldBe` assertion, it would be really nice if it threw AssertionFailedError to get the "Click to see difference" button to show.

Kantis:
fixes #3236 

`AssertionFailedError` is JVM-specific, so had to move some stuff to expect/actuals.

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/desktopMain/kotlin/io/kotest/assertions/ErrorCollector.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/desktopMain/kotlin/io/kotest/assertions/ErrorCollector.kt
@@ -1,6 +1,23 @@
 package io.kotest.assertions
 
+import io.kotest.mpp.stacktraces
 import kotlin.native.concurrent.SharedImmutable
 
 @SharedImmutable
 actual val errorCollector: ErrorCollector = NoopErrorCollector
+
+actual fun ErrorCollector.collectiveError(): AssertionError? {
+   val failures = errors()
+   clear()
+   return if (failures.isNotEmpty()) {
+      if (failures.size == 1) {
+         AssertionError(failures[0].message).also {
+            stacktraces.cleanStackTrace(it)
+         }
+      } else {
+         MultiAssertionError(failures).also {
+            stacktraces.cleanStackTrace(it) // cleans the creation of MultiAssertionError
+         }
+      }
+   } else null
+}
--- a/kotest-assertions/kotest-assertions-shared/src/desktopMain/kotlin/io/kotest/assertions/ErrorCollector.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/desktopMain/kotlin/io/kotest/assertions/ErrorCollector.kt
@@ -9,15 +9,5 @@ actual val errorCollector: ErrorCollector = NoopErrorCollector
 actual fun ErrorCollector.collectiveError(): AssertionError? {
    val failures = errors()
    clear()
-   return if (failures.isNotEmpty()) {
-      if (failures.size == 1) {
-         AssertionError(failures[0].message).also {
-            stacktraces.cleanStackTrace(it)
-         }
-      } else {
-         MultiAssertionError(failures).also {
-            stacktraces.cleanStackTrace(it) // cleans the creation of MultiAssertionError
-         }
-      }
-   } else null
+   return failures.toAssertionError()
 }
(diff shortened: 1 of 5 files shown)


myuwono (approved):

kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/com/sksamuel/kotest/ErrorCollectorTests.kt
```diff
@@ -0,0 +1,21 @@
+package com.sksamuel.kotest
+
+import io.kotest.assertions.AssertionFailedError
+import io.kotest.assertions.assertSoftly
+import io.kotest.assertions.throwables.shouldThrow
+import io.kotest.core.spec.style.FreeSpec
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldBe
+
+class ErrorCollectorTests : FreeSpec(
```
myuwono, line 10: ah right there's no module for jsTest..

kotest-assertions/kotest-assertions-shared/src/desktopMain/kotlin/io/kotest/assertions/ErrorCollector.kt
```diff
@@ -1,6 +1,23 @@
...
+import io.kotest.mpp.stacktraces
 import kotlin.native.concurrent.SharedImmutable
 
 @SharedImmutable
 actual val errorCollector: ErrorCollector = NoopErrorCollector
+
+actual fun ErrorCollector.collectiveError(): AssertionError? {
+   val failures = errors()
+   clear()
+   return if (failures.isNotEmpty()) {
```
matejdro: Could this be written cleaner as a when or else if statement instead of nested if?

matejdro (approved):
Tested on my local machine and it seems to work fine. Made a comment for minor readability improvement.

kotest-assertions/kotest-assertions-shared/src/desktopMain/kotlin/io/kotest/assertions/ErrorCollector.kt
Kantis: It could, yeah.. I'll do that.

kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/com/sksamuel/kotest/ErrorCollectorTests.kt
Kantis: Not sure what you meant with this 🤪  Did you mean that we should add a `jsTest` to check that I didn't break the js/desktop error collector?

kotest-assertions/kotest-assertions-shared/src/desktopMain/kotlin/io/kotest/assertions/ErrorCollector.kt
Kantis: Done :)

PR #3271, opened 2022-10-30
Update tags.md

AnouarDoukkali:
wrong annotation name


Kantis:
When you update docs, please also update the docs for future versions in `documentation/docs`, for instance [here](https://github.com/kotest/kotest/blob/master/documentation/docs/framework/tags.md) for this particular page. When the next version is released, any updated docs will be cut and marked as 5.6, 5.7, etc.

AnouarDoukkali:
ok i will

PR #3280, opened 2022-11-03
Fix misspelled property name

martiner:


Kantis:
Please also update the docs for next version in https://github.com/kotest/kotest/tree/master/documentation/docs 🙏🏼

martiner:
@Kantis updated!

PR #3291, opened 2022-11-16
Check enums using equals method instead of scanning their properties

genuss:
Hi team!
I believe I found a bug in `shouldBeEqualToComparingFields` function if compared objects have enums as their properties.
Right now enums are treated as complex objects and compared by their fields recursively which I suppose is wrong and they should only be compared by equals method or referential equality (which is equivalent).
Check out the minimal example:
```kotlin
import io.kotest.matchers.equality.shouldBeEqualToComparingFields

enum class SimpleEnum { ONE, TWO }
enum class EnumWithProperties(val value: String) { ONE("one"), TWO("two"), }
data class EnumWrapper<E : Enum<E>>(val enum: E)

fun main() {
  val oneNoProperties = EnumWrapper(SimpleEnum.ONE)
  val oneWithProperties = EnumWrapper(EnumWithProperties.ONE)
  val twoNoProperties = EnumWrapper(SimpleEnum.TWO)
  val twoWithProperties = EnumWrapper(EnumWithProperties.TWO)

  assertSoftly {
    oneNoProperties shouldBeEqualToComparingFields twoNoProperties
    oneWithProperties shouldBeEqualToComparingFields twoWithProperties
  }
}
```
This gives the following output on latest version:
```
Exception in thread "main" java.lang.AssertionError: Expected EnumWrapper(enum=ONE) to equal EnumWrapper(enum=TWO)
 Using fields: enum
 Value differ at:
 1) enum:	
		value:
			expected:<"two"> but was:<"one">
```
Notice that first assertion is not failed although it should.

This PR fixes this bug and also changes an error message a little bit which now contains enum names instead of their properties. It also takes into account java enums which are different from kotlin ones as far as I understand.


sksamuel (approved):

sksamuel:
Looks great

genuss:
Hi, @sksamuel ! Looks like my PR couldn't get merged as some changes appeared in master and I'm not able to run pipelines. Could you rerun them, so merge will succeed this time?

sksamuel (approved):

PR #3293, opened 2022-11-16
Add testScope variable to TestScope

### LINKED ISSUE #3226: Access to TestScope.backgroundScope
[Coroutines 1.6.4](https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.6.4) introduces `TestScope.backgroundScope`.

Unfortunately, kotest only takes `coroutineContext` from TestScope, it does not expose TestScope anywhere, making new `backgroundScope` inaccessible.

Could this be exposed somewhere?

matejdro:
this fixes #3226

I've tried implementing `TestScope` at first, but it turns out it's a sealed interface, so it cannot be implemented outside coroutines module. Next best thing seems to be just adding `testScope` variable to the `TestScope` scope.

It's a bit awkward, beacuse variable name is the same than the interface name, but I figure that TestScope interface is mostly used as a receiver, so most people would not notice it.


sksamuel:
testScope is experimental so I am cautious about exposing it. Every time we've done this it's ended in tears (eg durations).

Since the only thing we need is a background scope that is cancelled, we can just do this ourselves since we already create a scope for tests ?

matejdro:
Half of the things in the coroutine library are experimental. As much as I love the coroutines, it seems like they leave many useful things experimental for years and they work just fine. So I would not put that much stock on that.

The main idea here is that more things might be added to the coroutines TestScope in the future, other than just the background scope. And thus we would have to add all of those things manually to kotest TestScope, making this already big interface bigger.

But this is just a speculation, maybe nothing else will get added. If you want, I can change the PR to only expose background scope.

sksamuel:
> Half of the things in the coroutine library are experimental. As much as I love the coroutines, it seems like they leave many useful things experimental for years and they work just fine. So I would not put that much stock on that.

Except we've fallen foul of experimental stuff many many times. I would take a guess that 5% of the issues in this repo are "I upgraded kotlin from 1.x to 1.y and now [this] doesn't work".

It's happened with durations, coroutines, compiler plugins. I don't want to fall into the same trap again.

> The main idea here is that more things might be added to the coroutines TestScope in the future, other than just the background scope. And thus we would have to add all of those things manually to kotest TestScope, making this already big interface bigger.

Agreed, this isn't ideal. 

Perhaps we add background scope and do the same thing as TestScope, and when TestScope is no longer experimental switch to that?

matejdro:
What do you think about passing coroutines TestScope as a coroutine context element? That way we can get rid of most passing and extra code on the kotest TestScope and we can implement backgroundScope using extensions.

matejdro:
@sksamuel here is an alternative approach that does not expose TestScope directly: https://github.com/kotest/kotest/pull/3315. I can close this PR if you prefer the other one.

matejdro:
This is still a problem and PR is ready

matejdro:
I think this PR can be closed in favor of #3315?

Issue #3275, opened 2022-10-31
switching kotest.framework.dump.config to false  in kotest.properties has no effect


AnouarDoukkali:
**Which version of Kotest are you using**
5.5.3

 **property kotest.framework.dump.config=false**  still print the configuration

---

Change [DumpConfigInterceptor](https://github.com/kotest/kotest/blob/master/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/interceptors/DumpConfigInterceptor.kt#L26) to only dump config if `kotest.framework.dump.config` is set to `true` (ignoring case)

Kantis:
How are you setting the property?

AnouarDoukkali:
@Kantis in kotest.properties file

Kantis:
Where did you put the file? Can you verify that it's actually in the build/target resource directory?

Kantis:
Oh I just realized.. It's not a boolean property. Just by having the property exist the framework will dump the config. If you want to stop dumping just remove the property.

sksamuel:
Maybe we should change it to be boolean ?

AnouarDoukkali:
@sksamuel I think yes. since the documentation suggest to set the value to true

Kantis:
Updated the original description with the required change.

PR #3294, opened 2022-11-17
changing  DumpConfigInterceptor  to dump config when systemProperty set to true (#3275)

AnouarDoukkali:
Change DumpConfigInterceptor to only dump config if kotest.framework.dump.config is set to true


Kantis (changes requested):
Please add some tests. For instance, where the property exists but has no value, is set to false, is set to something not bool-ish, set to true.

AnouarDoukkali (commented):
hello @Kantis , i have some questions please if you don't mind  

- should i use Kotest when testing Kotest project itself ?

- when writing unit test for kotest  classes or objects that depends on other classes , should i mock dependencies with external lib like mockK ?

-how to test private function like syspropEnabled in this case ?

-Im i encouraged to change other contributors code to simplify testing or reduce inter-class dependencies ?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/interceptors/DumpConfigInterceptor.kt
```diff
@@ -23,6 +23,7 @@ internal object DumpConfigInterceptor : EngineInterceptor {
       return execute(context)
    }
 
-   private fun syspropEnabled() =
-      syspropOrEnv(KotestEngineProperties.dumpConfig) == true
+   fun syspropEnabled() =
+      syspropOrEnv(KotestEngineProperties.dumpConfig).toBoolean()
```
AnouarDoukkali, line 27: the function will now return only true if the system property is set to true  , or false when it is set to false or not set at all

Kantis:
> hello @Kantis , i have some questions please if you don't mind
> 
>     * should i use Kotest when testing Kotest project itself ?
> 

Sorry for the delayed answer.. It's not always easy to know how to test things.. I pushed a commit which is how I would test things, please have a look at it and see if you have further questions. Perhaps there's similar ways to test for JS and Native as well. Do you use kotlin multiplatform?

>     * when writing unit test for kotest  classes or objects that depends on other classes , should i mock dependencies with external lib like mockK ?

Mocks are generally not used. I tried to find some examples where they _were used_ but found next to none.. 

> 
> -how to test private function like syspropEnabled in this case ?
> 
If you check the commit, you can see we don't actually need to (and probably dont _want to_ either) test the private function, we can test the entire thing.. :) 

> -Im i encouraged to change other contributors code to simplify testing or reduce inter-class dependencies ?

Not sure, feel free to ask on a case by case basis.

AnouarDoukkali:
@Kantis thank you for the explanations 👍🏻 i see that your test is more effective and cover all possible values. i will try my best to learn :)

Kantis (approved):

PR #3300, opened 2022-11-24
Invoking After/before each for dynamic tests

### LINKED ISSUE #3290: beforeEach/afterEach not working with data driven testing
**Which version of Kotest are you using?** 5.3.2

---

According to the [data driven testing docs](https://kotest.io/docs/framework/datatesting/data-driven-testing.html#callbacks) (emphasis added by me):

> If you wish to have before / after callbacks in data-driven tests, then you can use the standard `beforeTest` / `afterTest` support. Every test created using data-driven testing acts the same way as a regular test, so **all standard callbacks work as if you had written all the test by hand**.

When I use `beforeTest`/`afterTest` and `beforeContainer`/`afterContainer`, the callbacks are invoked as expected. However, when I use `beforeEach`/`afterEach`, the callbacks are not invoked as expected:

```kt
class DataDrivenCallbackTest : DescribeSpec({
  beforeTest {
    // EXPECTED: Invoked 4 times
    // ACTUAL: Invoked 4 times
  }
  beforeContainer {
    // EXPECTED: Invoked 1 time
    // ACTUAL: Invoked 1 time
  }
  beforeEach {
    // EXPECTED: Invoked 3 times
    // ACTUAL: Invoked 0 times
  }

  afterEach {
    // EXPECTED: Invoked 3 times
    // ACTUAL: Invoked 0 times
  }
  afterContainer {
    // EXPECTED: Invoked 1 time
    // ACTUAL: Invoked 1 time
  }
  afterTest {
    // EXPECTED: Invoked 4 times
    // ACTUAL: Invoked 4 times
  }

  describe("My container") {
    withData(
      "My test 1",
      "My test 2",
      "My test 3"
    ) {
      // test goes here
    }
  }
})
```

Kantis:
fixes #3290 

Invokes `beforeEach` / `afterEach` between each case of a data-driven test (`withData`)

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestExtensions.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestExtensions.kt
@@ -77,7 +77,7 @@ internal class TestExtensions(private val registry: ExtensionRegistry) {
          }.mapError { ExtensionException.BeforeContainerException(it) }.exceptionOrNull()
       } + be.mapNotNull {
          runCatching {
-            if (testCase.type == TestType.Test) it.beforeEach(testCase)
+            if (testCase.type in arrayOf(TestType.Test, TestType.Dynamic) ) it.beforeEach(testCase)
          }.mapError { ExtensionException.BeforeEachException(it) }.exceptionOrNull()
       } + bt.mapNotNull {
          runCatching {
@@ -120,7 +120,7 @@ internal class TestExtensions(private val registry: ExtensionRegistry) {
          }.mapError { ExtensionException.AfterContainerException(it) }.exceptionOrNull()
       } + ae.mapNotNull {
          runCatching {
-            if (testCase.type == TestType.Test) it.afterEach(testCase, result)
+            if (testCase.type in arrayOf(TestType.Test, TestType.Dynamic)) it.afterEach(testCase, result)
          }.mapError { ExtensionException.AfterEachException(it) }.exceptionOrNull()
       }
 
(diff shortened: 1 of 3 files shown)


Kantis:
Can this go in 5.6?

myuwono (approved):

sksamuel:
One issue here might be that anyone using data driven tests could now see behavior change on them.

aSemy:
That's a good point @sksamuel.

I have also been thinking about property based tests, which has a separate listener system. It would be nice to unify the before/after checks, so I don't have to set up both the `PropTestListener` and data-driven listeners. Currently I use both in the same test class, even though the 'reset' task is the same. 

It would be nice only define an 'after' listener once, and this could be triggered on the conclusion of tests, regardless of type.

Perhaps instead, new listeners are created - `beforeDynamicTest {}` and `afterDynamicTest {}`. This would be a clean start, preventing behaviour changes. And conceptually it makes sense to me.

* `beforeTest {}` - to be triggered before an explicitly defined test (`FunSpec` example: `test("my defined test")`)
* `beforeDynamicTest {}` - to be triggered before a test that Kotest will dynamically based, either based on hardcoded data (data driven tests, `withData {}` ) or Arbs (property based tests `checkAll {}`)

Kantis:
I've thought about that as well, and I think in this case it's worth doing a change in behavior. Just in the past month there's been three reports that reached us where people have used `beforeEach/afterEach` together with `withData` and experienced unexpected behavior (before/after hooks not running).  I've had this issue at work as well, and we sort of avoided `withData` after that.

sksamuel:
I know there's been a few questions about this recently, but we can't just introduce such a breaking change as a bug fix release.

we could change it for 5.6 ?

Alternatively, we can just fix it properly in 6.0 ? I will start creating some issues for things we need in 6.0 anyway.

sksamuel:
I'd like to also have property testing become more of a first class citizen in kotest, but we need to find a way to make that work without meaning property tests _require_ kotest.

Kantis:
I agree, it should ideally go into a 6.0 release to do things properly and definitely not a patch release. 🙂

sksamuel:
Think we can probably close this @Kantis  ?

Kantis:
Made obsolete by #4599

PR #3313, opened 2022-12-12
Upgrade Kotlin

geekiyer:
[Kotlin 1.7.20](https://kotlinlang.org/docs/whatsnew1720.html)


sksamuel:
Kotest uses 1.60 to ensure that everyone can use it. There's no advantage to go to 1.7 right now I think.

Issue #3306, opened 2022-11-28
HTML report is generated outside of the build directory


aSemy:
I've configured the [Kotest HTML report](https://kotest.io/docs/extensions/html_reporter.html). However, the files are generated outside of the build directory.

<img width="306" alt="image" src="https://user-images.githubusercontent.com/897017/204262635-a8c9cc3e-7920-4895-a37e-cf06ebac24fb.png">

I tried configuring `HtmlReporter` to set the directory, but this didn't work

```kotlin
// https://kotest.io/docs/framework/project-config.html
internal class KotestConfig : AbstractProjectConfig() {

    override val specExecutionOrder = SpecExecutionOrder.Lexicographic

    override fun extensions(): List<Extension> = listOf(

        // https://kotest.io/docs/extensions/junit_xml.html
        JunitXmlReporter(
            includeContainers = false,
            useTestPathAsName = true,
        ),

        // https://kotest.io/docs/extensions/html_reporter.html
        HtmlReporter(
            // This doesn't seem to work:
            "./build/reports/test"
        ),
    )
    }
}
```

As a workaround I can configured the Gradle tasks to set the `gradle.build.dir` property.

```kotlin
// build.gradle.kts

tasks.withType<Test>().configureEach {

    // required by Kotest JUnit XML https://github.com/kotest/kotest/issues/3306
    systemProperty("gradle.build.dir", project.buildDir)

    // disable the Gradle reports - Kotest will generate them
    reports.html.required.set(false)
    reports.junitXml.required.set(false)
}
```

### Versions

* Kotlin 1.7.21
* Kotest 5.5.4
* Gradle 7.5.1

sksamuel:
Fixed in 5.5.5

PR #3326, opened 2022-12-29
Fix for issue-3306 - Removing default location for htmlReporter and using the default value from the constructor

arvarik:
## Context

#3306 

1) The `outputDir` customizable field in HTMLReporter does not work if the `gradle.build.dir` system property is not set.

2) Running tests using the kotest sidebar menu will produce reports in top level directory even when `gradle.build.dir` system property is set in `build.gradle` and a different `outputDir` is used (probably because the kotest sidebar execution context doesn't have the `gradle.build.dir` system property set).

## Root Cause

In HTMLReporter, if the `gradle.build.dir` system property [is not set](https://github.com/kotest/kotest/blob/master/kotest-extensions/kotest-extensions-htmlreporter/src/jvmMain/kotlin/io/kotest/extensions/htmlreporter/HtmlReporter.kt#L77), a DefaultLocation is used and ignores the `outputDir` field

## Changes Made
Removing the DefaultLocation variable and replacing it's usage with `outputDir`.

## Expected Result
When a user does not specify an `outputDir`, the default string from the constructor will be used and there is no change in logic.

When a user does specify an `outputDir`, the html report will be placed within the specified `outputDir`


sksamuel (approved):

sksamuel:
Thanks for fix!

Issue #2916, opened 2022-04-08
Use 'language injection' on assertions, so embedded languages are easier to use in IntelliJ


aSemy:
I'd like the JSON Path matchers to be easier to use. IntelliJ has a nice feature which will apply highlighting and formatting to a string if it's specially marked 
https://www.jetbrains.com/help/idea/using-language-injections.html

It would be very convenient if Kotest applied this to the JSONPath assertions.

#### Example language injection

```kotlin
import org.intellij.lang.annotations.Language

@Language("JSONPath")
val myJsonPath = "$.store.book[?(@.price < 10)].title"

// language=JSONPath
val myInvalidJsonPath = "$.store.book[(@.length-1)]title"
```

Note the invalid path has a red error highlight: it's missing a `.`.

<img width="471" alt="image" src="https://user-images.githubusercontent.com/897017/162420595-cde5cf0c-1d8f-46f0-92f7-b1445616d061.png">

#### Current usage / workaround

These can be applied locally, which is a bit more inconvenient because it requires repetitive commenting

```kotlin
response.body?.shouldContainJsonKey(   /* language=JSONPath */ "$.store.book[(@.length-1)].title")
response.body?.shouldNotContainJsonKey(/* language=JSONPath */ "$.store.book[(@.length-1)].age")
```

<img width="816" alt="image" src="https://user-images.githubusercontent.com/897017/162419065-d89270c6-939d-4d1a-ab33-18370c914a3c.png">

(I tried setting up something with `@Language("JsonPath")` but it didn't work - maybe it's possible with [external annotations](https://www.jetbrains.com/help/idea/external-annotations.html)?)

#### Suggestion

If the `@Language(...)` annotation is added to Kotest, then all users will automatically get the benefits (at least if their IDE supports it).

Change this method: 

https://github.com/kotest/kotest/blob/b832d4941ecc45074bcd6dea3f24d16553f2fc21/kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/keyvalues.kt#L10-L16

to be like this:

```kotlin
import org.intellij.lang.annotations.Language

inline fun <reified T> String?.shouldContainJsonKeyValue(
    @Language("JSONPath") path: String, 
    value: T,
) //...
```

(I'm not sure if adding the `// language=JSONPath` in Kotest will work the same as adding the annotation - the annotation seems more reliable from my brief experiments so I recommend that.)

[...]

PR #3328, opened 2023-01-03
IJ language injection

aSemy:
Partial fix for #2916 

Re-implements `@Language` to be Kotlin-multiplatform, [using workaround suggested by JetBrains](https://github.com/JetBrains/java-annotations/issues/34).

This injection makes the assertions easier to use, as IntelliJ will provide highlighting, autocompletion, and quick fixes ([read more about language injection](https://www.jetbrains.com/help/idea/using-language-injections.html)).

### TODO

- [ ] Verify that there is no clash on JVM between Kotest's `@Language` and IntelliJ's implementation

### Examples

image

image

### Further improvements

Languages aren't injected into 

* receivers https://youtrack.jetbrains.com/issue/KTIJ-12951/
* infix parameters 
   image

Perhaps the Kotest IJ plugin could correct implement the language injections to work with Kotlin.

### Related

* https://github.com/kotest/kotest-intellij-plugin/pull/218 - this PR tried to add the language injections on a per-function basis, but I don't think it worked once the plugin was released.

### DIFF
--- a/buildSrc/src/main/kotlin/kotest-multiplatform-library-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotest-multiplatform-library-conventions.gradle.kts
@@ -4,3 +4,14 @@ plugins {
    id("kotest-native-conventions")
    id("kotest-publishing-conventions")
 }
+
+kotlin {
+   sourceSets {
+      all {
+         languageSettings {
+            optIn("kotlin.RequiresOptIn")
+            optIn("io.kotest.common.KotestInternal")
+         }
+      }
+   }
+}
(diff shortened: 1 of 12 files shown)


sksamuel:
Can we just remove the @Language kotest annotation ?

aSemy:
> Can we just remove the `@Language` kotest annotation ?

Sorry, I don't understand...

IntelliJ will inject a language with Strings annotated with `@org.intellij.lang.annotations.Language`, but that annotation is JVM only. Fortunately, IntelliJ only matches the annotation based on the package + class name. So, to make `@Language` compatible with Kotlin Multiplatform, I copy-pasted it into Kotest ([as suggested by JetBrains](https://github.com/JetBrains/java-annotations/issues/34)).

An alternative to this approach would be 

* Create an annotation, named something like `@KotestLangugage`, in `src/commonMain/kotlin/io/kotest/common/`
* update the Kotest IntelliJ plugin to use the same language-injection logic as with the original `@Language` annotation.

However I'm reluctant to do that, because making something work across two projects is fiddly, and the current Kotest IJ Plugin [has language injection](https://github.com/kotest/kotest-intellij-plugin/pull/218) that doesn't seem to work 🤷‍♀️

sksamuel:
I'm happy to try this.

sksamuel (approved):

PR #3334, opened 2023-01-14
Replace BigDecimal double constructor usages with toBigDecimal extension

### LINKED ISSUE #3282: `Arb.bigDecimal` should generate decimals which terminate with all kinds of terminal digit
Currently the `BigDecimal` arb generates decimals that _always_ terminate the decimal part with 0 or 5. I think it would be good if all digits can occur as terminal digit.

> Why does this happen?

IEEE double point precision specifies the decimal part as the sum of (1/2)^-n, where n is [1, 52], which will always terminate with a 5 (if included) or 0 if excluded.

The `BigDecimal` docs seem to suggest that using the `BigDecimal(val: String)` constructor is the preferred way to go, which circumvents this by converting the double to a String first. Using Kotlin's `Double.toBigDecimal()` will do that.
```kotlin
/**
 * Returns the value of this [Double] number as a [BigDecimal].
 *
 * The number is converted to a string and then the string is converted to a [BigDecimal].
 */
@SinceKotlin("1.2")
@kotlin.internal.InlineOnly
public inline fun Double.toBigDecimal(): BigDecimal = BigDecimal(this.toString())
```

IvanPavlov1995:
Closes #3282


sksamuel (approved):
Nice!

PR #3336, opened 2023-01-16
Fix sequence matchers for constrained sequences

### LINKED ISSUE #3335: Sequence.shouldContainAll fails incorrectly on constrained once sequences
The following code fails with "This sequence can be consumed only once." message instead of human-readable one.
```kotlin
test("Arb.bigDecimal should generate random terminal digit") {
   fun BigDecimal.lastDigit() = stripTrailingZeros().toString().last().digitToInt()

   Arb.bigDecimal()
      .take(100)
      .map { it.lastDigit() }
      .shouldContainAll(0)
}
```

This happens because matcher tries to consume sequence twice while generating message https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt#L426

Note: the generated message is "Sequence should contain all of " plus a string based on `value` instead of `ts` which seems not correct

IvanPavlov1995:
Closes #3335


sksamuel (approved):

sksamuel:
Very thorough thanks!

sksamuel:
Looks like a test is failing

IvanPavlov1995:
@sksamuel Do you know any idea why this test can be affected?
```
> Task :kotest-property:jvmTest

com.sksamuel.kotest.property.CheckCoverageTest[jvm]

  Test should throw if coverage did not meet expectations[jvm] FAILED

  java.lang.AssertionError: Expected a throwable, but nothing was thrown.

  Caused by: java.lang.AssertionError: Expected a throwable, but nothing was thrown.
```
On my machine, this test seems non-deterministic or just flaky in both this branch and master branch
image
image

Edit: probably the test is not affected by changes and just fails sometimes

Edit: fyi I just wrapped a test in `repeat` block to launch it multiple times

IvanPavlov1995:
After additional investigation it seems like I found the issue.
I've added simple printing statement to the test
`println("IntNum: $a; zeroClassifications: ${classifications()["zero"]}; >zeroClassifications:${classifications()[">zero"]}")`
The last output for failing tests is
`IntNum: 1061516830; zeroClassifications: null; >zeroClassifications:1000`
There was no zero generated and it was not classified at all. Seems not related to the current PR, but definitely a bug

IvanPavlov1995:
I decided to move fix to a separate PR https://github.com/kotest/kotest/pull/3344

PR #3344, opened 2023-01-17
Make checkCoverage checking against provided pairs

IvanPavlov1995:
Issue: if there was no classification at all, it is not checked against the expectations.
This is caused by the fact that iteration is performed over `context.classifications()` which doesn't have an entry if there was no classification at all for some label.

Proposed solution: 
Iterate over `pairs` and check classification from `context.classifications()`. It should be safe to do that, because previous solution ignores all keys not included into `pairs` due to calculation algorithm. `actual` is non-negative and `required` is zero, so `actual < required` is always `false` for such cases


sksamuel (approved):

sksamuel:
Nice!

PR #3348, opened 2023-01-19
Support {  } lambdas as lazy clues

### LINKED ISSUE #3342: withClue({ "lazy message" }) { ... } logs "() -> kotlin.String"

```kotlin
withClue({ "lazy message" }) {
    2 shouldBe 3
}
```

fails with `() -> kotlin.String`, and `lazy message` is never printed:

```
() -> kotlin.String
expected:<3> but was:<2>
Expected :3
Actual   :2
<Click to see difference>
```

The reason is that the compiler calls `withClue(Any, ...)` function, so `() -> String` is printed.

I would say in most of the cases, lazy clue is needed so it is computed only in case of failure, and the current API makes is hard to use.

`withClue(lazy { ... }) { ... }` is too verbose.

WDYT of making `withClue(Lazy)` as deprecated, and replacing it with `withClue(() -> Any, ...)`?

vlsi:
Fixes https://github.com/kotest/kotest/issues/3342

The PR adds

```kotlin
withClue({ "message: $..." }) { ...assertions... }
{ "message: $:..." }.asClue { ...assertions... }
```

### DIFF
--- a/documentation/docs/assertions/clues.md
+++ b/documentation/docs/assertions/clues.md
@@ -4,9 +4,11 @@ slug: clues.html
 ---
 
 :::note
-Clues only work if you are using the Kotest test framework in conjuction with the assertions library
+Clues only work if you are using the Kotest assertions library
 :::
 
+A rule of thumb is that a failing test should look like a good bug report.
+In other words, it should tell you what went wrong, and ideally why it went wrong.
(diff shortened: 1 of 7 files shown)


sksamuel (approved):
nice!

PR #3352, opened 2023-01-19
Skip quoting top-level strings to make outputs copy-paste friendly

### LINKED ISSUE #3346: idea: avoid quoting strings to make outputs copy-paste friendly
## Current behaviour

```kotlin
"hello\nline2" shouldBe "world\nline3"
```

fails with

```
expected:<"world
line3"> but was:<"hello
line2">
<Click to see difference>

io.kotest.assertions.AssertionFailedError: expected:<"world
line3"> but was:<"hello
line2">
```

kotest adds excessive quotes to assertion output

Then, when I click "to see difference", the diff UI includes quotes around the strings.
That makes copy-pasting "actual" output harder, as I need to explicitly remove the quotes.

## Expected behaviour

If I use JUnit5 Assertions, then the quotes are not there, and I can select full contents and copy it:

junit5 shows nice assertion UI

vlsi:
This makes "expected: hello got: world" easier to see in "show diff" in IDE

Fixes https://github.com/kotest/kotest/issues/3346


vlsi:
Another observation: the current `StringPrint.printed` might produce weird results when the source string contains quotes.

What do you think if quoting is completely removed?
An alternative option might be keep quoting, however, escape special characters (e.g. convert newline to `\n`, `"` to `\"`, `\` to `\\` and so on)

sksamuel:
> 

How would it look with quoting completely removed? Do you have an example output ?

vlsi:
```
io.kotest.assertions.AssertionFailedError: expected:<expected:<"Some name that is wrong"> but was:<"John Doe">> but was:<expected:<Some name that is wrong> but was:<John Doe>>

io.kotest.assertions.AssertionFailedError: expected:<expected:<"b"> but was:<"a">> but was:<expected:<b> but was:<a>>

io.kotest.assertions.AssertionFailedError: expected:<expected:<1> but was:<"1">> but was:<expected:<1> but was:<1>>

  io.kotest.assertions.AssertionFailedError: Throwable should have message:
  "a" should not equal "a"

  Actual was:
  a should not equal a
  expected:<"a" should not equal "a"> but was:<a should not equal a>

  io.kotest.assertions.AssertionFailedError: expected:<Values differed at keys a
  expected:<{
    "a" = "expected"
  }> but was:<{
    "a" = "actual"
  }>> but was:<Values differed at keys a
  expected:<{
    a = expected
  }> but was:<{
    a = actual
  }>>

  io.kotest.assertions.AssertionFailedError: expected:<Values differed at keys 1
  expected:<{
    "1" = [("2", [("3", "bar")])]
  }> but was:<{
    "1" = [("2", [("3", [("4", ["foo"])])])]
  }>> but was:<Values differed at keys 1
  expected:<{
    1 = [(2, [(3, bar)])]
  }> but was:<{
    1 = [(2, [(3, [(4, [foo])])])]
  }>>

  io.kotest.assertions.AssertionFailedError: expected:<Ship(
    class         =  "Queen Elizabeth"
    displacement  =  65000L
    leadShip      =  true
    name          =  "HMS Queen Elizabeth"
    pennant       =  "R08"
  )> but was:<Ship(
    class         =  Queen Elizabeth
    displacement  =  65000L
    leadShip      =  true
    name          =  HMS Queen Elizabeth
    pennant       =  R08
  )>
```

Frankly speaking, I think non-quoted strings look good.

sksamuel:
The original intent I think was to make it clear when 1 is a int vs a string, the "1" makes it clear.
It's also easier to copy n paste when it has the "" around it ?

vlsi:
> The original intent I think was to make it clear when 1 is a int vs a string, the "1" makes it clear.

It works only in case when comparing String with Int.
Quoting does not work when comparing String with StringBuilder, and it does not help when comparing Int with BigInteger.

>It's also easier to copy n paste when it has the "" around it ?

It does not.
You put quotes, and you ignore to escape the contents. That produces ambiguous output, and it is never useful.

Consider the following:

```kotlin
      """
      hello
      world
      "foo" "ba\r"
      """.trimIndent() shouldBe "tes\\t"
```

```
expected:<"tes\t"> but was:<"hello
world
"foo" "bar"">
```

The quotes around do not help since they miss escaping the contents.
On top of that, if you copy `"tes\t"`, then you would get a different string since `\t` would be interpreted as a tab character.

See `quotes` in https://truth.dev/comparison.html#failure-messages

sksamuel:
Ok I think in your example, the quotes don't help yeah.

In this example though, if you wanted to copy the data class completely, it's better with the quotes?

```kotlin
  io.kotest.assertions.AssertionFailedError: expected:<Ship(
    class         =  "Queen Elizabeth"
    displacement  =  65000L
    leadShip      =  true
    name          =  "HMS Queen Elizabeth"
    pennant       =  "R08"
  )> but was:<Ship(
    class         =  Queen Elizabeth
    displacement  =  65000L
    leadShip      =  true
    name          =  HMS Queen Elizabeth
    pennant       =  R08
  )>
```

Could we get the best of both worlds ?

Kantis:
I'll close this until we agree on how it should work. Once we have a ready proposal we can revisit this PR

PR #3354, opened 2023-01-19
style: simplify NeverNullMatcher

vlsi:
This is PR makes the code smaller and easier to understand.

By the way, can you clarify why do you use `invoke` methods on `companion object` instead of a regular constructor?


sksamuel:
> By the way, can you clarify why do you use `invoke` methods on `companion object` instead of a regular constructor?

Do you have a specific example ?

vlsi:
> Do you have a specific example ?


For instance, MatcherResult objects are created via invoke calls rather than constructor calls

sksamuel:
It might have just been habit, I don't think there was any particular reason behind it.

sksamuel (approved):

PR #3358, opened 2023-01-25
Support binding properties to specific Arbs when doing reflective binding

Kantis:
Adds the method: 
```kotlin
inline fun <reified T : Any> Arb.Companion.bind(builder: ProvidedArbsBuilder.() -> Unit): Arb<T>
```

Which lets us construct reflective arbs while providing arbs to be used for various classes **and properties**. Previously this capability existed but one could only provide a default arb per class. This becomes inconvenient for commonly used types like `Int`, `String`, etc.

The new method proposes a builder syntax, like:

```kotlin
   "Bind using properties" {

      data class Person(val id: Int, val name: String)
      data class Family(val name: String, val persons: List<Person>)

      checkAll(Arb.bind<Family> {
         // Create a binding for the `name` property on Family to always 
         // generate a String prefixed with Flanders 
         bind(Family::name to Arb.string().map { s -> "Flanders-$s" })
     
         // Bind the `id` property of `Person` to generate positive ints
         bind(Person::id to Arb.positiveInt())
  
         // Binding using class is also still possible. 
         // Included for completeness but completely useless for the test.
         bind(Double::class to Arb.double())
      }) { family ->
         family.name.shouldStartWith("Flanders-")
         family.persons.forAll {
            it.name.shouldNotStartWith("Flanders-")
            it.id shouldBeGreaterThan 0
         }
      }

   }
```

### Work-around 
If one wishes to achieve the same result with currently available methods, you would have to first create the `Arb<Person>` and explicitly bind it to `Person::class` during reflective binding, like:

```kotlin
   "Bind WITHOUT using properties" {

      data class Person(val id: Int, val name: String)
      data class Family(val name: String, val persons: List<Person>)
    
      val personArb = Arb.bind<Person>(mapOf(Int::class to Arb.positiveInt())
 
      checkAll(Arb.bind<Family>(mapOf(Person::class to personArb)) { family ->
         family.name.shouldStartWith("Flanders-")
         family.persons.forAll {
            it.name.shouldNotStartWith("Flanders-")
            it.id shouldBeGreaterThan 0
         }
      }

   }
``` 
In a scenario with deeper nesting or more type-reuse this might be incredibly inconvenient.

### DIFF
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/targetDefaultForClassName.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/targetDefaultForClassName.kt
@@ -16,6 +16,8 @@ import java.time.YearMonth
 import java.time.ZonedDateTime
 import java.util.Date
 import kotlin.reflect.KClass
+import kotlin.reflect.KParameter
+import kotlin.reflect.KProperty1
 import kotlin.reflect.KType
 import kotlin.reflect.full.isSubclassOf
 import kotlin.reflect.javaType
@@ -26,7 +28,11 @@ import kotlin.reflect.typeOf
 @Deprecated("This logic has moved to ArbResolver and this function will be removed in 5.6. Since 5.5")
 actual inline fun <reified A> targetDefaultForClass(): Arb<A>? = targetDefaultForType(type = typeOf<A>()) as Arb<A>?
 
-fun targetDefaultForType(providedArbs: Map<KClass<*>, Arb<*>> = emptyMap(), type: KType): Arb<*>? {
+fun targetDefaultForType(
+   providedArbs: Map<KClass<*>, Arb<*>> = emptyMap(),
+   arbsForProps: Map<KProperty1<*, *>, Arb<*>> = emptyMap(),
+   type: KType
+): Arb<*>? {
    when (type) {
       typeOf<Instant>(), typeOf<Instant?>() -> Arb.instant()
       typeOf<Date>(), typeOf<Date?>() -> Arb.javaDate()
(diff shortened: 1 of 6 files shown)


Kantis:
@myuwono: WDYT?  Would it be better to expand the `ArbResolver` interface with property-handling capabilities, and make it possible to supply that when creating arbs using reflection?

myuwono:
That's a really good idea @Kantis ! hmm I think as long as we can get the usage to be ergonomic enough this is going to be really helpful. Can you supply perhaps an example code for the usage in here? Keen to see how it looks like.

Kantis:
@myuwono I updated the original PR comment to describe the proposed API properly

Kantis:
Still interested in feedback on this one

myuwono:
I think this is really awesome!

myuwono (approved):

sksamuel:
Can we get this ready for 5.7, would love to include it.

Kantis:
On vacation for another week. I could probably finish it up first week in September if that's soon enough :)

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/targetDefaultForClassName.kt
```diff
@@ -26,7 +28,11 @@ import kotlin.reflect.typeOf
 @Deprecated("This logic has moved to ArbResolver and this function will be removed in 5.6. Since 5.5")
 actual inline fun <reified A> targetDefaultForClass(): Arb<A>? = targetDefaultForType(type = typeOf<A>()) as Arb<A>?
 
-fun targetDefaultForType(providedArbs: Map<KClass<*>, Arb<*>> = emptyMap(), type: KType): Arb<*>? {
+fun targetDefaultForType(
```
Kantis, line 31: Should this function be `@KotestInternal`?

PR #3369, opened 2023-02-02
Trying to remedy flaky test

Kantis:


sksamuel (approved):

kotest-tests/kotest-tests-test-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/Test1.kt
```diff
@@ -1,11 +1,35 @@
 package com.sksamuel.kotest.parallelism
 
+import com.sksamuel.kotest.parallelism.ProjectConfig.parallelism
 import io.kotest.core.spec.style.FunSpec
+import kotlinx.coroutines.sync.Semaphore
 
 class Test1 : FunSpec({
-   repeat(10) { k ->
+   threads = parallelism
```
Kantis: Should this be required? I noticed that w/o setting it, I'd just get 5-6 concurrent threads. 

The docs [here](https://kotest.io/docs/framework/project-config.html#parallelism) suggests that parallelism is used to run _specs_ in parallel, not tests?

Kantis: If I change the test to create 10 specs and attempt to run them in parallel, w/o affinity, then it keeps failing as well. Since we create 10 single-threaded dispatchers [here](https://github.com/kotest/kotest/blob/master/kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/concurrency/FixedThreadCoroutineDispatcherFactory.kt), instantiated from [here](https://github.com/kotest/kotest/blob/master/kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/concurrency/DefaultCoroutineDispatcherFactory.kt#L7), and we're selecting a dispatcher at random out of these 10 for each of the 10 specs, we'll obviously get some collisions where two specs attempt to use the same dispatcher

Kantis: @sksamuel would appreciate your input here 🙏🏼 

I think we need to either change the way we schedule things or just change the test so it doesn't expect everything to run in parallel just because the parallelism is high enough to actually do so.

sksamuel: Yeah I don't think there's any guarantees you'll 100% parallelism just by setting the thread count high enough. It just adds more dispatchers to the pool to select from.

I guess we could get cleverer by picking a dispatcher from a pool of available dispatchers instead of random / round robin.

PR #3375, opened 2023-02-04
Fix team city output for dynamic parents

### LINKED ISSUE #3341: Nested Data Driven Tests is not displayed as nested in Intellij
Kotest version: 5.5.4
Intellij Idea version: 2022.3.1 (Ultimate Edition)
Kotest intellij plugin: 1.2.63-223.4884.69-EAP-SNAPSHOT
Kotlin version:1.8.0

When I run the following nested data driven tests in intellij, the tests are not shown as nested in the `Run` view

```
package com.example

import io.kotest.datatest.withData

class MyTest: FunSpec({
    context("each service should support all http methods") {

        val services = listOf(
            "http://internal.foo",
            "http://internal.bar",
            "http://public.baz",
        )
        val methods = listOf("GET", "POST", "PUT")
        withData(services) { service ->
            withData(methods) { method ->
                method shouldNotBe ""
            }
        }
    }
})
```
Here is my `Run` view.
my

I expect the run view should be like the following as mentioned in https://kotest.io/docs/framework/datatesting/nested-tests.html
expected

sksamuel:
Fix #3341

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/TeamCityTestEngineListener.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/TeamCityTestEngineListener.kt
@@ -129,7 +129,7 @@ class TeamCityTestEngineListener(
       when (testCase.type) {
          TestType.Container -> {
             val p = testCase.parent
-            // we might have a container inside a dynamic parent, in which case we need to start it
+            // we might have a container inside a dynamic parent, in which case we need to start the dynamic parent
             if (p != null && p.type == TestType.Dynamic) {
                if (!started.contains(p.descriptor)) startTestSuite(p)
             }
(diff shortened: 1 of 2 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/TeamCityTestEngineListener.kt
```diff
@@ -143,7 +143,13 @@ class TeamCityTestEngineListener(
...
             startTest(testCase)
          }
-         TestType.Dynamic -> Unit
+         TestType.Dynamic -> {
+            val p = testCase.parent
+            // we might have a dynamic inside another dynamic parent, in which case we need to start the dynamic parent
+            if (p != null && p.type == TestType.Dynamic) {
+               if (!started.contains(p.descriptor)) startTestSuite(p)
+            }
+         }
```
Kantis, line 152: Should there be a `startTest` here as well?

Kantis (approved):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/TeamCityTestEngineListener.kt
sksamuel: No, with dynamics it's done later because you don't know if a dynamic is a test suite or a test.

PR #3384, opened 2023-02-04
Feature/3382

lgtout:
This is a draft PR showing work done for 13-arity, and part of 14-arity.  
Please let me know if it looks okay.  If so, I'll replicate  for the remaining arities.  
Thanks!


vlsi:
Is this code generated?
Can it rather be replaced with a single method that takes List of arguments?

The current PR is really impossible to review and test

lgtout:
@vlsi It's a draft PR.  See [this](https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1675546893648889) conversation in Slack.  It was written by hand, not generated.  What do you mean that it's impossible to review and test?  It's incomplete.  As per the Slack convo, it was just to get @sksamuel's feedback before moving on to the bulk of the work.

vlsi:
Have you considered adding a single vararg or List method instead?

What is the purpose of copy-pasting the repeated code for all the arities?

lgtout:
@vlsi That's something you should discuss with @sksamuel.  As per the Slack convo, this was the approach he asked for.  I'm agnostic to implementation.

vlsi:
Ah, vararg / List won't handle `property` arguments.

In that case, it would be nice to have the code generated to avoid off-by-one errors

lgtout:
If you propose a way to generate the code, and @sksamuel approves, I'll consider implementing it, depending on how much time I have available, and compared with how much time completing the manual approach will take.

sksamuel (approved):

PR #3394, opened 2023-02-08
Add shouldContainAllIgnoringFields

sherviiin:
Add shouldContainAllIgnoringFields() which compares collections ignoring one or more fields, in any order


Kantis (commented):
LGTM overall.

Consider handling large collections (`AssertionsConfig.maxCollectionEnumerateSize`)
Perhaps some KDoc on the assertions, as well as the kotest.io [docs](https://github.com/kotest/kotest/blob/master/documentation/docs/assertions/collections.md)

What do you think about null-handling? Should this pass?

```kotlin
data class Person(val name: String, age: Int)

val sammy = Person("Sammy", 20)
val sara = Person("Sara", 27)

listOf(sammy, sara, null).shouldContainAllIgnoringFields(
  listOf(null, sara.copy(age = 28), sammy),
  Person::age
)
```

sherviiin:
Thanks @Kantis ,
Added docs.
About `maxCollectionEnumerateSize` I'm wondering why it is not being used in other assertions 🤔 
On the null topic: I prefer to align the most possible with `shouldContainAll`. That one ignores null AFAIK.

Kantis:
> Thanks @Kantis , Added docs. About `maxCollectionEnumerateSize` I'm wondering why it is not being used in other assertions 🤔 

I'm not sure why it's not used everywhere. I think it's been added incrementally whenever we had someone who cared enough about lack of readability in outputs :)  but I would have to do some digging in git to get the whole picture.. Looking around a bit more I think we should probably use `maxCollectionPrintSize` in this case.

> On the null topic: I prefer to align the most possible with `shouldContainAll`. That one ignores null AFAIK.

```kotlin
listOf(1, 2, null, 3, null).shouldContainAll(1, 2, null) // passes

listOf(1, 2, 3).shouldContainAll(1, 2, null)
// fails: Collection should contain all of [1, 2, <null>] but was missing [<null>]
```

sksamuel:
@sherviiin do you mind adding maxCollectionPrintSize support and then we can merge.

sherviiin:
Hey @sksamuel ,
Added `maxCollectionPrintSize` in last commit.
Thanks for for reminding me about this.

sksamuel (approved):

PR #3395, opened 2023-02-08
Printing type when failing numeric or default comparison

Kantis:


Kantis:
Maybe we just want to handle BigInteger with a platform-specific print (similar to `java.io.File` and `java.nio.Path`)

sksamuel:
Love this.

sksamuel (approved):

PR #3398, opened 2023-02-09
Update Spring docs for autoscan.disable=true

ryanlewis:
When auto scan is disabled, `SpringAutowireConstructorExtension` is not loaded. This PR updates the documentation to give a clue that it will need manually loading in your project config if auto scan is disabled.


Kantis (approved):

Kantis:
Thanks @ryanlewis 🚀

PR #3397, opened 2023-02-09
re-implement language injection annotation

aSemy:
Fix #3387 

* Create `expect annotation class KotestLanguage` in commonMain
* In JVM, typealias `@KotestLanguage` to `org.intellij.lang.annotations.Language`
* In JS/Desktop, typealias `@KotestLanguage` to a copy-pasted `@Language` implementation

Additionally:

* opt-in to `@ExperimentalKotest` to reduce warnings in the build logs
* ~~introduce an integration test for an Android project~~ Android Integration test moved to #3401 

Language injection still works as expected 👍 

image

#### Related

* #3328

Issue #3387, opened 2023-02-06
Kotest's @Language is prioritized over IntelliJ's, prompting for an opt-in


RBusarow:
Kotest: 5.5.5
Kotlin: 1.7.21 and 1.8.10
Gradle: 7.5 and 7.6

The new overloaded `@Language` annotation is prioritized over the IntelliJ one, in the IDE and during Gradle builds.

```text
w: [...]Main.kt:4:34 This declaration needs opt-in. Its usage should be marked with '@io.kotest.common.KotestInternal' or '@OptIn(io.kotest.common.KotestInternal::class)'
```

![img](https://user-images.githubusercontent.com/9890900/216869531-488626aa-6fb8-4b9b-8b67-d3dd4980e4e5.png)

reproducer project: https://github.com/RBusarow/Kotest-Language-repro

gabrielittner:
Kind of related, we are getting build errors for Android instrumentation tests with 5.5.5
```
Execution failed for task ':module:checkDebugAndroidTestDuplicateClasses'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckDuplicatesRunnable
   > Duplicate class org.intellij.lang.annotations.Language found in modules annotations-24.0 (org.jetbrains:annotations:24.0.0) and kotest-common-jvm-5.5 (io.kotest:kotest-common-jvm:5.5.5)
     
     Go to the documentation to learn how to <a href="d.android.com/r/tools/classpath-sync-errors">Fix dependency resolution errors</a>.
```

Kantis:
@aSemy what do you think is best way forward?

aSemy:
Ah that's a shame. I'm not sure how to fix it, it was a bit of a hack to begin with (albeit one that was suggested by JetBrains). I also don't know how to create automated tests for problems...

So since it's causing errors, then remove it. I'll make the PR.

UPDATE: Ah I realise the mistake I made, I think I can fix it. 

1. change `expect annotation class Language` to `expect annotation class KotestLanguage`
2. in non-JVM platforms, copy-paste the JVM `@org.intellij.lang.annotations.Language` implementation
3. then in all platforms, use 
   ```kotlin
   actual typealias KotestLanguage = org.intellij.lang.annotations.Language
   ```

   Which in non-JVM targets will use the copy-pasted annotation, and in JVM targets the original annotation.

PR #3401, opened 2023-02-10
Android integration test

aSemy:
Integration test for #3387 

Split off from PR #3397 because it's a larger change, and not strictly necessary for the `@Language` fix.

##### Summary

* introduces a Gradle plugin `kotest-publishing-test.gradle.kts` so subprojects can depend on Maven Publications from other subprojects
* Adds a sample Android Library project
   Note that the `android-library/ANDROID_SDK/licenses` directory is committed so that the Android Gradle plugin can automatically download the SDK.

* The test should fail with the following error until #3397 is merged

   ```
     * What went wrong:
     Execution failed for task ':checkDebugAndroidTestDuplicateClasses'.
     > A failure occurred while executing com.android.build.gradle.internal.tasks.CheckDuplicatesRunnable
        > Duplicate class org.intellij.lang.annotations.Language found in modules annotations-13.0 (org.jetbrains:annotations:13.0) and kotest-common-jvm-5.6.0-LOCAL (io.kotest:kotest-common-jvm:5.6.0-LOCAL)
       
          Go to the documentation to learn how to <a href="d.android.com/r/tools/classpath-sync-errors">Fix dependency resolution errors</a>.
     
     * Try:
     > Run with --debug option to get more log output.
     > Run with --scan to get full insights.
   ```

#### Utility

I'm undecided on the value of the Android integration test - how often are there issues? I also think that it might be fragile, because the Android implementation changes a lot and I suspect the automatic SDK download might not be stable (I've no evidence/experience, it's just a gut feeling).

However, the `kotest-publishing-test.gradle.kts` has some good potential, especially in the integration tests for the Kotest Gradle Plugin. This currently does not follow Gradle Project Isolation by directly referencing the tasks of other subprojects

https://github.com/kotest/kotest/blob/ed015bd0e061dc1854f7e0702038b569b7e0a766/kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts#L60-L68

By using `kotest-publishing-test.gradle.kts` this would be fixed.

I originally intended to do this safe-sharing in #3163


aSemy:
Ah... Android requires JVM 11

```
  * What went wrong:
  An exception occurred applying plugin request [id: 'com.android.library', version: '7.3.0']
  > Failed to apply plugin 'com.android.internal.library'.
     > Android Gradle plugin requires Java 11 to run. You are currently using Java 1.8.
        Your current JDK is located in /usr/lib/jvm/temurin-8-jdk-amd64/jre
        You can try some of the following options:
         - changing the IDE settings.
         - changing the JAVA_HOME environment variable.
         - changing `org.gradle.java.home` in `gradle.properties`.
```

Kantis:
Can `jvmToolchain` and included builds be used to facilitate testing with a different JDK?

aSemy:
Good idea! I've added toolchain config for Java 11

Kantis:
I pushed an example of what I meant, feel free to revert it.
Must be some easier way to get gradle testkit to resolve dependencies from the root build though..

aSemy:
I don't think making the integration tests an included-build help too much here. The actual test android-library project is isolated from the main Kotest repo because it's run via Gradle TestKit. And making it an included build makes it very difficult for the integration-test project to use dependencies from the main Kotest project ([build substitutions don't work when the intended replacements use `maven-publish`](https://docs.gradle.org/current/userguide/composite_builds.html#included_build_substitution_limitations), so if the integration-test project wants to test the latest Kotest code it needs other complicated workarounds).

> Must be some easier way to get gradle testkit to resolve dependencies from the root build though..

You mean the `kotest-publishing-test` plugin? That's a consequence of [how Gradle shares files between subprojects unfortunately](https://docs.gradle.org/current/userguide/cross_project_publications.html#sec:variant-aware-sharing). The Gradle API is comprehensive, but not so elegant! It does allow for parallel processing, work-avoidance, and keeping projects isolated though. 

Instead of publishing artifacts to a local Maven directory, they could be published to Maven Local instead. This saves a little complexity.

PR #3403, opened 2023-02-11
Bumping to Kotlin 1.8.10

Kantis:
Started fiddling with 1.8

There's possibly things to throw away in kotlin/native since there was an overhaul of the memory model in 1.7 that made freezing stuff redundant.


sksamuel:
https://github.com/kotest/kotest/pull/3468 is ready

PR #3404, opened 2023-02-11
kotest-framework-api: Fix KDoc links, code blocks, spellings

aSemy:
Docs only changes

* Fixed code links to work when the class isn't imported
* Added code fence backticks to multiline code blocks
* minor spelling fixes

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
@@ -8,6 +8,7 @@ import io.kotest.core.listeners.ProjectListener
 import io.kotest.core.names.DuplicateTestNameMode
 import io.kotest.core.names.TestNameCase
 import io.kotest.core.spec.IsolationMode
+import io.kotest.core.spec.Spec
 import io.kotest.core.spec.SpecExecutionOrder
 import io.kotest.core.test.AssertionMode
 import io.kotest.core.test.TestCaseOrder
@@ -18,7 +19,7 @@ import kotlin.time.Duration
 
 /**
  * Project-wide configuration. Extensions returned by an
- * instance of this class will be applied to all [Spec] and [TestCase]s.
+ * instance of this class will be applied to all [io.kotest.core.spec.Spec] and [io.kotest.core.test.TestCase]s.
  *
  * Create an object or class that is derived from this class and place it in your source.
  *
@@ -26,11 +27,12 @@ import kotlin.time.Duration
  *
  * For example, you could create this object and place the source in `src/main/kotlin/my/test/package`.
  *
+ * ```
  * object KotestProjectConfig : AbstractProjectConfig() {
  *    override val failOnEmptyTestSuite = true
  *    override val testCaseOrder = TestCaseOrder.Random
  * }
- *
+ * ```
  */
 abstract class AbstractProjectConfig {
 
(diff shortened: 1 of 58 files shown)


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
```diff
@@ -18,19 +19,20 @@ import kotlin.time.Duration
 
 /**
  * Project-wide configuration. Extensions returned by an
- * instance of this class will be applied to all [Spec] and [TestCase]s.
+ * instance of this class will be applied to all [io.kotest.core.spec.Spec] and [io.kotest.core.test.TestCase]s.
```
Kantis:
```suggestion
 * instance of this class will be applied to all [Spec] and [io.kotest.core.test.TestCase]s.
```

Since you added an import for Spec.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/ProjectConfiguration.kt
```diff
@@ -259,22 +260,28 @@ class ProjectConfiguration {
...
     *
     * And if disabled, the output would be:
     *
+    * ```text
     * this test 1
     *    test this 2
+    * ```
     *
-    * Defaults to null, which is to use the default.
+    * Defaults to `null`, which is the default.
```
Kantis:
```suggestion
    * Defaults to `null`, which is to let the [Spec] style determine whether to include affixes.
```

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/listeners/TestListener.kt
```diff
@@ -5,6 +5,9 @@ import io.kotest.core.test.TestCase
 @Deprecated("Renamed to IgnoredSpecListener. Deprecated since 5.0")
 typealias SpecIgnoredListner = IgnoredSpecListener
 
+@Deprecated("Renamed to IgnoredSpecListener. Deprecated since 5.0")
+typealias SpecIgnoredListener = IgnoredSpecListener
+
```
Kantis: Not sure there's a point in adding this typealias, unless `SpecIgnoredListener` existed in Kotest 4?

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
aSemy: good spot!

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/listeners/TestListener.kt
aSemy: Ah yeah, I forgot to mention this.

The existing deprecated type alias `SpecIgnoredListner` has a typo, `Listner`. The new type alias fixes it, `Listener`. 

Options:

1. keep the misspelling and don't add a fixed version (it's been deprecated for a while, everyone has probably already migrated already)
2. fix the spelling (might break anyone who still uses `SpecIgnoredListner`?)
3. keep the misspelled version and add a fixed version (it might help people migrate?)

I went with 3 because I wanted to have a discussion in the PR. Option 1 is probably best.

Kantis: Had to do some digging.. The previously existing class was typod, so the type alias must also be typod. 

https://github.com/kotest/kotest/blob/v4.6.1/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/listeners/spec.kt#L49

aSemy: I didn't even consider that the original was misspelled! I've removed my addition and added a comment.

Issue #3405, opened 2023-02-11
Confusing name of composing matchers


pientaa:
_Data Class Matchers_ name and its documentation suggests it's dedicated for `data class` only, which is not true. IMHO it should be renamed to simple _Class Matchers_ or even _Composed matchers_, because it's also possible to compose a matcher for an interface and use this matcher e.g. with an anonymous object

pientaa:
@sksamuel this one can be closed since #3406 is already merged

pientaa:
it was already solved in #3406, which was merged recently

Issue #3343, opened 2023-01-17
idea: composite matchers, MatcherResultBuilder


vlsi:
Currently, every matcher has to return exactly one result.

What do you think if a matcher was allowed to return several findings?

For instance, suppose I wanted adding a matcher that would verify if a database table is empty.

Here's a draft code

```kotlin
interface Table {
    val name: String
}

fun emptyTable(con: Connection) =
    object: Matcher<Table> {
        override fun test(value: Table): MatcherResult {
            con.createStatement().use { st ->
                val rs = st.executeQuery("select count(*) from ${st.enquoteIdentifier(value.name, false)}")
                // it would be nice to reuse "resultset must contain row" matcher here, but it's not possible
                rs.next()
                val count = rs.getInt(1)
                return MatcherResult(
                    count == 0,
                    { "Table ${value.name} is not empty, it contains $count rows" },
                    { "Table ${value.name} is empty" }
                )
            }
        }
    }
```

The usage would be

```kotlin
val tableUnderTest == ....

// verify that the table is empty in the given DB
tableUnderTest shouldBe emptyTable(con = databaseConnection)
```

So far so good, however, note how I can't really reuse "resultset must contain row" matcher (see https://github.com/kotest/kotest/issues/3339#issuecomment-1385336335).
On top of that, if the table is missing in the database, `executeQuery` would fail, and the assertion would be completely screwed.

However, suppose later I want to verify multiple database connections, not just one of them:

```kotlin
val tableUnderTest == ....

val databasesUnderTest: List<Connection>
// verify that the table is empty in the given DB
tableUnderTest shouldBe emptyTable(conections = databasesUnderTest)
```

I do not see what is the intended way to squeeze multiple assertion results into a single `MatchResult`.

Have you considered something as follows?

In other words, `MatcherResultBuilder` that will enable `Matcher` implementations to build composite result based on several checks.
At the same time, it could have its own `withClue` so the clues are added accordingly to the match results.

```kotlin
interface Table {
    val name: String
}

interface MatcherResultBuilder {
    fun failure(message: String, throwable: Throwable)
    fun addMatcherResult(result: MatcherResult)

    fun withClue(clue: () -> Any, thunk: () -> Unit)
}
fun buildMatcherResult(builder: MatcherResultBuilder.() -> Unit): MatcherResult {
    TODO("")
}

fun isEmpty(connections: List<Connection>) =
    object: Matcher<Table> {
        override fun test(value: Table): MatcherResult =
[...]

PR #3406, opened 2023-02-12
Rename Data Class Matchers to Composed Matchers

pientaa:
Implements #3405 

I changed the name of _Data Class Matchers_, to _Composed Matchers_ according to my proposition [here](https://github.com/kotest/kotest/issues/3343#issuecomment-1426854265)


Kantis:
What do you think about also making the composition work for any `(T) -> Any?`, instead of just properties?
Something like:

```kotlin
@Suppress("UNCHECKED_CAST")
fun <T : Any?> Matcher.Companion.compose(
   vararg pairs: Pair<Matcher<*>, (T) -> Any?>
): Matcher<T> = object : Matcher<T> {
   override fun test(value: T): MatcherResult {
      val results = pairs.mapIndexed { i, (matcher, prop) ->
         runCatching {
            (matcher as Matcher<Any?>).test(prop(value))
         }.onFailure {
            fail("Mismatching types in argument ${i+1} for composed matcher:\n\n${it.message}")
         }.getOrThrow()
      }

      return MatcherResult(
         results.all { it.passed() },
         { results.joinToString(separator = "\n") { it.failureMessage() } },
         { results.joinToString(separator = "\n") { it.negatedFailureMessage() } },
      )
   }
}
```

It's backwards compatible, adds flexibility in what you use for matching.. Not really sure what I'd use it for though

pientaa:
That's a very nice idea @Kantis!

I've tried to use it in some other way (than using actual class properties) and it's not possible because of `prop(value)`, so I'd rather leave the compose function using reflection as it is and I'll add another compose function which will provide more freedom:


```kotlin
@Suppress("UNCHECKED_CAST")
fun <T : Any> Matcher.Companion.compose(
   vararg matchers: Matcher<T>
): Matcher<T> = object : Matcher<T> {
   override fun test(value: T): MatcherResult {
      val results = matchers.mapIndexed { i, matcher  ->
         runCatching {
            (matcher as Matcher<Any?>).test(value)
         }.onFailure {
            fail("Mismatching types in argument ${i + 1} for composed matcher:\n\n${it.message}")
         }.getOrThrow()
      }

      return MatcherResult(
         results.all { it.passed() },
         { results.joinToString(separator = "\n") { it.failureMessage() } },
         { results.joinToString(separator = "\n") { it.negatedFailureMessage() } },
      )
   }
}
```

Thanks to that #3343 will be solved, because it'll be possible to compose matchers without any class inference. Simple example:

```kotlin
val passwordMatcher = Matcher.compose(
   containADigit(), contain(Regex("[a-z]")), contain(Regex("[A-Z]"))
)
```

pientaa:
I came up with another idea to make composing matchers more flexible, but apparently it's breaking the compose matcher API, so I'll wait for your opinion @Kantis on that - if we want to go this way

Firstly I thought that it sucks that composing matchers will work on JVM only (because of reflection I used before). So I thought that common API (in common module) for composing matchers will be much better (I already pasted that code above, but I'll do it here once more):

```kotlin
@Suppress("UNCHECKED_CAST")
fun <T : Any?> Matcher.Companion.compose(
   vararg matchers: Matcher<T>
): Matcher<T> = object : Matcher<T> {
   override fun test(value: T): MatcherResult {
      val results = matchers.mapIndexed { i, matcher ->
         runCatching {
            (matcher as Matcher<Any?>).test(value)
         }.onFailure {
            fail("Mismatching types in argument ${i + 1} for composed matcher:\n\n${it.message}")
         }.getOrThrow()
      }

      return MatcherResult(
         results.all { it.passed() },
         { results.joinToString(separator = "\n") { it.failureMessage() } },
         { results.joinToString(separator = "\n") { it.negatedFailureMessage() } },
      )
   }
}
```
And to use the reflection to build composed matcher on a class can be done by creating following function:

```kotlin
@Suppress("UNCHECKED_CAST")
fun <T : Any> havePropertyMatching(pair: Pair<Matcher<*>, KProperty1<T, *>>): Matcher<T> = object : Matcher<T> {
   override fun test(value: T): MatcherResult {
      val (matcher, prop) = pair
      return runCatching {
         (matcher as Matcher<Any?>).test(prop(value))
      }.onFailure {
         fail("Mismatching types for havePropertyMatching matcher:\n\n${it.message}")
      }.getOrThrow()
   }
}
```

But then composed matcher for `Person` (example from documentation) would look like this:

```kotlin
val matcherResult = Matcher.compose(
   havePropertyMatching(nameMatcher to Person::name),
   havePropertyMatching(ageMatcher to Person::age),
   havePropertyMatching(addressMatcher to Person::address)
)
```

IMO it's an improvement, because you can still do this:
```kotlin
val passwordMatcher = Matcher.compose(
   containADigit(), contain(Regex("[a-z]")), contain(Regex("[A-Z]"))
)
```
and composing matchers is multiplatform instead JVM-specific, but WDYT?

Kantis:
@pientaa I like the idea, but I would like to hear some more opinions. Mulitplatform support and greater flexibility are two good upsides of doing the change.

Perhaps `Matcher.compose(..)` -> `Matcher.all(m1, m2)` and `Matcher.any(m1, m2)`?

pientaa:
@Kantis do you think I should do the following
1. Make `Matcher.compose` multi-platform
2. Implement `Matcher.all(m1, m2)` and `Matcher.any(m1, m2)`

 in current PR? 
 
IMO it will be better to leave the current PR as it is (unless you see some potential improvements) and apply mentioned above changes in another PR (I can create an issue for each of enhancement I can see in composing matchers), because these changes will break the API.

I also thought about some changes regarding #3343, but it's going to break the API anyway, so I wanted to take some time to think about it all and play a little bit with the code, because right now I have several ideas of what can be improved there and I'm not sure which one is the best. And I wanted the documentation to be fixed already (with this little improvement you proposed in a very first comment here).

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/compose/compose.kt
```diff
@@ -0,0 +1,56 @@
...
+   vararg matchers: Matcher<T>
+): Matcher<T> = object : Matcher<T> {
+   override fun test(value: T): MatcherResult {
+      val results = matchers.mapIndexed { i, matcher  ->
+         runCatching {
+            (matcher as Matcher<Any?>).test(value)
+         }.onFailure {
+            fail("Mismatching types in argument ${i + 1} for composed matcher:\n\n${it.message}")
+         }.getOrThrow()
+      }
```
Kantis:
```suggestion
      val results = matchers.map { matcher -> matcher.test(value) }
```

Can be simplified, since we have type constraints on both `value` and `matcher`. Trying to compose matchers for different types gives a warning: (Will fail completely from next Kotlin version)
image

If you explicitly state the type of the matcher, it will fail compilation: 
image

Kantis (commented):
I agree, better keep the breaking changes for another PR targeting Kotest 6.0.

pientaa:
Thanks @Kantis! I created following issues: #3431 #3432 for what we've discussed before. I can implement both of them

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/reflection/compose.kt
```diff
@@ -1,28 +0,0 @@
-package io.kotest.matchers.reflection
-
-import io.kotest.matchers.Matcher
-import io.kotest.matchers.MatcherResult
-import kotlin.reflect.KProperty1
-
-@Suppress("UNCHECKED_CAST")
-fun <T : Any?> Matcher.Companion.compose(
```
Kantis, line 8: Maybe we should keep this one with a deprecation and replacewith pointing to the new location (`io.kotest.matchers.compose`)

Since the package changes, imports will stop working without notice otherwise. We can remove this one in 6.0 I think

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/compose/compose.kt
```diff
@@ -0,0 +1,50 @@
...
+import io.kotest.matchers.Matcher
+import io.kotest.matchers.MatcherResult
+import kotlin.reflect.KProperty1
+
+/**
+ * Compose matchers along with the property to extract to test against.
+ * All matchers must be the same type `T`. If any matcher fails, composed matcher fails, too.
+ */
+@Suppress("UNCHECKED_CAST")
+fun <T : Any?> Matcher.Companion.compose(
```
sksamuel: Can we add tests for this variation ?

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/compose/compose.kt
```diff
@@ -0,0 +1,50 @@
...
+   vararg pairs: Pair<Matcher<*>, KProperty1<T, *>>
```
sksamuel: There's nothing enforcing that the property extracts a type compatible with the matcher.

pientaa: actually we do 
```kotlin
         runCatching {
            (matcher as Matcher<Any?>).test(prop(value))
         }.onFailure {
            fail("Mismatching types in argument ${i + 1} for composed matcher:\n\n${it.message}")
         }.getOrThrow()
```
below

pientaa: I'll add additional test to prove it's working correctly

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/compose/compose.kt
pientaa: tests are already added

pientaa:
@Kantis @sksamuel notice that I renamed new `compose` to `all`, as far as we'd like to provide `Matcher.any`, too. The old `compose` is deprecated

sksamuel (approved):

PR #3413, opened 2023-02-16
Update test-coroutine-dispatcher docs

osipxd:
`coroutineTestScope` was added to `AbstractProjectConfig` in v5.5.5, so I've updated docs according to #3383.


sksamuel (approved):

sksamuel:
👍🏻

PR #3416, opened 2023-02-16
Update shouldBeEqualToComparingFields doc

ktrueda:
As shouldBeEqualToComparingFieldsExcept is deprecated after 5.4 doc should be updated.


Kantis:
Maybe we should keep it with a strike-through and refer to the replacement (`shouldBeEqualToIgnoringFields`)?

ktrueda:
I think it's not  `shouldBeEqualToIgnoringFields` but `shouldBeEqualToComparingFields` . Right?

Kantis:
`shouldBeEqualToComparingFieldsExcept` (allowing you to specify what fields to ignore) was replaced by `shouldBeEqualToIgnoringFields` (comparing by fields, excluding the fields provided as `ignored`)

ktrueda:
According to https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt#L400-L400


`shouldBeEqualToComparingFieldsExcept` (allowing you to specify what fields to ignore) was replaced by `shouldBeEqualToComparingFields`

not
`shouldBeEqualToComparingFieldsExcept` (allowing you to specify what fields to ignore) was replaced by `shouldBeEqualToIgnoringFields` (comparing by fields, excluding the fields provided as `ignored`)


Should I update https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt#L400-L400 too?

Kantis:
@ktrueda It seems you _can_ use `shouldBeEqualToComparingFields` as a replacement, but you would need to supply a `FieldEqualityCheckConfig` excluding the properties you previously excluded..

E.g.:

```
// deprecated
Foo.shouldBeEqualToComparingFieldsExcept(other, ignorePrivateFields = true, Foo::id, Foo::created)

// suggested replacement
Foo.shouldBeEqualToComparingFields(other, FieldEqualityCheckConfig(propertiesToExclude = listOf(Foo::id, Foo::created))

//alternative replacement
Foo.shouldBeEqualToIgnoringFields(other, ignorePrivateFields = true, Foo::id, Foo::created)
```

I think we should indeed update the deprecation and refer to `shouldBeEqualToIgnoringFields` as replacement in docs, but interested to hear more opinions on this

ktrueda:
@Kantis Thank you for your comment.

In my opinion both `shouldBeEqualToIgnoringFields` and `shouldBeEqualToComparingFields` are useful. Consistency is important. If you decide the which is preferable, I'll update it.

sksamuel:
@Kantis do you have a preference on this ?

Kantis:
My preference is:
* Both methods are documented
* Deprecation suggests using the method that takes the same parameters as the deprecated method

ktrueda:
Thank you.

I updated the Deprecated annotation and the doc.

Kantis (approved):

Issue #3032, opened 2022-06-02
Set<Path> comparison no longer works as expected


AarjavP:
``` kotlin
    @Test
    fun `compare set of paths`() {
        val one = Paths.get("foo")
        val two = Paths.get("bar")
        val lhs = setOf(one, two)
        val rhs = setOf(one, two)
        lhs shouldBe rhs
    }
```
Using `io.kotest:kotest-assertions-core-jvm:5.0.3` this passes.

In version `5.1.0` and onwards, it fails with the message 
```
java.lang.AssertionError: Disallowed nesting iterator bar (WindowsPath) within bar (WindowsPath); (use custom test code instead)
```
(The message is constructed [here](https://github.com/kotest/kotest/blob/v5.3.0/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/IterableEq.kt#L154))
Note that `Path` implements `Iterable<Path>`. 
I do not know what the proper fix would be as I can definitely see the need to error out for custom Iterable types. Is there any additional arguments that be passed to say compare the elements of the set using the equals method?

sksamuel:
Are you looking for something like `lhs should be equal using .equals rhs` ?

AarjavP:
You're right, I could be doing `lhs == rhs shouldBe true` but Ideally I should be using something like `lhs shouldContainExactly rhs` so that kotest can print on the set differences (missing or extra expected elements)
I will try out that function when I get on again but I think it had the same issue.

since `Path` is a jdk type I would like it to work out of the box, but if its not possible or a good idea then maybe we can have something like `lhs.shouldContainExactly(rhs, Path::equals)`?

sksamuel:
I think we should consider adding a `lhr shouldEqual rhs` which just uses .equals.

tbvh:
I've been tinkering with an alternative. I don't know if it fits the architecture of kotest.

In IterableEq the method `nestedIterator` triggers the failure, `checkEquality` on Iterables doesn't want to operate on nested iterables. So, either nestedIterator could throw an exception that is caught so that `equals(Iterable, Iterable)` can fall back on `.equals`, or nested Iterators should be identified before going into the method.

For some[1] nested Iterators this can be checked using reflection, `IterableEq.isValidIterable(it: Any)` can make a jvm specific call to:


    private val nothingType = typeOf<List<Nothing>>().arguments.first().type!!

    private fun isNestedIterable(self: Any): Boolean {
[...]

PR #3420, opened 2023-02-18
Remove Iterable checking logic from IterableEq

### LINKED ISSUE #3130: Jackson JsonNode comparisons are broken
`5.3.1`

When comparing two different `JsonNode`s, `shouldBe` is erroneously stating they are equal when they have different values at the same key.

Here's a simple test that reproduces this issue for me:
```kotlin
@Test
fun `it should return detect differences in values`() {
  val map1 = mapOf("test" to 1)
  val map2 = mapOf("test" to 2)
  val mapper = ObjectMapper()
  val jsonNode1 = mapper.valueToTree<JsonNode>(map1)
  val jsonNode2 = mapper.valueToTree<JsonNode>(map2)

  // Passes
  jsonNode1 shouldBe jsonNode2
  // Fails with:
  // expected:<"{"test":2}"> but was:<"{"test":1}">
  jsonNode1.toString() shouldBe jsonNode2.toString()
}
```

echo304:
Resolve #3130 #3032

It changes behavior of `IterableEq`. Although it still has the term "Iterable" but it does not mean that it support equality check for Iterables of any type. With this change, it only supports equality check for Iterables of the following types:
List, Set, Collection and additionally Arrays of any type.

- Removes `is Iterable` check from `IterableEq`
- Fix relevant test codes and update documentation


Kantis:
I think this is what we want to change, yes.

sksamuel:
I'm ok pulling this into 5.6
The existing behavior is just broken.
What do you think @Kantis ?

Issue #3418, opened 2023-02-17
assertSoftly should add a clue


CLOVIS-AI:
When writing a test like
```kotlin
val list = listOf(1, 2, 3)

assertSoftly(list) {
    it shouldContain 3
}
```

it is likely that the user is interested in the entire list in case the test fails. However, adding the entire list as a clue is quite verbose, making it harder to understand what the point of the test is:

```kotlin
val list = listOf(1, 2, 3)

assertSoftly(list) {
    withClue({ "Soft assertion: $it" }) {
        it shouldContain 3
    }
}
```

I think it would be beneficial if Kotest added the clue automatically when `assertSoftly` is called.

CLOVIS-AI:
If this helps, I can contribute a PR that adds a call to `withClue` in `assertSoftly`

sksamuel:
I think that's a good idea.

PR #3425, opened 2023-02-20
'assertSoftly' and 'all' imply clue

CLOVIS-AI:
> [Issue](https://github.com/kotest/kotest/issues/3418)

Adds a clue to `assertSoftly(t)` and `all(t)` calls, so the passed receiver is printed to the user in case of failure.


Kantis:
Checking the generated message, using:
```kotlin
   test("checking clues") {
      val x = listOf(1, 2, 3)
      assertSoftly(x) {
         x shouldContain 3
         x shouldContain 4
         x shouldContain 5
      }
   }
```

Getting:
```
io.kotest.assertions.MultiAssertionError: 
1) Soft assertion of: [1, 2, 3]
Collection should contain element 4 based on object equality; but the collection is [1, 2, 3]
	at com.sksamuel.kotest.eq.IterableEqTest$1$1.invokeSuspend(IterableEqTest.kt:50)
2) Soft assertion of: [1, 2, 3]
Collection should contain element 5 based on object equality; but the collection is [1, 2, 3]
	at com.sksamuel.kotest.eq.IterableEqTest$1$1.invokeSuspend(IterableEqTest.kt:51)
```

Previously it was:
```
io.kotest.assertions.MultiAssertionError: 
1) Collection should contain element 4 based on object equality; but the collection is [1, 2, 3]
	at com.sksamuel.kotest.eq.IterableEqTest$1$1.invokeSuspend(IterableEqTest.kt:50)
2) Collection should contain element 5 based on object equality; but the collection is [1, 2, 3]
	at com.sksamuel.kotest.eq.IterableEqTest$1$1.invokeSuspend(IterableEqTest.kt:51)
```

Personally I think something like this would be ideal
```
io.kotest.assertions.MultiAssertionError: 
The following 2 assertions of java.util.Arrays.ArrayList<[1, 2, 3]> failed:
1) Collection should contain element 4 based on object equality; but the collection is [1, 2, 3]
	at com.sksamuel.kotest.eq.IterableEqTest$1$1.invokeSuspend(IterableEqTest.kt:50)
2) Collection should contain element 5 based on object equality; but the collection is [1, 2, 3]
	at com.sksamuel.kotest.eq.IterableEqTest$1$1.invokeSuspend(IterableEqTest.kt:51)
```

What would it look like for a data class with 5-6 properties?

CLOVIS-AI:
Hi @Kantis, I agree with you that it would be better, however I do not have the knowledge of how to do it

sksamuel:
We can do this by not delegating to the existing clue function, but duplicating the logic and only outputting it once.

Kantis:
`assertSoftly(x)` now prepends a line informing that the subsequent assertions were performed on `x` 

example:
```kotlin
            shouldThrow<AssertionError> {
               assertSoftly("foo") {
                  length shouldBe 2
                  this[1] shouldBe 'o' // should pass
                  this shouldNotBe "foo"
               }
            }.run {
               // This line is new
               message should contain("The following 2 assertions for \"foo\" failed:")
 
               // Unchanged
               message should contain("1) expected:<2> but was:<3>")
               message should contain("2) \"foo\" should not equal \"foo\"")
            }
```

Also supports nesting

```kotlin
         "With nested receivers" - {
            data class Person(val name: String, val age: Int)

            "Prints both subjects" {
               shouldFail {
                  assertSoftly(Person("John", 20)) {
                     name shouldBe "Jane"
                     assertSoftly(age) {
                        it shouldBeGreaterThan 30
                        it.shouldBeOdd()
                     }
                  }
               }.message.shouldContainInOrder(
                  """The following 2 assertions for Person(name=John, age=20) failed:""",
                  """1) expected:<"Jane"> but was:<"John">""",
                  """2) The following 2 assertions for 20 failed:""",
                  """   1) 20 should be > 30""",
                  """   2) 20 should be odd""",
               )
            }
         }
```

adds some complexity to `ErrorCollector` (keeping track of current subject, depth) and `assertSoftly` (collecting intermediate errors to a `MultiAssertionError` on each level of nesting)

Issue #3227, opened 2022-10-07
Property Arb for Duration


aSemy:
I'd like to be able to generate arbitrary Durations.

I've created this code, but I don't have the time to integrate this into the project and create tests. Please feel free to use this as a basis :)

#### `Arb<Duration>`

```kotlin
import io.kotest.property.*
import io.kotest.property.arbitrary.ArbitraryBuilder
import kotlin.time.Duration
import kotlin.time.Duration.Companion.days
import kotlin.time.Duration.Companion.hours
import kotlin.time.Duration.Companion.minutes
import kotlin.time.Duration.Companion.seconds
import kotlin.time.Duration.Companion.nanoseconds
import kotlin.time.DurationUnit
import kotlin.time.toDuration


/** 
* Arbitrary [Duration]s.
* 
* @param[range] constrain the generated durations to be within this range.
*/
private fun Arb.Companion.duration(
   range: ClosedRange<Duration> = (-Int.MAX_VALUE.seconds)..Int.MAX_VALUE.seconds
): Arb<Duration> =
   ArbitraryBuilder.create { rs ->
   val unit = DurationUnit.values().random(rs.random)
   val value = rs.random.nextLong()
   value.toDuration(unit).coerceIn(range)
   }.withEdgecases(
      listOf(
         Duration.ZERO,
         Duration.INFINITE,
         Duration.INFINITE * -1,
      ).map { it.coerceIn(range) }
   ).withShrinker(DurationShrinker(range))
      .withClassifier(DurationClassifier(range))
      .build()
```

#### `DurationShrinker`

```kotlin
private class DurationShrinker(
   private val range: ClosedRange<Duration>
) : Shrinker<Duration> {

   override fun shrink(value: Duration): List<Duration> {
      val unsafeShrinks = unsafeShrinks.mapNotNull { op ->
         try {
            op(value)
         } catch (_: IllegalArgumentException) {
            null
         }
      }

      val componentShrinks = componentsShrinker.shrink(value)

      return (unsafeShrinks + componentShrinks)
         .map { it.coerceIn(range) }
         .takeIf { shrinks -> shrinks.all { it != Duration.ZERO } }
         ?.distinct()
         ?: emptyList()
   }

   /** Tries to shrink, but might through an [IllegalArgumentException] if the resulting duration is invalid */
   private fun interface ShrinkUnsafe : (Duration) -> Duration

   private val unsafeShrinks: List<ShrinkUnsafe> =
      listOf(
[...]

PR #3428, opened 2023-02-22
#3227 property arb for duration

pientaa:
Implements #3227


sksamuel (approved):

pientaa:
@sksamuel there were some tests crashing in the workflow, but I don't see any connection with my changes. Should I worry about it?

PR #3429, opened 2023-02-23
Fix `Codepoint.ascii()` to return arbitrary printable ASCII characters

### LINKED ISSUE #3426: Codepoint.ascii() does not cover the whole ASCII range.
**Kotest 5.5.4**

The generator only generates codepoints from " " (32) to "z" (122) which is not even the entire printable ASCII range!

According to [Wikipedia: ASCII](https://en.wikipedia.org/wiki/ASCII), the ASCII code points are 0..127 and the printable ones are 32..126.

echo304:
Resolve #3426 

Fix `Codepoint.ascii()` to return arbitrary printable ASCII characters.

as-is:
- Returns char from ASCII code 32 to 122

to-be:
- Accurately returns char from 32 to 126


sksamuel (approved):

sksamuel:
👍🏻

echo304:
@sksamuel 
jslrTest keeps failing due to below code blocks which is commented out. Can I just ignore those failing test?
스크린샷 2023-03-04 오후 2 21 50

fadimilio:
FYI: Since `.ascii()` is the default codepoint for the `Arb.string()` generator (which we use to generate data for our test models which are also written into our postgres database), this change broke a lot of our tests. Postgres does not like some control characters that are now being generated.
@sksamuel

PR #3441, opened 2023-03-04
Support superclass annotations when deciding whether a class should be isolated.

kozlov-aa:
This is a suggested change to improve behavior of `@Isolate` annotation.

### Context
At the moment, the `@Isolate` annotation on the parent class does not affect the isolation of child classes in any way.

```kotlin
   override val parallelism = 4
}

@Isolate
abstract class ShouldBeIsolated(body: FreeSpec.() -> Unit = {}) : FreeSpec(body)

@Isolate
interface ShouldAlsoBeIsolated

class MyTestClass1 : ShouldBeIsolated({

  "... still work in parallel" {
  }

})

class MyTestClass2 : FreeSpec({

  "... also still work in parallel" {
  }

}) : ShouldAlsoBeIsolated

```

### Changes
These changes will help make annotation search customizable as needed.
- added sealed interface `AnnotationSearchParameter`
- added two implementations `IncludingAnnotations` and `IncludingSuperclasses`
- method `Reflection#annotations` now accept a set of AnnotationSearchParameter instead `recursive` flag
- methods `KClass<*>#annotations` and `KClass<*>#hasAnnotation` by default still has just recursive behavior
- method `KClass<*>#isIsolate` in addition to recursive check, checks the presence of `@DoNotParallelize` or `@Isolate` in superclasses
- annotations caching still supported, but now cache key is `Pair<KClass<*>, Set<AnnotationSearchParameter>`


sksamuel:
This is great.

sksamuel (approved):

PR #3449, opened 2023-03-12
Add missing -jvm for dependency on JVM and Android

StefanLobbenmeierObjego:


StefanLobbenmeierObjego:
I copied the quickstart into my project and wondered why the dependency could not be found. For multiplattform, I understand if the version is kept generic, but for the general JVM case we should make it easier to get started in my opinion.

This also aligns Gradle/JVM with Maven/JVM.

StefanLobbenmeierObjego:
Nevermind, I had the error because I provided the version 5.5, not 5.5.5. My bad!

PR #3452, opened 2023-03-14
chore(doc): add missing html.required.set(false) flag to versioned docs

Stefanqn:
missed the versioned docs


sksamuel:
Thanks again

PR #3458, opened 2023-03-18
Add link to Collections Matchers from top-level Matchers pages in docs

avarun42:
Updated latest version of documentation as well as documentation for upcoming version


Kantis:
Collection matchers are part of the core matchers. All other listed modules are in different dependencies, so I think we should reconsider this change. Perhaps it's better to make collections a clearer part of the core matchers?

avarun42:
Either way there should be some way to get to them from the [Matchers](https://kotest.io/docs/assertions/matchers.html) page in the docs. Currently there is no way, which is extremely confusing. 

It would be reasonable to consolidate the Collections matchers page into the Core matchers page, but that can be part of a larger documentation cleanup effort. As it is, it doesn't make sense to leave the Collections page as difficult as it is to find.

sksamuel:
It's not consistent atm for sure, but there's no harm in adding the link given collection matchers are already a separate page?

PR #3461, opened 2023-03-20
Expose discovered specs as TestDescriptors during discovery and add support for unique IDs

marcphilipp:
This PR makes Kotest compatible with the requirements that are [documented in the JUnit User Guide](https://junit.org/junit5/docs/current/user-guide/#test-engines-requirements) for interoperability with build tools and IDEs.

In particular the following two properties were not met prior to this change:

> * Unless to indicate an empty discovery result, the `TestDescriptor` returned from `TestEngine.discover()` _should_ have children rather than being completely dynamic. This allows tools to display the structure of the tests and to select a subset of tests to execute.
> * A TestEngine _must_ be able to discover `UniqueIdSelectors`
>  for any unique ID that it previously generated and returned from `TestEngine.discover()`. This enables selecting a subset of tests to execute or rerun.

For Gradle Enterprise, we are relying on these properties in order to select a subset of the `TestPlan`  for [Predictive Test Selection](https://gradle.com/gradle-enterprise-solutions/predictive-test-selection/) and in order to split it into subsets that can be distributed with [Test Distribution](https://gradle.com/gradle-enterprise-solutions/test-distribution/). Therefore, our customers could not use neither of these features when using Kotest. However, we’re convinced other tools will also benefit from Kotest being more inline with other `TestEngine` implementations such as JUnit Jupiter or Spock to provide a consistent user experience for their users.


Kantis:
Looks good to me, but it's a bit beyond me. Would like to have @sksamuel eyes on this

marcphilipp:
@sksamuel Please let me know if you have any questions. I'd be happy to schedule a call with you if that makes it easier for you to review these changes.

sksamuel:
I'll review this weekend just been slammed at work.

sksamuel (approved):
So the specs are added at discovery time. I've looked over the engine and I don't think we do any filtering after that point so it looks good to me.

Thanks for such a comprehensive PR.

marcphilipp:
@sksamuel Thanks for the review and merging it! Should this PR be assigned to a milestone? Will it be released with 5.6? If so, do you have a rough estimate when that will be shipped?

sksamuel:
Yes it'll go out with 5.6 later this week.

Issue #3463, opened 2023-03-23
Default variety of generated floats is very biased towards high exponents


rubenpieters:
**Which version of Kotest are you using**
5.5.4

**Description**

I noticed when using the default float generator for `checkAll`, that it is very biased towards floats with high exponents. In the following code snippet I just keep track of all the floats generated and count all of them higher than `1e+34f` or lower than `-1e+34f`.

```kotlin
val l: MutableList<Float> = mutableListOf()

checkAll(1_000_000) { f: Float -> l.add(f) }

println(l.count())
println(l.filter { it < -1e+34f }.count())
println(l.filter { it > 1e+34f }.count())
```

The result is:

```
1000000 // total
492357 // lower than -1e+34f
494961 // higher than 1e+34f
```

Am I using something incorrectly? If not, I think preferably the floats would have a better variety across the entire range of possible values? This is very noticeable when using the default number of generated values.

rubenpieters:
I think it's an artifact from using the `Double` randomization method [here](https://github.com/kotest/kotest/blob/master/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/floats.kt#L71). Compare the distribution in the OP with choosing a random int and interpreting it as a float.

```kotlin
val rng = Random(42)
val l = (1 .. 1_000_000).map { _ -> Float.fromBits(rng.nextInt(Integer.MIN_VALUE, Integer.MAX_VALUE)) }
println(l.count())
println(l.filter { it < -1e+34f }.count())
println(l.filter { it > 1e+34f }.count())
```
result:
```
1000000
29563
29230
```

Kantis:
The set of numbers in `-1e34f..1e34f` is approximately 0.01% of all numbers in `Float.MIN_VALUE .. Float.MAX_VALUE`. Since all numbers have an equal probabilty of being drawn (as it stands), there's a much higher likelihood of numbers with high exponent to be drawn as there's more of them..

When I'm repeating your experiment, I'm only getting 29 numbers in the range. You're probably getting some hits from the  edge-cases included in your expressions? 

```kotlin
println(
    Arb.float(range = Float.MIN_VALUE..Float.MAX_VALUE)
        .take(1_000_000)
        .count { it in -1e34..1e34 }
) //  -> 29
```

I agree that an even distribution among exponents would likely yield better tests, as most numbers will be very similar as it stands.

rubenpieters:
> When I'm repeating your experiment

[...]

PR #3465, opened 2023-03-28
Fix float bias towards high exponents (#3463)

younghoonk17:
fix for #3463


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/floats.kt
```diff
@@ -68,7 +68,7 @@ fun Arb.Companion.numericFloat(
    min: Float = -Float.MAX_VALUE,
    max: Float = Float.MAX_VALUE
 ): Arb<Float> = arbitrary((numericEdgeCases.filter { it in min..max } + listOf(min, max)).distinct(), FloatShrinker) {
-   it.random.nextDouble(min.toDouble(), max.toDouble()).toFloat()
+   Float.fromBits(it.random.nextInt(min.toInt(), max.toInt()))
```
myuwono, line 71: I wonder if that's right? This would transform the float into an int, i.e. `3.0f` would become `3` instead of `1077936128`.

In theory we probably want to get the floats in the bits representations instead. Floats have a `Float.toBits()` extensions that should do that if I remember correctly. If that's right roundtripping that should work.

```kotlin
Float.fromBits(it.random.nextInt(min.toBits(), max.toBits()))
```

additionally, there's also another `Arb.Companion.float` in this file that is still using random.nextDouble. That would need to have the same logic.

myuwono: Ah thinking about this again.. i think we need to also think of the exponent because I'm not sure if currently it will generate the right float.. I wonder if there is a better way for us to do this. One thing that comes to mind would be to keep using double random, with some configurable probability of sampling from -minFloat and +minFloat if the arguments are in the range. Or perhaps utilising the edgecase random function to do that.

younghoonk17: Hey @myuwono I think this is the only way, I can't think of any elegant way of doing this except this
```
   val sign = it.random.nextInt(0, 2)
   val exponent = it.random.nextInt(0, 256)
   val mantissa = it.random.nextBits(23)
   val bits = (sign shl 31) or (exponent shl 23) or mantissa
   Float.fromBits(bits)
```

I also found a new way to draw the graph better
```
   test("draw distribution graph") {
      fun exponent(value: Float): Int {
         if (value == 0f) {
            return Int.MIN_VALUE
         }
         val bits = java.lang.Float.floatToIntBits(value)
         val exp = (bits shr 23) and 0xff
         return exp - 127
      }

      val generatedFloat = Arb.Companion.numericFloat(Float.MIN_VALUE, Float.MAX_VALUE * 100)
         .generate(RandomSource.seeded(1235), EdgeConfig(0.0))
         .take(1_000_000)
         .map { it.value}
         .groupBy{ number ->
            exponent(number)
         }
         .toSortedMap()

      fun intToGraph(count: Int): String{
         val stringNum = ceil(count.toDouble()/100)
         var result = ""
         (1..stringNum.toInt()).forEach {
            result += '-'
         }
         return result
      }

      generatedFloat.forEach {
         println("Exponent ${it.key} count ${intToGraph(it.value.count())} ")
      }
   }
```

sksamuel:
@younghoonk17 can you fix the tests then we can merge. Thanks!

PR #3469, opened 2023-04-02
expose random seed in property context

### LINKED ISSUE #3046: Provide RandomSource in PropertyContext
For certain tests it might be useful to have access to a deterministic random source. For example when working on (large) lists with expensive computations, where it is too expensive to verify all combinations exhaustively. Or simply when one element or a subset must be selected.

It would be helpful to be able to simply call:
`val element = theCompleteList.random(random)`

Omitting the random argument results in a test that is not repeatable.

Since the properties execute in the context of PropertyContext, it would be possible to proved the randomSource as a public member.

myuwono:
# In this PR
- introduce support for random source propagation within property context. 
- updated `shrink` to make sure the generated arbitraries yields consistent values when shrinking.
- add support for `Arb<A>.bind(): A` in checkAll and forAll which coordinates the generation of both edgecases / samples, correctly adhering to EdgeConfig passed within `PropertyContext`.
- updated test failure report to include the generated arbitraries

fixes #3046 

```kotlin
 checkAll(Arb.int(), Arb.int()) { a, b ->
    val intValue: Int = Arb.int(0..4).bind()
    val longValue: Long = Arb.long().bind()
    
    intValue.shouldBeBetween(0, 3)
 }
```

Reports the following:
```kotlin
Property test failed for inputs

0) 1398555929
1) -40476173
2) -4770396750018200643L (generated within property context)
3) 4 (generated within property context)

Caused by java.lang.AssertionError: 4 should be between (0, 3) at ...

Attempting to shrink arg 1398555929
Shrink #1: 0 fail
Shrink result (after 1 shrinks) => 0

Caused by java.lang.AssertionError: 4 should be between (0, 3) at ...

Attempting to shrink arg -40476173
Shrink #1: 0 fail
Shrink result (after 1 shrinks) => 0

Caused by java.lang.AssertionError: 4 should be between (0, 3) at ...


Property failed after 2 attempts
java.lang.AssertionError: Property failed after 2 attempts

	Arg 0: 0 (shrunk from 1398555929)
	Arg 1: 0 (shrunk from -40476173)
	Arg 2: -4770396750018200643L
	Arg 3: 4

Repeat this test by using seed -5391409397101461926

Caused by: 4 should be between (0, 3)
```

# Out of Scope / Further discussion required
For the time being I've included this function as a candidate to implement proper shrinking for dynamic arbs. 
Since we have to track contextual sample generations after test run, this means shrinking theoretically can be made possible. However, the solution to this might need to be discussed further and I'm keen to get feedback from both @sksamuel and @Kantis on how might we do this within the framework. I do vote to approach that in a separate PR as implementing such algorithm might be quite a challenging task on its own.

```kotlin
/**
 * The contextual shrinking is  only used for output purposes for the time being.
 * Contextual arb shrinking is theoretically possible. A strategy that may work is to treat shrinks
 * as a standard Vector of arbitrary dimension. However, this will need to be discussed and addressed separately.
 */
internal suspend fun PropertyContext.doContextualShrinking(
   mode: ShrinkingMode,
   test: suspend PropertyContext.() -> Unit
): List<ShrinkResult<*>> = generatedSamples().map {
   ShrinkResult(it.value, it.value, null)
}
```


sksamuel:
Definitely agree to do this in stages, and leave the dynamic seed generation to a future implementation.

kotest-property/src/commonMain/kotlin/io/kotest/property/context.kt
```diff
@@ -25,6 +27,32 @@ class PropertyContext(val config: PropTestConfig = PropTestConfig()) {
...
+
+   internal fun generatedSamples(): List<Sample<*>> = generatedSamples
+
+   /**
+    * Returns a [RandomSource] that can be used within this [PropertyContext]
+    */
+   fun randomSource(): RandomSource = contextualRandomSource ?: RandomSource.default()
+
+   /**
+    * Extracts a sample or edgecase value from an [Arb] using the contextual [RandomSource].
```
myuwono, line 48: @sksamuel I'm trying to find the correct place to put `randomSource()` and `Arb<A>.bind()` under the web documentation. Do you have a suggestion?

myuwono:
@sksamuel This PR is now ready to be reviewed. Please let me know if you'd like me to update anything.

myuwono:
@sksamuel would it be possible for this change to be included in the 5.6 release?

sksamuel:
Yep I'll review and merge before I release 5.6 this weekend

sksamuel:
Looks good merging in. One suggestion though, is we could move the setup to inside the property context, rather than supply it after it's been created.

sksamuel (approved):

Issue #3391, opened 2023-02-06
Mockito-kotlin + kotest?


rchugunov:
Kotest 5.5.4
Mockito 4.9.0
Mockito-Kotlin 4.1.0
JDK 11.0.17

I see no examples on the internet of using Kotest with Mockito-Kotlin. We already have JUnit5 and Mockito-Kotlin in our project but wanted to give a try to Kotest. I have a test where I do mocking like this:

    internal class MyTestSpec : FreeSpec({

        extensions(RxUiKotestExtension, ContextWrapperKotestExtension)

        lateinit var target: MyPresenter

        val view = mock<MyView>()
        
        "GIVEN some data" - {

              val houseName = "houseName"
              val streetName = "streetName"

Once I run it - it fails with the following:

java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at kotlin.reflect.jvm.internal.calls.CallerImpl$Constructor.call(CallerImpl.kt:41)
	at kotlin.reflect.jvm.internal.KCallableImpl.call(KCallableImpl.kt:108)
	at io.kotest.engine.spec.InstantiateSpecKt.javaReflectNewInstance(instantiateSpec.kt:51)
	at io.kotest.engine.spec.InstantiateSpecKt.createAndInitializeSpec(instantiateSpec.kt:30)
	at io.kotest.engine.spec.InstantiateSpecKt.instantiate(instantiateSpec.kt:11)
	at io.kotest.engine.spec.SpecRefKt.instance(SpecRef.kt:14)
	at io.kotest.engine.spec.SpecExecutor.createInstance-gIAlu-s(SpecExecutor.kt:128)
	at io.kotest.engine.spec.SpecExecutor.access$createInstance-gIAlu-s(SpecExecutor.kt:49)
	at io.kotest.engine.spec.SpecExecutor$referenceInterceptors$innerExecute$1.invokeSuspend(SpecExecutor.kt:86)
[...]

PR #3470, opened 2023-04-02
Move constant now to new module; bump junit to 1.8.2

### LINKED ISSUE #3433: JUnit Platform Version Clash
**Version**: 5.5.5 (currently latest)

**To reproduce**: Gradle build with JUnit JVM runner `io.kotest:kotest-runner-junit5-jvm` during testing throws (does not break the build)
```
java.lang.NoSuchMethodError: 'org.junit.platform.engine.ConfigurationParameters org.junit.platform.launcher.TestPlan.getConfigurationParameters()
```

**Reason**: Currently, used JUnit platform version (e.g. in JVM runner artifact) is `1.7.2` but it clashes with version that is transitively provided from `org.junit.jupiter:junit-jupiter` (version `5.8.2`) that uses `1.8.2`. So in the end, there are various platform JARs in the classpath (e.g. engine in `1.8.2` while launcher `1.7.2` etc...)

**How To Fix**: Upgrade JUnit platform version in [libs.versions.toml](https://github.com/kotest/kotest/blob/master/gradle/libs.versions.toml#L10) to `1.8.2` (or even newer `1.9.2`)

**Wokraround**: Specify (override) version of JUnit platform libraries in the project that uses kotest with JUnit runner, e.g.:
```
testImplementation("org.junit.platform:junit-platform-commons:1.8.2")
testImplementation("org.junit.platform:junit-platform-launcher:1.8.2")
testImplementation("org.junit.platform:junit-platform-engine:1.8.2")
testImplementation("org.junit.platform:junit-platform-suite-api:1.8.2")
```

Side note, also consider excluding `junit:junit` (legacy JUnit4) from `kotest-extensions-jvm` that have it transitively via `io.mockk:mockk-jvm:1.13.1`

sksamuel:
Fixes #3433 #3391

Bumps junit to 1.8.2
Moves the constant now extensions to a new module that avoids bringing mockk into the regular extensions.


i-walker (approved):

obecker:
Apparently `MutableClock` has been silently removed. I don't see any connection to the linked issues. Could you please share what the proper replacement is? Thank you.

sksamuel:
The test clock extensions module.

obecker:
Ah, ok. It's the `TestClock` in https://github.com/kotest/kotest-extensions-clock
Thanks

sksamuel:
The mutable clock was only added as an internal implementation, and it was a mistake for us to make it public. The test clock repo is a better way to provide this.

PR #3476, opened 2023-04-02
Implement shouldThrowSoftly

### LINKED ISSUE #3297: shouldThrow does not respect globalAssertSoftly
kotest 5.4.2
I configure globalAssertSoftly=true, and it mostry works well, except in combination with shouldThrow.
This only reports 1 failure, and never prints "2":
```
    "should fail twice" {
        shouldThrow<NullPointerException> { println("1") }
        shouldThrow<NullPointerException> { println("2") }
    }
```

This only reports failures 1 and 2:
```
    "should fail 4 times" {
        1 shouldBe 0
        2 shouldBe 0
        shouldThrow<NullPointerException> { println("3") }
        4 shouldBe 0
    }
```
3 is printed, but the failure to throw an NPE is not reported
The final assert is not executed at all

sksamuel:
Fixes #3297


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/throwables/CovariantThrowableHandling.kt
```diff
@@ -140,6 +143,45 @@ inline fun <reified T : Throwable> shouldThrow(block: () -> Any?): T {
...
+ *
+ * Use this function to wrap a block of code that you'd like to verify whether it throws [T] (or subclasses) or not.
+ *
+ * This function will include subclasses of [T]. For example, if you test for [java.io.IOException] and
+ * the code block throws [java.io.FileNotFoundException], the test will pass.
+ *
+ * If you wish to test for a specific class strictly (excluding subclasses), use [shouldThrowExactly] instead.
+ *
+ * This function is identical to [shouldThrow] except is used inside soft assertion blocks. When inside such a block,
+ * errors are only throw at the end of the block, rather than immediately. Therefore, the signature cannot return
```
myuwono: nit - typo: errors are only thrown*

myuwono (approved):

i-walker (approved):

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/throwables/CovariantThrowableHandling.kt
sksamuel: Updated

PR #3479, opened 2023-04-03
fix(JunitXmlReporter): resolve paths with irrelevant directories

Stefanqn:


Stefanqn:
this fix is needed for maven projects where no `build` dir exists

sksamuel (approved):

sksamuel:
👍🏻

PR #3487, opened 2023-04-17
add language annotation to json matchers

abendt:
this is a follow-up to https://github.com/kotest/kotest/pull/3438

this adds the `@KotestLanguage` annotation to the `expected` param of the matchers `shouldEqualJson` and `shouldNotEqualJson`. This is to enable proper syntax hightlighting in IntelliJ.

* also add a test for shouldNotEqualJson matcher as apparently there was none


Kantis:
This will sadly not work, but might be worthwhile to add anyway for the day when https://youtrack.jetbrains.com/issue/KTIJ-1001/infix-function-parameter-annotated-with-Language-has-no-language-injected is fixed.

abendt:
yes, i have noticed that too. thanks for linking the bug report.

however this could be useful already as the highlighting works when you call the infix function with the normal syntax

image

Kantis (approved):

Kantis:
Good point @abendt 🙂 thanks for the contribution!

PR #3489, opened 2023-04-18
Using generated accessors for kotlin targets

Kantis:

### DIFF
--- a/documentation/docs/proptest/index.mdx
+++ b/documentation/docs/proptest/index.mdx
@@ -84,7 +84,7 @@ Add the following dependency to your build.
 ```kotlin
 kotlin {
   sourceSets {
-    val commonTest by getting {
+    commonTest {
       dependencies {
         implementation("io.kotest:kotest-property:$version")
       }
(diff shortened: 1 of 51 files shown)


aSemy:
While of course I think this change is great, it has unfortunately broken the `jvmOnly` flag. 

When `jvmOnly` is true, the Kotlin/JS and Native targets don't get enabled.

https://github.com/kotest/kotest/blob/ee33820b7d7f9382474853d7c0c75067600597ad/buildSrc/src/main/kotlin/kotest-js-conventions.gradle.kts#L6-L16

This means that Gradle won't generate any of the accessors, and so this code 

https://github.com/kotest/kotest/blob/ee33820b7d7f9382474853d7c0c75067600597ad/kotest-extensions/kotest-extensions-http/build.gradle.kts#L31-L36

will fail because even though it's wrapped in an `if` statement, the `build.gradle.kts` can't be compiled because there's no `jsMain` value or function.

Kantis:
Ah, gotcha. Let's revert

PR #3496, opened 2023-04-24
fix map edgecases

myuwono:
# in this PR
- remove recently added includeEmpty arg in favour of edgecase derived from minSize. 
- modify edgecase generation logic to be determined by minSize instead
- This ensures sane generation of edgecases based on the parameter provided in the map

# Context

The recently added includeEmpty arg was introduced in 5.6.0 and was causing the regression


sksamuel (approved):

sksamuel:
lgtm

PR #3502, opened 2023-04-25
Adding kotlin BCV

### LINKED ISSUE #3430: Use Kotlin Binary compatibility validator to prevent unwanted breaking binary changes
Currently we have no automated way of telling how/if the API of Kotest changes. By having automatic tests on the API surface, we can easily see if a PR contributes breaking changes.

## Proposed way of working
Using https://github.com/Kotlin/binary-compatibility-validator,  GH workflow would run `apiCheck` on PRs, failing builds where the API changes. 

In order to submit a PR with API changes, contributor would have to run `./gradlew apiDump` to update the API snapshot with their changes.

Kantis:
In this PR:

- Adding Binary compatibility validator 
- Updating Contribution guidelines with info on how to update the API dumps.

Fixes #3430

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,3 +26,12 @@ We follow the [Kotlin Coding Conventions](http://kotlinlang.org/docs/reference/c
 
 * Minimize mutability
 * Choose self-explanatory names
+
+
+Binary compatibility
+------------------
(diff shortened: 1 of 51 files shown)


Kantis:
Decided to go with everything in a single PR (instead of doing some modules to try it out) since I think it'll be easier to evaluate when it has been used consistently. It'll also be easier to revert if we decide to do so.

PR #3520, opened 2023-05-06
Issue/3463 bias floats

younghoonk17:
fix for https://github.com/kotest/kotest/issues/3463

Using bit manipulation to create default float, I think this will result in most distributed generation


younghoonk17:
@myuwono How about something like this?

myuwono:
This is amazing @younghoonk17 !! thank you

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/floats.kt
```diff
@@ -68,7 +72,15 @@ fun Arb.Companion.numericFloat(
...
    max: Float = Float.MAX_VALUE
 ): Arb<Float> = arbitrary((numericEdgeCases.filter { it in min..max } + listOf(min, max)).distinct(), FloatShrinker) {
-   it.random.nextDouble(min.toDouble(), max.toDouble()).toFloat()
+   if (min == -Float.MAX_VALUE && max == Float.MAX_VALUE) {
+      val sign = it.random.nextBits(1)
+      val exponent = it.random.nextInt(0,254)
+      val mantissa = it.random.nextBits(23)
+      val bits = (sign shl 31) or (exponent shl 23) or mantissa
+      Float.fromBits(bits)
+   }else {
```
myuwono, line 81: nit - formatting

myuwono (dismissed):

Kantis:
Wont this have the issue with uneven distribution as soon as you supply a min or max value? That would feel kinda unexpected IMO

myuwono:
actually - that's a good point - let me check the distribution quickly

myuwono:
you're right @Kantis the population has became non-uniform

this is a simple test
```kotlin
   test("should be uniformly distributed") {
      // expected percentiles for a uniformly distributed floats
      val p25 = -Float.MAX_VALUE / 2 
      val p50 = 0.0f
      val p75 = Float.MAX_VALUE / 2
      val p100 = Float.MAX_VALUE
      

      val numbers = Arb.float().take(10_000_000).toList()
      println(numbers.count { it < p25 })
      println(numbers.count { it < p50 })
      println(numbers.count { it < p75 })
      println(numbers.count { it < p100 })
   }
```

that printed
```
0
4999160
10000000
10000000
```

I guess the effect that this change has to the statistical distribution is that it now uniformly randomizes the mantissa and exponents, instead of the real numeric representation.

This is going to change the behaviour of floats in a rather unexpected way.

myuwono:
@rubenpieters I think we need to get a better understanding of the use case in order to proceed. I don't think we can change `Arb.float()` in this way as it might cause rather unexpected behaviour change as @Kantis mentioned. @younghoonk17 's solution would work for your application but it is a rather special float distribution. We might need to think of:

1.  a reason why we want to introduce it in kotest, and 
2. if there is a good reason for 1 - a good name for it.

Kantis:
@myuwono my feedback was aimed at this line:
```kotlin
if (min == -Float.MAX_VALUE && max == Float.MAX_VALUE) {
```

IMO we need to make the uniformity in exponent + mantissa work even when you supply a min or max value. If you try your uniformity-test again by just adding  `Arb.float(max = Float.MAX_VALUE - 1.0)` you would likely be back to having a uniformity by value, rather than exponent/mantissa.

IMO uniformity by exponent + mantissa is desirable so we get tests which tries both big and small numbers. The existing Kotest implementation almost solely focus on very large numbers. See https://github.com/kotest/kotest/issues/3463#issuecomment-1481830704

myuwono:
Yeah I understand but this is what currently is in my mind..

This is tricky.. even if we can make it work, it does not really make sense for users that was trying to test for `Arb.float(1f..10f)`. In this case I think your assessment is actually very relevant. 
> The set of numbers in -1e34f..1e34f is approximately 0.01% of all numbers in Float.MIN_VALUE .. Float.MAX_VALUE.

If I were the user, when I defined `Arb.float(1f..10f)` I would likely want a uniform sampling by value between `1f` and `10f`.

It's a different story if a user wanted to test multi-modal distributions. e.g. numbers between `0f..1f` (or even smaller exponents, e.g. `1e-34`) and `1f..1_000_000f`, with equal probability.. won't this be a use case for `Arb.choice(...)`?

```kotlin
Arb.choice(
  Arb.float(0f..1f),
  Arb.float(0f..1e-34f),
  ...
  Arb.float(1f..1_000_000f)
)
```

I guess we can also simulate it like so with enough steps. But if we decide to do this, how would we decide which step would we take?

I'm trying to imagine a generalized solution, but the more I think about it, the more this becomes very problem specific. Maybe I'm missing something.

Another alternative might be to assign this as an edgecase logic instead?

myuwono:
For instance:

```kotlin
 test("should sample both small and large exponents") {
    val numbers = Arb.choice(
       Arb.float(-1e-34f..1e-34f),
       Arb.float(-Float.MAX_VALUE..Float.MAX_VALUE)
    ).take(1_000_000).toList()

    println(numbers.count { it in -1e34..1e34 })
    println(numbers.count { it < -1e34 || it > 1e34 })
 }
```

prints
```
499876
500124
```

myuwono:
Another idea: defining a distribution parameter
```kotlin
Arb.float(0f..1_000_000f, distribution = NumericDistribution.Logarithmic(10)) // default would be NumericDistribution.Uniform
```

this would generate floats which are uniformly distributed in the logarithmic scale, i.e. `0, .. , 10^-1 - 10^0, 10^0 - 10^1, 10^1 - 10^2, ...`

Perhaps this is a better approach? wdyt @Kantis @younghoonk17  ?

Kantis:
A distribution parameter sounds like an interesting approach. 

I'm still not sure the current behavior would be a good default, though.  
* It misses testing of very small numbers
* What software actually deals with things that are 1e34+ in scale? 

But given what we learned about changing behaviour of arbs (re: the ascii stuff), we should probably not change defaults outside of a major release.

Maybe we can reason a bit about in which scenarios a developer actually want uniformity by value, and not by logarithmic scale?

myuwono:
I'm not too sure on whichever approach is more appropriate than which. But yeah, I 100% agree kotest's default behaviour for full-range floats isn't too useful. 

We need to think about the specific problem space in question. Applications that processes high-precision sensors data may care about high-precision floating points within a particular bound, while some other applications e.g. payment processing may focus more on the larger numeric ranges and protection for numeric overflows.

The current Kotest's behaviour that assumes uniform distribution in values is a particularly fair assumption when the user explicitly want a known bounded range of uniformly distributed numbers. However, this assumption may not be entirely too useful, when the range selected was the entire float range and that the requirement now becomes a lot more ambiguous. The emphasis becomes all classes of floats, which log scale solves it well.

By the way, looking at various property testing libraries for some data points:
- Kotlin: current `kotest` uses uniformly distributed floats,
- Java: `junit-quickcheck` uses [uniformly distributed floats](https://github.com/pholser/junit-quickcheck/blob/master/core/src/main/java/com/pholser/junit/quickcheck/random/SourceOfRandomness.java#L223-L255), similarly to the current kotest approach 
- F# / C# / .NET `fscheck` uses [uniformly distributed floats](https://github.com/fscheck/FsCheck/blob/633c064ca96337a3689559b56ae8cb93cf1d7589/src/FsCheck/Arbitrary.fs#L450-L452), similarly to kotest and junit-quickcheck 
- Scala: `scalacheck` uses [bit randomization](https://github.com/typelevel/scalacheck/blob/a57a41055ff1abb608967c4535872a70d0ab0633/core/shared/src/main/scala/org/scalacheck/Arbitrary.scala#L132-L148) similar to the approach that @younghoonk17 's proposed. It needs to be noted that scalacheck does not have the ability to customize the range though. 
- I can't seem to distill Typescript's fast-check and Python's hypothesis yet.

younghoonk17:
yea you guys are right my approach is not uniform when given the range, let me see if I can find a way to make do the bit manipulation with given range... no promises 😭 

Like @myuwono said everyone seems to settle with the below approach, and im pretty sure that is not uniform
```
min + (max - min) * nextDouble()
```

myuwono:
@younghoonk17 if you can get bit manipulation to work with ranges, that'll be superb. Note it may be very challenging due to the way IEEE-754 bits are structured. 

In case you're completely blocked, one idea would be emulate log-scale with piecewice linear interpolation. Since this splits the generation by exponents, float varieties of various exponents should be better represented. This make it possible to avoid usage of `pow` and make an arb that is still relatively performant for end users.

```kotlin
fun Arb.Companion.float(range: ClosedFloatingPointRange<Float> = ...): Arb<Float> {
   /**
    * piecewise linear splits the input range into a list of log-scale separated ranges.
    * E.g. an -1023.0f .. 1003.0f range will in theory yield
    * [
    *   -1023.0f .. -100.0f,
    *   -100.0f .. -10.0f,
    *   -10.0f .. -1.0f,
    *   // notice the low exponent splits is balanced depending on the number of ranges > 1 
    *   // This is to ensure that the probability to sample floats <1 and >1 are equally split
    *   -1.0f .. -1e-15f,
    *   -1e-15f .. -1e-30f,
    *   -1e-30f .. -Float.MIN_FLOAT,
    *   Float.MIN_FLOAT .. 1e-30f,
    *   1e-30f .. 1e-15f,
    *   1e-15f .. 1.0f,
    *   1.0f .. 10.0f,
    *   10.0f .. 100.0f,
    *   100.0f .. 1003.0f,
    * ]
    **/
   val piecewiseLinearLogarithmicRanges: List<ClosedFloatingPointRange<Float>> = ...
   return arbitrary(
      ...
   ) { 
      val chosenRange = piecewiseLinearLogarithmicRanges.random(it.random)
      it.random.nextDouble(chosenRange.start.toDouble(), chosenRange.endInclusive.toDouble()).toFloat()
   }
}
```

PR #3523, opened 2023-05-08
Adding support for partial permutations

mickeelm:
Adding length parameter (defaults to list size) to the permutations function, providing the opportunity of creating exhaustives of partial permutations/k-permutations.

Also, changed the assertion of the full permutations test from shouldBe to shouldContainExactlyInAnyOrder, as the order in which the permutations come are not (should not be) relevant. This also makes the test less dependent on the underlying implementation.

This PR does break binary compatibility, due to adding the length parameter.


mickeelm:
Converting to draft as partial permutations of length zero aren't correctly handled, needs to be fixed.

Kantis:
@mickeelm this needs an updated apiDump :)

sksamuel (approved):

PR #3529, opened 2023-05-16
fix reflective bind failing on sealed class with object member

### LINKED ISSUE #3528: JVM Arb reflective bind unable to generate sealed class with objects
Kotest Version:
5.6.2

When using reflective bind in the jvm, it seems it can't generate arbitraries for sealed class that contains singleton objects as its member.

```kotlin
sealed class Pet {
  object Dog: Pet()
  object Cat: Pet()
}

val petArb: Arb<Pet> = Arb.bind<Pet>()
petArb.take(100).toList()
```

this fails with: `Could not locate a primary constructor for Pet.Cat`

myuwono:
# In this PR
- support object instance in reflective bind

fixes #3528


Kantis (approved):
Nice ⭐️

PR #3530, opened 2023-05-17
fix(Docs): Fix documented annotation as `@RequireTags` which should be `@RequiresTag`

zakhenry:


Kantis:
Thanks for the fix! 👍 💯

PR #3532, opened 2023-05-18
fix(RequiresTagInterceptor): Add fix to handle scenario where no tags are present in the runtime tags, but some specs are annotated with @RequiresTag()

### LINKED ISSUE #3531: @RequiresTag does not work as expected
5.6

Reopening #3274 (marked as bug but closed due to inactivity).

*Expected behavior*:
Adding a `RequiresTag()` annotation to a spec class prevents that class from being run *unless* the tag is in the list of active tags in the argument (i.e. not negated, or no tags specified)

*Observed behavior*:
Adding a `RequiresTag()` annotation to a spec class has no effect *unless* some other tag is specified (i.e. TagExpression is not null)

concrete example:
I have a tag `object EndToEnd: Tag()` that I don't want to run by default when `./gradlew myproject:test` is run. I've added `@RequiresTag("EndToEnd")` to my e2e spec classes. 

workaround: I could negate it with `./gradlew myproject:test -Dkotest.tags="!EndToEnd"` however that seems to defeat the purpose of `RequiresTag()` annotation

zakhenry:
Fixes #3531


Kantis:
@zakhenry I would like to check when the behaviour of `@RequiresTag` regressed (assuming it did). The strategy of testing in b551ce1 relies on `RequiresTagInterceptor` which is newer than the original issue, so I can't cherry-pick the test into older versions to see when it started failing. If you don't mind, I propose an alternate strategy in 01ea26c which should be backportable.

Edit: Had to update api dump as well

zakhenry:
Yea I guess that is fine for determining when this happened though the fact that you had to disable the test isn't ideal so it might be good to restore the explicit `RequiresTagInterceptor` tests as they cover the use case more directly by constructing the TagExpression

zakhenry:
Bump

Kantis (approved):

PR #3538, opened 2023-05-22
Update signature on collections' documentation

PaoloSimonazziit:
The methods do not reflect the real signature of the methods which can be found in the library...


sksamuel (approved):

sksamuel:
Thanks!

Issue #3541, opened 2023-05-25
add enumsExcept to Exhaustive


vladeekkk:
There is 
```
inline fun <reified T : Enum<T>> Exhaustive.Companion.enum(): Exhaustive<T> {
   val constants = enumValues<T>().asList()
   return exhaustive(constants)
}
```
in `package io.kotest.property.exhaustive`

Can I make a PR and add something like:
```
inline fun <reified T : Enum<T>> Exhaustive.Companion.enumsExcept(vararg items: T): Exhaustive<T> {
    val constants = enumValues<T>().toMutableList().apply { removeAll(items.toSet()) }
    return exhaustive(constants)
}
```
or
```
inline fun <reified T : Enum<T>> allEnumsExcept(vararg items: T): Exhaustive<T> {
    val constants = enumValues<T>().toMutableList().apply { removeAll(items.toSet()) }
    return exhaustive(constants)
}
```
So that it will be convinient to use in tests. For instance:
1) Add a test for some enum value X
2) Add a test for all other enums except X 

`checkAll(Exhaustive.enumsExcept(SOME_PRODUCT_ENUM)) { productId -> and so on`

vladeekkk:
`checkAll(Exhaustive.enumsExcept(SOME_PRODUCT_ENUM)) { productId -> and so on`

PR #3552, opened 2023-06-02
add exhaustive enumsExcept

vladeekkk:
`Exhaustive.Companion.enum()` helps to get an Exhaustive of all enums of a enum class.

However, sometimes we need to test that some predicate is `true` for exact enums of a enum class and `false` for remaining enums. Here `enumsExcept` function may help

P.S: is it okay that `./gradlew apiDump` didn't add anything?


vladeekkk:
https://github.com/kotest/kotest/issues/3541

sksamuel (approved):

sksamuel:
Nice :)

Issue #3503, opened 2023-04-25
Arbitrary for Date in JVM


Jacks0N23:
I saw at at [Generators List](https://kotest.io/docs/proptest/property-test-generators-list.html) there no Arb for Date in JVM
Could you add it?

Kantis:
I think we can add this

sksamuel:
Are we talking java.util.Date ?

Kantis:
I assumed so. Is java.sql.Date much used?

@Jacks0N23 please clarify 😄

Jacks0N23:
Yeah, sorry, for delay, of course I mean java.util.Date

sksamuel:
PR welcome.

PR #3554, opened 2023-06-03
Add Arb<Date> support

vladeekkk:
1) added Arb for java.util.Date
2) wrote test for it

### DIFF
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
@@ -1,6 +1,7 @@
 package io.kotest.property.arbitrary
 
 import io.kotest.property.Arb
+import java.text.SimpleDateFormat
 import java.time.Instant
 import java.time.LocalDate
 import java.time.LocalDateTime
@@ -15,6 +16,7 @@ import java.time.ZonedDateTime
 import java.time.temporal.ChronoUnit
 import java.time.temporal.TemporalQueries.localDate
 import java.time.temporal.TemporalQueries.localTime
+import java.util.Date
 import kotlin.random.Random
 
 /**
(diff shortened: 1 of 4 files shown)


kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
```diff
@@ -247,3 +248,17 @@ fun Arb.Companion.zonedDateTime(
    localDateTime(minValue, maxValue),
    zoneId
 ) { time, zone -> time.atZone(zone) }
+
+/**
+ * Arberates a stream of random [Date]
+ */
+fun Arb.Companion.date(
+   minDate: String = "1970-01-01",
+   maxDate: String = "2050-12-31",
```
vladeekkk, line 258: working with java.util.Date is inconvenient, is it okay that default params are Strings?

Kantis: What do you think about provide functions to use either Date or String in the API. 

`Arb.date()` feels like a very prioritized API for a class that probably shouldnt be used a lot (`java.util.Date`). What do you think about calling it `Arb.javaDate()`?

vladeekkk: Yep, renaming to javaDate seems better ✔️ 

Also added second `Arb.javaDate()` function

vladeekkk:
Some checks failed because `LocalDate.ofInstant()` is only in Java 9 and later, so I used `ZonedDateTime.ofInstant()` instead in `DateTest.kt`.

vladeekkk:
https://github.com/kotest/kotest/issues/3503

Kantis (approved):

Kantis:
Thank you for the contribution @vladeekkk 🙏🏼

PR #3562, opened 2023-06-17
Update build to use Gradle toolchains and modern JDK

lwasyl:
I've checked out Kotest repository and couldn't build it because I'm using JDK20. This PR is a suggestion on how this could be fixed: now contributors can run the build using any JDK they want, and Gradle will download JDK used for the build itself. This ensures all developers and CI will use the same JDK to build the project, and the resulting artifacts will always target the same bytecode

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -26,7 +26,7 @@ jobs:
             uses: actions/setup-java@v3
             with:
                distribution: 'temurin'
-               java-version: '8'
+               java-version: '17'
 
          -  uses: gradle/gradle-build-action@v2
 
@@ -57,7 +57,7 @@ jobs:
             uses: actions/setup-java@v3
             with:
                distribution: 'temurin'
-               java-version: '8'
+               java-version: '17'
 
          -  uses: gradle/gradle-build-action@v2
 
@@ -105,7 +105,7 @@ jobs:
          -  name: Setup JDK
             uses: actions/setup-java@v3
             with:
-               java-version: '8'
+               java-version: '17'
                distribution: 'temurin'
 
          -  uses: gradle/gradle-build-action@v2
@@ -142,7 +142,7 @@ jobs:
          -  name: Setup JDK
             uses: actions/setup-java@v3
             with:
-               java-version: '8'
+               java-version: '17'
                distribution: 'temurin'
 
          -  uses: gradle/gradle-build-action@v2
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -7,6 +7,10 @@ repositories {
    gradlePluginPortal()
 }
 
+kotlin {
+   jvmToolchain(17)
+}
+
 dependencies {
    implementation(libs.testlogger.gradle.plugin)
    implementation(libs.kotlin.gradle.plugin)
--- a/gradle.properties
+++ b/gradle.properties
@@ -25,6 +25,8 @@ kotlin.native.ignoreDisabledTargets=true
 systemProp.kotlin.native.disableCompilerDaemon=true
 org.gradle.configureondemand=false
 kotlin.mpp.stability.nowarn=true
+# https://youtrack.jetbrains.com/issue/KT-58063 (should be fixed in Kotlin 1.9)
+org.gradle.configuration-cache=false
 
 android.useAndroidX=true
 android.enableJetifier=true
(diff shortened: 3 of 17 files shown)


.github/workflows/PR.yml
```diff
@@ -26,7 +26,7 @@ jobs:
             uses: actions/setup-java@v3
             with:
                distribution: 'temurin'
-               java-version: '8'
+               java-version: '17'
```
lwasyl, line 29: The JDK used to run the build is no longer determining the target artifacts bytecode, so it's safe to use a more recent JDK.

This could easily be `'20'` as well, but I decided to use an LTS version until I get feedback that more recent versions are OK too. Fwiw I've been running JDK20 for a long time without issues (and JDK19 and 18 before that). Personally I'd suggest to always use the newest JDK for both CI and for the toolchains

gradle.properties
```diff
@@ -25,6 +25,8 @@ kotlin.native.ignoreDisabledTargets=true
 systemProp.kotlin.native.disableCompilerDaemon=true
 org.gradle.configureondemand=false
 kotlin.mpp.stability.nowarn=true
+# https://youtrack.jetbrains.com/issue/KT-58063 (should be fixed in Kotlin 1.9)
+org.gradle.configuration-cache=false
```
lwasyl, line 29: Configuration cache is now stable, so it's reasonable that people will enable it in `~/.gradle/gradle.properties`. Unfortunately, some multiplatform tasks are not yet compatible with it

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
```diff
@@ -92,6 +93,19 @@ gradlePlugin {
    }
 }
 
+kotlin {
```
lwasyl, line 96: This module doesn't use convention plugin for some reason, so I copied the configuration here assuming that using convention plugin here is tricky

buildSrc/build.gradle.kts
```diff
@@ -7,6 +10,18 @@ repositories {
    gradlePluginPortal()
 }
 
+kotlin {
+   jvmToolchain(17)
+}
+
+tasks.withType<KotlinCompile>().configureEach {
+   compilerOptions.jvmTarget.set(JvmTarget.JVM_1_8)
```
lwasyl: Btw if the reason to use JDK8 bytecode is android, then you should know Android supports JDK11 libraries for quite some time now. So unless there are other reasons to generate Java 8 bytecode, this can probably be bumped to 11 too

.github/workflows/PR.yml
Kantis: I think staying on LTS versions makes sense for this project. I dont think we would bother to bump these versions every 6 months, so it's likely we would end up on an unsupported version. Besides, I don't think we gain much from being on latest JDK?

Kantis (approved):

buildSrc/build.gradle.kts
sksamuel: We can bump this to 11 then.

.github/workflows/PR.yml
sksamuel: I agree. Compatibility is more important than anything else so lets stick to the LTS 17. We can go to 21 when that's out.

sksamuel (approved):

.github/workflows/PR.yml
lwasyl: > Besides, I don't think we gain much from being on latest JDK?

Generally, you do — newer JDKs bring performance improvements and bug fixes. Some are really noticeable (e.g. a metaspace leak that was transparently handled by Gradle but incurred massive memory cost, see https://youtrack.jetbrains.com/issue/KT-55831) but there are also smaller performance fixes, improvements to GC algorithms etc. This may not matter for CI, but this is related to:

> Compatibility is more important than anything else so lets stick to the LTS 17

This PR is specifically fixing incompatibility with JDK20, which is stable and is a valid JDK to run builds with, but Kotest build was failing. So building with JDK17 on CI is not guaranteeing further compatibility with JDK20. It's of course valid to say that Kotest is only _compatible_ with JDK17, but then you would force your consumers to use an old JDK (note that this PR had to change some _tests_ too, which means the consumers face the same issues). So if anything, it'd make sense to test the build on both LTS and current JDKs.

Same goes for toolchains, using newer JDK gives performance improvements for the cost of updating a number in couple of places every half a year. Of course that's your decision (I'm just happy I can now have deterministic output regardless of JDK I use to run the build). I just wanted to point out new JDKs _do_ have benefits and using LTS _doesn't_ solve any compatibility issues in a world where developers use non-LTS releases. Personally I'd say it's more reasonable to expect people to use most recent stable releases of software, even if LTS exists. In case of Java I think LTS releases make much more sense for something like servers where updating runtime is problematic

PR #3570, opened 2023-07-17
Feature/improve arrow documentation

rprominski:
Few change in documentation in arrow assertions documentation :

1. Changed link to maven central that you can use to search for kotest-assertions-arrow. Current one lists old version first, and if you use it with kotest version > 5.0 you get nasty exception at runtime. Took me a while to figure out that I actually need to use second library from the top in maven central, so I changed the link, and now only one (correct) library is listed - it's less confusing.

2. Changed tables formatting and corrected one preposition - mostly to get rid of warnings in intellij


Kantis (approved):

Kantis:
Thanks @rprominski  🙌🏼

PR #3573, opened 2023-07-19
Fix memory leak in CollectingTestEngineListener

### LINKED ISSUE #3572: Memory Leak in CollectingTestEngineListener
`CollectingTestEngineListener` stores a collection of

```
   val tests = concurrentHashMap<TestCase, TestResult>()
```
where 

```
data class TestCase(
   ...
   // the spec instance that contains this testcase
   val spec: Spec,
   ...
)
```

for all executed tests. It leads to retention of all test spec instances during the tests suite execution.
If user specs have some beans injected from the application context, it may lead to the retention of all ApplicationContexts created in tests.
This happens, for example, if a Spec implementation injects Micronaut httpClient for one of an application controllers.

**KoTest: 5.6.2**

Spikhalskiy:
This PR removes unnecessary retention of all the Spec instances by the `CollectingTestEngineListener` that leads to a memory leak and increased GC activity / OOM with more executed tests.
This is a proposed fix and potentially Closes #3572


sksamuel:
Nice

sksamuel (approved):

PR #3587, opened 2023-07-24
Fixing TestAbortedException not being marked as Ignored

### LINKED ISSUE #3586: TestAbortedException is not flagged as Ignored

Kantis:
Fixes #3586


sksamuel (approved):

sksamuel:
FYI you can do emptyList() instead of listOf() and that returns a preallocated empty list.

PR #3596, opened 2023-07-29
Add expect block to skip tests

### LINKED ISSUE #3582: Pleasant mechanism for ignoring/skipping a test from within the test code.
Kotest provides this feature already:

```kotlin
val disableDangerOnFridays: EnabledIf = { !(it.name.startsWith("danger") && isFriday()) }

"danger Will Robinson".config(enabledIf = disableDangerOnFridays) {
  // test here
}

"safe Will Robinson".config(enabledIf = disableDangerOnFridays) {
 // test here
}
```

However, I personally find the syntax quite cumbersome compared to one of the ways that Junit has achieved the same thing with their concept of "assumtions":

```kotlin
@Test
fun dangerWillRobinson() {
    assumeTrue(isFriday())

    // test here
}
```

Not that in the junit case, the assumption can be added or removed by adding/deleting a line with no additional formatting changes or line breaks/joins required, where as the kotest version requires the insertion of the `config` function in the chain. Depending on the length of your test name this will also require a bit of formatting and is generally a bit of a pain (I've assumed that 99.9% of tests are not using the `config` function already and it has to be added just to use `enabledIf`).

Furthermore, the machanism that makes the Junit version work is by throwing a special exception called `TestAbortedException` I think. This makes it very extensible since it's just plain old code, nothing fancy like annotations that have to be processed later. There is no obvious way to extend the kotest version to develop a custom DSL for conditional test execution.

Note that `Tag`s suffer from a similar syntax issue. In order to add tags to a test the syntax is also a bit cumbersome.

```kotlin
"danger Will Robinson".config(tags = setOf(Tag1, Tag2)) {
  // test here
}
```

However, that's beside the point of this feature request.

It would be nice to have a way to dynamically skip/ignore tests by pasting a single line of code, on a new line with no other formating changes required. Junit's mechanism via a special exception has always worked well for me in the past but perhaps there are other ways that the config syntax can be improved. For example by bypassing the `config` function:

```kotlin
"danger Will Robinson"
    .enabledIf(disableDangerOnFridays))
    .tags(Tag1, Tag2) {
  // test here
}
```

or (using BehaviorSpec as example this time):

```kotlin
then("name").config(enabledIf = foo, tags = bar) {
}
//becomes
then(
    "name",
    enabledIf = foo,
    tags = bar
) {
}
```

I appreciate this might be a case of personal preference and you have to be opinionated about your API and can't please everyone so thanks for your time in reading this if you got this far.

sksamuel:
Closes #3582

This adds the ability to do `expect { someboolean() }` in a test block, which, if false, marks the test as skipped.

Happy to rename "expect" to something else:
- require(s) clashes with the kotlin sdk method
- assume clashes with property test assumptions.


Kantis:
I like it. Tried to come up with some naming alternatives but it's hard.. @mickeelm you're pretty good at spitting out names, any thoughts on this one? 😄

Kantis:
Btw would this be compatible with having a second variant?:

```kotlin
inline fun <reified T> TestScope.expect(thunk: () -> T) = runCatching {
   thunk()
}.mapError {
   ExpectFailureException
}.getOrThrow()

val user = expect { createUser() } 
```

sksamuel:
Yeah that's better.

Kantis:
- `assumeSuccess {}`
- `assumeSuccessful {}`
- `expectSuccess {}`

Might make the intent a bit more obvious?

sksamuel:
precondition or condition or prereq are other alternatives.

mickeelm:
I'll try to live up to that introduction :)

I like `expect`, but I think that a potential problem with `expect` is the notion of `expect a to be b`, i.e. all in all the same as `shouldBe` (or `assert`) and most importantly, you're likely to expect the test to throw, not to be ignored. There might be a confusion risk here.

I think `precondition`/`condition`/`prereq` are closest to hinting that a `false` evaluation will result in an ignored test rather than a failed test.

Some other suggestions:

`runIf {}`
`runOnlyIf {}`
`onlyIf {}`
`conditional {}`
`conditionalOn {}`
`precheck {}`

Kantis:
I think I prefer `prereq`, but I dont feel very strongly about it.

sksamuel:
I like onlyIf or runIf

sksamuel:
I can't get two  variants to work, it always wants to resolve to the Boolean one. So we might need another name for the one that returns a value.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/ExpectFailureException.kt
```diff
@@ -0,0 +1,10 @@
+package io.kotest.core.test
+
+/**
+ * This exception is thrown when a test fails an expect block.
+ */
+object ExpectFailureException : Exception()
+
+fun TestScope.runIf(thunk: () -> Boolean) {
```
Kantis, line 8: I think we could make this `runIf(condition: Boolean)`, then we can have rhe one taking a function return `T`

PR #3599, opened 2023-07-29
Added see-difference formatting for shouldHaveSingleElement

### LINKED ISSUE #3515: Collection.shouldHaveSingleElement is missing  IntelliJ's Click to see difference
When calling `.shouldHaveSingleElement(..)` on a collection, resulting error message is just printed in one very long line, without "Click to see difference" IntelliJ helper.

As a workaround, `.shouldHaveSize(1).first().shouldBe(...)` can be used instead, but is more verbose.

**Which version of Kotest are you using:** 5.5.4

sksamuel:
Fixes #3515


Kantis (approved):

Kantis:
Took the liberty to update the API dump for you. Looks like it would be a binary breaking change due to moving the matchers to a new file

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/should.kt
```diff
@@ -47,9 +47,9 @@ fun <T> invokeMatcher(t: T, matcher: Matcher<T>): T {
       when (result) {
          is ComparableMatcherResult -> errorCollector.collectOrThrow(
             failure(
-               Expected(Printed(result.expected())),
-               Actual(Printed(result.actual())),
-               result.failureMessage()
+               expected = Expected(Printed(result.expected())),
+               actual = Actual(Printed(result.actual())),
+               prependMessage = result.failureMessage() + "\n"
```
Kantis, line 52: This newline breaks some tests. I'm pushing a commit that updates the tests, but feel free to drop it and resolve in another way.

sksamuel:
> Took the liberty to update the API dump for you. Looks like it would be a binary breaking change due to moving the matchers to a new file

The package names are the same, so the effect to the end user is zero, as the imports would stay the same. The signature of the methods has remained identical. I think it's a no op.

PR #3601, opened 2023-07-29
Use the default spec for the first test in InstancePerTest

### LINKED ISSUE #3490: Non-default IsolationMode skips some extensions
Kotest version: **5.6.1** (tested, but behavior is older).

First of all, thanks for maintaining this framework, it is a joy to use!

I hope I can describe this adequately, sorry for not being able to provide a more concise summary.

### Description

Running tests with either `InstancePerLeaf` oder `InstancePerTest` isolation modes causes the `body` block's execution once, without spec-based extensions being called every time. This not wrong, per se, but causes unexpected (or at least not trivially to work around) behavior.

### Reproduction

To reproduce the problem, the kotest source code can be used. The `TempFileTest` (`kotlin/com/sksamuel/kotest/engine/spec/TempFileTest.kt`) is green. Now add the following line to the top of the block constructor parameter:

```
   isolationMode = InstancePerTest
```

Now the test will fail at an after project error:

```
File /var/folders/1b/ftl3z7zd0zv45cyfc11y_qjm0000gp/T/com.sksamuel.kotest.engine.spec.TempFileTest1309413316694372959.tmp should not exist
```

### Identified Cause

As far as I can tell, the test now creates two temp directories, but only one gets deleted (the one that is created during the actual test leaf run). Since both are checked in the `afterProject` block, the still existing directory now causes an error. The `afterSpec` handler of the created directory is only called once (which makes sense, since there was no spec for the `afterSpec` block to run after).

Still, this means that changing the isolation mode probably leaves a lot of empty directories laying around, in the best case. Other resources would be equally affected if they need cleaning up. I suspect this issue is also the cause for resource-intensive brittle tests on my side, but have no proof at this time.

### Workaround

I could not find any. Neither `afterContainer` nor `afterAny` get called for that spec-less body invocation. I think `afterProject` could be abused in cooperation with `afterSpec` and an atomic flag (tracking the deletion), but that handler is not available in the `TestConfiguration` context the method is implemented in.

It's easy to switch single tests over to no isolation, but identifying test cases that might be affected is not entirely trivial.

sksamuel:
Fixes #3490


kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/spec/tempfile.kt
```diff
@@ -4,6 +4,7 @@ import io.kotest.core.TestConfiguration
 import java.io.File
 
 fun TestConfiguration.tempfile(prefix: String? = null, suffix: String? = null): File {
+   println("Creating file!!")
```
LeoColman, line 7: oopsie

LeoColman (changes requested):

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/spec/tempfile.kt
LeoColman:
```suggestion
  
```

LeoColman (changes requested):

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/spec/runners/InstancePerTestSpecRunner.kt
```diff
@@ -105,15 +106,30 @@ internal class InstancePerTestSpecRunner(
...
+   /**
+    * The first time we run a root test, we can use the already instantiated spec as the instance.
+    * This avoids creating specs that do nothing other than scheduling tests for other specs to run in.
+    * Eg, see https://github.com/kotest/kotest/issues/3490
+    */
+   private suspend fun executeInCleanSpecIfRequired(test: TestCase, defaultSpec: Spec): Result<Spec> {
+      return if (defaultInstanceUsed.compareAndSet(false, true)) {
+         Result.success(defaultSpec).flatMap { executeInGivenSpec(test, it) }
+      } else {
+         createInstance(test.spec::class).flatMap { executeInGivenSpec(test, it) }
```
Kantis:
```suggestion
         executeInCleanSpec(test)
```

Duplicating the code of `executeInCleanSpec` here, I think?

PR #3603, opened 2023-07-30
Added compiler plugin for generating clues

sksamuel:
This requires some clean up and is pretty basic at the moment.

But tl;dr

```kotlin
data class Person(val name: String)

val p = Person("sam")

class PropertyTest : FunSpec({

   test("foo") {
      p.name
         .shouldBe("wobble")
         .shouldHaveLength(3)
   }

})
```

Automatically adds the context of the caller:

```kotlin
property 'name'
expected:<"wobble"> but was:<"sam">
Expected :"wobble"
Actual   :"sam"
```


Kantis:
I think this is a great idea :)
How much does it add to compilation time?

In your given example, would it still identify `name` as the property if the first of the chained assertions pass and the second fails?

PR #3604, opened 2023-07-30
Fix assertions mode when using coroutines in another thread

### LINKED ISSUE #3022: `AssertionMode` incorrectly considers test to have no assertions if it launches coroutines that make assertions from other threads
v5.3.0

Example test:

```kotlin
class AssertionModeTest : FunSpec() {
   override fun assertionMode() = AssertionMode.Error

   init {
      test("make an assertion from another thread") {
         withContext(Dispatchers.Default.limitedParallelism(1)) {
            launch { 1 shouldBe 1 }
         }
      }
   }
}
```

This test fails with the following error when run on the JVM, even though an assertion is made in the test:

```
Test 'make an assertion from another thread' did not invoke any assertions
io.kotest.engine.test.interceptors.ZeroAssertionsError: Test 'assertion counter should count assertions in child coroutines' did not invoke any assertions
```

sksamuel:
Bit of coroutine trickery here @charleskorn 
It swaps in the thread locals used by the assertion counters as coroutines are launched, but with test name affinity.

Fixes #3022

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/counter.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/counter.kt
@@ -5,11 +5,11 @@ actual val assertionCounter: AssertionCounter = ThreadLocalAssertionCounter
 
 object ThreadLocalAssertionCounter : AssertionCounter {
 
-   private val context = object : ThreadLocal<Int>() {
+   val values = object : ThreadLocal<Int>() {
       override fun initialValue(): Int = 0
    }
 
-   override fun get(): Int = context.get()
-   override fun reset() = context.set(0)
-   override fun inc() = context.set(context.get() + 1)
+   override fun get(): Int = values.get()
+   override fun reset() = values.set(0)
+   override fun inc() = values.set(values.get() + 1)
 }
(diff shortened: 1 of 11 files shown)


LeoColman (approved):

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/counter.kt
```diff
@@ -5,11 +5,11 @@ actual val assertionCounter: AssertionCounter = ThreadLocalAssertionCounter
...
+   val values = object : ThreadLocal<Int>() {
       override fun initialValue(): Int = 0
    }
 
-   override fun get(): Int = context.get()
-   override fun reset() = context.set(0)
-   override fun inc() = context.set(context.get() + 1)
+   override fun get(): Int = values.get()
+   override fun reset() = values.set(0)
+   override fun inc() = values.set(values.get() + 1)
```
charleskorn, line 14: Is this safe if multiple coroutines are running for the same test simultaneously?

If all we need to track is "an assertion did / did not occur", perhaps an atomic boolean would be easier to reason about?

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/test/interceptors/AssertionModeThreadLocalContextInterceptor.kt
```diff
@@ -0,0 +1,60 @@
...
+      val counter = testNameCounters.getOrPut(testName) { 0 }
+      ThreadLocalAssertionCounter.values.set(counter)
+      // we track the state in our backing map so this can be ignored
+      return -1
+   }
+
+   // this is invoked after coroutine has suspended on current thread
+   override fun restoreThreadContext(context: CoroutineContext, oldState: Int) {
+      // need to put the current thread-local value into our backing map before the coroutine is switched out
+      testNameCounters[testName] = ThreadLocalAssertionCounter.values.get()
```
charleskorn, line 57: What happens if a test launches multiple coroutines and makes assertions in both? 

What happens if multiple coroutines launch at the same time but the last one to finish doesn't make any assertions? Won't we end up with a count of 0 assertions for the test?

sksamuel:
Closing in favour of #3623

PR #3611, opened 2023-08-04
Fix duration shrinker so it shrinks

dump247:
The duration shrinker was producing values greater than the value provided to the shrinker. This is not only a violation of the shrinking contract (it's not shrinking), it can result in infinite loops when a test fails and shrinks are generated.


sksamuel (approved):

sksamuel:
nice!

Issue #3613, opened 2023-08-10
Test-specific extension is overridden by project-level extention


OliverO2:
Kotest master, as of commit 45df9ae04e0d3077c3732d8c7ed6947eeae24d7f

Following up from https://github.com/kotest/kotest/pull/3602#issuecomment-1672994097:

I am currently working on enabling nested invocations for the `BlockHound` extension. The following will not work:
```kotlin
@DoNotParallelize
class BlockHoundSpecTest : FunSpec({
   extension(BlockHound())

   test("detects for spec") {
      shouldThrow<BlockingOperationError> { blockInNonBlockingContext() }
   }

   // other tests...

   test("nested configuration").config(extensions = listOf(BlockHound(BlockHoundMode.DISABLED))) {
      shouldNotThrow<BlockingOperationError> { blockInNonBlockingContext() }
   }
})
```

In the above case, the test-specific extension `BlockHound(BlockHoundMode.DISABLED)` will be overridden by the project-level extension `BlockHound()`. The cause seems to be an incorrect extension processing order. I am just preparing a PR to fix this.

sksamuel:
This is fixed by #3614

PR #3602, opened 2023-07-29
Fixed spec lifecycle ordering for JS

sksamuel:
Fixes #3340

PR #3614, opened 2023-08-10
Fix extension processing order, prioritize test case config (#3613)

OliverO2:
The test is actually Sam's work (#3602), which I shamelessly copied and adapted for this issue's purpose.


sksamuel (approved):

sksamuel:
👍🏻

PR #3621, opened 2023-08-17
Add KProperty0 assertions

### LINKED ISSUE #3547: Add KProperty shouldHaveValue assertion
When comparing several fields from big data classes:

```kotlin
assertSoftly(myDataClass) {
    firstProperty shouldBe true
    secondProperty shouldBe false
    thirdProperty shouldBe 1
}
```

error messages would not be very helpful, it would never say in the error message what property exactly failed:

```
1) expected:<true> but was:<false>
	at ...
2) expected:<false> but was:<true>
	at ...
3) expected:<1> but was:<2>
	at ...
```

However, since we are comparing properties directly, we could do something like this instead:

```kotlin
assertSoftly(myDataClass) {
    ::firstProperty shouldHaveValue true
    ::secondProperty shouldHaveValue false
    ::thirdProperty shouldHaveValue 1
}
```

This could generate much more helpful error message, with property names included:

```
1) firstProperty
expected:<true> but was:<false>
	at ...
2) secondProperty
expected:<false> but was:<true>
	at ...
3) thirdProperty
expected:<1> but was:<2>
	at ...
```

Additionally, we could add `withValue` to get access to other properties, such as null checks:

```kotlin
::firstProperty.withValue { shouldBeNull() }
```

I propose something like this (and I can create a PR for it, if desired):

```kotlin
infix fun <T> KProperty0<T>.shouldHaveValue(value: T) {
    withClue(name) {
        eq(this.get(), value)?.let(errorCollector::collectOrThrow)
    }
}

inline fun <T> KProperty0<T>.withValue(block: T.() -> Unit) {
    withClue(name) {
        block(get())
    }
}
```

matejdro:
This fixes #3547


sksamuel (approved):

sksamuel:
👍🏻

sksamuel:
You will need to run the api dump gradle task and commit that too.

Kantis:
Will this work as desired?

```kotlin

class FooSpec : FunSpec({
  test("some test")  {
    withClue("some clue") {
        point::x shoudHaveValue 5
    }
  }
})
```

matejdro:
Yes, clues stack, so both clues will be printed.

Kantis:
still, it feels wrong that this is not implemented as an actual matcher. @sksamuel are there no big downsides with that?

From the top of my head, it will mean that matcher composition is unavailable, and using `foo::x should haveValue(5)`-style also wont work.

sksamuel:
Yep, what you said is the downside @Kantis but we can add that after this PR is merged as it won't change the contract of these methods.

sksamuel:
See https://github.com/kotest/kotest/pull/3637

PR #3622, opened 2023-08-17
Bind BlockHoundMode setting to coroutines, lifting restrictions

OliverO2:
`BlockHound` testing is currently limited due to a single, global mode.

This PR enables a coroutine-bound mode, which means BlockHound testing can be used with different settings in concurrent coroutines, even across threads. `@DoNotParallelize` is also no longer necessary.


sksamuel:
This is great

sksamuel (approved):

sksamuel:
Needs the api dump running

PR #3623, opened 2023-08-17
Refactor concurrency

### LINKED ISSUE #3022: `AssertionMode` incorrectly considers test to have no assertions if it launches coroutines that make assertions from other threads

OliverO2:
This PR covers a number of issues regarding concurrency and multi-threading. It is based on [the recent BlockHound PR](https://github.com/kotest/kotest/pull/3622) and offers an alternative to https://github.com/kotest/kotest/pull/3604. Although this is a lot, I thought it would be better to put it in the open for everyone to consider and discuss.

I am currently examining the further evolution of concurrent testing around existing, but less documented parameters like `concurrentSpecs` and `concurrentTests`. This PR intends to provide a sound basis for that.

Details addressed in this PR:

* Avoid the use of `runBlocking` inside coroutines. Schedule all coroutines below the test engine layer via regular dispatchers, bringing the test environment more in line with actual production environments.

* Remove direct use of Java's `Executors`, delegating to kotlinx.coroutines-provided dispatchers instead.

* Preserve coroutine context (which includes lots of Kotest configuration) in multithreaded test executions.

* Remove hard-coded 1-day time limit for multithreaded test executions.

* Change tests relying on single-threading from a coroutine model to a thread model, removing coroutine invocations (`delay`), replacing those with thread invocations (`Thread.sleep`).

* Avoid leaking threads by closing dispatchers after use.

* Keep the `assertionCounter` synchronized with thread-switching coroutines.

* Remove `fun <K, V> concurrentHashMap(): MutableMap<K, V>`, which was not thread-safe.

* Make `FixedThreadCoroutineDispatcherFactory.dispatcherAffinity` thread-safe.

* Revert a change in `ConcurrentTestSuiteScheduler.schedule` by commit c316bbd33fbd68657a3e3cb8b51ed49ceb76254f, which replaced `launch` with `async` plus `joinAll`. This was functionally equivalent, but `async` was not needed, as there were no values returned, and `joinAll` was already performed by the enclosing `coroutineScope`.

Fixes #3022


OliverO2:
Test failure seems unrelated to this PR:
```
> Task :kotest-property:jvmTest
com.sksamuel.kotest.property.arbitrary.BindShrinkTest[jvm]

  Test Shrinks all components to minimum value[jvm] FAILED

  java.lang.AssertionError: "Property test failed for inputs
  
  0) MaximumComponents(a=1000, b=1000, c=1000, d=1000, e=1000, f=1000, g=1000, h=1000, i=1000, j=1000, k=1000, l=1000, m=1000, n=1000)
  
  Caused by java.lang.AssertionError: 1000 should be < 100 at
  	com.sksamuel.kotest.property.arbitrary.BindShrinkTest$1$2$stdout$1$1$1.invokeSuspend(BindShrinkTest.kt:52)
  	com.sksamuel.kotest.property.arbitrary.BindShrinkTest$1$2$stdout$1$1$1.invoke(BindShrinkTest.kt)
  	com.sksamuel.kotest.property.arbitrary.BindShrinkTest$1$2$stdout$1$1$1.invoke(BindShrinkTest.kt)
  	io.kotest.property.internal.ProptestKt$proptest$3$2.invokeSuspend(proptest.kt:49)
  
  " should include substring "Shrink result (after 45 shrinks) => MaximumComponents(a=0, b=0, c=0, d=0, e=0, f=0, g=0, h=0, i=0, j=0, k=0, l=0, m=100, n=0)"
```

sksamuel:
Prob some very intermittent prop test failure

OliverO2:
Seems like this one is causing the test failure:
```kotlin
inline fun captureStandardOut(fn: () -> Unit): String {
   val previous = System.out
   val buffer = ByteArrayOutputStream()
   System.setOut(PrintStream(buffer))
   try {
      fn()
      return String(buffer.toByteArray())
   } finally {
      System.setOut(previous)
   }
}
```

`PrintStream(buffer)` creates a stream without auto-flushing, so before returning the captured output, `flush()` should be invoked. I could correct this and other invocations in this PR.

sksamuel:
This is green and ready to merge but I'd like to take a few days to read through and understand the changes if that's ok.

OliverO2:
That sounds fantastic!

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
```diff
@@ -22,9 +22,7 @@ val errorCollectorContextElement: CoroutineContext.Element
    get() = ErrorCollectorContextElement(threadLocalErrorCollector.get())
 
 
-private val threadLocalErrorCollector = object : ThreadLocal<CoroutineLocalErrorCollector>() {
-   override fun initialValue() = CoroutineLocalErrorCollector()
-}
+private val threadLocalErrorCollector = ThreadLocal.withInitial { CoroutineLocalErrorCollector() }
```
sksamuel, line 25: Is this just a more convenient way of doing the former ?

kotest-common/src/jvmMain/kotlin/io/kotest/mpp/replay.kt
```diff
@@ -15,23 +15,15 @@ actual suspend fun replay(
...
-      val error = AtomicReference<Throwable>(null)
-      for (k in 0 until times) {
-         executor.submit {
-            runBlocking {
-               try {
-                  action(k)
-               } catch (t: Throwable) {
-                  error.compareAndSet(null, t)
+      @OptIn(DelicateCoroutinesApi::class)
+      newFixedThreadPoolContext(threads, "replay").use { dispatcher ->
```
sksamuel, line 19: What's the change here - just moving away from JVM executors to kotlin coroutine dispatchers?

kotest-framework/kotest-framework-concurrency/src/jvmTest/kotlin/io/kotest/framework/concurrency/EventuallySpec.kt
```diff
@@ -15,22 +15,23 @@ import io.kotest.matchers.longs.shouldBeGreaterThanOrEqual
...
 import java.util.concurrent.CountDownLatch
-import java.util.concurrent.Executors
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicInteger
 
 private fun Int.seconds(): Long = Duration.ofSeconds(this.toLong()).toMillis()
 private fun Int.milliseconds(): Long = this.toLong()
 
 @ExperimentalKotest
+@OptIn(DelicateCoroutinesApi::class)
```
sksamuel, line 34: Are these methods likely to change in future coroutine updates?
We've definitely been burnt many times by the kotlin sdk changing things on us.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/ConcurrentTestSuiteScheduler.kt
```diff
@@ -54,31 +53,32 @@ internal class ConcurrentTestSuiteScheduler(
...
-   ) = coroutineScope { // we don't want this function to return until all specs are completed
-      val coroutineDispatcherFactory = defaultCoroutineDispatcherFactory(context.configuration)
-      val semaphore = Semaphore(concurrency)
-      val collector = CollectingTestEngineListener()
-      specs.map { ref ->
-         logger.log { Pair(ref.kclass.bestName(), "Scheduling coroutine") }
-         async {
-            semaphore.withPermit {
-               logger.log { Pair(ref.kclass.bestName(), "Acquired permit") }
+   ) = defaultCoroutineDispatcherFactory(context.configuration).use { coroutineDispatcherFactory ->
```
sksamuel, line 56: Can you give me a tl;dr on the change here

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
```diff
@@ -10,8 +9,8 @@ import kotlin.reflect.KClass
 
 class CollectingTestEngineListener : AbstractTestEngineListener() {
 
-   val specs = concurrentHashMap<KClass<*>, TestResult>()
-   val tests = concurrentHashMap<TestCaseKey, TestResult>()
+   val specs = mutableMapOf<KClass<*>, TestResult>()
+   val tests = mutableMapOf<TestCaseKey, TestResult>()
```
sksamuel, line 15: Are we worried about race conditions if we just use mutable map ?

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/concurrency/FixedThreadCoroutineDispatcherFactory.kt
```diff
@@ -31,22 +32,27 @@ import kotlin.reflect.KClass
 class FixedThreadCoroutineDispatcherFactory(
    threads: Int,
    private val affinity: Boolean,
-) : CoroutineDispatcherFactory {
+) : CoroutineDispatcherFactory, Closeable {
```
sksamuel, line 35: Are the changes in this class mostly nomenclature ?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/provokeThreadSwitch.kt
```diff
@@ -0,0 +1,10 @@
+package com.sksamuel.kotest.engine.coroutines
+
+/**
+ * Provoke the dispatcher to switch coroutine threads by keeping the coroutine alive for a sufficient time period.
+ */
+@Suppress("RedundantSuspendModifier")
+suspend fun provokeThreadSwitch() {
```
sksamuel: How come this is suspend ?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/threads/BeforeTestConcurrentInstancePerLeafTest.kt
```diff
@@ -22,7 +21,7 @@ class SpecThreadBeforeTestConcurrentInstancePerLeafTest : FunSpec({
       if (isLockAcquired) {
          lock.lock()
          try {
-            delay(300)
+            Thread.sleep(300)
```
sksamuel, line 24: How come this was changed ?

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
OliverO2: Yes. Being more concise, I thought it would better communicate intent.

kotest-common/src/jvmMain/kotlin/io/kotest/mpp/replay.kt
OliverO2: On the surface, yes. Beneath, the idea is to move testing as close as possible to common production conditions. kotlinx.coroutine dispatchers still use Executors under the hood, but there are subtle differences, e.g.

* the various Executor flavors schedule differently at their discretion (creating threads more or less eagerly),
* directly using `Executor.shutdown()` can trigger a BlockHound exception due to blocking synchronization,
* `newFixedThreadPoolContext` explicitly creates daemon threads.

The change is meant to be a safe default which avoids having successful tests, but then seeing the same code fail in production due to such subtle differences.

kotest-framework/kotest-framework-concurrency/src/jvmTest/kotlin/io/kotest/framework/concurrency/EventuallySpec.kt
OliverO2: In 2018, these were marked `@ObsoleteCoroutinesApi`, thought to be replaced by dispatcher views (https://github.com/Kotlin/kotlinx.coroutines/issues/2919, https://github.com/Kotlin/kotlinx.coroutines/issues/2943), but with the latter becoming available in 2021, `newSingleThreadContext` and `newFixedThreadPoolContext` were "promoted" to `@DelicateCoroutinesApi` (though still "experimental"). Given that history, and the fact that they are used in current documentation, I'd say they'll be here to stay.

(They are marked as delicate because these dispatchers need to be closed, which is now consistently done in Kotest. Formerly, executors were not always shut down, which lead to leaking threads.)

My tendency is that each Kotest release should consider defaulting to the most common practice of its time. This might imply shifting towards dispatcher views by default in upcoming releases.

So I guess for now the above is one of the best options to run test scenarios as close to production as possible. An alternative would be to start using `Dispatchers.IO.limitedParallelism()` right away, but that's also marked as experimental. Other ideas on that?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/ConcurrentTestSuiteScheduler.kt
OliverO2: Sure. First is what I mentioned in the PR description: "Revert a change in ConcurrentTestSuiteScheduler.schedule by commit c316bbd33fbd68657a3e3cb8b51ed49ceb76254f, which replaced launch with async plus joinAll. This was functionally equivalent, but async was not needed, as there were no values returned, and joinAll was already performed by the enclosing coroutineScope."

Second, the factory is now closed at the end, releasing threads.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
OliverO2: Seems like we should be. The API with public properties is problematic in this regard, and Kotest doesn't use atomicfu so far, so I'll add a commit introducing a listener-wide Mutex for mutations.

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/concurrency/FixedThreadCoroutineDispatcherFactory.kt
OliverO2: Not really. The `close()` is essential to help avoid leaking threads. The rest is (hopefully) about communicating intent (needed this for myself to better understand things).

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/provokeThreadSwitch.kt
OliverO2: Sure, technically it does not. The idea is to communicate that this function is meant to be invoked only in a coroutine context. So more like a poor man's code coloring. Would you favor a change or added comment?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/threads/BeforeTestConcurrentInstancePerLeafTest.kt
OliverO2: If I understand this test correctly, it tries to fire up three threads and uses thread-based Java mechanisms exclusively. A `delay` would most probably just suspend and run the neighboring tests on the same thread (this test doesn't check for that). `Thread.sleep` seems to be the most reliable way of making an executor schedule a Runnable on a different thread.

kotest-framework/kotest-framework-concurrency/src/jvmTest/kotlin/io/kotest/framework/concurrency/EventuallySpec.kt
sksamuel: Definitely want to avoid experimental - I tried that in the past and it caused issues. I'd rather be conservative. This seems ok.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
sksamuel: Any reason we'd not just use the concurrent hash map ?

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/concurrency/FixedThreadCoroutineDispatcherFactory.kt
sksamuel: So in the whole class, it's close + nomenclature ?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/provokeThreadSwitch.kt
sksamuel: I guess if we run it outside of suspend it doesn't matter, maybe make it internal, and add a comment ?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/threads/BeforeTestConcurrentInstancePerLeafTest.kt
sksamuel: Because the thread is still allocated so the executor is unable to reuse it ?

sksamuel:
Couple more follow up questions added then we can merge in time for 5.7 release.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
OliverO2: Yes. The previous `concurrentHashMap` was never thread-safe, but actually returned a `MutableMap`. There is no `ConcurrentHashMap` implementation in stdlib common.

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/concurrency/FixedThreadCoroutineDispatcherFactory.kt
OliverO2: Yes, plus using a kotlinx.coroutines dispatcher instead of Executor.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
sksamuel: Yeah I see. It could use a concurrent hash map on the JVM actual impl though.

OliverO2: Indeed that could be an option, since we don't require special `ConcurrentHashMap` methods here. Side effect: Some methods (like `testIgnored()`) mutate multiple structures, which would no longer be atomic then. If results were looked up during ongoing test runs, the maps could be briefly inconsistent (one contains a test, the other does not). What would be the benefit of using expect/actual and `ConcurrentHashMap` instead?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/provokeThreadSwitch.kt
OliverO2: Like so?

```kotlin
/**
 * Provoke the dispatcher to switch coroutine threads by keeping the coroutine alive for a sufficient time period.
 *
 * This function is not suspending, but intended to be invoked in a coroutine exclusively.
 */
internal fun provokeThreadSwitch() {
   Thread.sleep(50)
}
```

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
sksamuel: This whole class was really just created in order to test kotest itself. It's not meant to be used outside of kotest.
I think ConcurrentHashMap is probably good enough for our purposes?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/provokeThreadSwitch.kt
sksamuel: I like it

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/threads/BeforeTestConcurrentInstancePerLeafTest.kt
OliverO2: Seems like that. I haven't looked into the Executor's internals. Might be they use a spin lock and re-use the same thread if just a short time period elapses. If it takes longer or a thread is actually parked, they'd use another thread. But that's all implementation-dependent and could change. The comment for `Executor.execute()` says
> Executes the given command at some time in the future. The command may execute in a new thread, in a pooled thread, or in the calling thread, at the discretion of the Executor implementation.

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/provokeThreadSwitch.kt
OliverO2: Done via 63c927a9a4e52bb193a865e8995707d865cb63ae

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/threads/BeforeTestConcurrentInstancePerLeafTest.kt
sksamuel: Could also write our own executor implementation that always creates new threads for the purposes of testing.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
OliverO2: I've tried: `concurrentHashMap()` with 1 "expect" and 3 "actuals" (like formerly but correct on the JVM). Then we're still missing a correct concurrent implementation on Native, and a concurrent implementation for `val names = mutableListOf<String>()`. Lots more code, less correct. Thoughts?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/threads/BeforeTestConcurrentInstancePerLeafTest.kt
OliverO2: Sure, I also have more thoughts on that. Probably too much for this iteration, but briefly: Testing as close to common practice (Kotlin default/promoted ways of using concurrency) could be the default in Kotest. Then we should accept that coroutines will always hop threads, even if parallel execution limits may be in effect. We should also cease to rely on thread guarantees, since we deal with coroutines, not threads.

Then, of course, we could offer other mechanisms for special purposes, e.g. a dispatcher with a strictly round-robin thread pool. But we should be clear about what our goals are when asking for special-purpose dispatchers.

sksamuel: I'm happy with the thread.sleep as its in improvement on delay.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
sksamuel: Maybe use concurrentHashMap on jvm, and mutableMap on the others, which is better than what we had before?

OliverO2: Better than originally, yes. Better than `CollectingTestEngineListener`'s Mutex implementation, which should just work on all platforms? What's wrong with that? What am I missing?

OliverO2: Or maybe commit 3281e9c0f826841aae771cae4c0596bf1fc42bea was just overlooked?

sksamuel: Yep missed the mutex.
I do wonder if the mutex on the JVM is mapped to something that ensures the JVM MM is adhered to. I assume it is,

sksamuel (approved):

sksamuel:
nice work :)

OliverO2:
Thanks for the excellent cooperation. Way quicker than I had envisioned, given the size.

sksamuel:
I'm eager to get 5.7 out this weekend :)

PR #3630, opened 2023-08-25
Rewrite System Extensions docs intro for more clarity

ajalt:
I recently tried to add the System Extensions to a project and found their docs a little confusing.

- It doesn't list the maven artifact, so I had to dig through maven central to find its name
- It doesn't explicitly mention that the extension is JVM-only. It makes sense why it is, but functions like `withEnvironment` could theoretically be multi-platform, so I think it's worth calling out.
- The scope of the extension were a little unclear from the intro, so I changed to wording to hopefully be more clear.

### DIFF
--- a/documentation/docs/extensions/system.md
+++ b/documentation/docs/extensions/system.md
@@ -10,10 +10,19 @@ slug: system_extensions.html
 
 ## System Extensions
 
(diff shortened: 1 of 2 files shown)


sksamuel:
This is great, thank you for clarifying.

Issue #3617, opened 2023-08-15
`Arb.offsetDateTime` (Ideally) Should Be Based On `Instant`


akefirad:
Looking at the implementation
```kotlin
fun Arb.Companion.offsetDateTime(
   minValue: LocalDateTime = LocalDateTime.MIN,
   maxValue: LocalDateTime = LocalDateTime.MAX,
   zoneOffset: Arb<ZoneOffset> = zoneOffset()
): Arb<OffsetDateTime> = Arb.bind(
   localDateTime(minValue, maxValue),
   zoneOffset
) { time, offset -> time.atOffset(offset) }
```
it seems more useful if the method accepts `min` and `max` as `Instant` and not `LocalDateTime`. The reason (at least in our case) is that when we're working with `OffsetDateTime` values, (often) we're interested in the moment in time (which cannot be represented by types like `LocalDateTime`) and the above implementation makes it hard to use the method. An example would be all "give me a stream of `OffsetDateTime` values in future". For that reason we had to write our own version of the above method:
```kotlin
fun offsetDateTime(
  minValue: Instant = Instant.MIN,
  maxValue: Instant = Instant.MAX
  zoneOffset: Arb<ZoneOffset> = zoneOffset()
): Arb<OffsetDateTime> = Arb.bind(
  instant(minValue, maxValue),
  zoneOffset,
) { t, o -> t.atOffset(o) }
```

sksamuel:
Fixed in 5.7

PR #3632, opened 2023-08-26
Arb.offsetDateTime based on Instant

pientaa:
Implements #3617 

Notice that I didn't put any default values to the new function's params, because this would be an API-breaking change:
image


sksamuel:
Need to add the results of running api dump from gradle.

sksamuel (approved):

PR #3634, opened 2023-08-26
Fix matchJsonResource should check file existence

### LINKED ISSUE #3633: The matchJsonResource function should throw an AssertionError instead of an NPE if the file does not exist.
kotest version : 5.6.2

## AS-IS 

When I call `shouldMatchJsonResource` with a non-existent resource will throw an NPE.

CleanShot 2023-08-26 at 22 55 46

## TO-BE

The `shouldMatchJsonResource` should throw AssertionsError instead of NPE.

kshired:
resolve #3633 

When I call shouldMatchJsonResource with a non-existent resource will throw an NPE.

Change it to proceed with a null check and throw an AssertionError for a safe assertion.


sksamuel (approved):

sksamuel:
thanks!

PR #3637, opened 2023-08-26
Reimplement KProperty0<T>.shouldHaveValue as matcher

sksamuel:
I have moved this to the right module, and implemented as a matcher, and added tests.


matejdro:
Does this new version have see-difference formatting? From what I can see, it just prints it as normal string?

PR #3638, opened 2023-08-26
Add includeNonFiniteEdgeCases flag for double and float arbitrary

### LINKED ISSUE #3561: Out Of Range Edge Cases for Double Arbitrary
5.5.5

The documentation of Arb.double says:
> Returns an Arb that produces Doubles in range. The edge cases are `Double.NEGATIVE_INFINITY`, , `-1.0`, `-Double.MIN_VALUE`, `-0.0`, `0.0`, `Double.MIN_VALUE`, `1.0`, ClosedFloatingPointRange.endInclusive, `Double.POSITIVE_INFINITY` and `Double.NaN` which are only included if they are in the provided range.

But looking at the implementation, this is not entirely accurate:
```kotlin
fun Arb.Companion.double(
   range: ClosedFloatingPointRange<Double> = -Double.MAX_VALUE..Double.MAX_VALUE
): Arb<Double> = arbitrary(
   (numericEdgeCases.filter { it in range } + nonFiniteEdgeCases  + listOf(range.start, range.endInclusive)).distinct(),
   DoubleShrinker
) {
   it.random.nextDouble(range.start, range.endInclusive)
}
```

Is that a documentation bug or is there away to check wither `nonFiniteEdgeCases` are in the rage or not? Maybe include them only if the start and/or end of the rage is min and/or max values?

kshired:
resolves #3561 

- add `includeNonFiniteEdgeCases` flag
- update function comments
- modify and add some tests


kshired:
I'm not sure if `includeNonFiniteEdgeCases` should default to false. If this value defaults to false, there is no way to generate edge cases for non-finite values in a reflective bind.

sksamuel:
The docs say they should be included, so lets default to true.

kshired:
@sksamuel 

I changed includeNonFiniteEdgeCases's default value to be true

sksamuel (approved):

PR #3642, opened 2023-08-26
Add allowOutOfOrderCallbacks

### LINKED ISSUE #3624: allowOutOfOrderCallbacks doesn't exist yet mentioned in exception
5.6.2

### Context
`ContainerScope` defines test functions *and* `afterTest` in the end of a body definition.

### Issue
Kotest throws
```
Cannot use afterTest after a test has been defined. To disable this behavior set the global configuration value allowOutOfOrderCallbacks to true
io.kotest.core.spec.InvalidDslException: Cannot use afterTest after a test has been defined. To disable this behavior set the global configuration value allowOutOfOrderCallbacks to true
```
But there is no such `allowOutOfOrderCallbacks` option to configure

sksamuel:
Fix #3624


Kantis (approved):
Nice! 👍

Issue #3480, opened 2023-04-04
Add support of Equality to all contain* matchers


neckbosov:
Currently only `contain` matcher supports passing custom `Equality` but other matchers (`containAll`, `containExactly`, `containExactlyInAnyOrder`, `containOnly`) are not, which is a bit confusing for me. I think it will be great to support it to make collections matching more flexible.

pientaa:
I can take this one

sksamuel:
Closing as 5.8.1 is being released with this change.

PR #3662, opened 2023-08-28
Add support of Equality to all contain* matchers

pientaa:
Implements #3480 

Add verifier parameter to all the following functions: containAll, containExactly, containExactlyInAnyOrder, containOnly


sksamuel:
I wonder if we should just use the Eq typeclass that we already have. Thoughts?

pientaa:
@sksamuel we already use `Equality` in `contain` and it differs a lot from `Eq`:
image
image

So I don't think that @neckbosov meant it in #3480

Maybe `Eq` and `Equality` should be merged in the future? But this will be a bigger change, not in scope of my changes.

sksamuel:
Yeah I'm worried if we add Equality everywhere it will be hard to replace it with Eq later.

Let me think on this one.

pientaa:
@sksamuel personally, I'd create another issue for merging `Eq` and `Equality`. I can pick this up if you don't mind

sksamuel:
Ok we can merge this in as it is, but first can you make overloads rather than adding default values.
This will keep the API dump the same for existing methods.

pientaa:
@sksamuel done

sksamuel (approved):

PR #3665, opened 2023-08-30
Add nullable type support to test functions

kshired:
Currently, `forAll`, `forNone` and `checkAll` do not generate `null` value when using implicit arbitrary, even with nullable types.

```kotlin
forAll<String?> { 
   it != null  // this test passed, even `String?` type
}


forAll(Arb.string().orNull()) {
  it != null // this test will not passed
}
```

So, for better usability adding nullable type supports to common platform types.


kshired:
This is just my suggestion for usability.

kshired:
@sksamuel Can I get some feedback of this PR?

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/NullableTest.kt
```diff
@@ -0,0 +1,57 @@
...
+         classify(num == null, "null", "non-null")
+         false
+      }.classifications()
+      classifications["null"]?.shouldBeBetween(300, 600)
+      classifications["non-null"]?.shouldBeBetween(300, 600)
+   }
+
+   test("checkAll with implicit nullable arbitraries") {
+      val iterations = 1000
+      val classifications = checkAll<Int?>(iterations) { num ->
```
sksamuel: If you specify a seed then you can set the values precisely.

kshired: @sksamuel I added seed !

sksamuel (approved):

PR #3700, opened 2023-09-17
Remove hardcoded emil path

sksamuel:


Kantis (approved):
It used to be sam! 🤪

sksamuel:
It was better with Emil, so much that we didn't notice since Feb 😂

PR #3707, opened 2023-09-22
Add see-difference formatting to shouldHaveValue

matejdro:
Change in #3637 caused a regression in shouldHaveValue, removing see-difference formatting. This PR adds it back.

Up until now, it appears that matchers could not generate see-difference formatting, because matchers would always throw `AssertionError` instead of `AssertionFailedError` needed for IntelliJ to show the click difference button. PR adds this capability.


sksamuel:
Nice !

sksamuel (approved):

PR #3710, opened 2023-09-27
Add range to reflective Set of Enum values

tbvh:
Creating a reflective Arb for a `Set<EnumType>` should not fail with 
`java.lang.IllegalStateException: the target size requirement of 60 could not be satisfied after 600 consecutive samples
`.

This change restricts the expected size of the Arb.set to match the number of enum values.


sksamuel (approved):

sksamuel:
Nice fix thanks.

PR #3717, opened 2023-10-01
Add backgroundScope in KMP #3708

### LINKED ISSUE #3708: `backgroundScope` is not accessible in KMP
`backgroundScope` is JVM only right now, should be available in KMP when using coroutine test scope.

sksamuel:
Fixes #3708


TadeasKriz:
I think it didn't actually fix it, because the interceptor isn't adding the `TestContainerScope` into the context. At least that's what we're seeing in our iOS tests. I compared the jvm vs iOS sources and the jvm one has `withContext(TestScopeContainer(this)) {`, but the iOS one doesn't.

Edit: I may have spoken too soon. I was looking at the sources in IDEA, but I need to find them in the repo first.

Edit 2: Okay, so it actually got fixed accidentally in https://github.com/kotest/kotest/commit/73017540b44aba4fdc3f5f48b7441175465221e0 on Nov 12. That's why I didn't see it in `5.8.0`.

PR #3719, opened 2023-10-02
Added support for sequence arbitraries

kshired:
## Added support for sequence arbitraries.
- add `Arb.sequence` function
- add `Arb.sequence` test

## FYI
At first, I tried to write a code without dependency with `Arb.list`. However, during "shirinking", "This sequence can be consumed only once" error occurs. 

So I used `Arb.list` function when I implement `Arb.sequence`.


kshired:
@sksamuel  Can you review this PR?

sksamuel (approved):

sksamuel:
👍🏻

PR #3720, opened 2023-10-03
Non-negative/positive `Long` assertions and property testers

RusticFlare:
These all exist for `Int` (which these `Long` versions are based on)

Are there any docs that need updating for this?


sksamuel:
Awesome thank you. The docs are in .md format and there's a list of matchers we could add this to.

kshired:
@sksamuel I think this PR merged without apiDump..

sksamuel:
Ah yes you're right.

PR #3728, opened 2023-10-07
Added support for array arbitraries

### LINKED ISSUE #3722: Support arrays in checkAll / forAll
https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1696350321521309

kshired:
resolves #3722 

- Added `Arb.Comapnion.array` function
- Add support for array reflective binding
- Add tests


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/array.kt
```diff
@@ -12,3 +12,15 @@ internal inline fun <P, A> Arb.Companion.toPrimitiveArray(
    is Arb -> generateArrayLength
    is Exhaustive -> generateArrayLength.toArb()
 }.flatMap { length -> Arb.list(generateContents, length..length) }.map { it.toArray() }
+
+inline fun <reified A> Arb.Companion.array(
+   gen: Gen<A>,
+   range: IntRange = 0..100,
+   crossinline toArray: Collection<A>.() -> Array<A> = { this.toTypedArray() }
+): Arb<Array<A>> {
+   check(!range.isEmpty())
```
LeoColman, line 21: Why should it be >= 0? Can't you have ranges from -5 to 5, for example?

kshired: @LeoColman 

`range` is the length of the array, so isn't it weird that -5 is allowed?

Also, `Arb.list` has same check logic.

https://github.com/kotest/kotest/blob/c32ca997517a6f7f02302913eb6a5cd00f044c69/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/collections.kt#L93

kshired: @LeoColman What do you think of my answer?

LeoColman: Sorry for not replying. Yeah, it's fine

LeoColman (approved):

Issue #3726, opened 2023-10-06
BeIn matcher should accept ClosedRange, OpenEndRange and Iterables 


LeoColman:
The syntax
```
0 shouldNotBeIn (1..10)
```

Should be valid, but it isn't. Be in checks only inside collections:

```
fun <T> beIn(collection: Collection<T>) = object : Matcher<T> {
   override fun test(value: T): MatcherResult {
      if (collection.isEmpty()) throwEmptyCollectionError()

      val match = value in collection

      return MatcherResult(
         match,
         { "Collection should contain ${value.print().value}, but doesn't. Possible values: ${collection.print().value}" },
         {
            "Collection should not contain ${value.print().value}, but does. Forbidden values: ${collection.print().value}"
         })
   }
}
```

But could also apply to plenty of other `Iterable` forms.

An example on a real scenario for Range:
```
LocalDate.of(1998, 2, 9) shouldBeIn (LocalDate.of(1997, 2, 9)..LocalDate.of(1999, 2, 9))
```

LeoColman:
Related #3723

AlexCue987:
can someone assign this issue to me?

LeoColman:
Absolutely!
Thanks for the contribution :)

AlexCue987:
first PR https://github.com/kotest/kotest/pull/3735

PR #3735, opened 2023-10-13
support-ranges

AlexCue987:
* partially resolve the following issue: https://github.com/kotest/kotest/issues/3726
* only ranges in this PR, Iterables will be dealt with in the next one
* did not update versioned_docs, as this is a new feature not in those versions.

### DIFF
--- a/documentation/docs/assertions/ranges.md
+++ b/documentation/docs/assertions/ranges.md
@@ -1,0 +1,14 @@
+---
+id: ranges
+title: Range Matchers
+slug: range-matchers.html
+sidebar_label: Ranges
+---
+
+This page describes the rich assertions (matchers) that are available for ClosedRange and OpenEndRange types.
+
+
+| Ranges                       |                                                                                           |
+|------------------------------|-------------------------------------------------------------------------------------------|
+| `value.shouldBeIn(range)`    | Asserts that an object is contained in range, checking by value and not by reference.     |
+| `value.shouldNotBeIn(range)` | Asserts that an object is not contained in range, checking by value and not by reference. |
--- a/documentation/docs/assertions/ranges.md
+++ b/documentation/docs/assertions/ranges.md
@@ -5,7 +5,7 @@ slug: range-matchers.html
 sidebar_label: Ranges
 ---
 
-This page describes the rich assertions (matchers) that are available for ClosedRange and OpenEndRange types.
+This page describes the rich assertions (matchers) that are available for [ClosedRange](https://kotlinlang.org/docs/ranges.html) and [OpenEndRange](https://kotlinlang.org/docs/ranges.html) types.
 
 
 | Ranges                       |                                                                                           |
--- a/gradle.properties
+++ b/gradle.properties
@@ -36,5 +36,3 @@ kotlin.incremental=true
 kotlin.incremental.js=true
 
 org.gradle.caching=true
-
-
--- a/gradle.properties
+++ b/gradle.properties
@@ -37,4 +37,4 @@ kotlin.incremental.js=true
 
 org.gradle.caching=true
 
-jvmOnly=true
+
--- a/gradle.properties
+++ b/gradle.properties
@@ -36,3 +36,5 @@ kotlin.incremental=true
 kotlin.incremental.js=true
 
 org.gradle.caching=true
+
+jvmOnly=true
--- a/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/ranges/bein.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/ranges/bein.kt
@@ -1,0 +1,65 @@
+package io.kotest.matchers.ranges
+
+import io.kotest.assertions.print.print
+import io.kotest.matchers.Matcher
+import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.should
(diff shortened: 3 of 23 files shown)


documentation/docs/assertions/ranges.md
```diff
@@ -0,0 +1,14 @@
+---
+id: ranges
+title: Range Matchers
+slug: range-matchers.html
+sidebar_label: Ranges
+---
+
+This page describes the rich assertions (matchers) that are available for ClosedRange and OpenEndRange types.
```
LeoColman:
```suggestion
This page describes the rich assertions (matchers) that are available for [ClosedRange](https://kotlinlang.org/docs/ranges.html) and [OpenEndRange](https://kotlinlang.org/docs/ranges.html) types.
```

gradle.properties
```diff
@@ -36,3 +36,5 @@ kotlin.incremental=true
 kotlin.incremental.js=true
 
 org.gradle.caching=true
+
+jvmOnly=true
```
LeoColman: I don't think we need this change commited. Could you remove it?

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/ranges/bein.kt
```diff
@@ -0,0 +1,65 @@
...
+ * Assertion to check that this element is in [range]. This assertion checks by value, and not by reference,
+ * therefore even if the exact instance is not in [range] but another instance with same value is present, the
+ * test will pass.
+ *
+ * An empty range will always fail. If you need to check for empty range, use [Iterable.shouldBeEmpty]
+ *
+ */
+fun <T: Comparable<T>> beIn(range: ClosedRange<T>) = object : Matcher<T> {
+   override fun test(value: T): MatcherResult {
+      if (range.isEmpty())    throw AssertionError("Asserting content on empty range. Use Iterable.shouldBeEmpty() instead.")
```
LeoColman:
```suggestion
      if (range.isEmpty()) throw AssertionError("Asserting content on empty range. Use Iterable.shouldBeEmpty() instead.")
```

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/ranges/bein.kt
```diff
@@ -0,0 +1,65 @@
...
+
+      val match = value in range
+
+      return MatcherResult(
+         match,
+         { "Range should contain ${value.print().value}, but doesn't. Possible values: ${range.print().value}" },
+         {
+            "Range should not contain ${value.print().value}, but does. Forbidden values: ${range.print().value}"
+         })
```
LeoColman:
```suggestion
         { "Range should not contain ${value.print().value}, but does. Forbidden values: ${range.print().value}" }
     )
```

LeoColman (changes requested):
Amazing PR!

Suggested a few changes, being **remove gradle.properties from commit the single blocker**

LeoColman:
Also, I think ranges are supported in Kotlin Common. Can we move this assertion to that package?

AlexCue987:
> Also, I think ranges are supported in Kotlin Common. Can we move this assertion to that package?

Can you please elaborate? Ranges themselves are in `package kotlin.ranges`. I've searched on `ClosedRange` in this repo and all I found was in `package io.kotest.matchers.bigdecimal`  - `fun beInClosedRange(range: ClosedRange<BigDecimal>)`. What am I doing wrong?

gradle.properties
AlexCue987: fixed. Sorry about that.

LeoColman:
BigDecimals are only available in Java, so it makes sense to put it only in JVM

Ranges, however, are part of the common library: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-closed-range/ 


For this reason, we don't need to limit to the 'jvm' module, we can move it up to 'commonMain', and users will be able to use these functions in `kotlin-js`, `kotlin-native`, etc

AlexCue987:
> BigDecimals are only available in Java, so it makes sense to put it only in JVM
> 
> Ranges, however, are part of the common library: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-closed-range/
> 
> For this reason, we don't need to limit to the 'jvm' module, we can move it up to 'commonMain', and users will be able to use these functions in `kotlin-js`, `kotlin-native`, etc

Got it, TY, will do!

AlexCue987:
> BigDecimals are only available in Java, so it makes sense to put it only in JVM
> 
> Ranges, however, are part of the common library: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-closed-range/
> 
> For this reason, we don't need to limit to the 'jvm' module, we can move it up to 'commonMain', and users will be able to use these functions in `kotlin-js`, `kotlin-native`, etc

Done

LeoColman (approved):
Very nice PR!

AlexCue987:
> Very nice PR!

Thanks a lot!

AlexCue987:
@LeoColman can you help me understand failing api_check step? TIA!

kshired:
@AlexCue987 

You have to dump api like this `./gradlew apiDump`

AlexCue987:
> @AlexCue987
> 
> You have to dump api like this `./gradlew apiDump`

Done - TY!

PR #3739, opened 2023-10-17
Fix TestFactory does not close AutoCloseable objects

### LINKED ISSUE #3738: TestFactory doesn't close registered `AutoCloseable` object
Kotest version: 5.7.2.
(Thank you for maintaining Kotest! I love many assertions and inspectors supported by Kotest. They are very useful!)

The TestFactory described in https://kotest.io/docs/framework/test-factories.html doesn't call the `close()` method, even when an `AutoCloseable` object is wrapped by the `autoClose` function defined in Kotest. Below is a minimal test case to reproduce this.

Is there a way to automatically call close when using TestFactory? If so, I would like to know!

``` kotlin
data class A(val text: String) : AutoCloseable {
    override fun close() {
        logger.info { "close $text" }
    }
    companion object : KLogging()
}

fun aTest() = shouldSpec {
    should("close is not called. why...?") {
        val a = autoClose(A("close is not called...")) // This |a|'s `close` is not called
        a.shouldNotBeNull()
    }
}

class Test : ShouldSpec({
    include(aTest())
    should("close is called") {
        val a = autoClose(A("close is called!!!"))
        a.shouldNotBeNull()
    }
})
```

kshired:
resolves #3738 

- Fix TestFactory does not close AutoCloseable objects
  - Add configuration field to `TestFactory` class
  - Add _parentConfiguration field to `TestConfiguration` for registering AutoCloseables to parent's configuration.
- Add test

### DIFF
--- a/kotest-framework/kotest-framework-api/api/kotest-framework-api.api
+++ b/kotest-framework/kotest-framework-api/api/kotest-framework-api.api
@@ -87,6 +87,7 @@ public abstract class io/kotest/core/TestConfiguration {
 	public final fun setAssertSoftly (Ljava/lang/Boolean;)V
 	public final fun setAssertions (Lio/kotest/core/test/AssertionMode;)V
 	public final fun setDefaultTestConfig (Lio/kotest/core/test/config/TestCaseConfig;)V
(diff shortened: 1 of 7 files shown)


sksamuel:
awesome

PR #3743, opened 2023-10-21
Improve Sequence.shouldContainExactly(...)

### LINKED ISSUE #3742: sequence.shouldContainExactly(...) should give a meaningful failure message
**Kotest 5.7.2**

This assertion
```kotlin
sequenceOf(1, 2, 3).shouldContainExactly(1, 2, 4)
```

gives this message:
```
Sequence should contain exactly kotlin.collections.ArraysKt___ArraysKt$asSequence$$inlined$Sequence$1@56bec4 but was kotlin.collections.ArraysKt___ArraysKt$asSequence$$inlined$Sequence$1@340c8b3c. (expected 4 at 2 but found 3)
```

This happens because the code in
https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt#L90
simply uses (implicitly) `toString()` on the involved sequences.

I can think of different ways to fix this (note, that sequence may not be traversed multiple times):
1. create a copy of the sequences as a first step of this matching logic, use this copy for testing and for reporting, **and** use the [provided](https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt#L9) (but currently not used) `toString(limit: Init)` function:
```Kotlin
   val actualCopy = value.toList().asSequence()
   val expectedCopy = expected.toList().asSequence()
   val actualIterator = actualCopy.withIndex().iterator()
   val expectedIterator = expectedCopy.withIndex().iterator()

   var passed = true
   var failMessage = "Sequence should contain exactly ${expectedCopy.toString(10)} but was ${actualCopy.toString(10)}."
   // ... etc
```
(by the way, the creation of the `failMessage` should better be delayed until it is actually needed)

2. or don't print the sequences at all, just return the details (different item _xyz_ at index _i_ / missing item _xyz_ after index _i_ / extra item _xyz_ after index _i_) - in that case copying the sequences is not necessary

IMHO option 2 would be enough (but **please** also change the part of my example "expected 4 at 2 but found 3" into "expected 4 at _index_ 2 but found 3" :wink:)

I can provide a fix for either of these options.

obecker:
Previously, in the case of a failure, simply `toString()` on the involved sequences has been invoked to build the failure message. However, this just printed the (internal) sequence class name.

Now, the consumed and compared elements are collected in separate lists and then included in the failure message. So the failure contains all elements up the point, when a difference has been detected. This approach even works for infinite sequences.

Fixes https://github.com/kotest/kotest/issues/3742

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
@@ -6,8 +6,6 @@ import io.kotest.assertions.eq.eq
 import io.kotest.assertions.print.print
 import io.kotest.matchers.*
 
-private fun <T> Sequence<T>.toString(limit: Int = 10) = this.joinToString(", ", limit = limit)
-
 /*
 How should infinite sequences be detected, and how should they be dealt with?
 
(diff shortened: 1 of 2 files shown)


OliverO2 (approved):
This is excellent! Thank you!

PR #3745, opened 2023-10-21
Improve Sequence.shouldHaveSingleElement(...)

### LINKED ISSUE #3744: sequence.shouldHaveSingleElement(...) should give a meaningful failure message
**Kotest 5.7.2**

This assertion
```Kotlin
sequenceOf(1).shouldHaveSingleElement(2)
```

gives this message
```
Sequence should be a single element of 2 but has 1 elements
```
which is somewhat confusing to me.

The failure message should clearly state, what the failure is (in my case it's the wrong element, not the number of elements).

obecker:
When the asserted sequence contains a different single element than expected, then the failure message was misleading. Now the three cases: empty sequence, different element, and sequence with more than one element have specific failure messages. Also, this assertion now works on infinite sequences and is much faster on long sequences since it no longer converts the sequence into a list.

Fixes https://github.com/kotest/kotest/issues/3744

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
@@ -276,14 +276,22 @@ infix fun <T> Sequence<T>.shouldNotBeSortedWith(cmp: (T, T) -> Int) = this shoul
 infix fun <T> Sequence<T>.shouldHaveSingleElement(t: T) = this should singleElement(t)
 infix fun <T> Sequence<T>.shouldNotHaveSingleElement(t: T) = this shouldNot singleElement(t)
 
-fun <T> singleElement(t: T) = object : Matcher<Sequence<T>> {
+fun <T> singleElement(expectedElement: T) = object : Matcher<Sequence<T>> {
    override fun test(value: Sequence<T>): MatcherResult {
-      val valueAsList = value.toList()
-      val actualCount = valueAsList.count()
+      var failureMessage: String? = null
+      val iterator = value.iterator()
+      var actualElement: T?
+      if (!iterator.hasNext()) {
+         failureMessage = "Sequence should have a single element of $expectedElement but is empty."
+      } else if (eq(iterator.next().also { actualElement = it }, expectedElement) != null) {
+         failureMessage = "Sequence should have a single element of $expectedElement but has $actualElement as first element."
+      } else if (iterator.hasNext()) {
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
```diff
@@ -278,12 +278,20 @@ infix fun <T> Sequence<T>.shouldNotHaveSingleElement(t: T) = this shouldNot sing
 
 fun <T> singleElement(t: T) = object : Matcher<Sequence<T>> {
    override fun test(value: Sequence<T>): MatcherResult {
-      val valueAsList = value.toList()
-      val actualCount = valueAsList.count()
+      var failureMessage: String? = null
+      val iterator = value.iterator()
+      var currentValue: T?
```
OliverO2: Now that we have expected and actual elements and sequences, can we improve naming precision? Like `t` -> `expectedElement`, `currentValue` -> `actualElement`. (Unfortunately, renaming `value` would conflict with the base function signature.)

OliverO2 (approved):
One small adjustment, otherwise LGTM.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
obecker: Yeah, of course. Good point!

OliverO2:
Nicely done! Thanks again!

Side note: Reviewing incremental changes is easier with separate commits pushed to the same branch (without force-pushing). (This repo is configured to squash commits on merge, so even lots of single commits in a PR don't hurt.)

obecker:
@OliverO2 Yes, I understand that. I thought it's a small change that doesn't deserve its own commit. Moreover, github is showing a [compare link](https://github.com/kotest/kotest/compare/cf443cf36e5386415c89fb7fd79afd6a8307dca2..fe5f469c349ee41777547e3ae4630ec3ccc04ac4) next to my force-pushed commit, so there you can see the changes I made.

The other reason I've done this, is that I wanted a clear commit message right from the beginning. I have seen some squashed commits in this repository (like [this one](https://github.com/kotest/kotest/commit/eab0f88057ec7f15a98afbce06f93a0da7234932)), where the resulting commit message hasn't been cleaned up. And leaving all these intermediate messages in the final commit is IMHO rather confusing than helpful. Just my 2 cents.

OliverO2:
That's a good point! As I'm usually viewing PR diffs in IntelliJ IDEA, I overlooked the compare link on GitHub. And yes, it makes sense to clean up squashed commit messages, which I always hope to remember.

Issue #3754, opened 2023-11-02
Behavior of `collection.shouldHaveSingleElement { block }` doesn't match the documentation


m-burst:
I am using Kotest 5.7.2.

The documentation for `collection.shouldHaveSingleElement { block }` states:

> Asserts that the collection only contains a single element and that the element matches the given predicate.

However, the implementation actually checks that the collection has single element matching the predicate (but may have other elements):

```kt
fun <T> singleElement(p: (T) -> Boolean): Matcher<Collection<T>> = object : Matcher<Collection<T>> {
   override fun test(value: Collection<T>): MatcherResult {
      val filteredValue: List<T> = value.filter(p)
      return MatcherResult(
         filteredValue.size == 1,
         { "Collection should have a single element by a given predicate but has ${filteredValue.size} elements: ${value.print().value}" },
         { "Collection should not have a single element by a given predicate" }
      )
   }
}
```

kshired:
The first time this feature added, documentation was `Asserts that the collection contains a single element by a given predicate.`.

But wrongly changed at #2830

mirageoasis:
can I get this issue assigned?

LeoColman:
Go for it @mirageoasis !

LeoColman:
@Kantis Ping as author of #2830

PR #3757, opened 2023-11-08
fixed : rolled back to initial Readme.md

mirageoasis:
rolled back documentation of `collection.shouldHaveSingleElement { block }` Readme.md to match real implementation which was wrongly fixed in #2830 

Before
`Asserts that the collection only contains a single element and that the element matches the given predicate.`

After
`Asserts that the collection contains a single element by a given predicate.`


mirageoasis:
fixed all document as it was stated in pr template!

mirageoasis:
this is pr for issue #3754

sksamuel:
Thanks!

PR #3763, opened 2023-11-12
Fix race condition in beforeSpec

### LINKED ISSUE #3684: beforeSpec sometimes gets called multiple times on the same Spec instance
After trying to upgrade to either 5.7.1 or 5.7.2 we're seeing `beforeSpec` getting called twice on a single spec instance, but only sometimes. We weren't seeing this behavior with 5.6.2 or earlier.

How many times `beforeSpec` gets called in 5.7.x seems to be nondeterministic: on some machines it usually only gets called once, on others it'll more often be called twice. Other things can affect this too, like if all specs are being run, or only a subset, or even if the repo is clean vs has previously run the tests.

Given the nondeterminism, it feels like a race condition: like something is not realizing that another coroutine already ran (or started to run) `beforeSpec`, and so it runs it again.

This is causing a number of our tests to fail, as we often use `beforeSpec` to set up resources for tests, and our assumption has always been that it is only run once per spec instance.

sksamuel:
Fixes #3684


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/BeforeSpecListenerInterceptor.kt
```diff
@@ -10,8 +10,12 @@ import io.kotest.engine.spec.SpecExtensions
 import io.kotest.engine.spec.interceptor.ref.BeforeSpecState
 import io.kotest.engine.spec.interceptor.ref.beforeSpecStateKey
 import io.kotest.engine.test.createTestResult
+import kotlinx.coroutines.sync.Semaphore
+import kotlinx.coroutines.sync.withPermit
 import kotlin.time.Duration.Companion.seconds
 
+private val mutex = Semaphore(1)
```
OliverO2: Wouldn't a plain `Mutex()` better express intent?

OliverO2 (approved):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/BeforeSpecListenerInterceptor.kt
sksamuel: I didn't know that existed, so thanks!

sksamuel:
Need to figure out why this is hanging. Might just be because I've created a bottleneck on all beforeSpecs.

PR #3764, opened 2023-11-12
Allow display name formatter to return null

### LINKED ISSUE #3679: Allow a custom `DisplayNameFormatter` to fallback to project-configured defaults.
Currently, if a user wants to provide a custom display name for specs, they have two options:
- Use the `@DisplayName` annotation, but it is a manual process for every spec.
- Register their own `DisplayNameFormatterExtension`, but have to implement the formatter from the ground up.

As an example, let's say I want to change the display name of some classes only, and use Kotest's defaults for other
classes and test cases.
The `condition` might be either based on package, supertype, other annotations, etc.
The current interface for the `DisplayNameFormatter` requires me to do something like this:

```kotlin
class MyCustomFormatter(
    private val fallback: DisplayNameFormatter = DefaultDisplayNameFormatter()
): DisplayNameFormatter {

    override fun format(testCase: TestCase): String =
        fallback.format(testCase)

    override fun format(kclass: KClass<*>): String =
        if(condition) TODO("Custom format") else fallback.format(kclass)
}
```

**The problem:**

The only available implementation for the `DisplayNameFormatter` is the `DefaultDisplayNameFormatter`.
Since the formatter is also required at the test discovery stage, before any extensions (even the project ones) can run, there is no point at which I can get a reference to the final configuration, meaning I must use the no-args constructor, which uses the default values and _ignores any configuration options_ present in an `AbstractProjectConfig` that a sourceset may have defined.

I can think of two possible workarounds:

**Possibility A: Let the formatter return `null`s**

By setting the `format` functions to return `String?` we can model the behaviour to use the custom formatter unless it
returns `null`, in which case fallback to the `DefaultDisplayNameFormatter` initialized by the engine, and therefore
using the correct settings.

```kotlin
class MyCustomFormatter: DisplayNameFormatter {

    override fun format(testCase: TestCase): String? = null

    override fun format(kclass: KClass<*>): String? =
        if(condition) TODO("Custom format") else null
}
```

**Possibility B: Supply the configured default formatter in the `DisplayNameFormatterExtension`**

Consider the following:

```kotlin
interface DisplayNameFormatterExtension : Extension {
   fun formatter(fallback: DisplayNameFormatter): DisplayNameFormatter
}
```

Then, a user might be able to register a custom extension using similar logic as the first snippet:

```kotlin
class MyCustomFormatterExtension : DisplayNameFormatterExtension {
    override fun formatter(fallback: DisplayNameFormatter) = MyCustomFormatter(fallback)
}
```

I hope my description makes sense, let me know if clarification is needed.

sksamuel:
Fixes #3679

### DIFF
--- a/kotest-extensions/kotest-extensions-junit5/src/jvmMain/kotlin/io/kotest/extensions/junit5/KotestExtensionContext.kt
+++ b/kotest-extensions/kotest-extensions-junit5/src/jvmMain/kotlin/io/kotest/extensions/junit5/KotestExtensionContext.kt
@@ -4,7 +4,7 @@ import io.kotest.core.config.ProjectConfiguration
 import io.kotest.core.descriptors.toDescriptor
 import io.kotest.core.spec.Spec
 import io.kotest.core.test.TestCase
-import io.kotest.engine.test.names.getDisplayNameFormatter
+import io.kotest.engine.test.names.getFallbackDisplayNameFormatter
 import org.junit.jupiter.api.TestInstance
 import org.junit.jupiter.api.extension.ExtensionContext
 import org.junit.jupiter.api.extension.TestInstances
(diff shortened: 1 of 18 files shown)


sksamuel:
@OliverO2 I believe changing the interface from String to String? is fine from an implementer of the interface point of view. It doesn't throw up compile errors for me.

OliverO2:
Yes, that's also my understanding: Narrowing in derivatives is always fine.

PR #3766, opened 2023-11-12
CoroutineDispatcher is not a TestDispatcher on JS

### LINKED ISSUE #3575: IllegalStateException: CoroutineDispatcher is not a TestDispatcher [WindowDispatcher@1] in JS
**Kotest 5.6.2**

Whenever I use the property `testCoroutineScheduler` with the `UnconfinedTestDispatcher` the JS test fails with this exception,
while Native and JVM succeed without a problem.

```
IllegalStateException: CoroutineDispatcher is not a TestDispatcher [WindowDispatcher@1]
	at <global>.get_testCoroutineScheduler(/tmp/_karma_webpack_12398/commons.js:48626)
	at <global>.get_testCoroutineScheduler(/tmp/_karma_webpack_12398/commons.js:48617)
	at protoOf.doResume_5yljmg(/tmp/_karma_webpack_12398/commons.js:136553)
	at protoOf.resumeWith_7onugl(/tmp/_karma_webpack_12398/commons.js:95693)
	at protoOf.resumeWith_s3a3yh(/tmp/_karma_webpack_12398/commons.js:95740)
	at protoOf.run_mw4iiu(/tmp/_karma_webpack_12398/commons.js:126843)
	at protoOf.process_mza50i(/tmp/_karma_webpack_12398/commons.js:131223)
	at <global>.<unknown>(/tmp/_karma_webpack_12398/commons.js:131331)
```

For example, the following code:

```
"simple test" {
    val flow = MutableSharedFlow<Int>(2)

    val job = launch(UnconfinedTestDispatcher(testCoroutineScheduler)) {
        delay(10)
        flow.take(2).collect {
            it % 2 shouldBe 0
        }
    }

    flow.emit(2)
    flow.emit(4)
    job.join()
}
```

Fails in JS with the above exception.
But if I try not to specify the scheduler, using only ` val job = launch(UnconfinedTestDispatcher())`, then, all the platform fails with the exception:

`Detected use of different schedulers. If you need to use several test coroutine dispatchers, create one 'TestCoroutineScheduler' and pass it to each of them.`

Am I missing something or I'm misusing the scheduler in an unintended way?

I've also noted that problems arise when inside a test the is a `delay` or a method that uses it in the body.
In fact, from the previous example (without `testCoroutineScheduler`) if I remove the delay the test succeeds.
But, if I use `testCoroutineScheduler` even without `delay`, the JS exception above arises.

sksamuel:
Fixes #3575


OliverO2:
The change looks good to me. As I'm a non-user of `runTest` (always doing my own interleaving/sequencing trickery), two questions:
* The `desktopMain` implementation of `TestCoroutineInterceptor` is missing `withContext(TestScopeContainer(this)) { ... }`. Is that intentional?
* Since all actuals of `TestCoroutineInterceptor` look identical, wouldn't one implementation in `commonMain` suffice, getting rid of expect/actual entirely?

sksamuel:
> The change looks good to me. As I'm a non-user of `runTest` (always doing my own interleaving/sequencing trickery), two questions:
> 
> * The `desktopMain` implementation of `TestCoroutineInterceptor` is missing `withContext(TestScopeContainer(this)) { ... }`. Is that intentional?
> * Since all actuals of `TestCoroutineInterceptor` look identical, wouldn't one implementation in `commonMain` suffice, getting rid of expect/actual entirely?

Good spot

OliverO2 (approved):
I love those deletion counts!

sksamuel:
yep :)

PR #3768, opened 2023-11-12
Updated to kotlin 1.9.20

sksamuel:


OliverO2:
Is this ready to test? I'm just exercising the 1.9.20 with the K2 compiler, so I'd be happy to try.

sksamuel:
I'm just prepping the PR. I think after we merge the other bunch of fixes I've queued up as 5.8.1 we could release 5.9.0 with 1.9.20 support ?

OliverO2:
No problems releasing that stuff later. But I'd just check the compiler plugin based on this PR. The current Kotest release fails with K2/Js.

sksamuel:
The compiler plugin tests should fail then as part of this PR.

sksamuel:
Usual gradle annoyances, I'll come back to this:

```
* What went wrong:
An exception occurred applying plugin request [id: 'kotest-multiplatform-library-conventions']
> Failed to apply plugin class 'org.gradle.api.plugins.JavaPlugin'.
   > Could not create an instance of type org.gradle.jvm.component.internal.DefaultJvmSoftwareComponent.
      > Cannot create multiple instances of DefaultJvmSoftwareComponent with source set name 'main'.
```

OliverO2:
Interestingly, overe here with Kotest 5.8.0, 1.9.20/K2/JS compilation fails with
```
java.lang.IllegalStateException: Cannot find io.kotest.engine.TestEngineLauncher class reference
	at io.kotest.framework.multiplatform.embeddablecompiler.Transformer$launcherClass$2.invoke(Transformer.kt:110)
	at io.kotest.framework.multiplatform.embeddablecompiler.Transformer$launcherClass$2.invoke(Transformer.kt:108)
```

OliverO2:
Since I would not expect any issues with the rest of the Kotest framework, I was already thinking about unbundling the Kotest compiler plugin to better keep up with Kotlin compiler dev releases. Ever considered going down that path?

sksamuel:
> Since I would not expect any issues with the rest of the Kotest framework, I was already thinking about unbundling the Kotest compiler plugin to better keep up with Kotlin compiler dev releases. Ever considered going down that path?

Seems like a good idea to me. The launcher interface is stable. We can also put a minimum kotest version on each of the compiler plugin releases.

OliverO2:
Yep. So that would be win-win: Not force new library versions on folks who don't need them, and still offer up-to-date compiler support.

Would require some time to set it up properly. Let's check this integrated PR first. As soon as this one is done, I could use it as a basis and see how far I can make the unbundling happen.

sksamuel:
So you're saying lets do 1.9.20 as a full release then unbundle the
plugins?

OliverO2:
Depends on the timing: We could also try to get 5.8.1 out the door, check how this PR works and then unbundle with the 5.9.0 release on 1.9.20. I'd just not want to delay 1.9.20 support because of some unexpected difficulties with unbundling (though I don't really expect any).

sksamuel:
Yeah that works. You can start unbundling in parallel if you want ?

OliverO2:
Sure, just tell me if this PR is ready for integration. Then I'll base my stuff on this one.

sksamuel:
I don't think this is ready, I can't even get gradle to recognize the versions :)

OliverO2:
OK, so then I'll try the unbundling independently. And you could continue fighting the elephant. :)

sksamuel:
lucky me :)

OliverO2:
Wasn't it that folks in your area were specially trained on [stuff like that](https://en.wikipedia.org/wiki/The_Relic)?

sksamuel:
Whoa never heard of that. Filmed at the field museum too.

sksamuel:
This got into a messy state, closing in favour of https://github.com/kotest/kotest/pull/3776

PR #3775, opened 2023-11-14
Regular file passes the shouldBeEmptyDirectory() assertion

### LINKED ISSUE #3774: A regular file passes the `shouldBeEmptyDirectory()` assertion, but it should not
**Kotest version 5.8.0 (current master)**

Looking at this code:
https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/file/matchers.kt#L21-L30

```Kotlin
fun beEmptyDirectory(): Matcher<File> = object : Matcher<File> {
   override fun test(value: File): MatcherResult {
      val contents = if (value.isDirectory) value.safeList() else emptyList()
      return MatcherResult(
         contents.isEmpty(),
         { "$value should be an empty directory but contained ${contents.size} file(s) [${contents.joinToString(", ")}]" },
         { "$value should not be a non empty directory" }
      )
   }
}
```

it appears that any file that is not a directory passes the `shouldBeEmptyDirectory()` assertion.

I don't think that this behavior is consistent. I would rather expect that `shouldBeEmptyDirectory()` is only true, if the file is also a directory.

sksamuel:
Fixes #3774

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/file/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/file/matchers.kt
@@ -20,12 +20,20 @@ fun File.shouldBeEmptyDirectory() = this should beEmptyDirectory()
 fun File.shouldNotBeEmptyDirectory() = this shouldNot beEmptyDirectory()
 fun beEmptyDirectory(): Matcher<File> = object : Matcher<File> {
    override fun test(value: File): MatcherResult {
-      val contents = if (value.isDirectory) value.safeList() else emptyList()
-      return MatcherResult(
-         contents.isEmpty(),
-         { "$value should be an empty directory but contained ${contents.size} file(s) [${contents.joinToString(", ")}]" },
-         { "$value should not be a non empty directory" }
-      )
+      return if (value.isDirectory) {
+         val contents = value.safeList()
+         MatcherResult(
+            contents.isEmpty(),
+            { "$value should be an empty directory but contained ${contents.size} file(s) [${contents.joinToString(", ")}]" },
+            { "$value should not be a non empty directory" }
+         )
+      } else {
+         MatcherResult(
+            false,
+            { "$value should be an empty directory but was a file" },
+            { "$value should not be a non empty directory" }
+         )
+      }
    }
 }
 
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/file/matchers.kt
```diff
@@ -20,12 +20,20 @@ fun File.shouldBeEmptyDirectory() = this should beEmptyDirectory()
...
+            { "$value should be an empty directory but contained ${contents.size} file(s) [${contents.joinToString(", ")}]" },
+            { "$value should not be a non empty directory" }
+         )
+      } else {
+         MatcherResult(
+            false,
+            { "$value should be an empty directory but was a file" },
+            { "$value should not be a non empty directory" }
+         )
+      }
```
LeoColman, line 36:
```suggestion
      if(vaue.isDirectory) return MatcherResult(false{ "$value should be an empty directory but was a file" }, { "$value should not be a non empty directory" })
      
      val contents = value.safeList()
      return MatcherResult(contents.isEmpty(), { "$value should be an empty directory but contained ${contents.size} file(s) [${contents.joinToString(", ")}]" },
            { "$value should not be a non empty directory" } )

```

I think an early return is better here

LeoColman (approved):

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/file/matchers.kt
sksamuel: Isn't that basically the same thing ?
Personally I think early returns are a bit of a smell and I prefer the functional style of returning the last value. But I know that's not particular kotlinesque.

LeoColman: Yes, same thing.
I personally don't like nesting things. I just like to remove all possible {} xD

obecker: Wow, that was fast. 😀

Actually, I wondered if the shortest solution (having the fewest curlies) is a combination with `aDirectory()` like this
```Kotlin
fun beEmptyDirectory(): Matcher<File> = aDirectory() and object : Matcher<File> {
   // etc ...
}
```

sksamuel: Oh could have been. Well the tests have us covered now :)

PR #3776, opened 2023-11-14
Gradle 8.4 and kotlin 1.9.20

sksamuel:


sksamuel:
@OliverO2 I think you could branch from this for the basis of the unbundling.

estivensh:
> @OliverO2 I think you could branch from this for the basis of the unbundling.

@OliverO2 @sksamuel  what is the status of this pr?

buildSrc/src/main/kotlin/kotest-native-conventions.gradle.kts
```diff
@@ -20,13 +20,11 @@ kotlin {
...
-      watchosX86()
       watchosX64()
       watchosSimulatorArm64()
 
       iosX64()
       iosArm64()
-      iosArm32()
       iosSimulatorArm64()
 
       sourceSets {
```
OliverO2, line 30: Could we just replace the entire `sourceSets` block with `applyDefaultHierarchyTemplate()`? I'd say the [full hierarchy template﻿](https://kotlinlang.org/docs/multiplatform-hierarchy.html#see-the-full-hierarchy-template) has got us covered and the intermediate `desktopMain` is no longer necessary (will be covered by `nativeMain` automatically).

OliverO2 (approved):
Source set configuration could be simplified in `kotest-android-native-conventions.gradle.kts`, `kotest-native-conventions.gradle.kts`, `kotest-watchos-device-conventions.gradle.kts`. Would also resolve warnings:
```
The Default Kotlin Hierarchy Template was not applied to 'project ':kotest-common'':
```

Otherwise looks good!

sksamuel:
> > @OliverO2 I think you could branch from this for the basis of the unbundling.
> 
> @OliverO2 @sksamuel what is the status of this pr?

This will be merged in as soon as we release 5.8.1

Jadarma:
> This will be merged in as soon as we release 5.8.1

@sksamuel Any updates on the 5.8.1 release? Is there something blocking it or are you just waiting for more items to be merged?

kshired:
@sksamuel 

Yesterday gradle 8.5 was released, how about updating to gradle 8.5 and kotlin 1.9.21?

OliverO2:
Seems like compiler plugin unbundling is unnecessary: #3799.

What is the use case for updating Kotest beyond its current Kotlin version of 1.8.10? We'd force users to newer library versions, so what would be the upside to do this now?

This would change, obviously, as new targets are introduced. `wasmJs` and `wasmWasi` seem possible now.

sksamuel:
The one that springs to mind is that TimeSource went stable so we can use that, but it's not a huge deal.

If Kotest 5.8 works with 1.9.x then happy to leave us on 1.8.x

Kantis:
Master is on Gradle 8.6 now, and I'm not sure if we plan to bump Kotlin to 1.9 anymore? Should this be closed @sksamuel ?

sksamuel:
Yeah closing

PR #3783, opened 2023-11-16
Fix orNull with nullProbability=1.0 produce non-nulls

### LINKED ISSUE #3777: `orNull` doesn't respect `nullProbability=1.0` and still produce non-nulls.
kotest version 5.7.2

`Arb<Double>.orNull()` produces non-null values even with `nullProbability=1.0`

kshired:
resolve #3777

There was a bug where the `edgecase` function implemented by the `orNull` function could produce a non-null value even though the `nullProbability` is 1.0. I fixed this bug.


myuwono (approved):

myuwono:
Thank you @kshired :raised_hands:

myuwono (approved):

myuwono:
This is a good change. Fyi @sksamuel I'll merge this

sksamuel:
Thanks!

PR #3786, opened 2023-11-17
Fix continually does not work with null values

### LINKED ISSUE #3780: continually fails with "kotlin.IllegalStateException: No successful result"
5.8.0

The following code fails, because `something()` and therefore `shouldBe` returns `null` and there is [this new elvis operator](https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/continually.kt#L56)
```
 continually(500.milliseconds) {
    something() shouldBe null
}
```

kshired:
resolve #3780

Instead of checking if result is null, I changed to checking using the `Result` class.


myuwono (approved):

sksamuel:
👍🏻

PR #3788, opened 2023-11-18
Provide overriddable defaults for eventually

### LINKED ISSUE #3643: Ability to Register Global Config in Kotest
There was a discussion on [PR #3597](https://github.com/kotest/kotest/pull/3597) regarding the enhancement of Kotest's configuration system.

The specific use-case discussed was about providing a global configuration for functions like eventually, which currently require an explicit configuration for each call. This would be particularly useful for setting defaults that could be shared across tests.

SkSamuel responded positively but mentioned the need for accessing the framework's project context within the assertions block, which currently does not have access to it. This might require moving the project context into the shared API and could be considered for a follow-up PR.

sksamuel:
Fixes #3643


Kantis (approved):
neat! 👍🏼

PR #3789, opened 2023-11-19
Avoid use of test types in junit listener

sksamuel:
if I can do the same for the intellij runner, then we can strip test types out completely, and avoid dynamic/test/container.


Kantis:
Not sure I understand what's going on here tbh :) Are we trying to automatically resolve whether we need to register a `container`, `test`, or `container_and_test` without relying on differentiating by container / test / spec in kotest-terms?? Sounds cool 👀

sksamuel:
I was trying to make it so we can ditch the entire concept of test / container but then I realized we need this in order to make isolation modes work. If we don't specify in the DSL if something is leaf or not, then it doesn't know if it should make a new instance or not before running it.

We should be able to scrap dynamic though.

Issue #3736, opened 2023-10-13
More support for ranges


AlexCue987:
Examples:
* `(0..2) shouldOverlapWith (1..3)`
* `(2..3) shouldBeIn (1..4)`
* `(0..1) shouldNotOverlapWith (2..3)`
* `(4..5) shouldNotBeIn (1..3)`

should support both ClosedRange and OpenEndRange.

I'd like to do it if this feature is needed.

mirageoasis:
I think add support shouldBeIn range would be hard. But maybe creating new function can be solution.

AlexCue987:
> I think add support shouldBeIn range would be hard. But maybe creating new function can be solution.

got it. how about this `(2..3) shouldBeInsideOf (1..4)`

mirageoasis:
> > I think add support shouldBeIn range would be hard. But maybe creating new function can be solution.
> 
> got it. how about this `(2..3) shouldBeInsideOf (1..4)`

think we have to specify which type we should accept only range? or other list types

mirageoasis:
if maintainer says yes maybe I can add method to support this proposal

AlexCue987:
> > > I think add support shouldBeIn range would be hard. But maybe creating new function can be solution.
> > 
> > 
> > got it. how about this `(2..3) shouldBeInsideOf (1..4)`
> 
> think we have to specify which type we should accept only range? or other list types

only ranges. It does not make sense for lists

sksamuel:
What about

```
(0..2) shouldIntersect (1..3)
(2..3) shouldBeWithin (1..4)
```

AlexCue987:
> What about
> 
> ```
> (0..2) shouldIntersect (1..3)
> (2..3) shouldBeWithin (1..4)
> ```

let's use these function names

sksamuel:
Assigned to you

mirageoasis:
I think there should be some careful approaches about ranges since it has backward range

mirageoasis:
> I think there should be some careful approaches about ranges since it has backward range

and think we need to add support for until 
or maybe can I get until feature assigned?

sksamuel:
> I think there should be some careful approaches about ranges since it has backward range

backwards or forwards, still comprises a set of numbers, so within should be fine.

LeoColman:
I'd also enforce that we should try to go for an strictly correct mathematical definition.

Sets theory is very well defined, ie
- A set within a set
- An open ended range
- A close ended
- etc

[...]

PR #3792, opened 2023-11-27
Add `intersect` matcher to ranges

AlexCue987:
Implement intersect for the following issue: https://github.com/kotest/kotest/issues/3736
Another matcher named within will be implemented in another PR.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/ranges/Range.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/ranges/Range.kt
@@ -1,0 +1,70 @@
+package io.kotest.matchers.ranges
+
+data class Range<T: Comparable<T>>(
+    val start: RangeEdge<T>,
+    val end: RangeEdge<T>
+) {
+   init {
+      require(start.value <= end.value) {
+         "${start.value} cannot be after ${end.value}"
+      }
+   }
+
+   override fun toString(): String {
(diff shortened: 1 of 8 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/ranges/Range.kt
```diff
@@ -0,0 +1,70 @@
...
+    val end: RangeEdge<T>
+) {
+   init {
+      require(start.value <= end.value) {
+         "${start.value} cannot be after ${end.value}"
+      }
+   }
+
+   override fun toString(): String {
+      return "${if(start.edgeType == RangeEdgeType.INCLUSIVE) "[" else "("}${start.value}, ${end.value}${if(end.edgeType == RangeEdgeType.INCLUSIVE) "]" else ")"}"
```
LeoColman, line 14: Complicated, but I like being mathematically correct! Kudos for using the symbols correctly.

I've also seen "[1, 3[" being used instead of "[1,3)". I think either way is fine

LeoColman (approved):
The tests can probably be generalized to a common case using property testing and actually asserting on sets mathematical properties.

I don't think it's necessary to get this merged, but could be a future improvement.

AlexCue987:
> The tests can probably be generalized to a common case using property testing and actually asserting on sets mathematical properties.
> 
> I don't think it's necessary to get this merged, but could be a future improvement.

will do!

AlexCue987:
> The tests can probably be generalized to a common case using property testing and actually asserting on sets mathematical properties.
> 
> I don't think it's necessary to get this merged, but could be a future improvement.

done https://github.com/kotest/kotest/pull/3824

Issue #10, opened 2016-04-22
Add tolerance matcher for doubles/floats


sksamuel:

russel:
Some notion of "almost equal" is needed for doubles and floats (and BigDecimal?) and perhaps should be the default for all non-integral value testing?

a == b or assertEqual(a, b) is a big problem where a and b are floating point, and yet far too few programmers actually appreciate this.

sksamuel:
Yep. I've added a shouldBe b within c

sksamuel:
Maybe i could emit a warn if you test doubles or floats without a tolerance.

russel:
Issuing a warning on detection of use of == or assertEqual or similar would be a huge USP for a testing framework!

sksamuel:
Agreed.

PR #3796, opened 2023-12-01
Compare lists in more detail (#10)

AlexCue987:
add new matcher `shouldMatchList` that compares two lists, tries to find sublists that match, and tries to find elements that match only partially. Example outputs are this one:
```
Mismatch:
actual[0] = Fruit(name=pear, color=green, taste=sweet)

Match:
expected[0] == actual[1]: Fruit(name=apple, color=green, taste=sweet)
expected[1] == actual[2]: Fruit(name=apple, color=red, taste=sweet)

Mismatch:
expected[2] = Fruit(name=pear, color=green, taste=sweet)

Possible matches:
actual[0] == expected[2], is: Fruit(name=pear, color=green, taste=sweet)"""
```
and this one:
```
               Mismatch:
               actual[0] = Fruit(name=onion, color=green, taste=sweet)

               Match:
               expected[0] == actual[1]: Fruit(name=apple, color=green, taste=sweet)
               expected[1] == actual[2]: Fruit(name=apple, color=red, taste=sweet)

               Mismatch:
               expected[2] = Fruit(name=pear, color=green, taste=sweet)

               Possible matches:
               actual[0] = Fruit(name=onion, color=green, taste=sweet) is similar to
               expected[0] = Fruit(name=apple, color=green, taste=sweet)

               "name" expected: apple,
                but was: onion
               "color" = green
               "taste" = sweet

               actual[0] = Fruit(name=onion, color=green, taste=sweet) is similar to
               expected[2] = Fruit(name=pear, color=green, taste=sweet)

               "name" expected: pear,
                but was: onion
               "color" = green
               "taste" = sweet
```

this is a new matcher, so it won't break any existing tests, and it allows users to opt-in, rather than change existing behavior.


Kantis:
The API and code footprint of this PR is huge. It would be good if the exposed API is reduced to the minimum possible.

It's also not entirely clear to me how this matcher is supposed to work, even after reading the docs and examples. Is this supposed to be an enhanced version of some existing matcher?

AlexCue987:
> The API and code footprint of this PR is huge. It would be good if the exposed API is reduced to the minimum possible.

Oops, my bad. Fixed: only the matchers are left public, so the change to public API is as small as possible. Sorry about that.

AlexCue987:
> It's also not entirely clear to me how this matcher is supposed to work, even after reading the docs and examples. Is this supposed to be an enhanced version of some existing matcher?

It's a completely new matcher, developed from scratch. It was inspired by `git diff`, and intends to describe differences in a similar way. Can you show me how to convey this message clearly? TIA!

Kantis:
>`shouldMatchList` that compares two lists, tries to find sublists that match, and tries to find elements that match only partially.

This sounds exactly like what `shouldContainInOrder` would test for? Perhaps we could use your improved failure messages to improve the output of that assertion instead?

AlexCue987:
> shouldContainInOrder

I concur that the output of `shouldContainInOrder` can be improved, and maybe it should be improved, eventually. 

But this is not what I want to accomplish this time - this matcher improves the output for the simplest, the most common for me use case, the exact match such as `shouldBe`. Which is what we use most of the time, and which is why I wrote all this code: to make it easier to see what exactly is different for `shouldBe`. 

WDYT?

AlexCue987:
> > `shouldMatchList` that compares two lists, tries to find sublists that match, and tries to find elements that match only partially.
> 
> This sounds exactly like what `shouldContainInOrder` would test for? Perhaps we could use your improved failure messages to improve the output of that assertion instead?

I concur, opened another PR https://github.com/kotest/kotest/pull/3849 - it modifies `shouldContainExactly`
Closing this one

AlexCue987:
> > `shouldMatchList` that compares two lists, tries to find sublists that match, and tries to find elements that match only partially.
> 
> This sounds exactly like what `shouldContainInOrder` would test for? Perhaps we could use your improved failure messages to improve the output of that assertion instead?

thank you for the feedback!

PR #3798, opened 2023-12-02
Fix TestWithConfigBuilder#config using parsed name when clear TestDslState

### LINKED ISSUE #3795: Unable to fully define a `DescribeSpec` test with a multiline description that invokes the `TestWithConfigBuilder#config` method
### Expected Behavior
The successful execution of a`DescribeSpec` test with multiline description that invokes the `TestWithConfigBuilder#config` method.

### Current Behavior
Failure of a`DescribeSpec` test with multiline description that invokes the `TestWithConfigBuilder#config` method, due to the test not being "fully defined".

### Context
A `DescribeSpec` test with a multiline description, that invokes the `TestWithConfigBuilder#config` method, fails with the following exception: 

```
java.lang.IllegalStateException: Test was not fully defined: [DescribeSpecContainerScope] -- prevents test from executing
```

### Steps to Reproduce
**Minimal reproducible example**
```kotlin
package com.example

import io.kotest.core.NamedTag
import io.kotest.matchers.booleans.shouldBeTrue

class KoTestMultilineTestNameWithConfigBugSpec: DescribeSpec({
    val fooTag = NamedTag("foo")

    describe("Multiline tagged test name") {
        it("""prevents test
            | from executing""".trimMargin()).config() {
            true.shouldBeTrue()
        }
    }
})
```

### Your Environment
- **Build tool:** Gradle 8.5
- **Kotest version:** 5.8

kshired:
resolve #3795

- Refactor `startTest` and `clear` functions  of the `TestDslState` class from taking a `TestDescriptor` class as a parameter to taking `String` class.
- Fix TestWithConfigBuilder#config using parsed name when clear TestDslState
  - The name field of the `TestWithConfigBuilder` class is a `TestName` class, which parses and stores the user's input during initialization. This caused a bug.


LeoColman:
This pull request is fundamentally changing how we name tests. 
This kind of change will have changes on all downstream dependencies that are unacceptable. 

Test naming use the whole path for design reasons, and that can't be changed without a lot of discussion. We are not willing to make that kind of change at the moment.

kshired:
@LeoColman 
But isn't the current situation of putting unparsed names in `startTest` and parsed names in `clear` also weird?

LeoColman:
Absolutely!
I believe that's how JUnit maps their tests, and thus we use it the same way for compatibility.

I agree with you it's weird

kshired:
@LeoColman 

But kotest is not a "kotlin version of JUnit", is there any reason to guarantee compatibility even it makes bug?

sksamuel:
I think this is fine. This is not changing the output in any way, it's just changing the way we internally check if a test was started but not completed.

sksamuel:
@kshired we do need to change this to use the full path though, otherwise this would report success as the second "foo" will mark the first "foo" as completed, since it uses a set.

```kotlin
class Test : FreeSpec() {
   init {
      "a" - {
         "foo".config()
      }
      "b" - {
         "foo".config() {}
      }
   }
}
```

sksamuel:
@LeoColman @kshired 
I am going to merge this in its in broken state as I want to release 5.8.1 tonight with this fix, so I will build off this with the fix to use the full path.
This approach is fine, as the name change only affects how we check for unfinished tests. It's not changing the way the name is presented in any outputs or anything.

PR #3805, opened 2023-12-13
Add wasmJs target

### LINKED ISSUE #3598: Support wasm target
WASM is a new experimental target from JetBrains from 1.8.20. They a planning to move from JS to WASM eventually. 

The support would be especially beneficial for library authors whoes common tests are depending on kotest, but who want to support the new platform. 

I, myself, primarily interested in support for kotest-assertions.

OliverO2:
Includes:
* Migrating to Kotlin 1.9.23 (with support for 2.0.0-Beta2), Gradle 8.6
* Using kotlinx.coroutines 1.8.0
* Simplifying the source set hierarchy in build scripts
* Adding a common `jsHosted` source set for JS-hosted targets (`js` and `wasmJs`)
* Removing `watchosX86` and `iosArm32` targets
* Deprecating `kotest-extensions-http`, meanwhile restricting it to JVM

Fixes #3598.


OliverO2:
Current status: Waiting for resolution or workaround regarding [KT-64357](https://youtrack.jetbrains.com/issue/KT-64357) K/Wasm: "CompileError: WebAssembly.Module(): Invalid types for ref.test: local.get of type externref has to be in the same reference type hierarchy as (ref ...)"

OliverO2:
Status update: There is a problem with `wasmJs` test invocations. While Mocha functions like `describe` and `it` are available with `js` tests, they are not with `wasmJs`. This prevents execution of Node-based and browser-based tests for `wasmJs`. The problem can be reproduced independently of Kotest ([kotlinlang Slack](https://kotlinlang.slack.com/archives/CDFP59223/p1702761028987939)). Once that problem is resolved, I'm expecting this PR to become ready for review.

eygraber:
@OliverO2 is there anything to track in terms of this getting unblocked?

OliverO2:
@eygraber No, AFAIK there is nothing blocking this technically, it's just a matter of review resources, since this is not exactly a tiny PR. Also, we'd possibly have to unbundle the Ktor extension in a separate step, as Ktor/Wasm currently requires the upcoming Ktor 3.0 version.

buildSrc/src/main/kotlin/Ci.kt
```diff
@@ -2,7 +2,7 @@ object Ci {
 
    // this is the version used for building snapshots
    // .buildnumber-snapshot will be appended
-   private const val snapshotBase = "5.9.0"
+   private const val snapshotBase = "5.8.999"
```
Kantis: Should this be reverted before merging?

buildSrc/src/main/kotlin/kotest-native-conventions.gradle.kts
```diff
@@ -20,74 +21,31 @@ kotlin {
...
-         val watchosArm32Test by getting { dependsOn(nativeTest) }
-         val watchosArm64Test by getting { dependsOn(nativeTest) }
-         val watchosX86Test by getting { dependsOn(nativeTest) }
-         val watchosX64Test by getting { dependsOn(nativeTest) }
-         val watchosSimulatorArm64Test by getting { dependsOn(nativeTest) }
-
-         val tvosX64Test by getting { dependsOn(nativeTest) }
-         val tvosArm64Test by getting { dependsOn(nativeTest) }
-         val tvosSimulatorArm64Test by getting { dependsOn(nativeTest) }
+      // FIXME: WORKAROUND https://github.com/Kotlin/kotlinx.coroutines/issues/3968
```
ajalt: @OliverO2 FYI coroutines 1.8.0 is out now with a fix for this bug, so the workaround is no longer necessary

OliverO2: Yep. Thanks for the reminder. I'll get back to this as soon as someone is ready for a complete review (but no sooner than March, 4th due to other commitments). Looking forward to updating this and possibly other dependencies then.

buildSrc/src/main/kotlin/kotest-android-native-conventions.gradle.kts
```diff
@@ -10,22 +11,15 @@ kotlin {
...
-         val androidNativeArm64Main by getting { dependsOn(desktopMain) }
 
-         val commonTest by getting
-         val nativeTest by getting { dependsOn(commonTest) }
-         val androidNativeX86Test by getting { dependsOn(nativeTest) }
-         val androidNativeX64Test by getting { dependsOn(nativeTest) }
-         val androidNativeArm32Test by getting { dependsOn(nativeTest) }
-         val androidNativeArm64Test by getting { dependsOn(nativeTest) }
+      // TODO: The "desktop" intermediate source set can be integrated into "native". In this case
+      //     the following block can be replaced with `applyDefaultHierarchyTemplate()`.
```
Kantis, line 16: Do you reckon we should get rid of desktop now or in a separate PR?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestEngine.kt
```diff
@@ -57,8 +57,8 @@ class TestEngine(private val config: TestEngineConfig) {
                config.configuration.concurrentSpecs ?: config.configuration.parallelism,
                context,
             )
-            Platform.JS -> SequentialTestSuiteScheduler(context)
-            Platform.Native -> SequentialTestSuiteScheduler(context)
+
+            else -> SequentialTestSuiteScheduler(context)
```
Kantis: I like keeping exhaustive whens, forces us to come back here if we ever need to add another platform.
```suggestion

            Platform.JS,
            Platform.Native,
            Platform.WasmJs -> SequentialTestSuiteScheduler(context)
```

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestEngineLauncher.kt
```diff
@@ -192,6 +191,7 @@ class TestEngineLauncher(
    }
 
    fun withJs(): TestEngineLauncher = withPlatform(Platform.JS)
+   fun withWasmJs(): TestEngineLauncher = withPlatform(Platform.WasmJs).withTeamCityListener()
```
Kantis, line 194: Why do we add the `TeamCityListener` here, and only for wasm?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/sorters.kt
```diff
@@ -22,7 +22,7 @@ interface SpecSorter {
 /**
  * An implementation of [SpecExecutionOrder] which will order specs in lexicographic order.
  */
-object LexicographicSpecSorter : SpecSorter {
+val LexicographicSpecSorter = object : SpecSorter {
```
Kantis: Breaking change i reckon. Do we want to make it `internal` while we're at it?

kotest-framework/kotest-framework-engine/src/wasmJsMain/kotlin/KotlinTestApi.kt
```diff
@@ -0,0 +1,7 @@
+// The JS launcher set up by the Kotlin Gradle plugin calls this function in addition to `main()`.
+@OptIn(ExperimentalJsExport::class)
+@JsExport
+internal fun startUnitTests() {
+   // The Kotlin compiler would insert test invocations here for kotlin-test.
+   // This mechanism is not used with Kotest.
```
Kantis, line 6: Do we have any documentation on how this works in Kotest instead? I keep forgetting. :(

kotest-framework/kotest-framework-multiplatform-plugin-gradle/test-project/build.gradle.kts
```diff
@@ -68,3 +76,17 @@ if (useNewNativeMemoryModel.toBoolean()) {
       }
    }
 }
+
+// Use a Node.js version current enough to support Kotlin/Wasm
+
+rootProject.plugins.withType<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin> {
+   rootProject.extensions.configure<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootExtension> {
+      nodeVersion = "22.0.0-v8-canary20231213fc7703246e"
```
Kantis: Should this be updated/changed before merge?

Kantis (approved):
Nice work @OliverO2 🙌🏼 

Some small things to clean up, but I think it would be great to get this out when you have time again.

buildSrc/src/main/kotlin/Ci.kt
OliverO2: Yes.

buildSrc/src/main/kotlin/kotest-android-native-conventions.gradle.kts
OliverO2: This should go into a separate PR, as it will affect many paths.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestEngine.kt
OliverO2: Fine with me.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestEngineLauncher.kt
OliverO2: It is remarkably complicated: The TeamCityListener formats test output in the "TeamCity" format. While on the JS platform, the same is accomplished by the Kotlin test infra (via Mocha), this is not the case for wasmJs (which does not use Mocha). Without the extra TeamCityListener, `gradlew test` in `kotest:kotest-framework:kotest-framework-multiplatform-plugin-gradle` fails. For details, see [this Slack thread](https://slack-chats.kotlinlang.org/t/16155931/adding-a-wasm-target-to-kotest-i-m-observing-strange-differe).

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/sorters.kt
OliverO2: Sure, makes sense.

kotest-framework/kotest-framework-engine/src/wasmJsMain/kotlin/KotlinTestApi.kt
OliverO2: I wouldn't call it documentation, but it can be inferred from sources in `kotest-framework-multiplatform-plugin-embeddable-compiler/src/jvmMain/kotlin/io/kotest/framework/multiplatform/embeddablecompiler`, in particular `entry.kt`.

kotest-framework/kotest-framework-multiplatform-plugin-gradle/test-project/build.gradle.kts
OliverO2: The regular 22.x Node.js release is scheduled for April. Until then, we must carefully select a pre-release which works with the Kotlin compiler infra (newest 22.0.0 canaries no longer support the '--experimental-wasm-gc' option). Meanwhile, I'll update to a compatible Node.js release mentioned in KT-65864.

OliverO2:
@Kantis, @sksamuel
Before merging this, can we decide what to do with `kotest-extensions-http`? It builds with `kotest-js-conventions`, which would include Wasm/JS. It also uses Ktor dependencies, which have Wasm support only for a Beta release. `kotest-extensions-http` does not appear in documentation and is probably not used that much.

So should we
1. Drop `kotest-extensions-http` entirely?
2. Make it build for Jvm only?
3. Craft a special build for JS, but not Wasm/JS?
4. Move `kotest-extensions-http` to a separate repo?
5. Do something else?

Kantis:
I believe 4 is appropriate

OliverO2:
Moving `kotest-extensions-http` to a separate repo should happen in a separate PR. For now, I have reverted the changes that would require Ktor 3.0.0-wasm2 and restricted the extension to build for JVM only. I have found that its main functionality is JVM-only anyway (requiring resource file access) and it has not been updated to work with current Ktor 2 versions.

So this seems to be the best intermediate step for now, enabling this PR to get merged.

Kantis:
I've begun setting up the extension [here](https://github.com/kotest/kotest-extensions-http) 

But let's proceed like you suggested..

Is everything ready for final review now?

OliverO2:
Yes. This one is complete and ready. How best to merge it given Sam's intention of getting out 5.8.1 this week (see kotlinlang Slack)?

OliverO2:
If there are no objections, I'd merge this within 24 hours so that other PRs can align.

sksamuel:
👍🏻

kotest-assertions/kotest-assertions-shared/src/jsHostedMain/kotlin/io/kotest/assertions/print/printers.kt
```diff
@@ -0,0 +1,5 @@
+package io.kotest.assertions.print
+
+// js doesn't support full reflection so we just need to delegate to the instances toString
+actual fun <A : Any> dataClassPrint(): Print<A> = ToStringPrint
+actual fun <A : Any> platformPrint(a: A): Print<A>? = null
```
sksamuel, line 5: What is jsHosted?

OliverO2: `jsHosted` is the common source set for Kotlin/JS and Kotlin/Wasm/JS, both of which operate within a JS host environment. In the near future, there will be Kotlin/Wasm/WASI (not JS-hosted).

OliverO2: Defined here: https://github.com/kotest/kotest/pull/3805/files#diff-8eba8e3034dadebe1b9bae3190de6c5de838ea157d15ac6fe48d54291883bf0a

kotest-common/src/desktopMain/kotlin/io/kotest/mpp/sysprop.kt
```diff
@@ -1,7 +1,9 @@
 package io.kotest.mpp
 
+import kotlinx.cinterop.ExperimentalForeignApi
 import platform.posix.getenv
 import kotlinx.cinterop.toKString
 
 actual fun sysprop(name: String): String? = null
+@OptIn(ExperimentalForeignApi::class)
```
sksamuel, line 8: Whats ExperimentalForeignApi do

kotest-assertions/kotest-assertions-shared/src/jsHostedMain/kotlin/io/kotest/assertions/print/printers.kt
sksamuel: By within a js host, does that mean executes within a JS engine/vm?
And Kotlin/Wasm/WASI will not require a js engine ?

kotest-common/src/desktopMain/kotlin/io/kotest/mpp/sysprop.kt
OliverO2: Don't remember exactly, seems to be required when calling this one (which was there before):
```kotlin
/**
 * @return the [kotlin.String] decoded from given zero-terminated UTF-8-encoded C string.
 */
@ExperimentalForeignApi
public fun CPointer<ByteVar>.toKString(): String = this.toKStringFromUtf8()
```

kotest-assertions/kotest-assertions-shared/src/jsHostedMain/kotlin/io/kotest/assertions/print/printers.kt
OliverO2: 1. Yes, a JS host means that a JS engine is available, providing access to JS APIs (though the bridging process slows things down). The JS engine also shares its garbage collector with GC-based languages targeting Wasm (like Kotlin).
2. Yes. Pure WASI runtimes like [Wasmtime](https://github.com/bytecodealliance/wasmtime) or [WasmEdge](https://wasmedge.org/) have no JS engine at all. These seem to become the fast (microsecond startup times), lightweight, language-independent, universal virtualization runtimes of the future (no more Docker). (There's also Node.js, which does have a V8-based WASI runtime.)

OliverO2: Ticket to track: [KT-60278](https://youtrack.jetbrains.com/issue/KT-60278) Make Kotlin/Wasm suitable for standalone Wasm VMs

kotest-common/src/desktopMain/kotlin/io/kotest/mpp/sysprop.kt
ajalt: All of the foreign language APIs (`CPointer` etc.) [are experimental][1].

[1]: https://kotlinlang.org/api/latest/jvm/stdlib/kotlinx.cinterop/-experimental-foreign-api/

PR #3810, opened 2023-12-21
docs: fix EnabledCondition example (interface, not class)

jhult:
Relates to #3259 but applied to all doc versions


sksamuel:
Nice spot thank you

PR #3814, opened 2023-12-25
Fix `BigDecimal.ZERO.shouldBePositive()` returning true

pubiqq:


Kantis (approved):
Thanks for the contribution. I think this can be considered a bug fix. Anyone relying on the old behaviour (zero being considered positive) could switch assertion to `bd.shouldNotBeNegative()`

PR #3818, opened 2024-01-02
Add `shouldNotBePositive` and `shouldNotBeNegative` matchers for BigDecimal

pubiqq:
> Anyone relying on the old behaviour (zero being considered positive) could switch assertion to `bd.shouldNotBeNegative()`

_Originally posted by @Kantis in https://github.com/kotest/kotest/pull/3814#pullrequestreview-1798445484_

--

Okay, I got the hint 😂.


Kantis (approved):
Thanks. Thought for sure that we had these 😆

Issue #3820, opened 2024-01-04
shouldContainExactlyInAnyOrder doesn't fully support custom Equality comparisons


gwtjxyz:
**Which version of Kotest are you using**
5.8.0

When supplying a custom Equality verifier to `containExactlyInAnyOrder()`, our test result is false even though the verifier's `areEqual()` returns true. The culprit seems to be this code block:

https://github.com/kotest/kotest/blob/8fb0b73b3b24b24b23064cc5614c41c0db5a4431/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactlyInAnyOrder.kt#L121-L124

The `filterKeys` block uses the custom verifier to match the item, but then extracts the map value out of it through a normal comparison with `[k]`, meaning that if the normal comparison returns false and the custom equality one returns true, the overall result will still be false.

Doing something like this seems to fix this problem, but I haven't fully tested it with an absent Equality verifier and there's probably a more elegant solution out there than this:

```kotlin
val passed = expectedGroupedCounts.size == valueGroupedCounts.size
    && expectedGroupedCounts.all { (k, v) ->
        val equalKey = valueGroupedCounts.keys.find { verifier?.verify(k, it)?.areEqual() ?: (k == it) }
        return valueGroupedCounts.filterKeys { verifier?.verify(k, it)?.areEqual() ?: (k == it) }[equalKey] == v
   }
```

sksamuel:
The bug described here (using `filterKeys { ... }[k]` with standard equality after custom-equality filtering) was fixed in PR #4160. The current implementation in `containExactlyInAnyOrder.kt` uses a proper `countMismatch` function that finds the matching key via the custom verifier first (`equalKeyInExpected`) and then looks it up in the map using that key — avoiding the original problem entirely. There is also a test at `ShouldContainExactlyTest.kt` covering the custom verifier case. Closing as fixed.

PR #3821, opened 2024-01-04
Fix custom equality verifier-related logic and add tests for it

gwtjxyz:
Fix for https://github.com/kotest/kotest/issues/3820, added some tests for custom Equality too to make sure it works properly

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactlyInAnyOrder.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactlyInAnyOrder.kt
@@ -120,7 +120,9 @@ fun <T, C : Collection<T>> containExactlyInAnyOrder(
 
    val passed = expectedGroupedCounts.size == valueGroupedCounts.size
       && expectedGroupedCounts.all { (k, v) ->
-      valueGroupedCounts.filterKeys { verifier?.verify(k, it)?.areEqual() ?: (k == it) }[k] == v
+      // account for the case when a key might pass verifier equality but not default equality
+      val key = valueGroupedCounts.keys.find { verifier?.verify(k, it)?.areEqual() ?: (k == it) }
+      valueGroupedCounts.filterKeys { key == it }[key] == v
    }
 
    val missing = expected.filterNot { t ->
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactlyInAnyOrder.kt
```diff
@@ -120,7 +120,9 @@ fun <T, C : Collection<T>> containExactlyInAnyOrder(
 
    val passed = expectedGroupedCounts.size == valueGroupedCounts.size
       && expectedGroupedCounts.all { (k, v) ->
-      valueGroupedCounts.filterKeys { verifier?.verify(k, it)?.areEqual() ?: (k == it) }[k] == v
+      // account for the case when a key might pass verifier equality but not default equality
+      val key = valueGroupedCounts.keys.find { verifier?.verify(k, it)?.areEqual() ?: (k == it) }
+      valueGroupedCounts.filterKeys { key == it }[key] == v
```
AlexCue987, line 125: If `actual` is `listOf("Apple", "apple")`, and value is `listOf("APPLE","APPLE")`, then `expectedGroupedCounts` is `mapOf("Apple" to 1, "apple" to 1)`, and `valueGroupedCounts` is `mapOf("APPLE" to 2)`. If the verifier matches strings in case-insensitive way, this fix won't work - these two collections should match and they won't.

AlexCue987: the code to reproduce:
```
            val caseInsensitiveStringEquality: Equality<String> = object : Equality<String> {
               override fun name() = "Case Insensitive String Matcher"

               override fun verify(actual: String, expected: String): EqualityResult {
                  return if(actual.uppercase() == expected.uppercase())
                     EqualityResult.equal(actual, expected, this)
                  else
                     EqualityResult.notEqual(actual, expected, this)
               }
            }
            listOf("apple", "Apple") should containExactlyInAnyOrder(listOf("APPLE", "APPLE"), caseInsensitiveStringEquality)
```

LeoColman:
Closing as stale

PR #3827, opened 2024-01-08
Make `shouldBe expected` equivalent to `should be(expected)` for non-matcher objects

pubiqq:
This PR:
- Makes `x shouldBe y` and `x should be(y)` equivalent (for any object except `Matcher`).
- Makes the implementations of `x shouldBe y` and `x shouldNotBe y` similar.
- Improves the error message when using `x should be(y)`.

--

P.S. It might also be worth doing something with `Matcher` objects, but I'm not sure what the expected behavior should be in this case:

1. `x shouldBe Matcher` and `x should be(Matcher)` should have different behavior.
2. `x shouldBe Matcher` and `x should be(Matcher)` should behave the same way (the same way as `x shouldBe Matcher` does now).
3. `x shouldBe Matcher` and `x should be(Matcher)` should behave the same way (the same way as `x should be(Matcher)` does now).


pubiqq:
Friendly ping

pubiqq:
Bump

pubiqq:
Bump

sksamuel (approved):

sksamuel:
lgtm

PR #3831, opened 2024-01-16
Support `Iterable` in `withData` on root level

### LINKED ISSUE #3830: `withData` on root level should support `Iterable` (like `withData` on container level)
Currently kotest allows me to use a range in `withData` when it is inside a `describe`
```kotlin
class ExampleTest : FunSpec({

    describe("numbers 1 to 10") {
        withData(1..10) {
            // ....
        }
    }
})
```

However, using the same `withData` on root level doesn't compile
```kotlin
class ExampleTest : FunSpec({

    withData(1..10) {
        // ....
    }
})
```

That is because `withData` on container level accepts an `Iterable`, whereas `withData` on root level accepts only a `Collection`.

This should be aligned so that `Iterable` can be consistently used in both contexts.

obecker:
Fixes #3830


OliverO2 (approved):
Good catch, excellent work as usual! Thank you for this contribution!

PR #3839, opened 2024-01-19
implement beWithin for ClosedRange and OpenEndRange

AlexCue987:
* implement beWithin for `ClosedRange` and `OpenEndRange`.
* handle differently for discrete numbers `Int` and `Long`, because for ranges of discrete numbers the logic is different. For example, while for contiguous numbers like Double `[0.0, 3.0)  shouldNotBeWithin [0.0, 2.0]` because 2.5 is in fisrt range but not in second, for Int and Long `[0, 3)  shouldBeWithin [0, 2]` because these are the same sets of Int.


kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
```diff
@@ -1862,6 +1871,8 @@ public final class io/kotest/matchers/ranges/Range {
 	public fun <init> (Lio/kotest/matchers/ranges/RangeEdge;Lio/kotest/matchers/ranges/RangeEdge;)V
 	public final fun component1 ()Lio/kotest/matchers/ranges/RangeEdge;
 	public final fun component2 ()Lio/kotest/matchers/ranges/RangeEdge;
+	public final fun contains (Lio/kotest/matchers/ranges/Range;)Z
+	public final fun contains (Lio/kotest/matchers/ranges/RangeEdge;)Z
```
Kantis: Why is this entire class not internal?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/ranges/bewithin.kt
```diff
@@ -0,0 +1,216 @@
...
+import io.kotest.assertions.print.print
+import io.kotest.matchers.Matcher
+import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldNot
+
+/**
+ * Verifies that this [ClosedRange] beWithin with another [ClosedRange].
+ *
+ * Assertion to check that this [ClosedRange] beWithin with another [ClosedRange].
```
Kantis: Are these AI generated? Feels like some more effort could be put into the wording here :)

* First sentence and 2nd sentence feels like they say the same thing with different words
* What does `beWithins` mean? Phrase it better and/or provide an example

kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
AlexCue987: > Why is this entire class not internal?

it's internal now, TY for calling it out

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/ranges/bewithin.kt
AlexCue987: added examples

Kantis (approved):

LeoColman (approved):

PR #3842, opened 2024-01-21
Added java.time.Year arbitrary support

kshired:
resolve #3838

PR #3844, opened 2024-01-22
Added java.time.Year supports for reflective arbitrary bind

kshired:
related with #3842


kshired:
@myuwono 

I forgot that I have to change reflective bind as well for supporting `Arb<Year>`

myuwono (approved):

myuwono:
Good one 👍

PR #3845, opened 2024-01-22
Should contain exactly in any order count mismatch draft (#19)

AlexCue987:
modify `containExactlyInAnyOrder` so that it explicitly prints out when count of duplicates is different, such as:
```
            shouldThrow<AssertionError> {
               listOf(1, 2, 2, 3).shouldContainExactlyInAnyOrder(listOf(1, 2, 3, 3))
            }.shouldHaveMessage(
               """
                  Collection should contain [1, 2, 3, 3] in any order, but was [1, 2, 2, 3]
                  CountMismatches: Key="2", expected count: 1, but was: 2, Key="3", expected count: 2, but was: 1
               """.trimIndent()
```

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactlyInAnyOrder.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactlyInAnyOrder.kt
@@ -129,13 +129,19 @@ fun <T, C : Collection<T>> containExactlyInAnyOrder(
    val extra = actual.filterNot { t ->
       expected.any { verifier?.verify(it, t)?.areEqual() ?: (t == it) }
    }
+   val countMismatch = countMismatch(expectedGroupedCounts, valueGroupedCounts)
 
    val failureMessage = {
       buildString {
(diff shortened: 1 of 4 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containExactlyInAnyOrder.kt
```diff
@@ -147,3 +153,22 @@ fun <T, C : Collection<T>> containExactlyInAnyOrder(
       negatedFailureMessage
    )
 }
+
+internal fun<T> countMismatch(expectedCounts: Map<T, Int>, actualCounts: Map<T, Int>) =
```
myuwono, line 157: Does this work for `List<T?>`? Is there a nested nullability issue?

AlexCue987: > Does this work for `List<T?>`? Is there a nested nullability issue?

it does work with nulls, added tests to verify that

myuwono (approved):

Issue #20, opened 2016-05-28
throwable does not cause test to fail


holgerbrandl:
Example:
```
class MyException() : Throwable()

class Playground : FlatSpec() { init {
    "it" should "do something" {
        throw MyException()
    }

    "it" should "do something else" {
    }
}
}
```
In Intellij/gradle this gives an incorrect picture like if the first test suceeded correctly.
![image](https://cloud.githubusercontent.com/assets/200952/15626107/141aa72a-24bc-11e6-888f-e99adf7f0c01.png)

If a `class MyRunException() : RuntimeException()` instance is thrown instead, the test is tagged as failed as expected. Why don't `Throwable`s cause tests to fail but require RuntimeException as base class? 

I thought it may be a junit feature/problem after all but it does not seem to be the case:
```

public class Dummy {

    @Test
    public void runTest() throws Throwable {
        throw new Throwable();
    }

    @Test
    public void runAnotherTest() throws Throwable {
    }
}
```
Here `runTest` fails as expected:
![image](https://cloud.githubusercontent.com/assets/200952/15626133/d036b444-24bc-11e6-8479-f38a593d8fad.png)

sksamuel:
:+1: Sounds like I just need to change Exception to Throwable.

PR #3846, opened 2024-01-23
ContainDuplicates should print them (#20)

AlexCue987:
`containDuplicates` should list duplicates that violate the assumption, for instance:
```
           shouldThrowAny {
               listOf(1, 2, 3, 4, 2, 1) shouldNot containDuplicates()
            }.shouldHaveMessage("Collection should not contain duplicates, but has some: [1, 2]")
```

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/duplicates.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/duplicates.kt
@@ -1,5 +1,6 @@
 package io.kotest.matchers.collections
 
+import io.kotest.assertions.print.print
 import io.kotest.matchers.Matcher
 import io.kotest.matchers.MatcherResult
 import io.kotest.matchers.should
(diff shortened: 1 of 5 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/duplicates.kt
```diff
@@ -35,10 +36,20 @@ fun <T> Collection<T>.shouldNotContainDuplicates(): Collection<T> {
...
+         { "Collection should contain duplicates" },
+         {
+            "Collection should not contain duplicates, but has some: ${duplicates.print().value}"
+         })
+   }
 }
+
+internal fun<T> Collection<T>.duplicates(): Collection<T> = this.groupingBy { it }
+   .eachCount().entries
+   .mapNotNull { when(it.value) {
```
myuwono: Can you add a test for `List<T?>` with duplicated null entries? I wonder if this will detect that?

AlexCue987: > Can you add a test for `List<T?>` with duplicated null entries? I wonder if this will detect that?

done

myuwono (approved):

Issue #3850, opened 2024-01-27
Gradle plugin `io.kotest.multiplatform` breaks project configuration


hezean:
**Which version of Kotest are you using**

5.8.0

```kotlin
plugins {
    kotlin("multiplatform") version "1.9.20"  //<-- https://kotlinlang.org/docs/multiplatform-hierarchy.html#default-hierarchy-template
    id("com.android.library") version "8.2.1"
    id("io.kotest.multiplatform") version "5.8.0"
}

kotlin {
    /* Unresolved reference: androidTarget */
    androidTarget {
        compilations.all {
            kotlinOptions {
                jvmTarget = "17"
            }
        }
    }

    sourceSets {
        /*
Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
public fun Project.dependencies(configuration: DependencyHandlerScope.() -> Unit): Unit defined in org.gradle.kotlin.dsl
         */
        commonTest.dependencies {
            implementation(libs.bundles.kotest)
        }
    }
}
```

Hi, I'm trying to use Kotest in my KMM project (android & iOS target), but adding the kotest multiplatform in the shared module would break the Gradle build.

As for not using `android()`:

<img width="899" alt="Screenshot 2024-01-28 at 15 29 25" src="https://github.com/kotest/kotest/assets/49837965/333e2e2a-2088-46fa-95f5-169dfa4f897c">

<details>
<summary>Build Output</summary>
<pre>
> Configure project :shared
w: Missing 'androidTarget()' Kotlin target in multiplatform project 'shared (:shared)'.
The Android Gradle plugin was applied without creating a corresponding 'android()' Kotlin Target:

```
plugins {
    id("com.android.library")
    kotlin("multiplatform")
}

kotlin {
    androidTarget() // <-- please register this Android target
}
```

e: file:///Users/hezean/Downloads/MyApplication/shared/build.gradle.kts:8:5: Unresolved reference: androidTarget
e: file:///Users/hezean/Downloads/MyApplication/shared/build.gradle.kts:9:9: Unresolved reference: compilations
e: file:///Users/hezean/Downloads/MyApplication/shared/build.gradle.kts:10:13: Unresolved reference: kotlinOptions
e: file:///Users/hezean/Downloads/MyApplication/shared/build.gradle.kts:11:17: Unresolved reference: jvmTarget
[...]

PR #3856, opened 2024-02-01
Changing the Kotlin Gradle plugin dependncy to compileOnly

Kantis:
Using it as an implementation dependency seems to add a possibly stale version of the Kotlin plugin to the build path. Normally a user would already have a possibly different version of Kotlin pulled in, and these could clash (as seen in issue #3850)

Could this cause regressions that I'm not thinking of?


sksamuel:
Seems fine, the gradle stuff is on the classpath from gradle itself.

sksamuel (approved):

PR #3858, opened 2024-02-02
Mention that one might want a platform-specific module

reftel:
Being new to kotest (and Kotlin), we misinterpreted the tip about adding `kotest-framework-datatest` to mean we should have `io.kotest:kotest-framework-datatest:5.8.0` as a Maven dependency. Took an inordinate amount of head-scratching before we figured out the reason `withData` didn´t get resolved was that we should have used `io.kotest:kotest-framework-datatest-jvm:5.8.0` instead. I´m guessing other beginners might make the same mistake, so I think the tip should mention platform modules.

### DIFF
--- a/documentation/docs/framework/datatesting/data_driven_testing.md
+++ b/documentation/docs/framework/datatesting/data_driven_testing.md
@@ -5,7 +5,8 @@ slug: data-driven-testing.html
 ---
 
 :::tip Required Module
-Before data-driven-testing can be used, you need to add the module `kotest-framework-datatest` to your build.
+Before data-driven-testing can be used, you need to add the `kotest-framework-datatest` module for your platform,
+e.g. `kotest-framework-datatest-jvm`, to your build.
 :::
 
 
--- a/documentation/versioned_docs/version-5.2/framework/datatesting/data_driven_testing.md
+++ b/documentation/versioned_docs/version-5.2/framework/datatesting/data_driven_testing.md
@@ -5,7 +5,8 @@ slug: data-driven-testing.html
 ---
 
 :::tip Required Module
-Before data-driven-testing can be used, you need to add the module `kotest-framework-datatest` to your build.
+Before data-driven-testing can be used, you need to add the `kotest-framework-datatest` module for your platform,
+e.g. `kotest-framework-datatest-jvm`, to your build.
 :::
 
 
--- a/documentation/versioned_docs/version-5.3/framework/datatesting/data_driven_testing.md
+++ b/documentation/versioned_docs/version-5.3/framework/datatesting/data_driven_testing.md
@@ -5,7 +5,8 @@ slug: data-driven-testing.html
 ---
 
 :::tip Required Module
-Before data-driven-testing can be used, you need to add the module `kotest-framework-datatest` to your build.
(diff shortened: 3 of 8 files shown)


Kantis:
This is only relevant for Maven, and is always the case when you're adding a Maven dependency for a multiplatform library. I've tried to check how other multiplatform library mentions this caveat to Maven users..

### Projects with Maven samples
* [Arrow](https://arrow-kt.io/learn/quickstart/#include-the-dependencies) includes Maven docs, but doesn't include -jvm suffix for the dependency. Unsure of how/if this actually works.
* [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization?tab=readme-ov-file#maven) includes Maven docs, but doesn't include the -jvm suffix

### Projects w/o Maven samples
* [Ktor](https://ktor.io/docs/getting-started-ktor-client.html#add-dependencies) doesn't include anything on how to use Maven 
* [Koin](https://insert-koin.io/docs/quickstart/kotlin/#setup) doesn't have any maven samples
* [SQLDelight](https://cashapp.github.io/sqldelight/2.0.1/jvm_sqlite/) only includes Gradle samples.

## Conclusion
I think merging this PR would create equal confusion for Gradle users, since they actually shouldn't include `-jvm` in their dependency. Given that neither arrow nor kotlinx.serialization includes the suffix, I'm starting to wonder if there's something we could improve to remove the need to specify the suffix for Maven users.

Side note: [Kotlin docs](https://kotlinlang.org/docs/maven.html) don't mention anything about this either

reftel:
I´m a complete beginner, so I can´t really have an opinion what the right way of explaining this to users would be, but at least I´d say that not saying anything is not the right solution.

Some library is going to be a user´s first encounter with multiplatform modules. In our case it was kotest, and I would guess for many others as well - trying out Kotlin (with kotest) for test code first and production code if the experiment is deemed a success seems to me like a strategy many would take. So, I think it´s fair for the documentation to expect that quite a few readers are unfamiliar with how to correctly pull in a multiplatform module in their build.

As an aside, do you think it would be possible to somehow let the build break loudly if you include the non-jvm in a Maven project? Now, the library seemed to get included correctly, and the error message was simply "Unresolved reference", which is far too general for it to be useful for searching for information about the issue. If it would be possible to get it to fail with something more specific, then the story would just be that we googled an error message and copied a solution from SO =)

Kantis:
I asked around and found out there's a semi-hacky workaround we can perform that removes the need to specify the `-jvm` suffix for Maven users. I'll raise a PR with the workaround.

I also raised a [YouTrack issue](https://youtrack.jetbrains.com/issue/KT-65529/As-a-multiplatform-library-dev-I-want-the-root-variant-to-include-JVM-dependencies) to discuss if that solution could be made standard to avoid causing this kind of issue.

sksamuel:
Is it a big deal to include -jvm ?
Like if we make it clear in the docs ?

Kantis:
@reftel starting with next release, Maven should be able to resolvbe both `kotest-framework-datatest` and `kotest-framework-datatest-jvm`.

reftel:
Awesome! Thank you _very_ much for your prompt and thoughtful replies, and for what seems to me as the best possible solution =)
I´m closing this PR, since it´s (happily) now redundant.
Again, thank you!

PR #3859, opened 2024-02-02
Publishing JVM jar and pom in root variant

Kantis:
Solves the issue of Maven and Gradle <6.0 users being unable to resolve platform variants.

See discussion in https://youtrack.jetbrains.com/issue/KT-65529/As-a-multiplatform-library-dev-I-want-the-root-variant-to-include-JVM-dependencies and PR #3858


sksamuel (approved):

Kantis:
I've verified that resolution works for both root modules and jvm modules, so Maven users can migrate this gracefully.

E.g. both `kotest-runner-junit5` and `kotest-runner-junit5-jvm` works, starting with next release.

PR #3860, opened 2024-02-03
Fixing Arb.localDate when minDate == maxDate

### LINKED ISSUE #3829: LocalDate generator fails if minDate and maxDate are equal
When using the `localDate` generator, the following exception occurs if `minDate` and `maxDate` are equal:

```
Random range is empty: [0, 0).
java.lang.IllegalArgumentException: Random range is empty: [0, 0).
[... stack trace]
```

The exception comes from the code here:

https://github.com/kotest/kotest/blob/ce6b4feddf5364f3c6a9992d086c5229c50b046d/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt#L64C5-L64C5

A minimal example to reproduce the bug:

```kotlin
class BugExample : DescribeSpec({
  val failingGenerator : Arb<LocalDate> = Arb.localDate(minDate = LocalDate.MIN, maxDate = LocalDate.MIN)
  describe("The above generator") {
    it("Fails") {
      forAll(failingGenerator) { true }
    }
  }
})
```

And a more concrete example of real-world code that can make this happen, when the date generated by the first generator is equal to the default `minDate` :

```kotlin
class BugExample2 : DescribeSpec({
  val orderedDatesGenerator: Arb<Pair<LocalDate, LocalDate>> =
    Arb.localDate().flatMap { laterDate: LocalDate -> Arb.localDate(maxDate = laterDate).map { earlierDate -> Pair(earlierDate, laterDate) } }
  describe("The above generator") {
    it("sometimes fails, if laterDate is the default minDate") {
      forAll(failingGenerator) { true }
    }
  }
})
```

I'm not completely sure what the best way to address this is. Maybe before calling random, check if the range is empty and in that case return a constant ?

Kantis:
Now simply returns a constant Arb for the given date.

Fixes #3829

### DIFF
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
@@ -56,6 +56,7 @@ fun Arb.Companion.localDate(
    minDate: LocalDate = LocalDate.of(1970, 1, 1),
    maxDate: LocalDate = LocalDate.of(2030, 12, 31)
 ): Arb<LocalDate> {
+   if (minDate == maxDate) return Arb.constant(minDate)
 
    val leapYears = (minDate.year..maxDate.year).filter { isLeap(it.toLong()) }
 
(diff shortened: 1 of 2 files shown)


myuwono (approved):

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
```diff
@@ -56,6 +56,7 @@ fun Arb.Companion.localDate(
    minDate: LocalDate = LocalDate.of(1970, 1, 1),
    maxDate: LocalDate = LocalDate.of(2030, 12, 31)
 ): Arb<LocalDate> {
+   if (minDate == maxDate) return Arb.constant(minDate)
```
myuwono: Also i just realised maybe we should require maxdate >= mindate to fail early..

PR #3867, opened 2024-02-10
initial attempt at before and after hooks

### LINKED ISSUE #2792: Nested "before" lifecycle hook
I'm trying to use Kotest to write nested tests that have hooks that only run once per nest level.  This is how mocha works in JavaScript and how RSpec works in ruby.  In mocha the function is called `before`. I tried using `beforeContainer` but it runs once for *each* nested container instead of just once for the entire container including any nested containers within that parent container. This is hard to explain, so below is a set of tests that I would expect to pass, if a `before` method existed in kotest that behaved like mocha's or RSpec's.  In short, each `before` method should only be called once, and only when tests in its container run (e.g., the `before` method inside `"nested level 2"` shouldn't run until after the tests in `"nested level 1"` have run).

```kotlin
package foo.bar

class WidgetTest : DescribeSpec({
    var foo = "initial"
    var counterOne = 0
    var counterTwo = 0
    var counterThree = 0

    before {
        foo = "bar"
        counterOne++
    }

    it("foo should be bar") {
        foo shouldBe "bar"
        counterOne shouldBe 1
    }

    it("beforeSpec should have been called only once") {
        foo shouldBe "bar"
        counterOne shouldBe 1
    }

    it("counterTwo should be 0") {
        counterTwo shouldBe 0
    }

    it("counterThree should be 0") {
        counterThree shouldBe 0
    }

    describe("nested level 1") {
        before {
            foo = "buzz"
            counterTwo++
        }

        it("foo should be buzz") {
            foo shouldBe "buzz"
        }

        it("and counterOne should be 1") {
            counterOne shouldBe 1
        }

        it("and counterTwo should be 1") {
            counterTwo shouldBe 1
        }

        describe("nested level 2") {
            before {
                foo = "jazz"
                counterThree++
            }

            it("foo should be jazz") {
                foo shouldBe "jazz"
            }

            it("and counterOne should be 1") {
                counterOne shouldBe 1
            }

            it("and counterTwo should be 1") {
                counterTwo shouldBe 1
            }

            it("and counterThree should be 1") {
                counterThree shouldBe 1
            }
        }
    }
})
```

I can get close to what I want if I make the first `before` a `beforeSpec` and the other two `beforeContainer` (and move them above each nested container), but then `counterTwo` gets incremented twice: once for `"nested level 1"` and once for `"nested level 2"`.  I'm not sure if I'm just doing something wrong or if this capability simply doesn't exist in kotest.

A somewhat more realistic example:

```kotlin

class WidgetApiTest : DescribeSpec({
    describe("CRUD operations") {
        var widget: Widget? = Widget(0, "")

        describe("when we create a widget") {
            before { // This should run exactly once
                createWidget(1, "widget-a")
            }

            describe("when we fetch the widget") {
                before { // This should run exactly once
                    widget = getWidget(1) // assume this is a slow HTTP or database request
                }

                it("the widget should not be null") {
                    widget shouldNotBe null
                }

                it("and the widget should have a name") {
                    widget!!.name shouldBe "widget-a"
                }

                describe("when we update the widget's name") {
                    before { // the above "before" functions should not have run again in this test container
                        widget?.let { w ->
                            w.name = "new-name"
                            updateWidget(w)
                        }
                    }

                    it("then the re-fetched widget should be updated") {
                        widget = getWidget(1) // assume this is a slow HTTP or database request
                        widget!!.name shouldBe "new-name"
                    }
                }
            }
        }
    }
})

data class Widget(val id: Int, var name: String?)

val widgetContainer: HashMap<Int, Widget> = hashMapOf()

suspend fun createWidget(id: Int, name: String?): Widget {
    delay(200) // slow HTTP request or Database call
    val widget = Widget(id, name)
    widgetContainer[id] = widget
    return widget
}

suspend fun getWidget(id: Int): Widget? {
    delay(200) // slow HTTP request or Database call
    return widgetContainer[id]
}

suspend fun updateWidget(widget: Widget): Widget {
    delay(200) // slow HTTP request or Database call
    val id = widget.id
    widgetContainer[id] = widget
    return widget
}
```

dondod:
This PR is an attempt to fix #2792.  The `before` hook works, but I'm having trouble figuring out an approach for the `after` hook.  I have a failing test for the `after` hook.  In the current code, only the top-level `after` hook runs, so the assertion fails in that hook.  It seems that I can either get all the `after` hooks to run, but run more than once (like `afterEach`) _or_ I can get only one to run at all.  

Any ideas on how to get each `after` hook to run exactly once, and starting with the deepest-nested hook?


Kantis:
@dondod I adjusted your implementation slightly. Check my commit and see if it aligns with your interpretation of the issue.

Kantis:
@dondod api_check is failing, since the ABI doesn't match the dump. Please run `./gradlew apiDump` and commit the updated `.api` file(s).

See https://github.com/kotlin/binary-compatibility-validator/ for more info

dondod:
Thanks @Kantis - your changes do work.  So simple too.. 

I'll run the `apiDump` task and commit.

dondod:
Hi @Kantis - I don't see what is causing the test_linux jvmTest failure.  Anything I need to change?

Kantis:
What I think is happening is that one of the tests is using a Kotest extension that has been built using an older version of Kotest. That version of Kotest's `TestListener` did not include the newly added `BeforeListener`, and thus it hasn't inherited the default implementation for that method.

Either we need to push this feature to 6.0 and do a breaking change and update all extensions (including third-party ones), or we must introduce a new interface and deprecate the old one. @sksamuel are there other ways to fix this?

dondod:
Hi @Kantis - just wondering if there has been any more traction on this.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
```diff
@@ -171,6 +173,36 @@ interface ContainerScope : TestScope {
...
    }
 
+   fun before(f: Before) {
+      if (hasChildren()) throw outOfOrderCallbacksException
+      val thisTestCase = this.testCase
+      var hasRun = false
+
+      appendExtension(object : TestListener {
+         override suspend fun before(testCase: TestCase) {
+            if (!hasRun && thisTestCase.descriptor.isAncestorOf(testCase.descriptor)) {
```
sksamuel, line 183: What's the purpose of the hasRun ?

sksamuel:
I can take a look. I added a comment too.

sksamuel:
the other alternative, which I did here: https://github.com/kotest/kotest/pull/4285 (which I am happy to close in favour of this) was to have after just delegate to afterContainer and check that it's only running at the level where it's defined.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
dondod: That was to prevent the `before` hook from running more than once.  There's likely a better way to do that..

dondod:
@sksamuel - #4285 would be fine, if it works like this: https://github.com/kotest/kotest/pull/3867/files#diff-eda4bc7972601e1ea67e2dc389bfc13ec4746a6f3b664f17d2e70d7204a27a2eR79. (in addition to your tests: https://github.com/kotest/kotest/pull/4285/files#diff-f74e15c3f3598fdb48f5c42965c6cc56e41c27dd517fbcb1d78d762881dc10e6R54)

sksamuel:
Should I add your tests to mine or do you want to adjust yours?

dondod:
You can just add mine if that's ok with you. Or modify as you wish..

sksamuel:
Replaced by https://github.com/kotest/kotest/pull/4285

PR #3886, opened 2024-02-19
more detailed failure for startWith and endWith

AlexCue987:


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/StartWithEndWithTest.kt
```diff
@@ -46,12 +52,18 @@ class StartWithEndWithTest : WordSpec() {
          "print errors unambiguously"  {
             shouldThrow<AssertionError> {
                listOf(1L, 2L, 3L, 4L) should endWith(listOf(1L, 3L))
-            }.shouldHaveMessage("List should end with [1L, 3L] but was [3L, 4L]")
+            }.shouldHaveMessage("""
+               |List should end with [1L, 3L] but was [3L, 4L]
+               |Mismatched elements: value[2] != 1L, value[3] != 3L
+               """.trimMargin())
```
Kantis: Could we do something like:

```suggestion
            }.shouldHaveMessage("""
               |List should end with [1L, 3L] but was [3L, 4L]
               |The following elements failed: 
               |  [2] 3 => expected <1>, but was <3>
               |  [3] 4 => expected <3>, but was <4>
               """.trimMargin())
```

This would be much more consistent with how Inspectors failures look like. See `InspectorsTest` for  examles

Kantis (changes requested):

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/StartWithEndWithTest.kt
AlexCue987: done

Kantis (approved):

Issue #3857, opened 2024-02-02
Outdated badges in docs showing old snapshot version v5.6.0...-SNAPSHOT


MeikPiepmeyer:
There are a few places where the Kotest docs link to badges of the current stable version of Kotest und the latest snapshot version. The lastest snspshot version seems outdated, linking to a version 5.6.0<...>-SNAPSHOT The link from here https://kotest.io/docs/framework/framework.html goes to https://oss.sonatype.org/content/repositories/snapshots/io/kotest/ where the artifacts seem rather old.

This seems to be the case in the whole docs where the badge got included:
* At the former listed introduction of Kotest: https://kotest.io/docs/framework/framework.html
* At Kotest assertions: https://kotest.io/docs/assertions/assertions.html

This is independent of the version, meaning: This is the same when viewing the docs of Kotest 5.8 or the coming 5.9.

Kantis:
We migrated to a new Sonatype repository a while back which caused the badges to break. They should refer to s01.oss.sonatype.org now.

PR welcome 😄

ddaakk:
Hello, I would like to be assigned to the specific issue.

PR #3888, opened 2024-02-21
Fix the snapshot version badge in index markdowns

ddaakk:
#3857 
Started from this issue.

Kotest migrated to a new Sonatype repository a while back which caused the badges to break. 
it changed to s01.oss.sonatype.org now.


Kantis:
Thanks Martin! Looks good. However, it only fixes the _next_ version docs. If you don't mind, I think it would also be nice to update the versioned docs for at least v5.8 as well. See the PR template comment for details.

graceful-martin:
I've also updated the v5.8 documentation. Appreciate your response.

ddaakk:
> Thanks Martin! Looks good. However, it only fixes the _next_ version docs. If you don't mind, I think it would also be nice to update the versioned docs for at least v5.8 as well. See the PR template comment for details.

I've also updated the v5.8 documentation. Appreciate your response.

Kantis (approved):

PR #3891, opened 2024-02-22
Remove unnecessary delay when retry and fix error message

### LINKED ISSUE #3890: retry has unnecessary delay
5.8.0

The [retry](https://github.com/kotest/kotest/blob/01bd98ff10102c51b5a80b51fb375c68264f627d/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/Retry.kt#L63) function is like this:

```
while (end.hasNotPassedNow() && retrySoFar < config.maxRetry) {
    ...
      retrySoFar++
      delay(nextAwaitDuration)
      nextAwaitDuration *= config.multiplier
}
```

This causes `retry` to make one more `delay` than is necessary. It would make tests faster (i.e. not unnecessarily slower) if we exit early using `break` or with an `if`:

```
while (end.hasNotPassedNow() && retrySoFar < config.maxRetry) {
    ...
    retrySoFar++
    if (end.hasNotPassedNow() && retrySoFar < config.maxRetry) {
        delay(nextAwaitDuration)
        nextAwaitDuration *= config.multiplier
    }
}
```

kshired:
resolve #3890

Pre-check the number of retries to eliminate additional delays

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/Retry.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/Retry.kt
@@ -85,7 +85,7 @@ suspend fun <T> retry(
          // else ignore and continue
       }
       retrySoFar++
-      if (retrySoFar >= config.maxRetry) break
+      if (retrySoFar >= config.maxRetry || end.hasPassedNow()) break
       delay(nextAwaitDuration)
       nextAwaitDuration *= config.multiplier
    }
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/Retry.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/Retry.kt
@@ -85,6 +85,7 @@ suspend fun <T> retry(
          // else ignore and continue
       }
       retrySoFar++
+      if (retrySoFar >= config.maxRetry) break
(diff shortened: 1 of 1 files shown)


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/Retry.kt
```diff
@@ -85,6 +85,7 @@ suspend fun <T> retry(
          // else ignore and continue
       }
       retrySoFar++
+      if (retrySoFar >= config.maxRetry) break
```
klitoskyriacou: Why only checking `retrySoFar`? It could also be that the end time has passed, even if fewer than max retries have been taken.

klitoskyriacou: Also note the failure message is misleading: `Test failed after ${config.delay.toLong(DurationUnit.SECONDS)} seconds`. Here, `delay` is only the initial delay. It ignores the multiplier, and it ignores how many times we've delayed.

kshired: @klitoskyriacou 

Thank you for review. I've revised code to reflect your code review.

klitoskyriacou (approved):

klitoskyriacou (approved):

Kantis (approved):

Issue #36, opened 2016-06-17
Support Sets in properties natively


sksamuel:
See #35

PR #3892, opened 2024-02-23
override suspend fun beforeTest (#36)

AlexCue987:
`beforeTest` is now a `suspend` function - fix the docs accordingly


sksamuel:
ty

PR #3894, opened 2024-02-24
Quote expected and actual exception messages in assertions

sschuberth:
This makes the output more readable if the exception messages are full sentences.

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/throwables/CovariantThrowableHandling.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/throwables/CovariantThrowableHandling.kt
@@ -63,7 +63,7 @@ inline fun <reified T : Throwable> shouldThrowUnitWithMessage(message: String, b
    shouldThrowUnit<T>(block).let {
       when (it.message) {
          message -> it
-         else -> throw failure("Expected exception message $message but was ${it.message} instead.", it)
+         else -> throw failure("Expected exception message '$message' but was '${it.message}' instead.", it)
       }
    }
 
@@ -212,7 +212,7 @@ inline fun <reified T : Throwable> shouldThrowWithMessage(message: String, block
    shouldThrow<T>(block).let {
       when (it.message) {
          message -> it
-         else -> throw failure("Expected exception message $message but was ${it.message} instead.", it)
+         else -> throw failure("Expected exception message '$message' but was '${it.message}' instead.", it)
       }
    }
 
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/throwables/CovariantThrowableHandling.kt
```diff
@@ -63,7 +63,7 @@ inline fun <reified T : Throwable> shouldThrowUnitWithMessage(message: String, b
    shouldThrowUnit<T>(block).let {
       when (it.message) {
          message -> it
-         else -> throw failure("Expected exception message $message but was ${it.message} instead.", it)
+         else -> throw failure("Expected exception message '$message' but was '${it.message}' instead.", it)
```
Kantis: Switch to the failure-function which takes 3 params. It'll format the message so that we get `Click to see difference` option when it fails. 

```suggestion
         else -> throw failure(message, it.message, "Unexpected exception message: ")
```

Kantis (changes requested):

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/throwables/CovariantThrowableHandling.kt
sschuberth: I'm not sure how useful a diff on exception messages really is... anyway, I guess the `prependMessage` should not end with ":" but ".".

sschuberth: Also, that function takes arguments of types `Expected` / `Actual`, not strings.

Kantis: Wouldn't it be useful at least occasionally for finding typos and such? 

I think you could do `failure(Expected(message.printed()), Actual(it.message.printed()), "Unexpected exception message.")`

sschuberth: > Wouldn't it be useful at least occasionally for finding typos and such?

Maybe... I've added a new commit for it.

> anyway, I guess the `prependMessage` should not end with ":" but ".".

Actually, after seeing how it looks like in the test, I changed my mind and used your original wording.

Kantis (approved):

PR #3898, opened 2024-02-27
specify module name kotest-assertions-json

jlous:
Added import instructions to doc.  Not impossible to guess the full module name, but the docs should make it explicit. Some other modules are missing this too.


Kantis (commented):
Please add it to `documentation/docs/assertions/json/overview.md` as well

PR #3899, opened 2024-02-27
Don't consider kotlinx and javax types to be builtin

### LINKED ISSUE #3855: Kotest doesn't compare JAXBElement properly
5.8.0

When using shouldBeEqualToComparingFields for JAXBElement fields doesn't compare properly. Because the JAXBElement is marked as built in, but it doesn't have equals method by default

Kantis:
Fixes #3855


sksamuel:
We're probably better off just having a list of types we do support, and expand on that.
But I fear it'll break too many things.
Let's add an exemption for jaxb only for now, and then change it properly for 6.0 ?

Kantis:
> We're probably better off just having a list of types we do support, and expand on that. But I fear it'll break too many things. Let's add an exemption for jaxb only for now, and then change it properly for 6.0 ?

Yeah sounds good. :)

sksamuel:
@Kantis this will check only for jaxb, but I've disabled the tests since we're getting module errors. JAXBElement is not accessible, so we can't actually compare this way, but at least now people will get an error if they try.

sksamuel (approved):

PR #3900, opened 2024-02-29
Add includeFirst option to EventuallyConfiguration

### LINKED ISSUE #3889: EventuallyConfiguration should allow capturing only the lastError
### Problem

- Android Studio works really badly with a long Assertion Error stacktrace (you actually could see some un-decrypted string);
- Long stacktrace increases risk of OutOfMemoryError.

### Solution
- Add to `EventuallyConfiguration` new `includeFirst: Boolean = true` option, so we could turn it off. Really, most of the time we don't care about the first error - if tests fails, we probably want to know the last one. 
- Also some optimization could be done by comparing first and last error at `buildFailureMessage` -> if those are equal, there is no need to duplicate them in a stacktrace.

kin4496:
This PR resolves #3889

add includefirst option to EventuallyConfiguration data class


Kantis:
Please include tests also

Issue #43, opened 2016-06-24
Warning stops test execution


helmbold:
If I run all tests of the KotlinTest project, the execution (in IntelliJ 2016.1.3) stops with 109 of 110 tests. It appears the message:

> [WARN] When comparing doubles consider using tolerance, eg: a shouldBe b plusOrMinus c

![error in IntelliJ](http://i.imgur.com/gLoNPwe.png)

sksamuel:
I don't think the warn is anything to do with that, because that's just being emitted when I test plusOrMinus in `DoubleMatchersTest`. I get something similar, but all tests pass. I think perhaps the count is just off.

sksamuel:
Ah the issue is just that there is an ignored test in ConfigTest. It's there just to make sure the dsl doesn't change and for a visible check. I'm not sure how to check that an ignored test didn't actually run.

helmbold:
How does JUnit gets the number of tests? Maybe ignored tests can be subtracted there.

sksamuel:
That's certainly easy enough to do, but then it doesn't appear in the list.
What do you think is better.

helmbold:
It is good to see (and count) ignored tests.However there is some difference between ignored JUnit tests and ignored KotlinTest tests.

JUnit counts all tests and ignored tests:
![test count in JUnit](http://i.imgur.com/0YoYqk7.png)

KotlinTest doesn't count ignored tests:
![test count with KotlinTest](http://i.imgur.com/K07uRjg.png)

I have no idea how to do it, but I'd prefer the JUnit behavior and view.

sksamuel:
Agreed. I'll figure it out while you're doing the config stuff.

sksamuel:
Sorted

PR #3907, opened 2024-03-04
Contains in order more details draft (#43)

### LINKED ISSUE #3881: containAllInAnyOrder fails to detect mismatch in element counts, as long as the overall counts of elements match
**latest from master**
Presumably `containAllInAnyOrder` should detect mismatch in element counts, and sometimes it does, as follows:
```
        "detect count mismatch" {
            sequenceOf("apple", "apple", "orange") should containAllInAnyOrder(
                sequenceOf("apple", "apple", "orange", "orange")
            )
        }
-----------
Sequence should contain the values of [apple, apple, orange, orange] in any order, but was [apple, apple, orange]
```
However, the following test passes, which seems counter-intuitive to me:
```
        "fail to detect count mismatch" {
            sequenceOf("apple", "apple", "orange") should containAllInAnyOrder(
                sequenceOf("apple", "orange", "orange")
            )
        }
```
Interestingly, when I tried to fix the issue by comparing counts of elements, it broke two existing tests, as follows:
```
        succeed("for subset, same count with nulls") {
            sampleData.sparse.shouldContainAllInAnyOrder(sampleData.nulls)
         }
----
Sequence should contain the values of [null, null, null, null] in any order, but was [null, null, null, 3].
----
         fail("for subset, same count with nulls") {
            sampleData.sparse.shouldNotContainAllInAnyOrder(sampleData.nulls)
         }
------
Expected exception java.lang.AssertionError but no exception was thrown.
// this test matches [null, null, null, null] with [null, null, null, 3] - these seems different to me and as such the test should pass
```
My understanding is that these two tests are not validating correct behavior, and as such they should be fixed.
My understanding is that these two tests should not be passing in the first place.
If this is correct, I have a fix and can raise a PR.

AlexCue987:
fix the following bug - the test below was passing:
```
 "fail to detect count mismatch" {
            sequenceOf("apple", "apple", "orange") should containAllInAnyOrder(
                sequenceOf("apple", "orange", "orange")
            )
        }
```
* also fix two tests named "for subset, same count with nulls" - they were not validating correct behavior.
* also provide more details about what exactly did not match, such as:
```
Sequence should contain the values of [1, 1, 2] in any order, but was [1, 2, 2].
Count Mismatches: 
  For 1: expected count: <2>, but was: <1>
  For 2: expected count: <1>, but was: <2>
```

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/UnorderedCollectionsDifference.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/UnorderedCollectionsDifference.kt
@@ -1,0 +1,66 @@
+package io.kotest.matchers.sequences
+
+import io.kotest.assertions.print.print
+
+internal data class UnorderedCollectionsDifference<T>(
+   val missingElements: Set<T>,
+   val extraElements: Set<T>,
+   val countMismatches: List<CountMismatch<T>>
+) {
+   private val describedMismatches = sequenceOf(
+      DescribedMismatch(missingElements, "Missing Elements"),
+      DescribedMismatch(extraElements, "Extra Elements"),
+      DescribedMismatch(countMismatches, "Count Mismatches")
+   )
+
+   fun isMatch(): Boolean {
+      return describedMismatches.all { it.elements.isEmpty() }
+   }
+
+   override fun toString(): String {
+      return if (isMatch()) "" else "\n" + describedMismatches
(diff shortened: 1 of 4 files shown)


AlexCue987:
the corresponding issue: https://github.com/kotest/kotest/issues/3881

LeoColman (approved):

AlexCue987:
@LeoColman can you help me understand what exactly went wrong? All I've managed to see at this itme is "13 successful and 1 skipped checks".

LeoColman:
Mac tests failed and I just reran them

PR #3915, opened 2024-03-10
Add support for running private classes

### LINKED ISSUE #3879: Private Test Classes
JUnit supports execution of private Kotlin test classes. For some reason IntelliJ dropped support for that, although it supported it in the past. 😝 The Kotest IntelliJ plugin supports private Kotlin test classes (all the necessary gutter icons are drawn), but Kotest with the JUnit runner fails. It complains that it cannot find any tests with that name.

Now, this could be a bug report for the Kotest IntelliJ plugin, just asking for not drawing the gutter icons when the class is private. Like IntelliJ does it since a while for JUnit. I'm not asking for that, because I believe that test classes should be private. They shouldn't show up in auto-completion, and they shouldn't be reusable.

sksamuel:
Closes #3879

### DIFF
--- a/documentation/docs/framework/config_props.md
+++ b/documentation/docs/framework/config_props.md
@@ -37,6 +37,9 @@ object KotestEngineProperties {
     */
    const val filterSpecs = "kotest.filter.specs"
 
+   /**
+    * If set to "true" then private classes (specs) will be included in the test suite.
+    */
    const val includePrivateClasses = "kotest.include.private.classes"
 
    const val propertiesFilename = "kotest.properties.filename"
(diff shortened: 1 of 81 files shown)


Kantis:
WDYT about just making this always enabled? I'm having a hard time seeing in what scenario someone might want to both define private `Spec`s _and_ disable this setting

sksamuel:
> WDYT about just making this always enabled? I'm having a hard time seeing in what scenario someone might want to both define private `Spec`s _and_ disable this setting

I think my reason was that we have private tests in kotest we don't want to execute.
But you're probably right that private should be on by default.
Let's do that for 6.0 and we can add an internal annotation to hide individual classes, as well as a setting.
I'll redo this PR for 6.0

PR #3920, opened 2024-03-10
Arb.distinct() should work for edgecases

### LINKED ISSUE #3861: Generator created by Arb.distinct() may still return duplicates if those come from edge cases
Setup: `Arb.string(codepoints = Codepoint.az()).distinct()` this generator still returns many strings `"a"`. This might be the 'expected' behavior, but it is not according to my expectations, and I believe not according to expectations of 'common sense'.

What I would expect, is that generators created by `.distinct()` **not** generate each value more than once (for a given run), irrespective of whether the value is generated completely randomly, or comes from one of the edge cases. While it is true that edge cases in general may be repeated multiple times for a given run, I think it makes sense to expect _unique_ values in this specific case when user indicated that they want distinct values.

So either the implementation of `distinct()` should be fixed, or its documentation should clearly indicate that edge cases are not part of the filtering (and it should better provide a good explanation of why).

**Which version of Kotest are you using** 5.8.0

sksamuel:
Fixes #3861

### DIFF
--- a/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/DistinctTest.kt
+++ b/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/DistinctTest.kt
@@ -3,23 +3,22 @@ package com.sksamuel.kotest.property.arbitrary
 import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.common.DelicateKotest
 import io.kotest.core.spec.style.FunSpec
+import io.kotest.matchers.collections.shouldBeUnique
 import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.shouldBe
 import io.kotest.property.Arb
 import io.kotest.property.EdgeConfig
 import io.kotest.property.RandomSource
 import io.kotest.property.arbitrary.arbitrary
-import io.kotest.property.arbitrary.arbitraryBuilder
 import io.kotest.property.arbitrary.distinct
 import io.kotest.property.arbitrary.int
 import io.kotest.property.arbitrary.take
-import kotlin.random.Random
 
 @OptIn(DelicateKotest::class)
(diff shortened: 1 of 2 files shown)


kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/DistinctTest.kt
```diff
@@ -1,24 +1,30 @@
...
 
    test("with enough entropy distinct should return required count") {
-      Arb.int(0..100000).distinct().take(100).toSet().shouldHaveSize(100)
+      Arb.int(0..100000).distinct().take(100).toList().shouldHaveSize(100)
    }
 
    test("without enough entropy distinct should throw") {
       shouldThrow<NoSuchElementException> {
-         Arb.int(1..9).distinct().take(10).toSet().shouldHaveSize(10)
+         Arb.int(1..9).distinct().take(10).toList().shouldHaveSize(10)
```
Kantis, line 26: Why did this change?

sksamuel: Because using to set would mask if distinct wasn't working

Kantis: Isn't it the other way around? 

If `distinct().take(2)` returns `[1, 1]` and we do `.toList() shouldHaveSize 2` it feels like distinct is broken but test would still pass.

sksamuel: Yeah it needs both, because toSet() will mask duplicates, and toList() will count duplicates.
So I've changed it to use shouldBeUnique() and shouldHaveSize to cover both

Issue #3911, opened 2024-03-07
Simplify shouldMatchInOrder and shouldMatchEach


gianninia:
The extension functions `shouldMatchInOrder` and `shouldMatchEach` are great to assert collections. However, the signature could be a bit improved in my view. Intuitively, if I want to customize the assertion of the collection elements, I will want to pass the collections and some lambda that receives the actual and expected elements to be asserted.

I mean a signature like this:
```kotlin
fun <T> List<T>.shouldMatchInOrder(expected: List<T>, asserter: (T, T) -> Unit) {}
fun <T> List<T>.shouldMatchEach(expected: List<T>, asserter: (T, T) -> Unit) {}
```

This allows you to do something like this:
```kotlin
val expectedList = listOf(-1, -2, -3)
val actualList = listOf(1, 3, 4, 5)

actualList.shouldMatchEach(expectedList) { actual, expected ->
    actual shouldBe -expected
}
```

Instead of:
```kotlin
    actualList.shouldMatchEach(expectedList.map { actual ->
        { expected ->
            actual shouldBe -expected
        }
    })
```

I mean, looking at both variants when they are there does feel kind of obvious and not so different from each other. But if you have to write something like this for the first time, it does require some elaboration until you get the twist of it, as you don't see this immediately by just looking at the signature. I believe with my suggested signatures it's immediately clear what it's supposed to do.

In my project, I have added the following extension functions to solve this problem. Would you maybe consider adding this to the library?
```kotlin
fun <T> List<T>.shouldMatchInOrder(expected: List<T>, asserter: (T, T) -> Unit) {
    this.shouldMatchInOrder(expected.map { actualElement ->
        { expectedElement: T ->
            asserter(actualElement, expectedElement)
        }
    })
}

fun <T> List<T>.shouldMatchEach(expected: List<T>, asserter: (T, T) -> Unit) {
    this.shouldMatchEach(expected.map { actualElement ->
        { expectedElement: T ->
            asserter(actualElement, expectedElement)
        }
    })
}
```

I also have overloads for Sequences:
```kotlin
fun <T> Sequence<T>.shouldMatchInOrder(expected: Sequence<T>, asserter: (T, T) -> Unit) = toList().shouldMatchInOrder(expected.toList(), asserter)
fun <T> Sequence<T>.shouldMatchEach(expected: Sequence<T>, asserter: (T, T) -> Unit) = toList().shouldMatchEach(expected.toList(), asserter)
```
[...]

PR #3929, opened 2024-03-11
Add simplified overloads to shouldMatchEach extensions (#3911)

gianninia:
* Added matchEach matcher that receives an expected list and an asserter to verify elements from both lists
* Added shouldMatchEach extensions for List, Iterable and Sequence


sksamuel:
What do you think @Kantis  ?

Kantis:
I think it looks useful, I like it 🙂

Kantis (approved):

PR #3939, opened 2024-03-15
add-similarity-to-contain (#53)

AlexCue987:
add search for similarity to `contain` for example:
```
         "print a similar key when no exact match" {
               mapOf(sweetGreenApple to 1, sweetRedApple to 2) should contain(sweetGreenPear, 1)
            }.message
            message shouldContain """
               | expected: Fruit(name=apple, color=green, taste=sweet),
               |  but was: Fruit(name=pear, color=green, taste=sweet),
               |  The following fields did not match:
               |    "name" expected: <"apple">, but was: <"pear">
            """.trimMargin()
         }
         "print entries with same value" {
               mapOf(sweetGreenApple to 1, sweetRedApple to 2) should contain(sweetGreenPear, 1)
            }.message
            message shouldContain """
               |Same value found for the following entries: [Fruit(name=apple, color=green, taste=sweet)=1]
            """.trimMargin()
         }
         "print entries with similar values" {
               mapOf(1 to sweetGreenApple, 2 to sweetRedApple) should contain(3, sweetGreenPear)
            }.message
            message shouldContain """
               | expected: Fruit(name=apple, color=green, taste=sweet),
               |  but was: Fruit(name=pear, color=green, taste=sweet),
               |  The following fields did not match:
               |    "name" expected: <"apple">, but was: <"pear">
            """.trimMargin()
         }
```


Kantis:
I feel it's kind of ambiguous what we're looking at in the failure messages. Was it the wrong key or value?

Perhaps we should produce a similarity-score for each Map.Entry and just print the closest one, taking both key and value into consideration?

AlexCue987:
> I feel it's kind of ambiguous what we're looking at in the failure messages. Was it the wrong key or value?
> 
> Perhaps we should produce a similarity-score for each Map.Entry and just print the closest one, taking both key and value into consideration?

I like the idea of finding the most similar Map.Entry, and I'll do that, and output it first. But I perceive good value in printing other distinct cases that may be helpful, such as "there is a very similar value with a completely different key".

AlexCue987:
> I feel it's kind of ambiguous what we're looking at in the failure messages. Was it the wrong key or value?
> 
> Perhaps we should produce a similarity-score for each Map.Entry and just print the closest one, taking both key and value into consideration?

At this time I know how to take "both key and value into consideration" only when they both are data classes. But for the cases when only the key or only the value is a data class, IMO we should keep the existing logic.

So let's merge this as is, and add support for taking "both key and value into consideration" later on. Sounds like a good plan?

sksamuel (approved):
Sounds like a plan

PR #3940, opened 2024-03-16
fix(docs): indentation in System Extensions

craigpastro:
This PR updates some of the indentations in the system extensions files in the docs. There is a mix of two space and four space indentations in the docs, but as Kotlin seems to recommend four spaces this is what I went for here.

No content has changed. I also didn't update all relevant versions. At least it will look nice for the next release.


sksamuel:
ty

PR #3943, opened 2024-03-18
Change Arb.filterIsInstance() signature to use single type parameter

### LINKED ISSUE #3942: `filterIsInstance()` for `Arb` requires both type arguments compared to stdlib `filterIsInstance()`
- 5.8.0

```kotlin
    @Test
    fun `contrived example`() = runTest { 
        val intArb = Arb.int()
        val longArb = Arb.long()
        val numberArb: Arb<Number> = intArb.merge(longArb)
        // 2 type arguments expected for inline fun <A, reified B : A> Arb<A>.filterIsInstance(): Arb<B
        // val derivedIntArb = numberArb.filterIsInstance<Int>()
        val derivedIntArb = numberArb.filterIsInstance<_, Int>()
        
        checkAll(derivedIntArb) { n -> 
            // ...
        }
    }
```

The type signature today is [`inline fun <A, reified B : A> Arb<A>.filterIsInstance()`](https://github.com/kotest/kotest/blob/0315867c89dae3aab54827fc558fd69c18f559f8/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/filter.kt#L42) which requires the call-site provide both base and derived type. Since `_` can be used as a type that can be inferred, this can be done fairly simply but it still doesn't mimic the convention currently built into the standard library

This compared to the [standard library `filterIsInstance()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/filter-is-instance.html) variants shows the slight difference in definition and usage.

```kotlin
fun <reified R> Array<*>.filterIsInstance(): List<R>
fun <reified R> Iterable<*>.filterIsInstance()
```

So, it looks like the tradeoff here is either 1 of better type safety or 1 or simplifying usage. Potentially, with compiler improvements (like if [KT-13394](https://youtrack.jetbrains.com/issue/KT-13394/) is fully supported or when context receivers are moved forward) could "fix" this.

I wanted to open this as I didn't see anything around it in other issues or the history.

kshired:
resolve #3942 

Change Arb.filterIsInstance() signature to use single type parameter.

This change is **breaking change.** But there's no way to add a new function, so I change the existing signature.

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/filter.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/filter.kt
@@ -6,7 +6,6 @@ import io.kotest.property.RTree
 import io.kotest.property.RandomSource
 import io.kotest.property.Sample
 import io.kotest.property.filter
-import kotlin.jvm.JvmName
 
 /**
  * Returns a new [Arb] which takes its elements from the receiver and filters them using the supplied
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/filter.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/filter.kt
@@ -6,6 +6,7 @@ import io.kotest.property.RTree
 import io.kotest.property.RandomSource
 import io.kotest.property.Sample
 import io.kotest.property.filter
+import kotlin.jvm.JvmName
 
 /**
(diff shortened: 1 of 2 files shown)


Kantis:
apiDump needs to be updated

kshired:
@Kantis I did. But there's no change. Also, `apiCheck` action was passed.

LeoColman:
Hmm... Interesting.

I believe it makes sense as Generics don't really exist, so the signature is kept the same

LeoColman (approved):

PR #3949, opened 2024-03-21
Better assertion failures for containJsonKeyValue

Kantis:
Now prints whether the key was missing, or if it had wrong value

### DIFF
--- a/kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/ContainJsonKeyValueTest.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/ContainJsonKeyValueTest.kt
@@ -37,6 +37,15 @@ class ContainJsonKeyValueTest : StringSpec({
       }
    """.trimIndent()
 
+   "Negated assertions" {
+      "".shouldNotContainJsonKeyValue("$.store.bicycle.color", "red")
+      "{}".shouldNotContainJsonKeyValue("$.store.bicycle.color", "red")
+      """{ "foo": "bar" }""".shouldNotContainJsonKeyValue("foo", "baz")
+      shouldFail {
+         """{ "foo": "bar" }""".shouldNotContainJsonKeyValue("foo", "bar")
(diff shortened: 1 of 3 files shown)


kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/ContainJsonKeyValueTest.kt
```diff
@@ -0,0 +1,89 @@
...
+      shouldFail {
+         """{ "foo": "bar" }""".shouldNotContainJsonKeyValue("foo", "bar")
+      }.message shouldBe """{ "foo": "bar" } should not contain the element foo = bar"""
+   }
+
+   "Failure message states if key is missing, when it's missing" {
+      shouldFail {
+         json.shouldContainJsonKeyValue("$.bicycle.engine", "V2")
+      }.message shouldBe """
+         Expected given to contain json key <'$.bicycle.engine'> but key was not found.
```
AlexCue987, line 53: Would it make sense to print partial matches, such as:
```
Expected given to contain json key <'$.store.book.weight'> but key was not found. 
Found the following sub-key: <'$.store.book'>.
```

Kantis: I suppose it would, but I don't believe we get that information out of the JsonPath library. How do you intend that we should calculate that subkey?

Kantis: I think perhaps we can merge this small improvement as-is, and we could make further improvements in a separate PR :) We can continue the discussion here even after PR is merged

AlexCue987: > I think perhaps we can merge this small improvement as-is, and we could make further improvements in a separate PR :) We can continue the discussion here even after PR is merged

sure - https://github.com/kotest/kotest/pull/4013

PR #3956, opened 2024-03-29
Update Sonatype snapshot URL

aSemy:
I couldn't resolve the latest Kotest snapshot release from https://oss.sonatype.org/content/repositories/snapshots. However, https://s01.oss.sonatype.org/content/repositories/snapshots worked.

This is a quick edit via the GitHub web gui. Do other parts of the docs need to be updated too?


OliverO2:
Whoa, there are snapshots! Why did I never read that page all the way down? Such a time saver.

sksamuel:
thanks!

PR #3968, opened 2024-04-11
Fix json assertions artifact coordinate in docs

TWiStErRob:
Affecting: https://kotest.io/docs/assertions/json/json-overview.html

https://mvnrepository.com/artifact/io.kotest/kotest-assertions-json, the other one apparently never existed (according to mvnrepository).

Fix https://github.com/kotest/kotest/pull/3898 cc @jlous


sksamuel (approved):

sksamuel:
Thanks!

PR #3971, opened 2024-04-17
Add kotlin-faker to extensions list

serpro69:
Hi,
I've been using kotest for many years now, and thought it would be nice to contribute to the kotest community by providing an extension for kotest property testing, which allows one to generate even a larger variety of data than the built-in Arbs. 

Should I also update other versions of the docs? Which ones?

Also maybe you want to consider adding a link to the extension somewhere in kotest property docs? E.g. https://kotest.io/docs/proptest/property-test-extra-arbs.html


Kantis (approved):

Kantis:
I added it in non-versioned docs as well, which is the basis for all future versioned docs :)

PR #3976, opened 2024-04-19
Fix verifier in contain exactly in any order 

### LINKED ISSUE #3840: containExactlyInAnyOrder does not handle custom verifier properly
**latest code from master**
the following code shows the issue. these two lists should match wrt this verifier, but they don't:
```kotlin
            val caseInsensitiveStringEquality: Equality<String> = object : Equality<String> {
               override fun name() = "Case Insensitive String Matcher"

               override fun verify(actual: String, expected: String): EqualityResult {
                  return if(actual.uppercase() == expected.uppercase())
                     EqualityResult.equal(actual, expected, this)
                  else
                     EqualityResult.notEqual(actual, expected, this)
               }
            }
            listOf("apple", "Apple") should containExactlyInAnyOrder(listOf("APPLE", "APPLE"), caseInsensitiveStringEquality)
```
we could:
* just remove custom verifier
* replace verifier with a mapper (value: T) -> mappedValue: U and map all elements in both collections before comparison
* try to fix it while keeping the signature, which might be challenging

What should we do?

AlexCue987:
use custom verifier properly in `containExactlyInAnyOrder`, making sure it is commutative,
fix https://github.com/kotest/kotest/issues/3840


sksamuel:
If we resolve conflicts I can take a look.

AlexCue987:
> If we resolve conflicts I can take a look.

done

AlexCue987:
there are so many conflicts, not worth resolving. starting from scratch: https://github.com/kotest/kotest/pull/4160

Issue #3981, opened 2024-04-25
Arb.list fails within edge cases in case of null values 


oliverblaha:
Test case: `Arb.list(Arb.constant(null), 1..100).edgecases().forAll { }`

Expected: Passes and generates lists of null with a variable length.
Actual: `java.util.NoSuchElementException: Collection is empty.` in collections.kt, Line 108.

Note: This passes if the lower boundary is set to `0`.

Currently, edge cases returning null will lead to the conclusion "there is no edge case for this arb" and subsequently fall back to use a random value. If this is null, too, then the particular edge case for the list is ignored. However, this is not only incorrect (either the edge case or the random value might actually be null), but causes no edge case to be left; therefore the random selection of edge cases fails with "Collection is empty."

I'll submit a PR to address this.

oliverblaha:
Fixed in 5.9.0.

Kantis:
Change being reverted due to performance impact. In 6.0 we will change the `Arb.edgecase(rs: RandomSource)` function to return a `Sample<A>?` which lets us handle the nested nullability problem (`null` can currently symbolize both the absence of an edgecase, and the `null` edgecase)

oliverblaha:
That's a pity, as this will cause some tests to break again :(
Do you have any example that illustrates the performance impact?

oliverblaha:
Ah - found the reported issue; I like addressing the nullability issue. However, how about going back to previous behavior, but with a workaround for the failing test issue?

What happened was: if the result was null, it would be treated as "there is no edge case", leading to an index exception.
We could relatively easily work around that: if the list of edge cases is empty (which then would cause the exception), just return a list with a single null value. Not perfect, but beats throwing an exception.

(Edit: obviously this shall be a temporary workaround until 6.0, assuming the latter will still take a while.)

Kantis:
Could you draft a PR to show what you mean? I'll create a separate issue for addressing nullability

oliverblaha:
Hi @Kantis, sorry for the delay.
If drafted a PR that shows what I meant (although it's not actually returning a list with a single null, but simply null instead).

This still passes the new regression test, as well as the tests I wrote for the original issues.
Still needs a proper solution for edge-case-nullability though.

PR #3982, opened 2024-04-25
Fix Arb.list failing within edge cases in case of null values

oliverblaha:
This shall fix the issue as described in https://github.com/kotest/kotest/issues/3981.
Instead of just assuming that a null value means "there is no edge case", this change adds explicitly checking that, effectively fixing the problem.


kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/CollectionsTest.kt
```diff
@@ -91,6 +92,11 @@ class CollectionsTest : DescribeSpec({
             it.shouldHaveAtMostSize(500)
          }
       }
+
+      it("not fail on edge cases for non-empty lists with null values") {
```
AlexCue987: as the success here is "it doesn't throw an exception", should we say so: `shouldNotThrowAny ...`
also do we need to add any other checks for the output?

oliverblaha:
shouldNoThrowAny sounds great.
Don't think a check on the output adds any benefit though, but I'll give it
another thought.

oliverblaha:
Updated: I've added some more scenarios and introduced some output checks.

Kantis (approved):

PR #3983, opened 2024-04-26
singleElement-more-info 

AlexCue987:
provide more info when `singleElement` fails. add one of the following:
* Element found at index(es): [1]
* Element not found in collection


sksamuel (approved):

sksamuel:
lgtm

PR #3984, opened 2024-04-26
shouldExist-more-info (#62)

AlexCue987:
print more information when `shouldExist` fails:
* before: `Collection [1, 2, 3, 2] should not contain an element that matches the predicate (kotlin.Int) -> kotlin.Boolean`
* this PR adds `, but elements with the following indexes matched: [1, 3]`


sksamuel (approved):

sksamuel:
lgtm

PR #3985, opened 2024-04-26
shouldHaveElementAt-more-info 

AlexCue987:
print more info when `haveElementAt` fails. examples:
* Collection [\"a\", \"b\", \"c\"] should contain \"b\" at index 3, but collection was shorter
* Collection [\"a\", \"b\", \"c\"] should contain \"b\" at index 2, but element was different. Expected: <\"b\">, but was <\"c\">


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/CollectionMatchersTest.kt
```diff
@@ -143,6 +143,16 @@ class CollectionMatchersTest : WordSpec() {
             tests should haveElementAt(0, TestSealed.Test1("test1"))
             tests.shouldHaveElementAt(1, TestSealed.Test2(2))
          }
+         "indicate than collection is too short" {
```
metters:
```suggestion
         "indicate that collection is too short" {
```

metters (approved):

sksamuel (approved):

PR #3987, opened 2024-04-27
Create CODE_OF_CONDUCT.md

sschuberth:
This is based on GitHub's Contributor Covenant template. Using @sksamuel's email address as the contact for now. Is there a better one that we could use?


sksamuel:
We don't have a generic kotest@kotest.io email or anything like that so I guess mine is fine for now.

PR #3995, opened 2024-05-02
Correct Collection Matchers Documentation For `collection.shouldContainAllInAnyOrder()`

u-ways-e:
Hello, we noticed a minor error in the documentation page: 

https://kotest.io/docs/assertions/collection-matchers.html

For:

```kotlin
collection.shouldContainAllInAnyOrder()
```

> Assert that a collection contains all the given values and nothing else, in any order.

However it should be:

> Assert that a collection contains all the given values, in any order.

As the correct matcher for that incorrect description is:

```kotlin
collection.shouldContainExactlyInAnyOrder()
```

Feel free to reject if incorrect.


sksamuel:
Thank you

PR #3998, opened 2024-05-03
Add Stacktrace Substring Matching to Throwable Matchers

neBM:
This PR introduces the ability to match substrings and regular expressions in the stacktrace of a `Throwable` in the Kotest assertions library. This feature can be useful when writing tests that need to assert on specific parts of a stacktrace.

Changes:

1. Added two new infix functions `shouldHaveStacktraceContaining` and `shouldNotHaveStacktraceContaining` to the `Throwable` class. These functions accept a `String` or `Regex` and use it to match against the stacktrace of the `Throwable`.
2. Added corresponding matcher functions `haveStacktraceContaining` that create a `Matcher<Throwable>` to perform the actual matching logic.
3. Added tests for these new functions in `ThrowableMatchersTest.kt`.

This enhancement provides more flexibility when writing tests involving exceptions and their stacktraces.


Kantis:
Could you please ensure that the failure message is tested as well?

E.g.
```kotlin
shouldFail {
  Throwable() shouldHaveStackTraceContaining "foobar"
}.message shouldBe """
   exact message here
"""
```

Kantis:
Also, exceptions spell it as "stackTrace", so IMO we should also capitalize the T.

neBM:
Hi @Kantis, thank you for your feedback. I've implemented your suggestions 👍🏻

I noticed that the message with all its stack trace frames can make the test for the message in it's entirety exceptionally lengthy. Considering this, I thought about [using a regex](https://github.com/kotest/kotest/pull/3998/files#diff-26374820461f909ea5760d4bfdd9aa11f795e371c2140884684b5fd9bd247e77R202) to assert on what follows the _actual_. What are your thoughts on this approach? Do you think it maintains the integrity of the test while addressing the length concern?

This is my first time contributing to open source, so any feedback is greatly appreciated!

sksamuel:
Looks good to me. Happy to merge if @Kantis is happy.

Kantis (approved):
Looks good. I added a small test for the failure messages for the regex matchers as well. Thanks for contributing @neBM 🙌🏼

sksamuel:
can this be merged @Kantis ?

PR #4000, opened 2024-05-04
fixing the kotest autoscan warning message to reference the right property name

stepanv:
The original warning message was referencing property:
```
kotest.framework.classpath.scanning.config.disable
```
which doesn't exist. 
With this change, the right property name is dynamically resolved and currently would resolve to:
```
kotest.framework.classpath.scanning.autoscan.disable
```


kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/config/applyConfigFromAutoScan.kt
```diff
@@ -17,12 +17,17 @@ import io.kotest.mpp.log
 internal actual fun applyConfigFromAutoScan(configuration: ProjectConfiguration) {
 
    // this property is used to disable class path scanning for configurations
-   if (System.getProperty(KotestEngineProperties.disableAutoScanClassPathScanning) == "true") {
+   if (System.getProperty(KotestEngineProperties.disableAutoScanClassPathScanning) == true.toString()) {
```
sksamuel: What's the reason for this?

stepanv: This is unnecessary indeed but it avoids doing the same mistake as the author did with the property name that is using string literals.

Let me know and I’ll revert this line.

I’d expect even `True` or `TRUE` to resolve to true and this is the first step towards that. (But I didn’t want to introduce a breaking change. So I ended up doing just a half step.)

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/config/applyConfigFromAutoScan.kt
```diff
@@ -22,7 +22,12 @@ internal actual fun applyConfigFromAutoScan(configuration: ProjectConfiguration)
       return
    }
 
-   println("Warning: Kotest autoscan is enabled. This means Kotest will scan the classpath for extensions that are annotated with @AutoScan. To avoid this startup cost, set autoscan to false by setting the system property 'kotest.framework.classpath.scanning.config.disable=true'. In 6.0 this value will default to true. For further details see https://kotest.io/docs/next/framework/project-config.html#runtime-detection")
+   println("Warning: Kotest autoscan is enabled. This means Kotest will scan the classpath for extensions that are "
+      + "annotated with @AutoScan. To avoid this startup cost, set autoscan to false by setting the system property "
+      + "'${KotestEngineProperties.disableAutoScanClassPathScanning}=${true}'. "
```
OliverO2: Could you change this back as well?
```suggestion
      + "'${KotestEngineProperties.disableAutoScanClassPathScanning}=true'. "
```

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/config/applyConfigFromAutoScan.kt
```diff
@@ -22,7 +22,12 @@ internal actual fun applyConfigFromAutoScan(configuration: ProjectConfiguration)
       return
    }
 
-   println("Warning: Kotest autoscan is enabled. This means Kotest will scan the classpath for extensions that are annotated with @AutoScan. To avoid this startup cost, set autoscan to false by setting the system property 'kotest.framework.classpath.scanning.config.disable=true'. In 6.0 this value will default to true. For further details see https://kotest.io/docs/next/framework/project-config.html#runtime-detection")
+   println("Warning: Kotest autoscan is enabled. This means Kotest will scan the classpath for extensions that are "
+      + "annotated with @AutoScan. To avoid this startup cost, set autoscan to false by setting the system property "
```
OliverO2: Grab the opportunity and clear up this wording as well?
```suggestion
      + "annotated with @AutoScan. To avoid this startup cost, disable autoscan by setting the system property "
```

OliverO2 (changes requested):
Small changes, otherwise LGTM.

This may be unrelated and maybe it's just me, but actually I never managed to turn off the autoscan warning using just Gradle with `gradle.properties` like
```
kotest.framework.classpath.scanning.autoscan.disable=true
kotest.framework.classpath.scanning.config.disable=true
```

stepanv:
Sure, it’s done. Thx for posting the exact suggestions.

OliverO2 (approved):
Thanks!

Issue #3848, opened 2024-01-25
shouldContainExactly does not support new JDK 21 SequencedSet/SequencedCollection abstractions


quadrrem:
**Which version of Kotest are you using**

5.8.0

Java Version 21 introduced new [`SequencedSet`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/SequencedSet.html)/[`SequencedCollection`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/SequencedCollection.html) abstractions to mark sets/collection with a well-defined iteration order.

Both provide a [`reversed()`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/SequencedCollection.html#reversed()) method, to get reverse-ordered view of the underlying Set/Collection. 

When trying to test such a reverse-ordered view with `shouldContainExactly` it fails with an Assertion Error.

```kotlin
    linkedSetOf("a", "b", "c").shouldContainExactly("a", "b", "c") // successfully
    linkedSetOf("a", "b", "c").reversed().shouldContainExactly("c", "b", "a") // fails
```

```
java.lang.AssertionError: Disallowed: Sets can only be compared to sets, unless both types provide a stable iteration order.
ReverseLinkedHashSetView does not provide a stable iteration order and was compared with ArrayList which is not a Set
```

A fix could be to update the `isOrderedSet` function to check against this new abstractions instead of the concrete Set types.

https://github.com/kotest/kotest/blob/ebcbd0e22fee429ddafaf9af38794816d347800d/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/eq/isOrderedSet.kt#L8

mirageoasis:
can i get this issue assigned?

mirageoasis:
Able to reproduce error and find out structure of the code

However, there is one question here!

I would like to know if there are codes to look around. Since I want to branch different code according to different java version (sorry I am beginner to this open source 😂😂)

problem in here is that there is totally different data structure called

`ReverseLinkedHashSetView` which implements interface `SequencedSet` newly introduced in java 21

```kotlin
actual fun isOrderedSet(item: Iterable<*>) =
   item is LinkedHashSet ||
      item is TreeSet ||
      (item is Set && item.size == 1)
```
here we can see that except LinkedHashSet and TreeSet if certain item is Set this item's length should be 1

so `ReverseLinkedHashSetView` in example above generates error

[...]

PR #4002, opened 2024-05-04
feat: added SequencedSet for isOrderedSet

### LINKED ISSUE #3996: UnmodifiableSequencedSet not supported

5.8.1

Related to: https://github.com/kotest/kotest/issues/3848

Java Version 21 also introduced unmodifiable wrapper for  `SequencedSet`

When trying to test such a reverse-ordered view with shouldContainExactly it fails with an Assertion Error.

```
linkedSetOf("a", "b", "c").shouldContainExactly("a", "b", "c") // succeeds
Collections.unmodifiableSequencedSet(linkedSetOf("a", "b", "c")).shouldContainExactly("c", "b", "a") // fails
Collections.unmodifiableCollection(linkedSetOf("a", "b", "c")).shouldContainExactly("c", "b", "a") // succeeds
```

The issue probably is that the fix for 3848 still checks for subclasses of `SequencedSet` (https://github.com/kotest/kotest/pull/3873/files#diff-2d15d04b11a0d4ac020e2406ffcaa01ce78d750e0222572a224b7f54e3478309R8), I guess in an attempt to be compatible with older version of Java. 

What I saw in other project is to perform different checks for different versions

mirageoasis:
adding `SequencedSet` interface for 
`isOrderedSet`

finally fixing
#3848 
and fix
#3996 

working on test code for now


kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/eq/isOrderedSet.kt
```diff
@@ -1,11 +1,11 @@
...
-import java.util.SortedSet
+import java.util.*
 
 /**
  * JVM-specific: Allows TreeSet to be considered ordered for iterable comparisons
  */
 actual fun isOrderedSet(item: Iterable<*>) =
    item is LinkedHashSet ||
-      item is SortedSet ||
+      item is SequencedSet ||
```
AlexCue987: do we need to add/change any tests to validate this code change?

mirageoasis: > do we need to add/change any tests to validate this code change?

well... for some java `linkedHashSet` I think so?

Kantis:
@mirageoasis i kidnapped this PR to try out adding targets for different JDK versions, but it's really hard to get it working correctly and I don't think it's worth the complexity added. 

Another thing that I realized is that perhaps we should just rely on reflection in this case. We can add a check like this in `isOrderedSet` for JVM:

```kotlin
      item::class.java.interfaces.any { it.name == "java.util.SequencedSet" }
```

and it should be compatible with any java-version and correctly identify subclasses of SequencedSet for newer versions

mirageoasis:
@Kantis well I finally understood what you have said

```kotlin
actual fun isOrderedSet(item: Iterable<*>) =
   item is LinkedHashSet ||
      item is SortedSet ||
      (item is Set && item.size <= 1) ||
         item::class.java.interfaces.any { it.name == "java.util.SequencedSet" }
```

the real problem is how should run test for jdk21

Is it ok for me implement subclass of `EnabledCondition` 

to run code on specific jdk version?

https://github.com/kotest/kotest/blob/17d1b3b189fe4d804f1d3c6ce14280f724b0f168/kotest-framework/kotest-framework-api/src/jvmMain/kotlin/io/kotest/core/annotation/enabledif/os.kt#L7

like this!

LeoColman:
Closing this PR as it's a little stale.
Let us know if we should reopen it :)

PR #4007, opened 2024-05-09
Handling generic data classes when determining identifier stability

### LINKED ISSUE #3947: `row` toString() considered "unstable" even if all type parameters are clearly stable
5.8.1

### Background

I was trying to use `withData` to spin off multiple very similar tests.

The documented way to use it is to make a data class for the row objects. I have a lot of tests, I am very lazy, and having to define the types of the row objects for every test is just tons of repetition. And these row types effectively never get to be reused, because the types inside them always differ. And worst of all, I have to come up with _names_ for them every single time. Which is even more repetition. Which is even more repetition.

So I started writing generic classes for them, like `Row1<A>`, `Row2<A,B>`, `Row3<A,B,C>`, `Row4<A,B,C,D>`. And then a bunch of factory functions all called `row(...)` so you didn't have to type the number of parameters over and over either.

And then I thought, wait a second, doesn't kotest already have something like this?

So I tried using the existing `io.kotest.data.row(...)`, which returns classes with the exact same names I was using, and which are defined pretty much the exact same way as well.

So, for example, one of my tests looks like:

```kotlin
    "argument" - {
        withData(
            row(-3, 2, "2.55359005004222568722", "top left quadrant"),
            row(-3, -2, "-2.55359005004222568722", "bottom left quadrant"),
            row(3, -2, "-0.58800260354756755125", "bottom right quadrant"),
            row(3, 0, "0.00000000000000000000", "positive real"),
            row(-3, 0, "3.14159265358979323846", "negative real"),
            row(0, 2, "1.57079632679489661923", "positive imaginary"),
            row(0, -2, "-1.57079632679489661923", "negative imaginary"),
            row(0, 0, "0.00000000000000000000", "zero"),
        ) { (x, y, expected, _) ->
            Complex(Real.valueOf(x), Real.valueOf(y)).argument shouldBeCloseTo expected
        }
    }
```

This seems like it should be perfectly fine, but...

### Problem

When using `row(...)` to specify the rows for data-driven tests, all the test names come out as just the class name for the row object. In this case, `io.kotest.data.Row4`.

In the logs, there are many of these:

```
Warning, type class io.kotest.data.Row4 used in data testing does not have a stable toString()
Warning, type class io.kotest.data.Row4 used in data testing does not have a stable toString()
WARN: Duplicated test name io.kotest.data.Row4. To disable this message, set DuplicateTestNameMode to Silent.
```

It seems that kotest tries to detect whether the `toString()` method is stable, i.e. returns the same value when called multiple times for the same test. Because `Row4<A,B,C,D>` is generic, any of its types _could_ be an array, or something else which doesn't `toString()` sensibly.

This is completely nuts for my case though, because I am using solely `Int` and `String`. There is no way in hell the `toString()` for those is unstable, and if I replace the data class with one where the types are hard-coded, it works fine.

### Potential Solutions

Option A. Add `@IsStableType` to `RowN` classes and call it a day. The issue is, someone _might_ feed in something like arrays, which was presumably the reason it was disabled in the first place. (This is probably what I will do for my own workaround - keep the boilerplate classes I have now, add the annotation, bam, all the tests will work. If I do something stupid like adding an array, it would break, so I'd just aim not to be stupid.)

Option B. Actually look at the types of the objects in the `RowN` object when detecting whether it's stable. Consider primitives and strings as stable and then think about whether other types are or not. Maybe consider anything annotated with `@IsStableType` as stable as well. I don't know whether some kind of reified types hack could be used to do this, or whether you'd just have to look at the runtime values. But I guess since the important thing is the behaviour of `toString()`, looking at the runtime values should be totally fine.

### A Side Note

If the basic `forAll` way of doing data tests had properly come up as separate test failures/successes in IDEA, I never would have needed to try `withData`. It sucks that the former didn't "Just Work" in this situation. It has been a real long day.

Kantis:
By inspecting the stability of the actual values, we can see beyond the generics of the data class.

Fixes #3947


kotest-common/src/commonMain/kotlin/io/kotest/mpp/stable.kt
```diff
@@ -51,7 +78,5 @@ private val allPlatformStableTypes = setOf(
    Byte::class,
    Short::class,
    Boolean::class,
-   Pair::class,
-   Triple::class,
```
Kantis, line 55: I noticed that these classes were sometimes incorrectly marked as stable.

kotest-common/src/commonMain/kotlin/io/kotest/mpp/stable.kt
```diff
@@ -20,11 +21,24 @@ fun isStable(type: KType) = when (val classifier = type.classifier) {
...
+ * Returns true if the given class is a "stable" class, ie one that has a consistent toString().
+ *
+ * A stable class is either:
+ *  - a Kotlin or Java primitive type or a String
+ *  - an enum class
+ *  - a data class that only contains stable classes as members
+ *  - a stable type on the executing platform
+ */
+@Deprecated("Internal class, will not be exposed in the future. Deprecated since 5.9")
+fun isStable(kclass: KClass<*>, t: Any? = null): Boolean {
```
Kantis: This function is probably public to allow usage from `kotest-framework-datatest` even though this function is in `kotest-common` module.

* isStable seems to be used by _only_ `kotest-framework-datatest`.  Could we move these functions there and mark them as `internal`? 
* Otherwise, we could mark the classes as `@KotestInternal`?

kotest-common/src/commonMain/kotlin/io/kotest/mpp/stable.kt
```diff
@@ -20,11 +21,24 @@ fun isStable(type: KType) = when (val classifier = type.classifier) {
...
    return when {
       allPlatformStableTypes.contains(kclass) -> true
       reflection.isEnumClass(kclass) -> true
-      reflection.isDataClass(kclass) && hasStableMembers(kclass) -> true
+      reflection.isDataClass(kclass) && hasStableMembers(kclass, t) -> true
```
AlexCue987, line 41: because the intent here is to "have a stable toString()", should we also verify that this class does not override `toString()`?

Kantis: How does overriding `toString()` affect stability?

AlexCue987: A contrived example:
```kotlin
class SampleTest: StringSpec() {
    init {
        "unstable toString" {
            val something = WithContrivedToString("something")
            something.toString() shouldNotBe something.toString()
        }
    }
}

data class WithContrivedToString(
    var name: String
) {
    var accessCount: Int = 0

    override fun toString(): String {
        return "${super.toString()}, accessed ${++accessCount} times."
    }
}
```

Kantis: Fair enough, that might be. However, we haven't made that distinction before, and I haven't seen any complaints about it so I think we could just leave it as-is for now.

kotest-common/src/commonMain/kotlin/io/kotest/mpp/stable.kt
sksamuel: I would if in 6.0 we should just merge datatest into the core module.
Are we gaining anything from having the two ?

kotest-common/src/commonMain/kotlin/io/kotest/mpp/stable.kt
sksamuel: It's perfectly possible for someone to have a data class with a random string implementation, but I think that's getting quite edge casey as @Kantis says.
There's always going to be avenues for this to fail. The isStable method is meant to be a quick check to rule out the majority of use cases.

kotest-common/src/commonMain/kotlin/io/kotest/mpp/stable.kt
Kantis: I think we could move it. But perhaps we should also have a look at "[table-driven tests](https://github.com/kotest/kotest/tree/master/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/data)" vs. "[data-driven tests](https://github.com/kotest/kotest/tree/master/kotest-framework/kotest-framework-datatest/src/commonMain/kotlin/io/kotest/datatest)". They're very similar and I think that makes it hard for users to know what to use. 

I also think there's problems with `forAll` being used for both [property testing](https://github.com/kotest/kotest/blob/f4cd449df1cb5867941ff4e259c7f9333e5ba048/kotest-property/src/commonMain/kotlin/io/kotest/property/propertyTest1.kt#L89) and [table-driven tests](https://github.com/kotest/kotest/blob/f4cd449df1cb5867941ff4e259c7f9333e5ba048/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/data/forAll1.kt#L6).

By moving it into the core module, we're bringing both into context which might exasperate the problem

Kantis: Moved to discussions https://github.com/kotest/kotest/discussions/4052

PR #4009, opened 2024-05-09
Deprecating inspecting.

Kantis:
A [code search on github](https://github.com/search?q=import+io.kotest.assertions.inspecting+NOT+is:fork&type=code) shows no usages aside from Kotest itself, albeit being in Kotest since 2018.

Kotlin provides the `with` scoping function which does basically the same thing.

Kotest also provides `assertSoftly` on an object which adds it to scope and executes all assertions before failing.


sksamuel (approved):

sksamuel:
good idea

PR #4013, opened 2024-05-10
Partial key (#65)

AlexCue987:
If a key is missing in json, try finding a valid subpath:

```kotlin
   "Failure message states if key is missing, shows valid subpath" {
      shouldFail {
         json.shouldContainJsonKeyValue("$.store.bicycle.engine", "V2")
      }.message shouldBe """
         Expected given to contain json key <'$.store.bicycle.engine'> but key was not found. Found shorter valid subpath: <'$.store.bicycle'>
      """.trimIndent()
   }
```

I had to add a few public methods, as I couldn't invoke internal ones from a public inline function. 

Also made sure that we distinguish between these two cases:
* key not found
* key found, the corresponding value is null

Ideally we short also find indexes out of bound such as 
```kotlin
      json.shouldContainJsonKeyValue("$.store.book[2].author", "Evelyn Waugh")

saying <'$.store.book'> has only 2 elements
```
But that might be done later in another PR. WDYT @Kantis ?


sksamuel:
Can you add @KotestInternal to the classes you exposed. That's our way of saying they shouldn't be public but have to be.

sksamuel (approved):

AlexCue987:
> Can you add @KotestInternal to the classes you exposed. That's our way of saying they shouldn't be public but have to be.

done

Kantis:
IMO I would prefer if the tests focused more on the matcher and less on the helper functions, but let's merge anyway

Kantis:
And yes, let's deal with array indices in a later PR :)

AlexCue987:
> And yes, let's deal with array indices in a later PR :)

https://github.com/kotest/kotest/pull/4046

PR #4017, opened 2024-05-13
Test discovery: Make classpath scanning fallback configurable

### LINKED ISSUE #3973: Tests get run multiple times when using maxParallelForks
Kotest version: 5.8.1

This is one of the weirder bugs I've seen.

[Repro here](https://github.com/mightyguava/kotest-bug-repro-tests-using-when/commits/main/), containing a single commit on top of the `kotest-examples-multiplatform` repo main.

If you run `./gradlew cleanJvmTest jvmTest`, you'll see that **every test** within the module is run twice. This is also visible from `build/reports/tests/jvmTest/index.html`. If `maxParallelForks` is set to 1, this issue does not occur.

If you copy the test case in `WhenTest`, each new copy of the test case will cause all tests to be run an additional time.

Based on trial and error, I narrowed this down to `when` clauses being added to the test that operate on an `enum`. Looking at the compiled classes, I suspect it has something to do with this `$WhenMappings` class that is generated, e.g.

```
build/classes/kotlin/jvm/test/io/kotest/examples/mpp/whendupe/BarTests$1$1$WhenMappings.class
```

Each test case that uses a `when` on an `enum` seems to generate a extra inner class... and my hypothesis is that Kotest's runner (or JUnit framework? or Gradle?) isn't deduping this with the outer class properly.

OliverO2:
Kotest will now by default use classpath scanning if and only if one or
more non-class selectors are present.

The new system property
`kotest.framework.discovery.classpath.fallback.enabled=true` can be
used to enable classpath scanning if no selectors are present.

Also makes Kotest's internal debug logging multi-process-aware,
naming log files like "kotest-PID.log", containing the respective
process id.

Fixes #3973.


sksamuel:
> Fixes #3973.
> 
> Introduces a system property "kotest.framework.discovery.classpath.scanning.enabled", defaulting to "true".
> 
> This exists to work around a Gradle bug which occurs with JunitPlatform and `maxParallelForks > 1` (process parallelization). In this case, depending on the number of CPU cores, Gradle may invoke all but one Kotest launchers with a non-empty list of class selectors, and the remaining Kotest launcher with an empty list of class selectors, triggering a classpath scan in Kotest. In that case, tests will execute twice (discovered once via a class selector, a second time via the classpath scan).
> 
> Also makes Kotest-internal debug logging multi-process-aware, naming log files like "kotest-PID.log", containing the respective process id.

If it sends all but one a non empty list, how does the maxParallelForks do anything? Does it just spawn multiple kotest's that don't do anything?

OliverO2:
> If it sends all but one a non empty list, how does the maxParallelForks do anything? Does it just spawn multiple kotest's that don't do anything?

It spawns multiple Kotests (each in its own JVM). Then almost all of them get a non-empty list of classes, so these Kotest's do their stuff. Just (the last?) one gets an empty list and (with this PR) does nothing (instead of doing a classpath scan and running all tests again). Seems like a classical "off by one" error in Gradle's Java test task.

More detailed description here: https://github.com/kotest/kotest/issues/3973#issuecomment-2107882004

Considerations here: https://github.com/kotest/kotest/issues/3973#issuecomment-2108409310

#3973 shows how the drama unfolds.

Kantis:
Should we consider moving to making classpath scanning disabled unless the `ClasspathRootSelector` is included? Perhaps adding a fat warning when there's no selectors at all stating that this default will be changing in 6.0?

OliverO2:
Following up on https://github.com/kotest/kotest/issues/3973#issuecomment-2147827343:

Given its history and current usage, I see two viable options regarding JUnit 5 discovery in Kotest:
1. Remove classpath scanning entirely.
2. Make it optional as this PR does, but flip the default from "enabled" to "disabled".

@Kantis, @sksamuel ?

OliverO2:
This should be safe to use (no configuration required for all known use cases) and ready for final review.

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/config/applyConfigFromSystemProperties.kt
```diff
@@ -56,8 +57,11 @@ internal fun invocationTimeout(): Long? =
 internal fun allowMultilineTestName(): Boolean? =
    sysprop(KotestEngineProperties.allowMultilineTestName)?.let { it.uppercase() == "TRUE" }
 
+internal fun discoveryClasspathScanningEnabled(): Boolean? =
+   sysprop(KotestEngineProperties.discoveryClasspathFallbackEnabled)?.toBoolean()
+
 internal fun disableTestNestedJarScanning(): Boolean? =
-   sysprop(KotestEngineProperties.concurrentSpecs)?.toBoolean()
```
Kantis, line 60: Uhh, that looks odd.. good catch 😁

kotest-framework/kotest-framework-discovery/src/jvmMain/kotlin/io/kotest/framework/discovery/Discovery.kt
```diff
@@ -43,15 +42,12 @@ class Discovery(
 
    private val requests = ConcurrentHashMap<DiscoveryRequest, DiscoveryResult>()
 
-   // the results of a classpath scan, lazily executed and memoized.
-   private val scanResult = lazy { classgraph().scan() }
-
```
Kantis, line 48: This was never closed? Could be cause of the heap issues?

kotest-framework/kotest-framework-discovery/src/jvmMain/kotlin/io/kotest/framework/discovery/Discovery.kt
```diff
@@ -95,98 +78,82 @@ class Discovery(
...
     * locations specified by the uris param.
     */
-   private fun scanUris(): List<KClass<out Spec>> {
-
-      log {
-         val uptime = ManagementFactory.getRuntimeMXBean().uptime
-         "Discovery: Starting test discovery scan... [uptime=$uptime]"
+   private fun specsFromClassGraph(): List<KClass<out Spec>> {
+      val start = System.currentTimeMillis()
+      val specs = classgraph().scan().use { scanResult ->
```
Kantis, line 144: 👍🏼

Kantis (approved):
Looks good to me, nice work 🙌🏼

kotest-framework/kotest-framework-discovery/src/jvmMain/kotlin/io/kotest/framework/discovery/Discovery.kt
OliverO2: Probably, but mostly in our own discovery tests. Classpath scanning was never part of known use cases in production (invocation via JUnit Platform by Gradle or Maven Surefire).

PR #4018, opened 2024-05-14
eventually fixed

### LINKED ISSUE #3988: `eventually` without customization never runs test method
Using Kotest `5.8.1`, the following test
```kotlin
class MyTestClass {
    @Test
    fun myTestMethod() = runTest {
        eventually {
            error("Whoops!")
        }
    }
}
```
(running in JUnit 4 - Android project)

immediately fails with
```
java.lang.AssertionError: Block failed after 0ms; attempted 0 time(s)
```

The test block is never run, and the output is really just the above, giving no further details about what went wrong.

---

I've dug into this and believe the issue is that the `duration` of the default `EventuallyConfig` is set to `INFINITE`:
https://github.com/kotest/kotest/blob/4ad80702f12c2de9c938ee1909dc42f128fb6534/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt#L118
which is then added to the start time:
https://github.com/kotest/kotest/blob/4ad80702f12c2de9c938ee1909dc42f128fb6534/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt#L208
which results in an overflow, I believe. The result is that the check against the end time is immediately failing
https://github.com/kotest/kotest/blob/4ad80702f12c2de9c938ee1909dc42f128fb6534/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt#L245

---

While I initially wanted to raise this as a documentation issue only, I do wonder whether a `duration` of `INFINITE` should be handled specifically to mean "retry forever". Currently it basically results in "attempt not even once". To clarify: While the default value could simply be changed, I'm instead suggesting that `EventuallyControl` should be changed to allow retrying forever.

mirageoasis:
will resolve #3988 

this is draft for `eventually`

I made some senario for this new eventually

1. duration is set to `positve` -> works well
2. duration is set to `minus`-> error (this pr)
3. retries is set to `minus` -> error (this pr)
4. user do not set duration -> which make infinite loop (this pr)

and there are some other feature I introduced in this pr

1. if duration millisecond overflowed it is set to `Duration.INFINITE` and prints out waring

~~by the way can we change retries to `Long`? I think it doesn't matter to change `Int` to `Long`~~

sorry found out reason why retries cannot be `Long` in `ExponentialIntervalFn.kt`


mirageoasis:
currently working on test code 
Any suggestions are welcome!

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -205,7 +211,7 @@ object NoopEventuallyListener : EventuallyListener {
 private class EventuallyControl(val config: EventuallyConfiguration) {
 
    val start = timeInMillis()
-   val end = start + config.duration.inWholeMilliseconds
+   val end = addSafely()
```
OliverO2: ~Have you tried to run this? The local `start` variable is not visible in `addSafely()`, right?~
[Edit: It is visible. My bad. Just needed to look at the whole file instead of the GitHub diff.]

What about expressing more clearly what happens, so that readers don't have to look up the implementation of `addSafely()`? For example, mirroring the `coerce...` functions in the stdilb, we could use:

```kotlin
val end = (start + config.duration.inWholeMilliseconds).orInfinityIfNegative()
```
with:
```kotlin
fun Long.orInfinityIfNegative() = if (this < 0) Duration.INFINITE.inWholeMilliseconds else this
```

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -242,7 +248,14 @@ private class EventuallyControl(val config: EventuallyConfiguration) {
       lastDelayPeriod = (timeInMillis() - delayMark).milliseconds
    }
 
-   fun hasAttemptsRemaining() = timeInMillis() < end && iterations < config.retries
+   fun hasAttemptsRemaining() : Boolean {
+      // If the duration is infinite, it is considered as default value, which means it will never stop
+      val isInfinite = config.duration == INFINITE
```
OliverO2: Isn't this case already covered by `isWithinDuration`?

mirageoasis: well 

> Isn't this case already covered by `isWithinDuration`?

`isWithinDuration` only checks time depletion
`isInfinte` checks if `config.duration` if is set to default which is `INFINITE`

mirageoasis: oh maybe since `Duration.INFINITE` is almost same as `INFINITE` I think I can rollback this expression

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
mirageoasis: thanks to your advice I refactored code! thanks a lot!

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -121,7 +121,7 @@ data class EventuallyConfiguration(
 }
 
 object EventuallyConfigurationDefaults {
-   var duration: Duration = Duration.INFINITE
+   var duration: Duration = INFINITE
```
AlexCue987: I am concerned over this approach in general. This thing is shared (because `object`) and mutable (because `var`). As such, there will be race conditions. We should refactor away the very need for shared mutable things. WDYT @OliverO2

mirageoasis: since this is default setting I think it is ok to change `var` to `val`

OliverO2: @AlexCue987 I agree.

_All_ properties should really be `val` (though I wonder how this would affect backward compatibility in this case). Global state is almost never a good idea and in particular with the option of running parallel tests and concurrent coroutines, any context-induced configuration should be coroutine-local. See https://github.com/kotest/kotest/pull/2791 for an example.

I also wonder why that object is part of the public API. It might better be private.

mirageoasis: I think its appropriate to make another pr and deal with this issue
on this pr I will only focus on eventually's function
I will raise another pr for this issue!

OliverO2: Yes, that makes sense. To make this PR ready for review, we'd need some test code covering the proposed fixes. Would also be good to update the initial PR comment with references to issues resolved by this PR.

AlexCue987: > I think its appropriate to make another pr and deal with this issue on this pr I will only focus on eventually's function I will raise another pr for this issue!

sure - this is a preexisting condition, not something introduced in this PR

mirageoasis: > Yes, that makes sense. To make this PR ready for review, we'd need some test code covering the proposed fixes. Would also be good to update the initial PR comment with references to issues resolved by this PR.

done with test!

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -260,6 +266,21 @@ private class EventuallyControl(val config: EventuallyConfiguration) {
...
+   fun Long.ensureInRange() : Long {
+      // Check if the duration is INFINITE which is default
+      if (config.duration == INFINITE) {
+         return INFINITE.inWholeMilliseconds
+      }
+
+      return when {
+         this in 0..INFINITE.inWholeMilliseconds -> this
+         else -> INFINITE.inWholeMilliseconds.also {
+            println("[WARN] end value $this is out of the valid range (0 to ${INFINITE.inWholeMilliseconds}), value is fixed to Duration.INFINITE.inWholeMilliseconds")
```
kshired:
```suggestion
            println("[WARN] end value $this is out of the valid range (0 to $it), value is fixed to Duration.INFINITE.inWholeMilliseconds")
```

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -260,6 +266,21 @@ private class EventuallyControl(val config: EventuallyConfiguration) {
          appendLine(this.stackTraceToString())
       }
    }.toString()
+
+
+   fun Long.ensureInRange() : Long {
+      // Check if the duration is INFINITE which is default
+      if (config.duration == INFINITE) {
```
kshired:
```suggestion
      if (config.duration.isInfinite()) {
```

kshired (commented):
@mirageoasis 

- I commented some suggestions.
- Why not modify the comment of EventuallyConfigurationBuilder.duration that the constraint has been added?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
sksamuel: Should we not just error out and throw if the duration is invalid?

mirageoasis: think throwing error is more clear then changing value since people will not set duration close to inifnite

sksamuel: And no one is going to spot the warn in the logs unless they look closely.
Failing the test makes more sense if its a bug, which having an incorrect duration is.

mirageoasis:
Can someone help me with test case to trigger Duration overflow?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/io/kotest/assertions/nondeterministic/EventuallyTest.kt
```diff
@@ -371,5 +371,38 @@ class EventuallyTest : FunSpec() {
...
+      test("raise error if retries is less than 0") {
+         shouldThrow<IllegalArgumentException> {
+            eventuallyConfig {
+               retries = -1
+            }
+         }
+      }
+
+      test("when duration is set to default it cannot end test until iteration is done") {
+         val finalCount = 10000
```
OliverO2: This would also work with a count of 10, and faster, right?

mirageoasis: yes

mirageoasis:
> Can someone help me with test case to trigger Duration overflow?

here's how it works

`Duration`'s single value cannot be Bigger than `2 ^ 63 - 1`

if we add two `Duration` and if it's value's result is bigger than `2 ^ 63 -1` then we set `Duration` to `Duration.INFINITE`

so

this 
```kotlin
val end = (start.milliseconds + config.duration).inWholeMilliseconds
```
is 100% safe!

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -205,7 +211,7 @@ object NoopEventuallyListener : EventuallyListener {
 private class EventuallyControl(val config: EventuallyConfiguration) {
 
    val start = timeInMillis()
-   val end = start + config.duration.inWholeMilliseconds
+   val end = (start.milliseconds + config.duration).inWholeMilliseconds
```
mirageoasis: here's how it works

Duration's single value cannot be Bigger than 2 ^ 63 - 1

if we add two Duration and if it's value's result is bigger than 2 ^ 63 -1 then we set Duration to Duration.INFINITE

so

this

val end = (start.milliseconds + config.duration).inWholeMilliseconds
is 100% safe!

I've come a long way round :(

mirageoasis:
#4114 
since this pr collides with this pr I will check what has changed!

mirageoasis:
think conflicting pr now resolves overflow issue I think job of this pr is to limit duration and retries from being negative

mirageoasis:
@aSemy 
thank to your PR duration overflow issue has been solved
but I think `retries` and `duration` should not be negative since it can cause some error
can you review if my fixed pr does not collide with pr of your's?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -115,7 +115,12 @@ data class EventuallyConfiguration(
    val listener: EventuallyListener,
    val shortCircuit: (Throwable) -> Boolean,
    val includeFirst: Boolean,
-)
+) {
+   init {
+      require(duration.inWholeMilliseconds >= 0) { "Duration must be greater than or equal to 0" }
```
aSemy:
```suggestion
      require(duration >= Duration.ZERO) { "Duration must be greater than or equal to 0" }
```

nitpick

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -115,7 +115,12 @@ data class EventuallyConfiguration(
...
+      require(retries >= 0) { "Retries must be greater than or equal to 0" }
+   }
```
aSemy:
```suggestion
   init {
      require(duration.inWholeMilliseconds >= 0) { "Duration must be greater than or equal to 0, but was $duration" }
      require(retries >= 0) { "Retries must be greater than or equal to 0, but was $retries" }
   }
```

Suggestion: Add the actual values to the messages, so the error messages are easier to understand.

aSemy:
> @aSemy thank to your PR duration overflow issue has been solved but I think `retries` and `duration` should not be negative since it can cause some error can you review if my fixed pr does not collide with pr of your's?

It looks like there are no conflicts 👍 

However, there's lots of formatting changes that could cause conflicts in other PRs, and it makes the changes in the PR more noisy and harder to review. Would you mind reverting the formatting changes?

E.g. Revert

```
val result =
   testEventually(5.days) {
```

back to 

```
val result = testEventually(5.days) {
```

image

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
AlexCue987: should be `but was $retries" }` not `but was retries" }`

aSemy: oops, thanks @AlexCue987 - fixed

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -133,7 +138,7 @@ object EventuallyConfigurationDefaults {
 class EventuallyConfigurationBuilder {
 
    /**
-    * The total time that the [eventually] function can take to complete successfully.
+    * The total time that the [eventually] function can take to complete successfully. Must be greater than or equal to 0.
```
AlexCue987: maybe just "Must be non-negative"?

mirageoasis: omg....
I will fix formatting on evening! 👍

mirageoasis: I will fix this on evening!

mirageoasis:
> > @aSemy thank to your PR duration overflow issue has been solved but I think `retries` and `duration` should not be negative since it can cause some error can you review if my fixed pr does not collide with pr of your's?
> 
> It looks like there are no conflicts 👍
> 
> However, there's lots of formatting changes that could cause conflicts in other PRs, and it makes the changes in the PR more noisy and harder to review. Would you mind reverting the formatting changes?
> 
> E.g. Revert
> 
> ```
> val result =
>    testEventually(5.days) {
> ```
> 
> back to
> 
> ```
> val result = testEventually(5.days) {
> ```
> 
> image



> > @aSemy thank to your PR duration overflow issue has been solved but I think `retries` and `duration` should not be negative since it can cause some error can you review if my fixed pr does not collide with pr of your's?
> 
> It looks like there are no conflicts 👍
> 
> However, there's lots of formatting changes that could cause conflicts in other PRs, and it makes the changes in the PR more noisy and harder to review. Would you mind reverting the formatting changes?
> 
> E.g. Revert
> 
> ```
> val result =
>    testEventually(5.days) {
> ```
> 
> back to
> 
> ```
> val result = testEventually(5.days) {
> ```
> 
> image

omg....
I will fix formatting on evening! 👍

mirageoasis:
think it is now okay to merge can someone approve my change?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -202,7 +207,7 @@ class EventuallyConfigurationBuilder {
 typealias EventuallyListener = suspend (Int, Throwable) -> Unit
 
 object NoopEventuallyListener : EventuallyListener {
-   override suspend fun invoke(iteration: Int, error: Throwable) {}
+   override suspend fun invoke(iteration: Int, error: Throwable, ) {}
```
aSemy:
```suggestion
   override suspend fun invoke(iteration: Int, error: Throwable) {}
```

formatting nitpick

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -156,7 +161,7 @@ class EventuallyConfigurationBuilder {
    var intervalFn: DurationFn? = EventuallyConfigurationDefaults.intervalFn
 
    /**
-    * The maximum number of invocations regardless of durations. By default, this is set to max retries.
+    * The maximum number of invocations regardless of durations. By default, this is set to max retries. And Must be greater than or equal to 0.
```
aSemy:
```suggestion
    * The maximum number of invocations regardless of durations. By default, this is set to max retries. Must be non-negative.
```

nitpick: match the KDoc of `duration`.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/io/kotest/assertions/nondeterministic/EventuallyTest.kt
```diff
@@ -431,6 +431,54 @@ class EventuallyTest : FunSpec() {
...
+      }
+
+      test("test eventually without configuration") {
+         // linked to issue #3988
+         var cnt = 0
+         eventually {
+            cnt += 1
+            cnt shouldBe 100
+         }
+         cnt shouldBe 100
```
aSemy:
```suggestion
         var count = 0
         eventually {
            count += 1
            count shouldBe 100
         }
         count shouldBe 100
```

nitpick; rename var for clarity

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -18,7 +18,7 @@ import kotlin.time.TimeSource
  * To supply more options to eventually, use the overload that accepts an [EventuallyConfiguration].
  */
 suspend fun <T> eventually(
-   test: suspend () -> T,
+   test: suspend () -> T
```
aSemy:
```suggestion
   test: suspend () -> T,
```

formatting nitpick

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -86,7 +86,7 @@ suspend fun <T> eventually(
 }
 
 fun eventuallyConfig(
-   configure: EventuallyConfigurationBuilder.() -> Unit,
+   configure: EventuallyConfigurationBuilder.() -> Unit
```
aSemy:
```suggestion
   configure: EventuallyConfigurationBuilder.() -> Unit,
```

formatting nitpick

aSemy (approved):
lgtm!

I have some minor nitpicks. Please double check them, and if they make sense to you, apply them before merging.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
aSemy: tbh I find "Must be greater than or equal to zero" easier to read, but "Must be non-negative" is fine too.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
mirageoasis: I personally like trailling comma but since it ruin's commit's visibility I will fix this!

mirageoasis:
should I fix docs or leave docs cause my change and limitation is too obvious?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
aSemy: I like trailing commas too, but only if the parameters are on separate lines :)

mirageoasis:
@Kantis Can you tell me why my test fails? is it my fault? or problem with configuration?

when I see logs it says some kind of kotlin dependency

OliverO2 (approved):
Looks good, ready to merge. Thank you for this contribution!

PR #4019, opened 2024-05-16
fix broken matcher documentation for dates

dzappold:
fix documentation, table for date matchers where broken


AlexCue987:
> fix documentation, table for date matchers where broken

can you share some details?

dzappold:
the table of the dates matchers weren't displayed as table, due to some syntax error in markdowns, see https://kotest.io/docs/assertions/core-matchers.html

Kantis (approved):

Issue #4016, opened 2024-05-13
Performance Degradation in 5.9.0 compared to 5.8.1


panphg:
### Description

Version 5.9.0 has introduced a performance degradation compared to version 5.8.1 when using a generator for a data structure with nested lists.

I have set up a small repository that reproduces this issue: https://github.com/panphg/kotest-performance
It contains a small Kotlin project with a single test [PerformanceTest](https://github.com/panphg/kotest-performance/blob/main/src/test/kotlin/PerformanceTest.kt). Using the gradle commands below, the test can be run with Kotest 5.8.1 and 5.9.0.

When run with Kotest 5.8.1, the test runs in a few seconds. When run with 5.9.0, the test runs for a long time (minutes, if not hours).

To run the test with Kotest 5.8.1, execute:
```bash
./gradlew test -Dversion=5.8.1
```

To run the test with Kotest 5.9.0, execute:
```bash
./gradlew test -Dversion=5.9.0
```
### Expected Behaviour
The test should finish in seconds when run with 5.9.0, as it does when run with 5.8.1.

Kantis:
Possibly related to https://github.com/kotest/kotest/pull/3982. I think it was the only change related to property testing in 5.9

panphg:
This might be the culprit. [Replacing](https://github.com/kotest/kotest/pull/3982/files#diff-07978b8f35a89840ce619edfba9841ebde2bd601673e949b9261e7e41c60a1d2R99) `gen.edgecase(rs)` with `gen.edgecases()` increased the amount of generated values by a factor of 100, the default value for the argument iterations. And calling `gen.edgecases()` twice and then twice again a couple of lines below certainly didn't help either. Maybe using `gen.edgecases(iterations = 1)` instead already fixes the performance issues, but I'm not sure about the implications of a change like this.

Our build uses lots of generators and went from taking about 6min with Kotest 5.8.1 to timing out after more than 2 hours with 5.9.0.

Kantis:
@myuwono @sksamuel I think the correct thing to do would be to add a wrapper around the [edgecase function's](https://github.com/kotest/kotest/blob/19ce4d3b4712ed04c23d9aad7ce56c064caaa503/kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt#L103) result. E.g. it should return either `Edgecase<A>?` or simply a `Sample<A>?`.That way we can distinguish between the lack of an edgecase, i.e.`null` result, vs the null edgecase

[...]

PR #4027, opened 2024-05-20
add-regression-test-for-performance-degradation 

AlexCue987:
Let's add a regression test for the following recent fix "Revert "Fix Arb.list failing within edge cases in case of null values": 

https://github.com/kotest/kotest/commit/8ba8975f319a1d94e640acee776809ef03e48bc3

This test passes off master branch, but it fails if I revert that fix: `git revert 8ba8975 --no-commit `


Kantis:
Nice. Perhaps it would be good with a reference to the issue explaining the performance problem as well?

AlexCue987:
> Nice. Perhaps it would be good with a reference to the issue explaining the performance problem as well?

here is the issue, if that's what you are asking for:

https://github.com/kotest/kotest/issues/3981

AlexCue987:
and this https://github.com/kotest/kotest/issues/4016

Kantis:
No I mean in the actual code. Perhaps in a comment, or preferably in the test name. That way it'll stick with us and be easy to find the history for that particular test

AlexCue987:
> No I mean in the actual code. Perhaps in a comment, or preferably in the test name. That way it'll stick with us and be easy to find the history for that particular test

done

Kantis (approved):

PR #4030, opened 2024-05-21
Add an Arbitrary to generate whitespace.

jedesroches:
This pull request adds a `whitespace` function to the Codepoints in order to generate strings with only whitespace characters (and compose more complex generators). I happened to need one for a project and thought I might share it.

Since this is my first contribution, I'm not sure of which docs I should update where (and also, I've never multiplatform'd in kotlin before, so I'm not sure if I need to do anything else than the code I've written for this to work everwhere).

Cheers :)


jedesroches:
For the record, in order for the `apiCheck` task to pass, I have run the `:kotest-property:apiDump` task as suggested in the logs. I hope this is expected ? Or is something else required when you change the API (I only added a function) ?

Kantis:
@jedesroches that should be it. Contribution looks good to me and there seems to be no relevant docs to add this to currently.

Kantis (approved):

PR #4038, opened 2024-05-27
Enabling experimental KLib API validation

Kantis:
Main benefit is that it also dumps inlined functions, plus per-target APIs.

`apiCheck` workflows are altered to run on macos so they can validate for all targets, even if this shouldn't be strictly necessary. See [here](https://github.com/Kotlin/binary-compatibility-validator?tab=readme-ov-file#klib-abi-dump-generation-and-validation-on-linux-and-windows-hosts)


OliverO2:
Wouldn't we need some way to have the CI generate updated API dumps on macOS if these had been previously created on a non-macOS host?

(Emphasis mine)

> When it comes to dump generation (apiDump task) on non-Apple hosts, binary compatibility validator attempts to infer an ABI from dumps generated for supported targets and an old dump from project's api folder (if any). **Inferred dump may not match an actual dump, and it is recommended to update a dump on hosts supporting all required targets, if possible.**

sksamuel:
Prob a bit out of the loop here, but why are we using macos for the non mac builds ?

OliverO2:
If my understanding is correct:
1. The change towards macOS is only used for `apiCheck` in this PR.
2. With KLib API validation, macOS is the only target that can correctly dump for all targets. Dumps on the other targets create only an "inferred" dump for macOS which may be incomplete. Maybe `apiCheck` is also more complete on macOS, but I'm not sure.

sksamuel:
Right yeah I see, api check only.
That makes sense, thanks!

Kantis:
Not sure how we would auto-update the API dump. Arrow is using same version and it looks like they've added a job that handles it: https://github.com/arrow-kt/arrow/blob/9894996dc38aa7d704f1b02f2bd977978ab6683a/.github/workflows/pull_request.yml#L74

For some reason it's run on ubuntu though, so they don't seem to care about having Apple targets in the dump. Want to go with that approach, but use a macos-runner?

OliverO2:
Maybe we could do an `apiDump` on macOS, but make that depend on a successful `apiCheck` on Linux. This way our dump would be always current for all operating systems while still requiring a more lenient API check which all PRs could easily satisfy.

Kantis:
Sounds like a good strategy @OliverO2. I'll try to add that in the workflows

sksamuel:
What do we wanna do with this @Kantis

LeoColman:
Closing this as it's stale.
Feel free to reopen

PR #4039, opened 2024-05-27
Use MethodSource for JUnit test cases

### LINKED ISSUE #4024: Test case names in JUnit 5 xml reports are
`ShouldSpec` and `BehaviorSpec` based tests when executed by Maven Failsafe Plugin or Maven Surefire Pluing (using JUnit5 runner) generate XML reports (in `target/failsafe-reports` or `target/surefire-reports` directory) with empty `name` fields in `testcase` elements.

Kotest version 5.9.0
Maven version 3.9.6
Maven Failsafe Plugin version 3.2.5
Maven Surefire Plugin version 3.2.5
JUnit Jupiter version 5.10.2

adangel:
Fixes #4024


sksamuel (approved):

kotest-runner/kotest-runner-junit5/src/jvmMain/kotlin/io/kotest/runner/junit/platform/JUnitTestEngineListener.kt
```diff
@@ -316,14 +319,20 @@ class JUnitTestEngineListener(
          id,
          formatter.format(testCase),
          t,
-         ClassSource.from(testCase.spec::class.java, null), // gradle-junit-platform hides tests if we don't send this
+         // gradle-junit-platform hides tests if we don't send a source at all
+         // surefire-junit-platform (maven) needs a MethodSource in order to separate test cases from each other
+         //   and produce more correct XML report with test case name.
+         getMethodSource(testCase.spec::class, id),
```
Kantis: Is this really the right thing to do? With `ClassSource` we can report the line of each test (which is available in the `TestCase.source`, not sure why we don't do that already). A `MethodSource` sounds to me that it should reference a method, which does not exist. 

To me, it feels like Surefire should be fixed to take into account the reported display name, even when they receive a test that uses `ClassSource` as you identified here. https://github.com/apache/maven-surefire/blob/790e332790b1d601cb58e566c29eef3e47353f7d/surefire-providers/surefire-junit-platform/src/main/java/org/apache/maven/surefire/junitplatform/RunListenerAdapter.java#L366

sksamuel: the changes of us getting surefire fixed are low.
Method source might be ok as a hack ?

adangel: I didn't test the gradle integration, so can't say if this change would negatively impact the junit-platform integration of gradle (see https://github.com/gradle/gradle/blob/master/platforms/jvm/testing-junit-platform/src/main/java/org/gradle/api/internal/tasks/testing/junitplatform/JUnitPlatformTestExecutionListener.java ). They also seem to use ClassSource and/or MethodSource in some cases.

Kantis: I have registered for an account at https://issues.apache.org/jira/projects/SUREFIRE, will post a report there and see

Kantis: I looked at it a bit more and they are actually correctly parsing the display name. They've just got the same concept of all tests being methods as sort of hard-wired into their models, so I guess that won't really change soon.

I guess we could merge this and see how it plays out?

sksamuel (approved):

kotest-runner/kotest-runner-junit5/src/jvmMain/kotlin/io/kotest/runner/junit/platform/JUnitTestEngineListener.kt
sksamuel: Can give it a go

sksamuel:
The JUnit tests are failing.

straxess:
@adangel, Hello, I understand (https://github.com/kotest/kotest/issues/4024) I need this PR to fix the test reports in my builds on TeamCity. Can you start working on the [failed tests](https://github.com/kotest/kotest/pull/4039#issuecomment-2241337904)?

straxess:
@sksamuel, hi! Is this PR ready to be merged, or do you plan to review it again?

PR #4040, opened 2024-05-27
Experimental K2 JS support for nesting at all levels

sksamuel:


OliverO2:
If I run `jsKotest` in [kotest-examples-javascript#36](https://github.com/kotest/kotest-examples-javascript/pull/36), then comment out the following line in [SsnTest.kt](https://github.com/kotest/kotest-examples-javascript/blob/e474e7ea14a640b108733e890c673b766b8825f2/src/jsTest/kotlin/io/kotest/examples/js/SsnTest.kt#L18),
```kotlin
            validateSocial("313-03-5310") shouldBe false
```

the next run of `jsKotest` produces `java.lang.IllegalStateException: config function was not declared in io.kotest.js`.

PR #4043, opened 2024-05-29
Better timeouts handling

### LINKED ISSUE #3703: Kotest uses virtual time for tests timeouts when Dispatchers.Main is overridden
**Version:** 5.7.2

If you have overridden `Dispatchers.Main` before a test, timeout uses virtual time and always fails with `TimeoutCancellationException`.

Kotest config:
```kotlin
class KotestConfig : AbstractProjectConfig() {
    override val coroutineTestScope: Boolean = true
    override val timeout = 5.seconds
}
```

Test code:
```kotlin
class MySpec : FeatureSpec({

    val dispatcher = StandardTestDispatcher()

    beforeTest { Dispatchers.setMain(dispatcher) }
    afterTest { Dispatchers.resetMain() }

    feature("...") {
        scenario("...") {
            // It will crash if Main dispatcher was not set before
            launch(Dispatchers.Main) {
                delay(100)
            }.join()

            dispatcher.scheduler shouldBe testCoroutineScheduler
        }
    }
})
```

**Actual result**
Test fails with timeout exception:
```stacktrace
Timed out after 5s of _virtual_ (kotlinx.coroutines.test) time. To use the real time, wrap 'withTimeout' in 'withContext(Dispatchers.Default.limitedParallelism(1))'
kotlinx.coroutines.TimeoutCancellationException: Timed out after 5s of _virtual_ (kotlinx.coroutines.test) time. To use the real time, wrap 'withTimeout' in 'withContext(Dispatchers.Default.limitedParallelism(1))'
	at app//kotlinx.coroutines.TimeoutKt.TimeoutCancellationException(Timeout.kt:191)
	at app//kotlinx.coroutines.TimeoutCoroutine.run(Timeout.kt:159)
	...
	at app//io.kotest.engine.test.interceptors.TestCoroutineInterceptor.intercept(TestCoroutineInterceptor.kt:34)
	at app//io.kotest.engine.test.TestInvocationInterceptor$runBeforeTestAfter$wrappedTest$1$1.invokeSuspend(TestInvocationInterceptor.kt:70)
	at app//io.kotest.engine.test.TestInvocationInterceptor$runBeforeTestAfter$wrappedTest$1$1.invoke(TestInvocationInterceptor.kt)
	at app//io.kotest.engine.test.TestInvocationInterceptor$runBeforeTestAfter$wrappedTest$1$1.invoke(TestInvocationInterceptor.kt)
	at app//io.kotest.engine.test.interceptors.InvocationTimeoutInterceptor$intercept$3.invokeSuspend(InvocationTimeoutInterceptor.kt:43)
	...
```

**Expected result**
Test runs without exception.

### LINKED ISSUE #3969: coroutineTestScope: timeout for TestDispatcher cannot be configured
5.8.1

Reproducer:
```kotlin
class KotestTest : StringSpec({
	addTest(TestName("test"), disabled = false, config = TestConfig(timeout = 120.seconds, coroutineTestScope = true, testCoroutineDispatcher = true)) {
		withContext(Dispatchers.Default) {
			delay(90.seconds)
		}

		println("If this point is reached, the timeout was respected :)")
	}
})
```

This fails with
```
After waiting for 1m, the test coroutine is not completing, there were active child jobs: [UndispatchedCoroutine{Active}@39c77ef3]
kotlinx.coroutines.test.UncompletedCoroutinesError: After waiting for 1m, the test coroutine is not completing, there were active child jobs: [UndispatchedCoroutine{Active}@39c77ef3]
```

Yet, I passed `TestConfig(timeout = 2.minutes)`, so the test should not have failed.

---

If I set `coroutineTestScope = false, testCoroutineDispatcher = false`, then the timeout is respected.

osipxd:
Addressed some issues related to timeouts:
- Use wall clock time for timeouts if `TestCoroutineScheduler` is present in coroutine context (fixes #3703)
- Pass the `timeout` parameter into `runTest` instead of `InvocationTimeoutInterceptor` if `coroutineTestScope` is enabled (fixes #3969)
- Add an original exception as a cause into `TestTimeoutException`

TODO:
- [x] Write tests

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/ResolvedTestConfig.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/ResolvedTestConfig.kt
@@ -1,12 +1,14 @@
 package io.kotest.core.test.config
 
 import io.kotest.core.Tag
+import io.kotest.core.config.Defaults
 import io.kotest.core.extensions.Extension
 import io.kotest.core.test.AssertionMode
 import io.kotest.core.test.Enabled
 import io.kotest.core.test.EnabledOrReasonIf
 import io.kotest.core.test.TestCaseSeverityLevel
 import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
 
 /**
  * Runtime resolved config attached to a [io.kotest.core.test.TestCase].
@@ -30,13 +32,13 @@ data class ResolvedTestConfig(
     *
     * To set a timeout per invocation see [invocationTimeout].
     */
-   val timeout: Duration?,
+   val timeout: Duration,
 
    /**
     * This timeout applies to individual invocations of a test case. If invocations is 1, then this
     * has the same effect as timeout. To set a timeout across all invocations then see [timeout].
     */
-   val invocationTimeout: Duration?,
+   val invocationTimeout: Duration,
 
    /**
     * [Tag]s that are applied to this test case, in addition to any tags declared on
@@ -81,6 +83,9 @@ data class ResolvedTestConfig(
       require(threads > 0) { "Number of threads must be greater than 0" }
    }
 
+   // note: the invocation timeout can't be larger than the test case timeout
+   val resolvedInvocationTimeout: Duration = minOf(timeout, invocationTimeout)
+
    companion object {
       val default = ResolvedTestConfig(
          { Enabled.enabled },
@@ -92,8 +97,8 @@ data class ResolvedTestConfig(
          assertSoftly = false,
          blockingTest = false,
          extensions = emptyList(),
-         timeout = null,
-         invocationTimeout = null,
+         timeout = Defaults.defaultTimeoutMillis.milliseconds,
+         invocationTimeout = Defaults.defaultInvocationTimeoutMillis.milliseconds,
          tags = emptySet(),
          severity = TestCaseSeverityLevel.TRIVIAL,
          failfast = false,
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/ResolvedTestConfig.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/ResolvedTestConfig.kt
@@ -83,9 +83,6 @@ data class ResolvedTestConfig(
       require(threads > 0) { "Number of threads must be greater than 0" }
    }
 
-   // note: the invocation timeout can't be larger than the test case timeout
-   val resolvedInvocationTimeout: Duration = minOf(timeout, invocationTimeout)
-
    companion object {
       val default = ResolvedTestConfig(
          { Enabled.enabled },
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TestCoroutineInterceptor.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TestCoroutineInterceptor.kt
@@ -6,7 +6,7 @@ import io.kotest.core.coroutines.TestScopeElement
 import io.kotest.core.test.TestCase
 import io.kotest.core.test.TestResult
 import io.kotest.core.test.TestScope
-import io.kotest.core.test.TestType
+import io.kotest.core.test.timeout
 import io.kotest.engine.test.scopes.withCoroutineContext
 import kotlinx.coroutines.test.runTest
 import kotlinx.coroutines.withContext
@@ -31,13 +31,7 @@ class TestCoroutineInterceptor : TestExecutionInterceptor {
       logger.log { Pair(testCase.name.testName, "Switching context to coroutines runTest") }
 
       // Handle timeouts here to avoid the influence of the default timeout set inside runTest
-      val timeout = if (testCase.type == TestType.Container) {
-         testCase.config.timeout
-      } else {
-         testCase.config.resolvedInvocationTimeout
-      }
-
-      runTest(timeout = timeout) {
+      runTest(timeout = testCase.timeout) {
          var additionalContext: CoroutineContext = TestScopeElement(this)
          if (testCase.spec.nonDeterministicTestVirtualTimeEnabled) {
             additionalContext += NonDeterministicTestVirtualTimeEnabled
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TestCoroutineInterceptor.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TestCoroutineInterceptor.kt
@@ -1,12 +1,13 @@
 package io.kotest.engine.test.interceptors
 
 import io.kotest.common.NonDeterministicTestVirtualTimeEnabled
+import io.kotest.core.Logger
 import io.kotest.core.coroutines.TestScopeElement
 import io.kotest.core.test.TestCase
 import io.kotest.core.test.TestResult
 import io.kotest.core.test.TestScope
+import io.kotest.core.test.TestType
 import io.kotest.engine.test.scopes.withCoroutineContext
-import io.kotest.core.Logger
 import kotlinx.coroutines.test.runTest
 import kotlinx.coroutines.withContext
 import kotlin.coroutines.CoroutineContext
@@ -28,7 +29,15 @@ class TestCoroutineInterceptor : TestExecutionInterceptor {
    ): TestResult {
       var result: TestResult = TestResult.Ignored()
       logger.log { Pair(testCase.name.testName, "Switching context to coroutines runTest") }
-      runTest {
+
+      // Handle timeouts here to avoid the influence of the default timeout set inside runTest
+      val timeout = if (testCase.type == TestType.Container) {
+         testCase.config.timeout
+      } else {
+         testCase.config.resolvedInvocationTimeout
+      }
+
+      runTest(timeout = timeout) {
          var additionalContext: CoroutineContext = TestScopeElement(this)
          if (testCase.spec.nonDeterministicTestVirtualTimeEnabled) {
             additionalContext += NonDeterministicTestVirtualTimeEnabled
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
@@ -52,7 +52,7 @@ internal class TimeoutInterceptor(
       } catch (t: CancellationException) {
          if (t is WallclockTimeoutCancellationException || t is TimeoutCancellationException) {
             logger.log { Pair(testCase.name.testName, "Caught timeout $t") }
-            TestResult.Error(mark.elapsedNow(), TestTimeoutException(timeout, testCase.name.testName))
+            TestResult.Error(mark.elapsedNow(), TestTimeoutException(timeout, testCase.name.testName, t))
          } else {
             throw t
          }
@@ -107,5 +107,10 @@ private class WallclockTimeoutCancellationException(message: String) : Cancellat
 /**
  * Exception used for when a test exceeds its timeout.
  */
-open class TestTimeoutException(val timeout: Duration, val testName: String) :
-   Exception("Test '${testName}' did not complete within $timeout")
+open class TestTimeoutException(val timeout: Duration, val testName: String, cause: Throwable? = null) :
+   Exception("Test '${testName}' did not complete within $timeout", cause) {
+
+   @Suppress("unused")
+   @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+   constructor(timeout: Duration, testName: String) : this(timeout, testName, cause = null)
+}
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
@@ -1,10 +1,10 @@
 package io.kotest.engine.test.interceptors
 
+import io.kotest.core.Logger
 import io.kotest.core.test.TestCase
 import io.kotest.core.test.TestResult
 import io.kotest.core.test.TestScope
 import io.kotest.engine.test.scopes.withCoroutineContext
-import io.kotest.core.Logger
 import kotlinx.coroutines.withTimeout
 import kotlin.time.Duration
 import kotlin.time.TimeMark
@@ -23,22 +23,20 @@ internal class TimeoutInterceptor(
       scope: TestScope,
       test: suspend (TestCase, TestScope) -> TestResult
    ): TestResult {
+      val timeout = testCase.config.timeout
 
-      // this timeout applies to the test itself. If the test has multiple invocations then
+      // This timeout applies to the test itself. If the test has multiple invocations, then
       // this timeout applies across all invocations. In other words, if a test has invocations = 3,
-      // each test takes 300ms, and a timeout of 800ms, this would fail, becauase 3 x 300 > 800.
-      logger.log { Pair(testCase.name.testName, "Switching context to add timeout ${testCase.config.timeout}") }
+      // each test takes 300ms, and a timeout of 800ms, this would fail, because 3 x 300 > 800.
+      logger.log { Pair(testCase.name.testName, "Switching context to add timeout $timeout") }
 
-      return when (val timeout = testCase.config.timeout) {
-         null -> test(testCase, scope)
-         else -> try {
-            withTimeout(timeout) {
-               test(testCase, scope.withCoroutineContext(coroutineContext))
-            }
-         } catch (t: Throwable) {
-            logger.log { Pair(testCase.name.testName, "Caught timeout $t") }
-            TestResult.Error(mark.elapsedNow(), TestTimeoutException(timeout, testCase.name.testName))
+      return try {
+         withTimeout(timeout) {
+            test(testCase, scope.withCoroutineContext(coroutineContext))
          }
+      } catch (t: Throwable) {
+         logger.log { Pair(testCase.name.testName, "Caught timeout $t") }
+         TestResult.Error(mark.elapsedNow(), TestTimeoutException(timeout, testCase.name.testName))
       }
    }
 }
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
@@ -5,7 +5,22 @@ import io.kotest.core.test.TestCase
 import io.kotest.core.test.TestResult
 import io.kotest.core.test.TestScope
 import io.kotest.engine.test.scopes.withCoroutineContext
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.DelicateCoroutinesApi
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.selects.select
+import kotlinx.coroutines.test.TestCoroutineScheduler
 import kotlinx.coroutines.withTimeout
+import kotlinx.coroutines.yield
+import kotlin.coroutines.coroutineContext
 import kotlin.time.Duration
 import kotlin.time.TimeMark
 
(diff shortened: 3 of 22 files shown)


kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/test/interceptors/BlockedThreadTimeoutInterceptor.kt
```diff
@@ -84,4 +84,10 @@ internal class BlockedThreadTimeoutInterceptor(
 /**
  * Exception used for when a test exceeds its timeout.
  */
-class BlockedThreadTestTimeoutException(timeout: Duration, testName: String) : TestTimeoutException(timeout, testName)
+class BlockedThreadTestTimeoutException(timeout: Duration, testName: String, cause: Throwable? = null) :
+   TestTimeoutException(timeout, testName, cause) {
+
+   @Suppress("unused")
+   @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+   constructor(timeout: Duration, testName: String) : this(timeout, testName, cause = null)
```
Kantis, line 94: 💯

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TestCoroutineInterceptor.kt
```diff
@@ -26,11 +27,21 @@ class TestCoroutineInterceptor : TestExecutionInterceptor {
...
          withContext(TestScopeElement(this)) {
             result = test(testCase, scope.withCoroutineContext(coroutineContext))
          }
       }
+
+      // Handle timeout here, not in InvocationTimeoutInterceptor
+      // to avoid the influence of default timeout set inside runTest
+      val timeout = testCase.config.invocationTimeout
+      if (timeout != null) {
+         runTest(timeout = timeout, testBody = testBody)
```
Kantis: Would this apply `invocationTimeout` for `TestType.Container` as well?

Kantis (commented):
Would appreciate 2nd opinions from @OliverO2 :) 

I think this overall looks good. I'm just wondering about if this breaks things for slow containers which have previously been ignored by `invocationTimeout`? On the other hand, doesn't it make sense to also prevent the test runner from hanging indefinitely if a container gets stuck somehow? 🤔

OliverO2:
Will do. Don't move 😉!

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/resolveConfig.kt
```diff
@@ -72,7 +72,11 @@ internal fun resolveConfig(
       threads = threads,
       invocations = invocations,
       timeout = timeout,
-      invocationTimeout = invocationTimeout,
+      // note: the invocation timeout cannot be larger than the test case timeout
+      invocationTimeout = when {
+         timeout != null && invocationTimeout != null -> invocationTimeout.coerceAtMost(timeout)
```
sksamuel: Should we just error here so the user is clear it's setup wrong ?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
```diff
@@ -32,19 +46,67 @@ internal class TimeoutInterceptor(
...
+         result
+      }
+      timeoutJob.onJoin {
+         blockDeferred.cancel()
+         throw KotestTimeoutCancellationException("Timed out waiting for $timeout")
+      }
+   }
+}
+
+private class KotestTimeoutCancellationException(message: String) : CancellationException(message)
```
sksamuel: could we add a comment on why we've added our own exception type

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TestCoroutineInterceptor.kt
osipxd: Oh, I completely forgot to check the test type. Thank you! I've added this check.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/resolveConfig.kt
osipxd: Done! But I'm not sure if I throw a correct exception (`IllegalStateException`) and if I do it in the proper place.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
osipxd: Added 👌

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
```diff
@@ -32,19 +46,68 @@ internal class TimeoutInterceptor(
...
          null -> test(testCase, scope)
          else -> try {
-            withTimeout(timeout) {
+            withAppropriateTimeout(timeout) {
                test(testCase, scope.withCoroutineContext(coroutineContext))
             }
-         } catch (t: Throwable) {
-            logger.log { Pair(testCase.name.testName, "Caught timeout $t") }
-            TestResult.Error(mark.elapsedNow(), TestTimeoutException(timeout, testCase.name.testName))
+         } catch (t: CancellationException) {
```
osipxd: Was there some reason to catch `Throwable` instead of a more specific exception type like `CancellationException` or `TimeoutCancellationException`?

OliverO2: Catching `Throwable` seems like a mistake to me. Shouldn't we be even more specific, catching just `TimeoutCancellationException` and `KotestTimeoutCancellationException` (which might be named `WallclockTimeoutCancellationException` to better communicate intent)?

OliverO2: OK, I see that narrowing the Cancellation types has been done inside the catch block.

OliverO2 (commented):
Overall looks good to me so far.

Still needs some more tests for final review, right?
And wouldn't this also fix #3969?

osipxd:
> Still needs some more tests for final review, right?

Yes, I hope I will have time to write these tests this week

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TestCoroutineInterceptor.kt
```diff
@@ -26,11 +28,21 @@ class TestCoroutineInterceptor : TestExecutionInterceptor {
...
       }
+
+      // Handle invocation timeout here, not in InvocationTimeoutInterceptor
+      // to avoid the influence of default timeout set inside runTest
+      val timeout = testCase.config.invocationTimeout
+      if (timeout != null && testCase.type != TestType.Container) {
+         runTest(timeout = timeout, testBody = testBody)
+      } else {
+         runTest(testBody = testBody)
+      }
```
OliverO2: I know that this is unrelated to this PR, but this `runTest` invocation occurs in common code, and its test result is discarded instead of being returned immediately to the encompassing (JS) test framework. [This is described to cause losing test errors.](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-test-result/)

osipxd: As I know, currently, it is impossible to use `TestCoroutinesInterceptor` with JS. It is added only if this condition is true:
```kotlin
val useCoroutineTestScope = platform != Platform.JS && testCase.config.coroutineTestScope
```
So it doesn't affect JS tests yet.

But I agree this issue should be addressed before `coroutineTestScope` is enabled for JS.

OliverO2: That makes sense. Thanks for pointing it out. Maybe you could grab the opportunity and [add a comment why `coroutineTestScope` is excluded on JS](https://github.com/kotest/kotest/blob/85f6370c608c955030260dd14d393730b3834829/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt#L83).

osipxd: Should I also create an issue to track this change?

OliverO2: That would be super nice!

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
```diff
@@ -32,19 +46,68 @@ internal class TimeoutInterceptor(
...
+
+   select {
+      blockDeferred.onAwait { result ->
+         timeoutJob.cancel()
+         result
+      }
+      timeoutJob.onJoin {
+         blockDeferred.cancel()
+         throw WallclockTimeoutCancellationException("Timed out waiting for $timeout")
+      }
```
osipxd: It doesn't work 🤔
The exception seems to be ignored. I've tried several other workarounds with no luck. So I need some time to investigate the reason.

osipxd (commented):
I've added tests.
Force-pushed the branch after it, so here is the [link to the correct diff](https://github.com/kotest/kotest/compare/64e26c0f80e3516e9908d869d61c627ce5e7d0de..4fafbe454e174654f93a40e6cb92cc71ea97557f)

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TestCoroutineInterceptor.kt
osipxd: Here it is: https://github.com/kotest/kotest/issues/4077

osipxd: Added this comment in a separate PR: https://github.com/kotest/kotest/pull/4078
To make this change visible in changelogs.

OliverO2:
@osipxd Could you please avoid [prematurely squashing/rebasing and force pushing your PRs](https://jacobtomlinson.dev/posts/2022/dont-prematurely-squash/rebase-and-force-push-your-prs/)? This basically invalidates all prior reviewer efforts and delays the process. Everything will be squashed on merge anyway.

osipxd:
> @osipxd Could you please avoid [prematurely squashing/rebasing and force pushing your PRs](https://jacobtomlinson.dev/posts/2022/dont-prematurely-squash/rebase-and-force-push-your-prs/)?

Oh, I'm sorry. I will not rebase this PR anymore. Here is [the link to the correct diff](https://github.com/kotest/kotest/compare/64e26c0f80e3516e9908d869d61c627ce5e7d0de..4fafbe454e174654f93a40e6cb92cc71ea97557f)

OliverO2:
Thanks! Your PRs are appreciated and avoiding force-pushes would help us to expedite the review process. Even if there have been no reviews yet, someone may already have looked into the PR at some stage, so incremental commits are important for tracking reviewed changes.

Do you think this one is ready for final review?

osipxd:
> Do you think this one is ready for final review?

Not yet. I haven't solved [the problem](https://github.com/kotest/kotest/pull/4043#discussion_r1631474161) with timeout ignore in `TimeoutInterceptor`.

sksamuel:
would be fantastic to get this over the line

osipxd:
@OliverO2, I've rebased this PR because almost everything has been changed since the last review, so it feels easier to review from scratch instead of reviewing diffs.

It is better to review commit by commit.

Things to be discussed:
- `BlockedThreadTimeoutInterceptor` has something in common with the logic I've added to `TimeoutInterceptor`. Maybe these two approaches could be merged into one
- `runTest` handles timeout itself and throws `UncompletedCoroutinesError`. This exception is a subclass of `AssertionError` so it is treated as a **failure**, compared to `TimeoutCancellationException` treated as an **error**.
- Container timeout is handled twice - in `TimeoutInterceptor` and `runTest` (`TestCoroutineInterceptor`). Maybe we can handle it only in `runTest`?

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/ResolvedTestConfig.kt
```diff
@@ -30,13 +32,13 @@ data class ResolvedTestConfig(
...
     */
-   val timeout: Duration?,
+   val timeout: Duration,
 
    /**
     * This timeout applies to individual invocations of a test case. If invocations is 1, then this
     * has the same effect as timeout. To set a timeout across all invocations then see [timeout].
     */
-   val invocationTimeout: Duration?,
+   val invocationTimeout: Duration,
```
osipxd, line 41: As I can see it is safe to make these fields non-nullable. The only place where these fields were `null` was `ResolvedTestConfig.default`, used only for Kotest tests and always replaced by a resolved configuration in real test execution.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/ResolvedTestConfig.kt
```diff
@@ -92,8 +97,8 @@ data class ResolvedTestConfig(
          assertSoftly = false,
          blockingTest = false,
          extensions = emptyList(),
-         timeout = null,
-         invocationTimeout = null,
+         timeout = Defaults.defaultTimeoutMillis.milliseconds,
+         invocationTimeout = Defaults.defaultInvocationTimeoutMillis.milliseconds,
```
AlexCue987, line 98: the following line says "miliseconds" twice: `Defaults.defaultInvocationTimeoutMillis.milliseconds` - is that redundant?

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/ResolvedTestConfig.kt
```diff
@@ -81,6 +83,9 @@ data class ResolvedTestConfig(
       require(threads > 0) { "Number of threads must be greater than 0" }
    }
 
+   // note: the invocation timeout can't be larger than the test case timeout
+   val resolvedInvocationTimeout: Duration = minOf(timeout, invocationTimeout)
```
osipxd: Feel free to propose a better naming for this field. Maybe, it should be internal.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/ResolvedTestConfig.kt
osipxd: The only problem is the [binary compatibility](https://jakewharton.com/public-api-challenges-in-kotlin/#binary-compatibility) of the data classes. I can provide a secondary constructor, copy method and destructuring operators with old signatures to maintain binary compatibility, but it is a pain.

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/timeout/ContainerTimeoutTest.kt
```diff
@@ -1,38 +1,42 @@
...
-         val collector = CollectingTestEngineListener()
-
-         val duration = testTimeSource().measureTime {
             TestEngineLauncher(collector)
+               .withConfiguration(config)
                .withClasses(NestedTimeout::class)
-               .async()
-         }
-
-         duration shouldBe 100.milliseconds
```
osipxd, line 28: After removing `coroutineTestScope` from the Spec itself, duration is not deterministic anymore, that's why I decided to remove this assertion.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TestCoroutineInterceptor.kt
```diff
@@ -28,7 +29,15 @@ class TestCoroutineInterceptor : TestExecutionInterceptor {
    ): TestResult {
       var result: TestResult = TestResult.Ignored()
       logger.log { Pair(testCase.name.testName, "Switching context to coroutines runTest") }
-      runTest {
+
+      // Handle timeouts here to avoid the influence of the default timeout set inside runTest
+      val timeout = if (testCase.type == TestType.Container) {
```
AlexCue987: should we delegate this logic to `testCase` so that we avoid its tight coupling with this interseptor?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
```diff
@@ -23,28 +38,79 @@ internal class TimeoutInterceptor(
...
-
-      return when (val timeout = testCase.config.timeout) {
-         null -> test(testCase, scope)
-         else -> try {
-            withTimeout(timeout) {
-               test(testCase, scope.withCoroutineContext(coroutineContext))
-            }
-         } catch (t: Throwable) {
+      // each test takes 300ms, and a timeout of 800ms, this would fail, because 3 x 300 > 800.
+      logger.log { Pair(testCase.name.testName, "Switching context to add timeout $timeout") }
```
AlexCue987, line 46: Nit: do we need a `Pair` to construct a text message to log?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/timeout/CoroutinesTimeoutTest.kt
```diff
@@ -88,3 +95,20 @@ private suspend fun someCoroutine() {
       }
    }
 }
+
+/** Delay using real (non-virtual) time. */
+suspend fun wallclockDelay(duration: Duration) {
```
AlexCue987: should we rename this function to something like `nonVirtualTimeDelay` to design away the need for this comment?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
```diff
@@ -23,28 +38,79 @@ internal class TimeoutInterceptor(
...
+): T {
+   return if (coroutineContext[TestCoroutineScheduler] != null) {
+      // withTimeout uses virtual time, which will hang.
+      withWallclockTimeout(timeout, block)
+   } else {
+      withTimeout(timeout, block)
+   }
+}
+
+private suspend fun <T> withWallclockTimeout(
```
AlexCue987: this function is quite large. does it have its own unit tests?

AlexCue987 (commented):
LGTM, some optional cosmetic suggestions

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/ResolvedTestConfig.kt
osipxd: Yes, `Defaults.defaultInvocationTimeoutMillis` is `Long`, while I need `Duration` here.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
osipxd: `Logger.log` takes `() -> Pair<String?, String>` as a parameter, so I think yes

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TimeoutInterceptor.kt
osipxd: It is tricky to test it. I have only a test `ContainerTimeoutOverriddenMainDispatcherTest` failing without this function, but I'll try to write a separate test.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/test/config/ResolvedTestConfig.kt
osipxd: Removed this field

AlexCue987 (approved):

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/timeout/ContainerTimeoutOverriddenMainDispatcherTest.kt
```diff
@@ -0,0 +1,45 @@
...
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.test.setMain
+import kotlinx.coroutines.withContext
+import kotlin.time.Duration.Companion.hours
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.minutes
+
+// Issue: https://github.com/kotest/kotest/issues/3703
+@OptIn(ExperimentalCoroutinesApi::class)
+class ContainerTimeoutOverriddenMainDispatcherTest : FunSpec({
```
OliverO2, line 19:
```suggestion
// Issue: https://github.com/kotest/kotest/issues/3703
@DoNotParallelize
@OptIn(ExperimentalCoroutinesApi::class)
class ContainerTimeoutOverriddenMainDispatcherTest : FunSpec({
```

Using/changing `Dispatchers.Main` depends on a global object. This does not play well with any type of concurrency. `@DoNotParallelize` will run this test in isolation, so that it cannot interfere with possible tests elsewhere which also happen to use/change `Dispatchers.Main`.

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/timeout/CoroutinesTimeoutOverrideTest.kt
```diff
@@ -0,0 +1,35 @@
+package com.sksamuel.kotest.engine.test.timeout
+
+import io.kotest.core.spec.style.FunSpec
+import kotlin.time.Duration.Companion.milliseconds
+
+class CoroutinesTimeoutOverrideTest : FunSpec({
```
OliverO2, line 6:
```suggestion
@DoNotParallelize
class CoroutinesTimeoutOverrideTest : FunSpec({
```

`System.setProperty` affects global state.

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/timeout/CoroutinesTimeoutTest.kt
```diff
@@ -88,3 +95,20 @@ private suspend fun someCoroutine() {
       }
    }
 }
+
+suspend fun nonVirtualTimeDelay(duration: Duration) {
```
OliverO2, line 99: At your discretion: Would `realTimeDelay` be a better fit for our human wetware, avoiding negativity?

OliverO2 (commented):
Thanks for the huge effort!

Some general aspects:
* Mixing virtual and real time (timeouts) is hard to reason about. I've tried to spot potential issues, but I'm not sure that I have a full understanding of all the implications.
* Normally, I would not expect a test running on virtual time to require any real-time timeout, as such tests typically complete quickly.
* Some real-time timeouts introduced or changed in this PR may result in flaky tests, in particular on our slow CI runners (macOS, Windows). We have switched most time-dependent tests to virtual time to avoid that. To stabilize tests with real-time timeouts may require large time intervals (seconds), which slows down tests. See https://github.com/kotest/kotest/issues/4113.

Hope this helps!

osipxd:
Oh, I've just noticed that I accidentally deleted the fork. I can’t reopen this PR even after restoring the repository, so I'm going to open a new one.

OliverO2:
I've also tried to re-open, but to no avail. It's 2024, and there's no Undo? Come on, GitHub! 🙃

osipxd:
Created: https://github.com/kotest/kotest/pull/4329

Fixed comments in the latest two commits.

PR #4046, opened 2024-05-31
Json path array match 

AlexCue987:
detect when index in json array is too high, find an existing element with lower index, such as
```kotlin
"Failure message states if key is missing, shows valid subpath with lower array index" {
      shouldFail {
         json.shouldContainJsonKeyValue("$.store.book[1].tags[2]", "humor")
      }.message shouldBe """
         Expected given to contain json key <'$.store.book[1].tags[2]'> but key was not found. Found valid path: <'$.store.book[1].tags[1]'>
      """.trimIndent()
   }
```


kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/keyvalues.kt
```diff
@@ -69,7 +69,7 @@ inline fun <reified T> containJsonKeyValue(path: String, t: T) = object : Matche
           }
           is JsonPathNotFound -> {
              val validSubPathDescription = findValidSubPath(value, path)?.let { subpath ->
-                " Found shorter valid subpath: <'$subpath'>"
+                " Found valid path: <'$subpath'>"
```
Kantis:
```suggestion
                " Closest valid path: <'$subpath'>"
```
Maybe?

kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/ContainJsonKeyValueTest.kt
```diff
@@ -55,11 +56,19 @@ class ContainJsonKeyValueTest : StringSpec({
       """.trimIndent()
    }
 
+   "Failure message states if key is missing, shows valid subpath with lower array index" {
+      shouldFail {
+         json.shouldContainJsonKeyValue("$.store.book[1].tags[2]", "humor")
+      }.message shouldBe """
+         Expected given to contain json key <'$.store.book[1].tags[2]'> but key was not found. Found valid path: <'$.store.book[1].tags[1]'>
```
Kantis, line 63: I'm wondering if we should take an alternate approach here?

If we're looking for an array element, and we fail to find that subpath, couldn't we just try to get an `Array<Any> using JsonPath for that path? If successful, we print: 
```
Expected array at path <'$currentSubPath'> to have $expectedElementCount, but actual count is 
$actualElementCount
```

If we fail to get an array at all, we could immediately remove last part of the path and move on.

AlexCue987: > I'm wondering if we should take an alternate approach here?
> 
> If we're looking for an array element, and we fail to find that subpath, couldn't we just try to get an `Array using JsonPath for that path? If successful, we print:
> 
> ```
> Expected array at path <'$currentSubPath'> to have $expectedElementCount, but actual count is 
> $actualElementCount
> ```
> 
> If we fail to get an array at all, we could immediately remove last part of the path and move on.

makes sense. will do

Kantis:
@AlexCue987 the self-assign was related to reviewing, not implementing. Just in case you've been holding off on this PR due to that. :)

AlexCue987:
> @AlexCue987 the self-assign was related to reviewing, not implementing. Just in case you've been holding off on this PR due to that. :)

you got it right. resuming work on this thing

AlexCue987:
re-implemented as suggested by @Kantis : https://github.com/kotest/kotest/pull/4118

PR #4047, opened 2024-05-31
Lazily apply pom details

Kantis:
I noticed that POM details were not applied correctly when releasing the Android native and watchosDeviceArm64. I suppose it typically works when doing the full release for _reasons_. 

Proposed change will make application of POM details happen lazily, so only when publishing actually comes into play. Meaning it'll work for any potential publication that is registered late will still get the correct stuff. At least that's my interpretation of things, but gradle typically work in mysterious ways.

I think we can keep this PR as a draft for a while. If the same issue arises when doing a full release next time we could consider merging.

### DIFF
--- a/buildSrc/src/main/kotlin/kotest-publishing-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotest-publishing-conventions.gradle.kts
@@ -119,33 +119,30 @@ publishing {
       }
    }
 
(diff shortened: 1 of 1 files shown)


sksamuel:
You should add yourself as a developer to the pom too. It's been a long time since it was just me :)

Kantis:
Duplicated and superseded by recent work from @aSemy

PR #4048, opened 2024-05-31
Lazily dealing with eq-results

### LINKED ISSUE #4041: Significant Performance bug with "shouldNotBe" in Property Tests
5.9.0

**Description**
When using the `shouldNotBe` in a medium-large testset for a property test, it takes unreasonably long. This is very likely a bug, as the performance issue does not occur when using `shouldBe` or even `shouldNotBeIn setOf(<the item>)` in the same setup.

**How to reproduce**
Make an exhaustive containing a large amount of strings (i.e. 50.000x "test"), call .checkAll on it and assert `shouldNotBe "notMe"` and see the slow performance. Then replace the assertion with `shouldNotBeIn setOf("notMe")` and observe the performance be fine again. Using `shouldBe "test"` also works fine.

**Stats**
When using `shouldNotBe`, the Test takes 2.5 seconds to run. With `shouldNotBeIn`, the Test takes ~100ms (25x less!). When profiling just the difference between the statements (and ignoring other test setup and overhead), we get roughly 50ms vs 2200ms in the profiler.
If I add additional `shouldNotBe`-Statements inside the `checkall` clause, the tests takes proportionally longer, so the `shouldNotBe` is a likely culprit and not other test mechanics/overheads.

**Conclusion**
I am not familiar with the inner workings, but intuitively I would guess the issue might be with the negation of the testcase potentially throwing+catching exceptions internally?

Kantis:
We can defer the cost of creating failure messages and throwing exceptions until as late as possible. When negating/inverting matchers, we actually want the failure and in those cases the cost of generating messages+exceptions is spent in vain.

Also deferring cleaning of stacktraces to top-level saves a lot when dealing with negated matchers.

Fixes #4041

---

- [ ] `EqResult` and `EqualityResult` should probably be merged into a single class before merging, since this needs to go in a major release either way
- [ ] IterableEq is probably going away / changing in 6.0? Check what needs to be done there...


AlexCue987:
IMO same thing should be done in several other places, such as collection matchers: if the expected outcome is failure, they shouldn't waste time describing differences

sksamuel:
What do we want to do with this ?

AlexCue987:
this problem persists in multiple places. this PR exposes just the tip of this iceberg. we can start making similar changes elsewhere

Kantis:
Agreed. I will close this PR. I tried rebasing it onto latest master but it's just too many conflicts with things I don't know anything about.

PR #4051, opened 2024-06-01
Cleaning stacktraces later

### LINKED ISSUE #4041: Significant Performance bug with "shouldNotBe" in Property Tests

Kantis:
When we throw an AssertionError and the matcher is actually used by an inverted matcher (i.e. we expect a failure), then the stacktrace is irrelevant.

Partially fixes #4041 
#4048 fixes rest, but this fix can go into next patch and fixes 2/3rds of the performance costs


sksamuel (approved):

sksamuel:
this good to go ?

PR #4053, opened 2024-06-02
Refactor team city listener to extract the writing methods

sksamuel:


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/teamcity/TeamCityWriter.kt
```diff
@@ -0,0 +1,178 @@
...
+import io.kotest.core.test.TestCase
+import io.kotest.core.test.TestResult
+import io.kotest.engine.test.names.FallbackDisplayNameFormatter
+import io.kotest.mpp.Logger
+import kotlin.reflect.KClass
+
+/**
+ * [TeamCityWriter] handles outputting in the team city format using a [TeamCityMessageBuilder].
+ */
+class TeamCityWriter(
```
Kantis: internal?

Kantis (approved):
Looks good :) 

When is the TC listener used? 
If we were to write our own gradle test plugin/tasks which adds the TC listener, would that mean we could get navigation to sources for nested tests? (Since we would not be hamstrung by Gradle's reliance on 'all tests are methods'?)

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/teamcity/TeamCityWriter.kt
sksamuel: I'll make it internal.

sksamuel:
> Looks good :)
> 
> When is the TC listener used? If we were to write our own gradle test plugin/tasks which adds the TC listener, would that mean we could get navigation to sources for nested tests? (Since we would not be hamstrung by Gradle's reliance on 'all tests are methods'?)

It's used right now by the kotest plugin.

I'm not sure where the issue is with the navigation as it doesn't work for the non-gradle runs too. I notice that there's two methods we're not implementing in the plugin's SMTestLocator. Maybe that is the missing ingredient. I'll play about. If that fixes it then yes, moving to our own test task should allow full navigation.

sksamuel:
> > Looks good :)
> > When is the TC listener used? If we were to write our own gradle test plugin/tasks which adds the TC listener, would that mean we could get navigation to sources for nested tests? (Since we would not be hamstrung by Gradle's reliance on 'all tests are methods'?)
> 
> It's used right now by the kotest plugin.
> 
> I'm not sure where the issue is with the navigation as it doesn't work for the non-gradle runs too. I notice that there's two methods we're not implementing in the plugin's SMTestLocator. Maybe that is the missing ingredient. I'll play about. If that fixes it then yes, moving to our own test task should allow full navigation.

I am able to get it to work now when we use the plugin and don't run the tests via gradle. Although running the full test even with the plugin goes via gradle. But now we know that gradle is the issue.

PR #4054, opened 2024-06-02
Removes need for test type differentiation in team city output

sksamuel:
Builds on #4053.

By not needing the test type enum in the TC listener, we have the possibility of removing test type entirely.
We would have to remove IsolationMode.InstancePerLeaf (since we don't know in advance if something is a leaf or not).

But it would mean we can remove StringSpec (as FreeSpec would do the same).
And it would mean we could arbitrarily nest property tests and data tests.


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/TeamCityTestEngineListener.kt
```diff
@@ -33,26 +31,11 @@ class TeamCityTestEngineListener(
    private var writer = TeamCityWriter(prefix, FallbackDisplayNameFormatter.default(ProjectConfiguration()))
 
    // once a spec has completed, we want to be able to check whether any given test is
-   // a container or a leaf test, and so this map contains all test that have children
-   private val children = mutableMapOf<Descriptor, MutableList<TestCase>>()
-
-   private val results = mutableMapOf<Descriptor, TestResult>()
+   // a suite or not, and so this map marks all tests that have children
+   private val parents = mutableSetOf<TestCase>()
```
Kantis, line 35: Doesn't this need to be a thread-safe type? Multiple tests in the same spec might run simultaneously if `concurrentTests > 1`

Will there be 1 TC-listener per Spec? Otherwise clearing this on specFinished sounds like it would break on any parallelism. Maybe we should maintain a map of Spec -> parents?

sksamuel: That's already how it works, this listener is wrapped in a thread safe listener. There's a comment to that effect in this file.

LeoColman (approved):

PR #4060, opened 2024-06-04
Add an `infix` variant of `shouldNotBeNull` that accepts a block

sschuberth:
This allows for some syntactic sugar when dealing with the non-null value.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/ShouldNotBeNullTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/ShouldNotBeNullTest.kt
@@ -50,6 +50,16 @@ class ShouldNotBeNullTest : WordSpec() {
                .shouldStartWith("fo")
                .shouldEndWith("oo")
          }
+
+         "accept a block" {
+            val a: String? = "foo"
+
+            a shouldNotBeNull {
+               shouldHaveLength(3)
+               shouldStartWith("fo")
+               shouldEndWith("oo")
+            }
+         }
       }
    }
 }
--- a/documentation/docs/assertions/core.md
+++ b/documentation/docs/assertions/core.md
@@ -27,7 +27,8 @@ Matchers provided by the `kotest-assertions-core` module.
 | `obj.shouldBeTypeOf<T>()`                   | Asserts that the given reference is exactly of type T. Subclass will fail. Ie, `1 should beOfType<Number>` would fail because although 1 _is_ a Number, the runtime type is not Number. |
 | `obj.shouldBeInstanceOf<T>()`               | Asserts that the given reference is of type T or a subclass of T. |
 | `obj.shouldHaveAnnotation(annotationClass)` | Asserts that the object has an annotation of the given type. |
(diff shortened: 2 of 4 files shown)


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/ShouldNotBeNullTest.kt
```diff
@@ -50,6 +50,16 @@ class ShouldNotBeNullTest : WordSpec() {
...
+
+         "accept a block" {
+            val a: String? = "foo"
+
+            a shouldNotBeNull {
+               shouldHaveLength(3)
+               shouldStartWith("fo")
+               shouldEndWith("oo")
+            }
+         }
```
Kantis, line 62: It's currently possible to do either:

```kotlin
// Typical method chaining
            a.shouldNotBeNull()
              .shouldHaveLength(3)
              .shouldStartWith("fo")
              .shouldEndWith("oo")

// or, using Kotlin features
            a.run {
               shoudNotBeNull()
               shouldHaveLength(3)
               shouldStartWith("fo")
               shouldEndWith("oo")
            }
```

Are there other use-cases where this syntax provides a bigger difference?

sschuberth: > Are there other use-cases where this syntax provides a bigger difference?

I don't thinks so. And you're right in saying that this is just syntactic sugar over using

    .shouldNotBeNull().run {

Mine is just slightly shorter an (arguably) a tad more readable due to the infix notation.

Kantis: You can do `run` before even? `shouldNotBeNull` has a compiler contract which makes the compiler narrow the type to non-null if `shouldNotBeNull` passes

sschuberth: > You can do `run` before even?

Not sure what you mean, can you give an example?

> `shouldNotBeNull` has a compiler contract

Yes, I'm aware of that. However, that contract does not work if `.shouldNotBeNull()` is used on a property that "is a public API property declared in a different module".

In my current case,

    receiveResponse.messages.shouldNotBeNull()
    receiveResponse.messages.forEach {  } // Smart cast impossible here.

does not work for that reason. Thus I prefer a syntax with a block in whose scope the property becomes non-nullable. So, 

    receiveResponse.messages.shouldNotBeNull().run { forEach {} }

works, but still

    receiveResponse.messages shouldNotBeNull { forEach {} }

reads a bit nicer, IMO.

Kantis (approved):

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/ShouldNotBeNullTest.kt
Kantis: Thanks, that's a good point 🙂

PR #4062, opened 2024-06-04
Build Config update & enable Configuration Cache

aSemy:
Minor build config updates, aiming to improve build configuration speed and stability.

- Re-order repositories by stability, and add snapshotsOnly() filter to Sonatype snapshots.
- Update Shadow plugin to 8.1.1 (better config cache support)
- remove deprecated `kotlin { targets {} }` config
- update gradle.properties
  - enable configuration cache - should work fine with KGP 1.9
  - re-enable KN compiler daemon - again, it should be more stable
  - remove ignoreIncorrectDependencies (it doesn't seem to be triggered any more - but if it was, it's an important warning so it shouldn't be ignored)
  - sort & group properties
  - remove `kotlin.code.style` - I don't think it does anything, and besides, the project has an .editorconfig file
- remove nested `gradle.properties` files (it's not clear whether it works, and the only option in the files was `org.gradle.parallel=false` - which only makes sense as a project-wide setting)
- Introduce BuildService for limiting parallel publishing, because Maven Central can't handle parallel uploads.
- Add a delay in `kotest-tests-spec-parallelism` test to try and make it more stable.
- Comment out non-functional `buildConfigDocs` task.
- Update GitHub workflows to make the Gradle flags closer to what's in the project's `gradle.properties`.

### DIFF
--- a/.github/workflows/release_base.yml
+++ b/.github/workflows/release_base.yml
@@ -17,7 +17,7 @@ env:
    OSSRH_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
    ORG_GRADLE_PROJECT_signingKey: ${{ secrets.SIGNING_KEY }}
    ORG_GRADLE_PROJECT_signingPassword: ${{ secrets.SIGNING_PASSWORD }}
-   GRADLE_OPTS: -Dorg.gradle.configureondemand=true -Dorg.gradle.parallel=false -Dkotlin.incremental=false -Dorg.gradle.jvmargs="-Xmx12g -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8"
+   GRADLE_OPTS: -Dorg.gradle.jvmargs="-Xmx12g -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8"
 
 permissions:
    contents: read
--- a/kotest-framework/kotest-framework-api/build.gradle.kts
+++ b/kotest-framework/kotest-framework-api/build.gradle.kts
@@ -40,59 +40,65 @@ kotlin {
    }
 }
 
-tasks.create("buildConfigDocs") {
-   // find config files
-   val fileNames = listOf("KotestEngineProperties.kt")
-
-   val foundFiles = File(project.rootDir.absolutePath).walk().maxDepth(25).map { file ->
-      if (fileNames.contains(file.name)) {
-         file
-      } else {
-         null
-      }
-   }.filterNotNull()
-      .toList()
-
-   if (foundFiles.size != fileNames.size)
-      throw RuntimeException("Fail to find files -> {$fileNames} in project, found only these files -> {$foundFiles}")
-
-   // replace in docs
-
-   val docName = "config_props.md"
-   val docsFolder = File(project.rootDir.absolutePath, "documentation/docs/framework")
-   val docFileFullPath = File(docsFolder.absolutePath, docName)
-
-   val configTemplate = """
----
-id: framework_config_props
-title: Framework configuration properties
-sidebar_label: System properties
-slug: framework-config-props.html
----
-
-   """.trimIndent()
-
-   val fileTemplate = """
-
-      ---
-      #### %s
-      ```kotlin
-      %s
-      ```
-
-   """.trimIndent()
-
-   val sb = StringBuilder(configTemplate)
-
-   foundFiles.forEach { file ->
-      val name = file.name
-      // intentionally use \n instead of System.lineSeparator to respect .editorconfig
-      val content = file.readLines().joinToString(separator = "\n")
-
-      sb.append(fileTemplate.format(name, content))
-   }
-
-   docFileFullPath.writeText(sb.toString())
-}
-
-tasks["jvmTest"].mustRunAfter(tasks["buildConfigDocs"].path)
+// TODO reimplement buildConfigDocs task
+//      - It is supposed to update `config_props.md` files, but it's not triggered by anything.
+//        (Replace `mustRunAfter()` with `dependsOn()`)
+//      - It does work in the configuration phase.
+//        (Add a `doLast {}` block)
+//      - Should properly register Task inputs/outputs.
+//tasks.create("buildConfigDocs") {
+//   // find config files
+//   val fileNames = listOf("KotestEngineProperties.kt")
+//
+//   val foundFiles = File(project.rootDir.absolutePath).walk().maxDepth(25).map { file ->
+//      if (fileNames.contains(file.name)) {
+//         file
+//      } else {
+//         null
+//      }
+//   }.filterNotNull()
+//      .toList()
+//
+//   if (foundFiles.size != fileNames.size)
+//      throw RuntimeException("Fail to find files -> {$fileNames} in project, found only these files -> {$foundFiles}")
+//
+//   // replace in docs
+//
+//   val docName = "config_props.md"
+//   val docsFolder = File(project.rootDir.absolutePath, "documentation/docs/framework")
+//   val docFileFullPath = File(docsFolder.absolutePath, docName)
+//
+//   val configTemplate = """
+//---
+//id: framework_config_props
+//title: Framework configuration properties
+//sidebar_label: System properties
+//slug: framework-config-props.html
+//---
+//
+//   """.trimIndent()
+//
+//   val fileTemplate = """
+//
+//      ---
+//      #### %s
+//      ```kotlin
+//      %s
+//      ```
+//
+//   """.trimIndent()
+//
+//   val sb = StringBuilder(configTemplate)
+//
+//   foundFiles.forEach { file ->
+//      val name = file.name
+//      // intentionally use \n instead of System.lineSeparator to respect .editorconfig
+//      val content = file.readLines().joinToString(separator = "\n")
+//
+//      sb.append(fileTemplate.format(name, content))
+//   }
+//
+//   docFileFullPath.writeText(sb.toString())
+//}
+//
+//tasks["jvmTest"].mustRunAfter(tasks["buildConfigDocs"].path)
--- a/kotest-framework/kotest-framework-standalone/build.gradle.kts
+++ b/kotest-framework/kotest-framework-standalone/build.gradle.kts
@@ -5,12 +5,6 @@ plugins {
    alias(libs.plugins.shadowjar)
 }
 
-kotlin {
-   targets {
-      jvm()
-   }
-}
-
 application {
    mainClass.set("io.kotest.engine.launcher.MainKt")
 }
--- a/kotest-framework/kotest-framework-standalone/build.gradle.kts
+++ b/kotest-framework/kotest-framework-standalone/build.gradle.kts
@@ -29,7 +29,11 @@ tasks {
       }
    }
    getByName("jvmJar") {
-      finalizedBy(getByName("shadowJar"))
+      finalizedBy(shadowJar)
+   }
+
+   startScripts {
+      dependsOn(shadowJar)
    }
 }
 
--- a/kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
+++ b/kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
@@ -5,6 +5,7 @@ import io.kotest.core.config.ProjectConfiguration
 import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.TimeMark
 import kotlin.time.TimeSource
+import kotlin.time.times
 
 object ProjectConfig : AbstractProjectConfig() {
 
@@ -21,9 +22,9 @@ object ProjectConfig : AbstractProjectConfig() {
 
    override suspend fun afterProject() {
       val duration = start.elapsedNow()
-      // There are 8 specs, and each one has a 100ms delay.
+      // There are 8 specs, and each one has a 500ms delay.
       // If parallel is working, they should all block at the same time.
-      if (duration > 700.milliseconds) {
+      if (duration > 7 * 500.milliseconds) {
          error("Parallel execution failure: Execution time was $duration")
       }
    }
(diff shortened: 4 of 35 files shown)


OliverO2:
I know it's coincidentally related, but after two failed runs it seems we'd have to relax the timing expectations in this case: https://github.com/kotest/kotest/actions/runs/9372917524/job/25808214714?pr=4062#step:5:1833

```
> Task :kotest-tests:kotest-tests-spec-parallelism:jvmTest
After Project Error[jvm]


  Test initializationError[jvm] FAILED
9 tests completed, 1 failed

  java.lang.IllegalStateException: Parallel execution failure: Execution time was 899
```

aSemy:
> I know it's coincidentally related, but after two failed runs it seems we'd have to relax the timing expectations in this case: https://github.com/kotest/kotest/actions/runs/9372917524/job/25808214714?pr=4062#step:5:1833
> 
> ```
> > Task :kotest-tests:kotest-tests-spec-parallelism:jvmTest
> After Project Error[jvm]
> 
> 
>   Test initializationError[jvm] FAILED
> 9 tests completed, 1 failed
> 
>   java.lang.IllegalStateException: Parallel execution failure: Execution time was 899
> ```

Hmmmm interesting!

Enabling configuration cache means that tasks within a single project can run in parallel, but these 8 tests are run in a single task. So that shouldn't have an impact.

Possibly it's something to do with forking the test process...

kotest-framework/kotest-framework-api/build.gradle.kts
```diff
@@ -40,59 +40,65 @@ kotlin {
...
-   docFileFullPath.writeText(sb.toString())
-}
-
-tasks["jvmTest"].mustRunAfter(tasks["buildConfigDocs"].path)
+// TODO reimplement buildConfigDocs task
+//      - It is supposed to update `config_props.md` files, but it's not triggered by anything.
+//        (Replace `mustRunAfter()` with `dependsOn()`)
+//      - It does work in the configuration phase.
+//        (Add a `doLast {}` block)
+//      - Should properly register Task inputs/outputs.
```
aSemy, line 48: The `buildConfigDocs` task is not correctly defined, so it never ran.

Is it still needed? If so, I can update it in a separate PR. But for now, I've disabled it.

kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
aSemy, line 1: I've refactored this test to try to make it more stable.

https://github.com/kotest/kotest/pull/4062#issuecomment-2148385498

When I ran `./gradlew jvmTest` I also saw the same error as on CI. This change fixes it. But please check my work!

.github/workflows/release_base.yml
aSemy, line 1: I think the release workflows could be simplified.

* Use a single machine to publish everything - `macos-latest` can build all Kotlin targets.
* Just trigger `./gradlew publish` instead of listing all of the publish tasks separately. 

In the build config I've introduced a BuildService with [limited concurrency](https://docs.gradle.org/8.7/userguide/build_services.html#concurrent_access_to_the_service) to try to avoid issues with Maven Central and parallel uploads. If it doesn't work, then the quickest and easiest fix is to add the flags `--no-configuration-cache org.gradle.parallel=false -Dorg.gradle.workers.max=1`.


For now I've just updated the Gradle options to be closer to what's in the project's `gradle.properties`.

kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
```diff
@@ -2,13 +2,19 @@ package com.sksamuel.kotest.parallelism
...
 
 object ProjectConfig : AbstractProjectConfig() {
 
-   private var start = 0L
+   private lateinit var start: TimeMark
 
    override suspend fun beforeProject() {
-      start = System.currentTimeMillis()
+      // Delay, so that tests can warm up (this is a wild guess - is it actually correct???)
+      delay(1.seconds)
```
OliverO2: If I understand the `beforeProject` semantics correctly, this delay would just prevent all specs from being executed for one second. So there is no warming up as nothing runs. If that's correct, we'd better omit it.

kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
```diff
@@ -17,10 +23,11 @@ object ProjectConfig : AbstractProjectConfig() {
    override val concurrentSpecs: Int = ProjectConfiguration.MaxConcurrency
 
    override suspend fun afterProject() {
-      val duration = System.currentTimeMillis() - start
-      // there are 8 specs, and each one has a delay
-      // if parallel is working they should all block at the same time
-      if (duration > 700)
+      val duration = start.elapsedNow()
```
OliverO2, line 24: Excellent! Reads much better,

kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
```diff
@@ -17,10 +23,11 @@ object ProjectConfig : AbstractProjectConfig() {
...
+      // There are 8 specs, and each one has a 1-second delay.
```
OliverO2: Do they? To me it seems they still block for 100 ms each.

kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
```diff
@@ -17,10 +23,11 @@ object ProjectConfig : AbstractProjectConfig() {
...
+      // If parallel is working, they should all block at the same time.
+      if (duration > 7.seconds) {
```
OliverO2: If the tests are at 100 ms each, total sequential execution time would be 800 ms so this could almost never fail.

Maybe a better approach would be to increase the blocking time inside the test threads from `Thread.sleep(100)` to `Thread.sleep(400)` to handle a slow CI and then check like `duration > (7*400).milliseconds`?

kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
OliverO2: That's a tricky one and I'm not sure we can get it totally stable, as the operating system's CPU scheduler will make no guarantees for our use case, but we can at least try.

.github/workflows/release_base.yml
OliverO2: > * Use a single machine to publish everything - macos-latest can build all Kotlin targets.
> * Just trigger ./gradlew publish instead of listing all of the publish tasks separately.

I think that's a great idea and would prevent incomplete publications when new targets (like `wasmWasi`) are introduced. I'm no publishing expert, but it seems safe to try for me.

OliverO2 (commented):
To me this is all on the right trajectory! 🎉

kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
aSemy: Oooooops yes, I misread 100 as 1000! I'll fix it.


Your proposal may well work, but just to note, I don't think the speed of the CI is the problem. When I ran `./gradlew jvmTest --rerun-tasks` I saw the same problem using my Macbook.

I'll fix the times and play around some more...

kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
aSemy: Thanks! My mistake, it's fixed.

kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
OliverO2: Weird things happen in real time. 🙃 Usually my machine is the slowest one around but it ran the 8 tests in 170 ms so well below the former 700 ms limit.

aSemy: I've bumped the blocking time - let's see if it helps!

Alternatively, it would probably work if `Thread.sleep()` was replaced with a coroutines `delay()` - but I don't know if using `Thread.sleep()` is the point of the test.

OliverO2: In this case I'd guess it makes no difference: Each test runs on its own thread via its own dispatcher. Since that dispatcher has no other jobs to schedule in case of a `delay()`, it would most probably do the same thing, i.e. invoke `Thread.sleep()`.

OliverO2: That apparently did the trick (at least for now):
```
> Task :kotest-tests:kotest-tests-spec-parallelism:jvmTest

SUCCESS: Executed 8 tests in 7s
```

But if this isn't really really slow I don't know what is...

aSemy: Nice!

Good point about each test having its own dispatcher. Maybe telling Kotest to use  [TestCoroutineScheduler](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/) would work? Then measure the virtual time, not the actual elapsed time.

aSemy: > Executed 8 tests in 7s

Why is it 7 seconds and not 3.5 seconds? 🤔

OliverO2: > Maybe telling Kotest to use TestCoroutineScheduler would work? Then measure the virtual time, not the actual elapsed time.

That would defeat the purpose: Each test would complete immediately. We'd have no way to detect actual cross-test parallelism. 

 > Why is it 7 seconds and not 3.5 seconds?
 
The actual execution time to expect would be slightly above 0.5 seconds (all 8 tests in parallel sleeping for 500 ms at the same time). The rest is just effects of CI system load (or artificial slowdown, I don't know if they do this).

kotest-framework/kotest-framework-api/build.gradle.kts
Kantis: Looks like it was added 4 years ago. It's likely been broken for quite some time. I think we can just do as you suggest, and maybe create a new issue to investigate (and potentially fix) it

Kantis: Besides, it wouldn't work if we do split the project, since docs and sources would reside in different repositories?

kotest-framework/kotest-framework-standalone/build.gradle.kts
```diff
@@ -29,7 +23,11 @@ tasks {
       }
    }
    getByName("jvmJar") {
-      finalizedBy(getByName("shadowJar"))
+      finalizedBy(shadowJar)
+   }
+
+   startScripts {
+      dependsOn(shadowJar)
```
Kantis, line 30: What is startScripts?

.github/workflows/release_base.yml
Kantis: Lets raise an issue and discuss this? I feel this PR is kind of big already?

Kantis (commented):
Nice work, looks good to me. I'll leave the approval to @OliverO2 who's been more involved

OliverO2 (approved):
All good from my side. Nicely done!

OliverO2:
@Kantis If your questions have been answered and the publishing part looks good to you, would you please merge? I'll be away from keyboards until Monday.

Kantis:
@aSemy  some configuration cache issue popped up on CI build. Do you know what it might be caused by?

image

aSemy:
Hmm that's strange... It might be a false positive CC error

https://youtrack.jetbrains.com/issue/KT-61102/Using-some-Kotlin-tasks-together-with-tasks-marked-as-notCompatibleWithConfigurationCache-causes-bogus-configuration-cache

https://github.com/gradle/gradle/issues/22481

kotest-framework/kotest-framework-api/build.gradle.kts
aSemy: Good point! I think this task can just be removed.

https://github.com/kotest/kotest/issues/4066

.github/workflows/release_base.yml
aSemy: https://github.com/kotest/kotest/issues/4067

kotest-framework/kotest-framework-standalone/build.gradle.kts
aSemy: The Gradle Application plugin creates a `.sh` and `.bat` scripts for the bundled application. `startScripts` is the task that generates those scripts.

The build config here is a bit hacky - instead of explicitly replacing the default JAR with the shadowed JAR, it changes the names of the JARs. However, Gradle doesn't realise this. `startScripts` tries to use the JAR produced by the default `jar` task, but because the name has changed, Gradle complains.

PR #4068, opened 2024-06-06
Report config-cache errors as warnings

aSemy:
prevent false-positive CC errors from failing the build https://github.com/kotest/kotest/pull/4062#issuecomment-2151725255

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -16,6 +16,7 @@ org.gradle.caching=true
 org.gradle.parallel=false
 org.gradle.daemon=true
 org.gradle.configuration-cache=true
+org.gradle.configuration-cache-problems=warn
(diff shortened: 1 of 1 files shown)


aSemy:
This PR is a quick fix - a more robust fix would be fixing `publishPlatformArtifactsInRootModule()` to be CC compliant, so that `notCompatibleWithConfigurationCache()` isn't necessary

 https://github.com/kotest/kotest/blob/7b4fe302109915a9c3d465ab35af82cacb1aefce/buildSrc/src/main/kotlin/publishingUtils.kt#L65-L69

Alternative, wait until Gradle fixes it https://github.com/gradle/gradle/issues/22481

Kantis (approved):

Issue #4067, opened 2024-06-06
Update release GitHub workflows


aSemy:
I think the release workflows could be simplified.

* Use a single machine to publish everything - `macos-latest` can build all Kotlin targets.
* Just trigger `./gradlew publish` instead of listing all of the publish tasks separately. 

In https://github.com/kotest/kotest/pull/4062 I introduced a BuildService with [limited concurrency](https://docs.gradle.org/8.7/userguide/build_services.html#concurrent_access_to_the_service) to try to avoid issues with Maven Central and parallel uploads. If it doesn't work, then the quickest and easiest fix is to add the flags `--no-configuration-cache org.gradle.parallel=false -Dorg.gradle.workers.max=1`.

We could also look into creating a staging repository before releasing - then Maven Central won't get confused about parallel uploads.

Kantis:
What are the problems with parallel uploads to Maven central? I've heard of it multiple times but don't know the details.

Kantis:
Should we also consider using https://github.com/vanniktech/gradle-maven-publish-plugin, I think it would automate the (otherwise) manual steps in Nexus as well?

aSemy:
> What are the problems with parallel uploads to Maven central? I've heard of it multiple times but don't know the details.

When publishing, Maven Central creates a private, temporary, staging repository to collect the files. It does this automatically when it receives the first artifact. However, if there are multiple artifacts being published from multiple machines, it may create multiple staging repositories. The result is the files are split across multiple staging repos, so no staging repo is valid.

https://stackoverflow.com/questions/15511484/mvn-releaseperform-creates-multiple-staging-repos

This can be avoided by not publishing in parallel, or manually creating the staging repo first (maybe there's an API to create a staging repo though?).

Possibly the new method of publishing to Maven Central is better? https://central.sonatype.org/register/central-portal/

> Should we also consider using https://github.com/vanniktech/gradle-maven-publish-plugin, I think it would automate the (otherwise) manual steps in Nexus as well?

It's not config-cache compatible :(

https://github.com/vanniktech/gradle-maven-publish-plugin/issues/259

aSemy:
It looks like the release is broken because of parallel uploads.

https://github.com/kotest/kotest/actions/runs/9427914203/job/25972687250

```
[...]

BUILD FAILED in 4m 24s
281 actionable tasks: 251 executed, 30 from cache
Configuration cache entry discarded with 6 problems.
Error: Process completed with exit code 1.

[...]

[...]

PR #4072, opened 2024-06-08
Refactor KMP POM rewriting

aSemy:
Update KMP POM re-writer to be CC compatible.

Instead of editing the POM in-memory, instead modify the POM file that the `generatePomFileForKotlinMultiplatformPublication` task produces.

Replace the Groovy XML utils with W3C Document classes, because the Groovy XML utils don't preserve comments.

I added a project-local Maven repository, for easier testing of the changes. Run `./gradlew publishAllPublicationsToRootBuildDirRepository` and check `build/maven-repo`.

Relates to 

* #4067
* #3164


Kantis (approved):

Kantis:
Nice work @aSemy 🚀

aSemy:
Thanks!

PR #4074, opened 2024-06-08
Create BuildService for collecting dependencies for kotest-bom

aSemy:
Create a custom [BuildService](https://docs.gradle.org/8.9/userguide/build_services.html) to collect subprojects that should be included in the kotest-bom.

This change is needed because currently the Kotest BOM entries are discovered by using discouraged Gradle methods (specifically, accessing the state of other subprojects via `rootProject` and `subprojects`). Gradle is planning to make this invalid - see [Isolated Projects](https://docs.gradle.org/8.9/userguide/isolated_projects.html).

Instead of reading other subprojects using `rootProject.subprojects`, the KotestBomService is used to share state. It will automatically collect the dependencies from any project that has the `kotest-publishing-conventions` convention plugin. This means that the kotest-bom subproject doesn't need to query the internal state of other subprojects.

Subprojects can opt out of being included in the BOM by setting `includeInKotestBom = false` in their `build.gradle.kts`. `kotest-bom` is excluded by default.

### Depends on 

- #4073


sksamuel:
Is this ready to be merged ?

aSemy:
yes

sksamuel (approved):

Issue #4077, opened 2024-06-08
Support coroutineTestScope option for JS platform


osipxd:
Currently, this option [is disabled for JS platform](https://github.com/kotest/kotest/blob/9e1398b2bc8af2948e517feb6060a14044796e5d/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt#L58-L63) specifically.

Known issues to be investigated before this option could be enabled:
- `runTest` [returns `TestResult`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-test-result/), which is a type alias for `Promise` on JS targets. It should be handled correctly.

CLOVIS-AI:
Ah, that's a shame.

How does Kotest execute `suspend` code in the JS world? `runTest` returns a `Promise` to interop with existing JS test frameworks, but since Kotest already supports `suspend` in JS tests, it must have another way to execute them. Presumably, `await`'ing the returned promise in Kotest's usual `suspend` scope should be enough to implement this?

PR #4078, opened 2024-06-08
Disable coroutineTestScope for WasmJs

osipxd:
It seems like this option was accidentally enabled for the Wasm JS platform. I've disabled it until #4077 is fixed

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
@@ -15,7 +15,6 @@ import io.kotest.engine.test.interceptors.AssertionModeInterceptor
 import io.kotest.engine.test.interceptors.BeforeSpecListenerInterceptor
 import io.kotest.engine.test.interceptors.CoroutineDebugProbeInterceptor
 import io.kotest.engine.test.interceptors.CoroutineLoggingInterceptor
-import io.kotest.engine.test.interceptors.TestEnabledCheckInterceptor
 import io.kotest.engine.test.interceptors.ExpectExceptionTestInterceptor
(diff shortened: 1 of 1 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
```diff
@@ -55,6 +55,11 @@ internal class TestCaseExecutor(
 
       val timeMark = MonotonicTimeSourceCompat.markNow()
 
+      // JS platforms require extra care when runTest is used, so skip it for now.
+      // Issue: https://github.com/kotest/kotest/issues/4077
+      val useCoroutineTestScope = (platform == Platform.JVM || platform == Platform.Native)
```
osipxd: Another option was `platform != Platform.JS && platform != Platform.WasmJs`, but it seems safer to restrict it to the allowed platforms, since the platform list could grow.

OliverO2: As the `runTest` limitation is specific to `Promise`-based JS platforms, wouldn't it be safer to use `platform != Platform.JS && platform != Platform.WasmJs`? Otherwise, we could accidentally introduce the limit to future non-JS targets like `wasmWasi`.

osipxd: I've replaced it with a `when` clause to prevent both accidental restriction and enabling for new platforms. Do you think it looks okay now?

OliverO2: That's OK for me.

OliverO2 (approved):
LGTM, thanks!

PR #4081, opened 2024-06-09
Tidy Range matchers

aSemy:
- Remove `@OptIn(ExperimentalStdlibApi::class)` now that Ranges are stable
- `@Suppress("UNCHECKED_CAST")` when casting ranges
- Use [`@PublishedApi`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-published-api/) instead of `@Suppress("NON_PUBLIC_CALL_FROM_PUBLIC_INLINE")`.
- Update ABI dumps, because `@PublishedApi` are technically part of the public ABI.
- Update kdoc to fix typos.
- Remove `If you need to check for empty range, use [ClosedRange.shouldBeEmpty]` from kdoc - no such function exists.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
+++ b/kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
@@ -1864,12 +1864,20 @@ public final class io/kotest/matchers/ranges/BeinKt {
 }
 
 public final class io/kotest/matchers/ranges/BewithinKt {
+	public static final fun beWithin (Lio/kotest/matchers/ranges/Range;)Lio/kotest/matchers/Matcher;
+	public static final fun beWithinRangeOfInt (Lkotlin/ranges/ClosedRange;)Lio/kotest/matchers/Matcher;
+	public static final fun beWithinRangeOfLong (Lkotlin/ranges/ClosedRange;)Lio/kotest/matchers/Matcher;
+	public static final fun resultForWithin (Lkotlin/ranges/ClosedRange;Lkotlin/ranges/OpenEndRange;Ljava/lang/Comparable;)Lio/kotest/matchers/MatcherResult;
 	public static final fun shouldBeWithin (Lkotlin/ranges/ClosedRange;Lkotlin/ranges/ClosedRange;)Lkotlin/ranges/ClosedRange;
 	public static final fun shouldBeWithin (Lkotlin/ranges/ClosedRange;Lkotlin/ranges/OpenEndRange;)Lkotlin/ranges/ClosedRange;
 	public static final fun shouldBeWithin (Lkotlin/ranges/OpenEndRange;Lkotlin/ranges/OpenEndRange;)Lkotlin/ranges/OpenEndRange;
+	public static final fun shouldBeWithinRangeOfInt (Lkotlin/ranges/OpenEndRange;Lkotlin/ranges/ClosedRange;)V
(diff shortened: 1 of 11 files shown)


kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
aSemy, line 1: I thought I could annotate the `@PublishedApi` functions with `@KotestInternal` to avoid updating the ABI dumps, but I see that `@KotestInternal` isn't excluded from the ABI dumps. Shouldn't it be?

https://github.com/kotest/kotest/blob/d4fc582d0cf0c60fbc0fcbd39096ceff3e4b6631/build.gradle.kts#L13-L21

Kantis: AFAIK there's no way to exclude API with a particular annotation. Or is it?

aSemy: There is:

```kotlin
apiValidation {
    /**
     * Set of annotations that exclude API from being public.
     * Typically, it is all kinds of `@InternalApi` annotations that mark
     * effectively private API that cannot be actually private for technical reasons.
     */
    nonPublicMarkers.add("my.package.MyInternalApiAnnotation")
}
```

Kantis: Oh, alright. Lets add that IMO.

aSemy: Agreed! I think it's a good idea

Kantis: Created separate issue for tracking that. Do you want to proceed with that before finishing up this PR, or should we look into merging this one as-is?

aSemy: Nice, thanks. 

Let's merge this one as-is.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/ranges/Range.kt
```diff
@@ -25,8 +25,8 @@ internal data class Range<T: Comparable<T>>(
       val endOfThis: T = this.end.value
       val startOfOther: T = other.start.value
       return when {
-         (this.end.edgeType== RangeEdgeType.INCLUSIVE && other.start.edgeType == RangeEdgeType.INCLUSIVE) -> (endOfThis < startOfOther)
-         else -> (endOfThis <= startOfOther)
+         (this.end.edgeType == RangeEdgeType.INCLUSIVE && other.start.edgeType == RangeEdgeType.INCLUSIVE) -> (endOfThis < startOfOther)
+         else                                                                                              -> (endOfThis <= startOfOther)
```
Kantis, line 29: What formatting rule is this? I don't think it's one we typically apply in this project :)

Kantis (approved):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/ranges/Range.kt
aSemy: Oh oops, I must have used my default formatter! I enabled 'align when branches in columns'.

Thanks for checking, I'll fix my config.

PR #4087, opened 2024-06-11
Suppressing usage of deprecated Kotest functionality

### LINKED ISSUE #4085: Building Kotest produces a lot of warnings, which make it difficult to identify problems
When I build Kotest I see a lot of warnings printed to the console. There are too many to be able to determine if any are significant, or any are irrelevant noise.

For example, some Kotest functions are deprecated. However, these functions themselves use deprecated functionality. It would make sense to suppress deprecation warnings for functions that are deprecated. 

I would like to fix these warnings over the course of several PRs.

In order to make sure that warnings don't creep in again, I also propose enabling [`allWarningsAsErrors`](https://kotlinlang.org/docs/gradle-compiler-options.html#attributes-common-to-jvm-js-and-js-dce).

Kantis:
@aSemy I accidentally started doing this all in one go. In the end, I don't think it's too big a deal to handle as a single PR. 

Partially addresses #4085

### DIFF
--- a/kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/EqualTest.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/EqualTest.kt
@@ -1,3 +1,5 @@
+@file:Suppress("DEPRECATION") // Remove when removing shouldEqualJson legacy options
+
 package com.sksamuel.kotest.tests.json
 
 import io.kotest.assertions.json.CompareMode
--- a/kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/EqualTest.kt
+++ b/kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/EqualTest.kt
@@ -1,4 +1,4 @@
-@file:Suppress("DEPRECATION") // Remove when removing shouldEqualJson legacy options
+
 
 package com.sksamuel.kotest.tests.json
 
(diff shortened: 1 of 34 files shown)


kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/EqualTest.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("DEPRECATION") // Remove when removing shouldEqualJson legacy options
```
aSemy: I'd suggest not doing file-level suppress. I know that suppressing every element is more verbose, but with a file-level suppress I think there's more of a chance that it'll be missed and not removed, and it might cause confusion later. But you added a comment, which I think is a good idea, so that will help.

I'll leave it up to you.

aSemy (approved):

aSemy:
> Fixes #4085

This PR doesn't entirely fix #4085 :) there are lots of other warnings

Kantis:
> > Fixes #4085
> 
> This PR doesn't entirely fix #4085 :) there are lots of other warnings

Good point. Adjusted accordingly

kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/EqualTest.kt
Kantis: I thought it would be required to get rid of the warnings from imports, but perhaps they don't cause warnings. I'll check how spammy it would be to get rid of.

aSemy: Ah true, I didn't consider imports. In which case, a file-level suppress makes sense.

aSemy (approved):

PR #4090, opened 2024-06-12
Remove the non-existent parameter includeEmpty in KDoc of Arb.Companion.map()

acanda:
The function [Arb.Companion.map(...)](https://github.com/kotest/kotest/blob/master/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/maps.kt#L25-L30) does not have a parameter `includeEmpty` (anymore).


OliverO2 (approved):
Good catch, thanks!

PR #4101, opened 2024-06-14
Refactor retry, and retry tests

aSemy:
The retry tests were unstable when running on master, so I thought I'd refactor them.

- Merge two `retry()` functions into a single function with a default argument (they were both the same, except for the `exceptionClass` parameter)
- Print duration in "Test failed after $delayedTime ..." instead of converting to exceptions, to avoid "Test failed after 0 seconds" message if the duration was less than 1 second.
- use `coroutineTestScope` to speed up tests (delays are emulated)
- assert the actual invocation times
- add TimeSource to Retry config (marked as `@KotestInternal`) so that the retry could be tested using the test scheduler TimeSource
- Replace `runSafely` with shouldThrow/shouldNotThrow assertions
- rename 'TestClass' to more descriptive name

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/RetryTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/RetryTest.kt
@@ -2,23 +2,37 @@ package com.sksamuel.kotest.assertions
 
 import io.kotest.assertions.retry
 import io.kotest.assertions.retryConfig
+import io.kotest.assertions.throwables.shouldNotThrow
+import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.StringSpec
-import io.kotest.matchers.longs.shouldBeGreaterThanOrEqual
+import io.kotest.core.test.TestScope
+import io.kotest.core.test.testCoroutineScheduler
+import io.kotest.matchers.collections.shouldBeEmpty
+import io.kotest.matchers.collections.shouldContainExactly
+import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.shouldBe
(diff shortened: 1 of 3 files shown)


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/RetryTest.kt
```diff
@@ -27,100 +41,185 @@ class RetryTest : StringSpec() {
...
-      } catch (exception: Exception) {
-         // Eating other exception
+   private class RetryTester(
+      private val readyAfter: Int,
+      private val timeMark: TimeMark,
+   ) {
+      val calledAtTimeInstance = mutableListOf<Duration>()
+
+      fun isReady(): Boolean {
+         calledAtTimeInstance += timeMark.elapsedNow()
```
OliverO2, line 209: I find it confusing if a function returns a value _and_ has a side effect as there is no indication of such behavior at call sites. This one is more like `advance()` plus `isReady()`. If splitting is not desirable, I'd consider something like `advance()` returning a result class like `RetryTester.Result(val isReady: Boolean)` a good compromise.

Just my thoughs. Leave it up to you to decide.

OliverO2 (approved):
Nice!

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/RetryTest.kt
aSemy: Yeah, I agree. I don't find the RetryTester class to be particularly clear. But I think it's a useful util, and I didn't want to completely re-write everything, so I just amended it.

I think it could probably be replaced with some `testRetry`/`testRetryFails` functions.

PR #4102, opened 2024-06-14
Fix testing on GitHub Windows runners

OliverO2:
Improve testing on GitHub Windows runners:
* PR tests on Windows: run all tests
* Speed up GitHub Windows runners by using the D: drive
* Relax elapsed real time expectations
* Relax thread count expectations for parallel tests


.github/workflows/run-gradle.yml
```diff
@@ -38,22 +47,25 @@ jobs:
...
+         -  name: Upload build reports
+            if: failure()
+            uses: actions/upload-artifact@v4
+            with:
+               name: build-reports-${{ runner.os }}-${{ github.action }}-${{ github.run_id }}
+               path: |
+                  **/build/reports/
+                  **/*.hprof
+                  **/*.log
+               if-no-files-found: ignore
```
aSemy, line 68:
```suggestion
         - name: Upload build reports
           if: failure()
           uses: actions/upload-artifact@v4
           with:
              name: build-reports-${{ runner.os }}-${{ github.action }}-${{ github.run_id }}
              path: |
                 **/build/reports/
                 **/*.hprof
                 **/*.log
              if-no-files-found: ignore
```

formatting nitpick

aSemy (approved):

.github/workflows/run-gradle.yml
OliverO2: Yeah, normally I'd do as suggested. But I actually prefer consistency and auto-formatting over slight personal preferences, and that change was what IJ wanted (me) to do. It is consistent with the other sections in that file.

Maybe we could change `.editorconfig` to accomodate a different style and re-format all `yaml`? Until then, I'd leave it at that for now.

PR #4103, opened 2024-06-15
replace deprecated getTimeMillis with posix clock_gettime

aSemy:
`getTimeMillis()` is deprecated https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.system/get-time-millis.html

The alternative in this PR was discussed in Slack: https://slack-chats.kotlinlang.org/t/13154128/https-kotlinlang-org-api-latest-jvm-stdlib-kotlin-system-get

Part of #4085

### DIFF
--- a/kotest-common/src/desktopMain/kotlin/io/kotest/mpp/time.kt
+++ b/kotest-common/src/desktopMain/kotlin/io/kotest/mpp/time.kt
@@ -1,5 +1,23 @@
 package io.kotest.mpp
 
-import kotlin.system.getTimeMillis
+import kotlinx.cinterop.ExperimentalForeignApi
+import kotlinx.cinterop.UnsafeNumber
+import kotlinx.cinterop.alloc
+import kotlinx.cinterop.convert
+import kotlinx.cinterop.memScoped
+import kotlinx.cinterop.pointed
+import kotlinx.cinterop.ptr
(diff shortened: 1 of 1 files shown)


Kantis:
Did you consider removing `getTimeMillis` and using `measureTime` or `TimeSource.Monotonic.markNow()` as suggested in the docs for `getTimeMillis`?

Kantis:
Ok, checked the surrounding code a bit and looks like this might be practical until we remove the old `continually` and `eventually` implementations.

Kantis (approved):

aSemy:
Ah yes, good point, maybe we don't need `getTimeMillis()` at all.

I'll have a look...

aSemy:
I think it'll make sense to remove `getTimeMillis()` in the same PR that removes the [TimeSource compatibility stuff](https://github.com/kotest/kotest/blob/0090ae49415fcae496a5cdaaacdc74c9a3118c59/kotest-common/src/commonMain/kotlin/io/kotest/common/timing.kt#L28).

I'll make another PR, and park this one for now...

Kantis:
@aSemy I started doing those changes locally, but I think you need to do breaking changes in `eventually` and `continually` so not sure it's feasible to do before 6.0.

aSemy:
Superseded by https://github.com/kotest/kotest/pull/4105

PR #4105, opened 2024-06-15
Remove TimeMarkCompat / MonotonicTimeSourceCompat / timeInMillis

aSemy:
TimeMarkCompat was only necessary because of a breaking change in Kotlin 1.7 - but that was 2 years ago and Kotest now requires more recent versions, so TimeMarkCompat can now be removed. [Slack discussion](https://slack-chats.kotlinlang.org/t/18926932/what-s-kotest-s-minimum-supported-kotlin-version-i-m-trying-#e406321c-881f-465f-9716-35925a980daf).

`timeInMillis()` is only used for measuring times, so it can be replaced with stdlib `measureTime {}`/`measureTimedValue {}`


sksamuel (approved):

sksamuel:
nice update

kotest-framework/kotest-framework-concurrency/src/commonMain/kotlin/io/kotest/framework/concurrency/continually.kt
```diff
@@ -38,28 +47,34 @@ class ContinuallyBuilder<T> {
...
+   val end = start.plus(duration)
    var times = 0
    var result: T? = null
 
-   while (timeInMillis() < end) {
+   while (end.hasNotPassedNow()) {
       try {
          result = f()
-         listener?.invoke(ContinuallyState(result, start, end, times))
+         listener?.invoke(ContinuallyState(result, 0L, duration.inWholeMilliseconds, times))
```
aSemy, line 70: Just to highlight, there's a potentially breaking change here. 

The start and end of a ContinuallyState will no longer reflect the system time, but start will always be 0, and end will always be the configured duration.

I think this change is acceptable because it's a minor change, and this code has been deprecated for ~3 years.

OliverO2 (approved):
Excellent!

kotest-framework/kotest-framework-concurrency/src/commonMain/kotlin/io/kotest/framework/concurrency/continually.kt
sksamuel: agreed, all deprecated functions are coming out for 6.0

sksamuel (approved):

PR #4106, opened 2024-06-15
remove unnecessary inline

aSemy:
Part of #4085


Kantis (approved):
I think this function is not used anymore, just remove it?

aSemy:
> I think this function is not used anymore, just remove it?

Ah yeah good point, it could be removed. Maybe it might get used again in the future? If not, then yeah, let's remove it.

PR #4109, opened 2024-06-16
Use test TimeSource in continually tests

aSemy:
Use a test TimeSource in `continually` tests.

- The test TimeSource uses a virtual time, so the tests run faster.
- The elapsed duration is reproducible, so the tests can assert the specific elapsed time.
- Use CoroutineContext to store & retrieve the TimeSource (This helps with testing both `continually(duration)` and `continually(config)` functions, and doesn't require adding an option in `ContinuallyConfiguration`)

### Dependencies

* #4105


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/continually.kt
```diff
@@ -122,3 +128,31 @@ private fun <T> ContinuallyConfigurationBuilder<T>.build(): ContinuallyConfigura
...
+       * For internal Kotest testing purposes the [TimeSource] can be overridden.
+       * For normal usage [TimeSource.Monotonic] is used.
+       */
+      internal suspend fun getContinuallyTimeSource(): TimeSource =
+         coroutineContext[KEY]?.timeSource
+            ?: TimeSource.Monotonic
+
+      internal val KEY = object : CoroutineContext.Key<ContinuallyTimeSource> {}
+   }
+}
```
aSemy: In the 'retry' tests PR https://github.com/kotest/kotest/pull/4101 I added a `val timeSource` to the RetryConfig.

Here I'm trying a different approach and storing the TimeSource in the CoroutineContext.

I think using the CoroutineContext is better, because it makes it easier to test any `continually()` functions without modifying the ABI, and it doesn't expose any options for users (even the option is guarded with `@KotestInternal`). 

I'd like to revisit the `retry` tests, and refactor the `eventually` tests, to do the same.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/io/kotest/assertions/nondeterministic/ContinuallyTest.kt
```diff
@@ -31,54 +44,103 @@ class ContinuallyTest : FunSpec() {
...
+            testContinually(3.seconds) {
                delay(10)
                (n++ < 10) shouldBe true
             }
          }
-         val r =
-            "Test failed after \\d+ms; expected to pass for 3000ms; attempted 100 times\nUnderlying failure was: 100 should be < 100".toRegex()
-         e.message?.matches(r) ?: (false shouldBe true)
+         n shouldBe 11
+         failure shouldHaveMessage "Test failed after 360ms; expected to pass for 3s; attempted 10 times\nUnderlying failure was: expected:<true> but was:<false>"
```
aSemy, line 100: Minor note: previously the regex assertion wasn't triggered, and wasn't correct.

I changed the expected message from `Underlying failure was: 100 should be < 100` to `Underlying failure was: 10 should be < 10`.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/io/kotest/assertions/nondeterministic/ContinuallyTest.kt
```diff
@@ -32,54 +44,103 @@ class ContinuallyTest : FunSpec() {
             intervalFn = DurationFn { 300.milliseconds }
          }
          var k = 0
-         continually(config) {
-            (System.currentTimeMillis() > 0) shouldBe true
+         val result = testContinually(config) {
+            1 shouldBe 1
             k++
          }
          k shouldBeIn 2..3
```
OliverO2: Test run indicates missing import for `shouldBeIn`.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/continually.kt
```diff
@@ -29,10 +32,9 @@ suspend fun <T> continually(
    config: ContinuallyConfiguration<T>,
    test: suspend () -> T,
 ): T {
-
    delay(config.initialDelay)
 
-   val start = TimeSource.Monotonic.markNow()
+   val start = getContinuallyTimeSource().markNow()
```
OliverO2: Just a style preference to consider as `get()` sounds rather imperative:
* top-level function `continuallyTimeSource()` or
* `ContinuallyTimeSource.current()`

OliverO2 (changes requested):
Missing import. Style consideration optional. Otherwise looks good!

OliverO2 (approved):

PR #4110, opened 2024-06-16
Use centralized repositories for build

aSemy:
Centrally define the repositories required for the build. This encourages a stable build, as the repositories are defined consistently, in the same order of priority, for all subprojects.

See: https://docs.gradle.org/current/userguide/declaring_repositories.html#sub:centralized-repository-declaration

### DIFF
--- a/buildSrc/settings.gradle.kts
+++ b/buildSrc/settings.gradle.kts
@@ -1,6 +1,19 @@
 rootProject.name = "buildSrc"
 
+pluginManagement {
+   repositories {
+      mavenCentral()
+      gradlePluginPortal()
+   }
+}
+
+@Suppress("UnstableApiUsage")
 dependencyResolutionManagement {
+   repositoriesMode = RepositoriesMode.PREFER_SETTINGS
+   repositories {
+      mavenCentral()
+      gradlePluginPortal()
+   }
    versionCatalogs {
       create("libs") {
          from(files("../gradle/libs.versions.toml"))
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -7,6 +7,65 @@ pluginManagement {
    }
 }
 
+@Suppress("UnstableApiUsage")
+dependencyResolutionManagement {
+   repositoriesMode.set(RepositoriesMode.PREFER_SETTINGS)
+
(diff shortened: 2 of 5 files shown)


buildSrc/settings.gradle.kts
```diff
@@ -1,6 +1,19 @@
 rootProject.name = "buildSrc"
 
+pluginManagement {
+   repositories {
+      mavenCentral()
```
OliverO2, line 5: Do we need `mavenCentral()` for plugins?

settings.gradle.kts
```diff
@@ -7,6 +7,65 @@ pluginManagement {
    }
 }
 
+@Suppress("UnstableApiUsage")
+dependencyResolutionManagement {
+   repositoriesMode.set(RepositoriesMode.PREFER_SETTINGS)
+
+   repositories {
+      google()
```
OliverO2: Do we use something from `google()`?

settings.gradle.kts
```diff
@@ -7,6 +7,65 @@ pluginManagement {
...
+      mavenCentral()
+      maven("https://maven.pkg.jetbrains.space/kotlin/p/wasm/experimental") {
+         name = "JetBrainsKotlinWasm"
+      }
```
OliverO2: This might be obsolete by now.

OliverO2 (approved):
Thanks for cleaning this up!

buildSrc/settings.gradle.kts
aSemy: No, however, Maven Central is more strict and slightly more stable than GPP, so putting it first helps (slightly).

settings.gradle.kts
aSemy: Good question, I removed the Google and Kotlin Wasm repos and the tests passed locally. Let's see if CI passes too...

OliverO2:
As long as we have flaky timeouts on the low-performing runners, what if you'd just bypass the merge queue and merge directly?

There's no value in re-running PR checks if it's just timeouts that we'll fix in upcoming PRs.

aSemy:
Good idea 👍

PR #4114, opened 2024-06-16
Use test TimeSource in eventually tests

aSemy:
Use a test TimeSource in `eventually` tests.

- The test TimeSource uses a virtual time, so the tests run faster.
- The elapsed duration is reproducible, so the tests can assert the specific elapsed time.
- Use CoroutineContext to store & retrieve the TimeSource (This helps with testing both `eventually(duration)` and `eventually(config)` functions, and doesn't require adding an option in `EventuallyConfiguration`)

Depends on: #4105 

Related: #4109


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -130,30 +133,30 @@ object EventuallyConfigurationDefaults {
...
 
    /**
-    * A delay that is applied before the first invocation of the eventually function.
+    * A delay that is applied before the first invocation of the [eventually] function.
     */
    var initialDelay: Duration = EventuallyConfigurationDefaults.initialDelay
 
    /**
-    * The delay between invocations. This delay is overriden by the [intervalFn] if it is not null.
+    * The delay between invocations. This delay is override by the [intervalFn] if it is not `null`.
```
OliverO2:
```suggestion
    * The delay between invocations. This delay is overridden by the [intervalFn] if it is not `null`.
```

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -130,30 +133,30 @@ object EventuallyConfigurationDefaults {
...
     */
    var interval: Duration = EventuallyConfigurationDefaults.interval
 
    /**
-    * A function that is invoked to calculate the next interval. This if this null, then the
+    * A function that is invoked to calculate the next interval. This if this `null`, then the
     * value of [interval] is used.
     *
-    * This function can be used to implement [fibonacci] or [exponential] backoffs.
+    * This function can be used to implement [fibonacci] or [exponential] backoff.
```
OliverO2, line 154:
```suggestion
    * This function can be used to implement a [fibonacci] or an [exponential] backoff.
```

OliverO2 (approved):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
aSemy: I much prefer the original one, omitting the articles reads more naturally and is more direct.

OliverO2: I don't know why as I usually can't explain language rules anyway, but the first variant seemed to long for a plural form "backoffs". But I don't mind.

PR #4115, opened 2024-06-16
commonize shouldCompleteWithin/shouldCompleteBetween/shouldTimeout

aSemy:
- Convert timeout assertions to common Kotlin
- Deprecate the JVM only versions
- Update tests to use virtual time
- Update failure messages to be clearer, and contain the expected/actual durations.
- De-duplicate `shouldTimeout`, which is present in both `kotest-assertions` and `kotest-assertions`
  - https://github.com/kotest/kotest/blob/cff27c578aebc384135321660a8d2d7ae5a93780/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/concurrent/suspension/concurrent.kt#L69
  - https://github.com/kotest/kotest/blob/cff27c578aebc384135321660a8d2d7ae5a93780/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/async/timeout.kt#L9


aSemy:
Hmmmm JVM tests work, but WasmJS fails

```
./gradlew :kotest-assertions:kotest-assertions-core:wasmJsTest --stacktrace

[...]

Disconnected (0 times) , because no message in 30000 ms.
java.lang.IllegalStateException: command '/Users/dev/.local/state/gradle/nodejs/node-v22.0.0-darwin-arm64/bin/node' exited with errors (exit code: 1)

> Task :kotest-assertions:kotest-assertions-core:allTests FAILED
```

OliverO2:
Seems like `wasmJsBrowserTest` fails with a single test exceeding the (presumably karma) timeout of 30 seconds. I could take a look cannot promise results today.

aSemy:
I found a similar issue on Slack https://slack-chats.kotlinlang.org/t/18859293/is-there-special-config-required-to-run-tests-with-wasmjs-i-, but the suggested workaround didn't help https://youtrack.jetbrains.com/issue/KT-67468

aSemy:
I got some help and we figured out the problem: the Kotest Gradle Plugin isn't applied to Kotest projects, so the tests can't be discovered or launched.

OliverO2:
Why does this lead to the timeout? I'd expect it the test run to finish on `wasmJsBrowserTest` without reporting any test. I'll try to investigate,

OliverO2:
The test times out because the dependency `implementation(projects.kotestFramework.kotestFrameworkEngine)` is missing in `commonTest`. For Wasm, the function `startUnitTests` is missing in that case:
https://github.com/kotest/kotest/blob/2f9215c2f0519e08fe94ee6b37431fde736c1424/kotest-framework/kotest-framework-engine/src/wasmJsMain/kotlin/io/kotest/engine/KotlinJsTestFramework.wasmJs.kt#L73-L79

The Kotlin test framework crashes as it tries to call the function on the headless browser before it could launch the test framework. Karma loses its patience and complains after 30 seconds without any response.

aSemy:
Thanks @OliverO2! However, the tests need `testCoroutineScheduler`, and I think this isn't available on Wasm yet? https://github.com/kotest/kotest/issues/4077

OliverO2:
That's correct, any JS platform test (including `wasmJs`) would fail if it requires a `runTest` invocation.

Kotest effectively does not fully self-test on non-JVM targets anyway due to the missing Kotest compiler plugin. While it appears to run non-JVM tests, without the plugin in place, test discovery actually does not take place.

This could be corrected, but would require our build scripts to use the build's own version of the Kotest compiler plugin. A previous compiler plugin version from a repo could be chosen, but this would be unstable in case of incompatible changes in the code generated by the plugin.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/concurrent/suspension/concurrent.kt
```diff
@@ -0,0 +1,99 @@
...
+       * For internal Kotest testing purposes the [TimeSource] can be overridden.
+       * For normal usage [TimeSource.Monotonic] is used.
+       */
+      internal suspend fun current(): TimeSource =
+         coroutineContext[KEY]?.timeSource
+            ?: TimeSource.Monotonic
+
+      internal val KEY = object : CoroutineContext.Key<ShouldCompleteBetweenTimeSource> {}
+   }
+}
```
Kantis: When would this typically be used?

buildSrc/build.gradle.kts
```diff
@@ -5,4 +5,5 @@ plugins {
 dependencies {
    implementation(libs.testlogger.gradle.plugin)
    implementation(libs.kotlin.gradle.plugin)
+   implementation("io.kotest:kotest-framework-multiplatform-plugin-gradle:5.9.1")
```
Kantis: Why?

Kantis (commented):
Looks good to me, but added a couple of questions for things I don't understand

buildSrc/build.gradle.kts
aSemy: Oops, I left this in my mistake. I've removed it.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/concurrent/suspension/concurrent.kt
aSemy: Ah yes, I forgot about this! Thanks for checking.

#4149 will simplify test TimeSource logic, which makes this config redundant. So I'll wait for #4149 to be merged and tidy up the test TimeSource config in this PR. 


To answer your question: you would use this whenever using the test coroutine dispatcher, in order to make the tests use virtual time (so that they run faster, since they don't have to wait for real life time, and so that they're more stable, since they aren't sensitive to the machine's load).

OliverO2: Please ping me if this is ready for a final review.

aSemy: @OliverO2 it's ready for review!

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/concurrent/suspension/concurrent.kt
```diff
@@ -0,0 +1,78 @@
...
+ * Note: It does not work well within [assertSoftly].
+ * If used within [assertSoftly] and this assertion fails, any later assertions won't run.
+ */
+suspend fun <A> shouldCompleteBetween(
+   durationRange: ClosedRange<Duration>,
+   operation: suspend () -> A,
+): A {
+   contract { callsInPlace(operation, EXACTLY_ONCE) }
+
+   val timeSource = nonDeterministicTestTimeSource()
```
OliverO2: As I understand it, the Kotest timeout matchers were all using `withTimeout()` before, which responds to virtual time. This would introduce a breaking change to real time for Kotest users, right?

aSemy: Ah good point... I didn't think about that.

So it should just be enough to set `coroutineTestScope = true` in the tests, and use `testTimeSource()` instead of `nonDeterministicTestTimeSource()`?

OliverO2: Exactly!

OliverO2 (approved):
Looks good!

PR #4117, opened 2024-06-17
Fix kotest-framework-multiplatform-plugin-gradle build

OliverO2:
* Define an input dependency for the task `updateKotestPluginConstants`, to prevent Gradle from skipping it with "NO-SOURCE".
* Avoid failing the build with an undefined symbol "KOTEST_COMPILER_PLUGIN_VERSION".
* Make the `clean` task compatible with the configuration cache.


aSemy:
Oh, interesting! Good catch. I guess that TextResource isn't compatible with configuration cache.

OliverO2:
With the dedicated task class, there was no longer a need for `TextResource`. The configuration cache complaints were caused by `$projectDir` in the `clean` task configuration.

aSemy:
Do you mind if I refactor it to use the dynamic task dsl? I find it more appropriate for tasks that are defined in a build script.

```kotlin
val updateKotestPluginConstants by tasks.registering {
   val kotestPluginConstants = """
      |// Generated file, do not edit manually
      |@file:org.gradle.api.Generated
      |
      |package io.kotest.framework.multiplatform.gradle
      |
      |const val KOTEST_COMPILER_PLUGIN_VERSION: String = "${Ci.gradleVersion}"
      |
   """.trimMargin()

   inputs.property("kotestPluginConstants", kotestPluginConstants)

   val outputDir = layout.buildDirectory.dir("generated/src/main/kotlin/")
   outputs.dir(outputDir).withPropertyName("outputDir")

   doLast {
      val kotestPluginConstantsFile = outputDir.get().asFile
         .resolve("io/kotest/framework/multiplatform/gradle/kotestPluginConstants.kt")

      kotestPluginConstantsFile.apply {
         parentFile.deleteRecursively()
         parentFile.mkdirs()
         writeText(kotestPluginConstants)
      }
   }
}

kotlin.sourceSets.main {
   kotlin.srcDir(updateKotestPluginConstants)
}
```

(GitHub won't let me add this as a code suggestion)


Otherwise, I'd prefer moving the class-based task into buildSrc.

OliverO2:
No, I don't mind at all. That's what I've tried to achieve first. Just could not find the `input.property` trick. I'm just a casual Gradle user after all... ;-)

aSemy (approved):

PR #74, opened 2016-07-13
Correct call to beforeEach()

hoangtranwork:
fix a mistake mentioned in https://github.com/kotlintest/kotlintest/issues/73

PR #4118, opened 2024-06-17
Json path array match take3 (#74)

AlexCue987:
provide more info when key not found, such as:

```kotlin
   "Failure message states if key is missing, shows json array index out of bounds" {
      shouldFail {
         json.shouldContainJsonKeyValue("$.store.book[2].category", "V2")
      }.message shouldBe "Expected given to contain json key <'$.store.book[2].category'> but key was not found. The array at path <'$.store.book'> has size 2, so index 2 is out of bounds."
   }
```

or 

```kotlin
   "Failure message states if key is missing, shows json array index out of bounds when array is empty" {
      shouldFail {
         json.shouldContainJsonKeyValue("$.store.book[1].comments[0]", "V2")
      }.message shouldBe "Expected given to contain json key <'$.store.book[1].comments[0]'> but key was not found. The array at path <'$.store.book[1].comments'> has size 0, so index 0 is out of bounds."
   }
```

### DIFF
--- a/kotest-assertions/kotest-assertions-json/api/kotest-assertions-json.api
+++ b/kotest-assertions/kotest-assertions-json/api/kotest-assertions-json.api
@@ -85,6 +85,19 @@ public final class io/kotest/assertions/json/FieldComparison : java/lang/Enum {
 	public static fun values ()[Lio/kotest/assertions/json/FieldComparison;
 }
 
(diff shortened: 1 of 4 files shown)


AlexCue987:
next step will be to do the same thing for `shouldContainJsonKey`

Kantis (approved):

PR #4119, opened 2024-06-17
Configure JVM test tasks with same args as main build

### LINKED ISSUE #4112: CI: test_linux(jvmTest) fails intermittently with OutOfMemoryError: Java heap space
Kotest 5.9.1/master at 846f0a66f795eb0ef419cc2fbebbe1c0c2c51647

[In the logs](https://github.com/kotest/kotest/actions/runs/9536396785/job/26283311789), we occasionally see

```
Warning: [107.675s][warning][gc,alloc] ClassGraph-worker-306: Retried waiting for GCLocker too often allocating 1026 words
```

followed by
```
> Task :kotest-runner:kotest-runner-junit5:jvmTest

com.sksamuel.kotest.runner.junit5.StringSpecEngineKitTest[jvm]
Test exception in beforeSpec override[jvm] FAILED
org.junit.platform.commons.JUnitException: TestEngine with ID 'kotest' failed to discover tests
Caused by: org.junit.platform.commons.JUnitException: TestEngine with ID 'kotest' failed to discover tests
Caused by: io.kotest.engine.config.ConfigurationException: io.github.classgraph.ClassGraphException: Uncaught exception during scan
Caused by: io.github.classgraph.ClassGraphException: Uncaught exception during scan
Caused by: java.lang.OutOfMemoryError: Java heap space
```

OliverO2:
Fixes #4112

### DIFF
--- a/kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
+++ b/kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
@@ -67,12 +67,6 @@ tasks.withType<Test>().configureEach {
    }
 
    useJUnitPlatform()
-
-   // Tests run in a separate JVM which does not inherit the JVM arguments from the main build:
-   // https://docs.gradle.org/current/userguide/java_testing.html
-   maxHeapSize = "3G"
-   jvmArgs = listOf("-XX:MaxMetaspaceSize=756m", "-XX:+HeapDumpOnOutOfMemoryError", "-Dfile.encoding=UTF-8")
-
    systemProperty("kotestVersion", Ci.publishVersion)
 
    testLogging {
--- a/kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
+++ b/kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
@@ -68,6 +68,11 @@ tasks.withType<Test>().configureEach {
 
    useJUnitPlatform()
 
+   // Tests run in a separate JVM which does not inherit the JVM arguments from the main build:
+   // https://docs.gradle.org/current/userguide/java_testing.html
+   maxHeapSize = "3G"
+   jvmArgs = listOf("-XX:MaxMetaspaceSize=756m", "-XX:+HeapDumpOnOutOfMemoryError", "-Dfile.encoding=UTF-8")
+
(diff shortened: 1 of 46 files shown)


kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
aSemy, line 1: Kotest Gradle Plugin is tested in a bit of a quirky way - Gradle is launched via `java.lang.ProcessBuilder`, so I don't think the memory settings of the project's Test task will be passed onto the launched process. It'd be better to keep the memory footprint smaller.

aSemy:
I thought that the JVM args used in Test tasks used `org.gradle.jvmargs` by default 🤔. I've fixed OOM in other projects by increasing Xmx in `org.gradle.jvmargs`. I'll do some testing...

UPDATE: I did some testing and I can see in the info log that this PR does correctly increase the Xmx from the default, which is too low. https://github.com/gradle/gradle/blob/6a61e6d812b2cefb7fb529f62ab4c867ca80c1d7/platforms/core-runtime/launcher/src/main/java/org/gradle/launcher/daemon/configuration/DaemonParameters.java#L47

aSemy:
On GitHub CI Xmx is set to 12g - maybe we should reduce that to allow more memory for the tests?

https://github.com/kotest/kotest/blob/265f0cba28ded95f601ec830df229b5f1a7d93d7/.github/workflows/run-gradle.yml#L76


Also, I think that setting `MaxMetaspaceSize` is a bit out of fashion these days - setting Xmx should be enough, and then Java will pick a sensible default for `MaxMetaspaceSize`, as a percentage of Xmx.

OliverO2:
I did not check how the JVM would actually do allocations (grab configured limits from the OS instantly or increase up to configured limits). I also did not check what might be required. So I have no better insights than you on this topic. Would it be OK if you'd just adjust the setting as you see fit (as part of this PR or in a separate one)?

aSemy:
I was hoping you had more expertise 😅. I've pushed an update that I think should work better though, so let's see...

kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
OliverO2: OK, let's try this.

PR #4121, opened 2024-06-17
CI: Harmonize master and PR tests

OliverO2:
* Make tests equivalent on all runners
* Run tests on fast primary runner first
* Run tests on secondary runners only if primary succeeds
* Fail fast on master

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -43,7 +43,8 @@ jobs:
                  fi
                done < files.txt
 
-   validate-abi:
+   validate-api:
+      name: Validate API
       needs: [ determine-workflows-to-run ]
       if: ${{ github.repository == 'kotest/kotest' && needs.determine-workflows-to-run.outputs.run_tests == 'true' }}
       uses: ./.github/workflows/run-gradle.yml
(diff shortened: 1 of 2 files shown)


.github/workflows/PR.yml
```diff
@@ -43,61 +43,42 @@ jobs:
                  fi
                done < files.txt
 
-   api_check:
+   validate-abi:
```
sksamuel: validate-api ?

OliverO2: Took that from master. API is probably correct. Will change.

aSemy: I changed the name to be `validate-abi` becuase it's more accurate. The BCV dumps have a `.api` extension, but this might be changed to `.abi`

https://github.com/Kotlin/binary-compatibility-validator/issues/198

However, if `validate-api` is more understandable, then that is more important 👍

OliverO2: Good point! The BCV's README says
> binary API of a JVM part of a Kotlin library

So we have it all. (Why am I thinking of "Cake by the Ocean" now?)

What if we'd follow their naming and go with API for now, then switch to ABI when they do? That way, no one needs to translate names.

aSemy: Sounds good to me!

sksamuel (approved):

PR #4123, opened 2024-06-18
Improve stacktraces of shouldMatchEach

sebek64:
When using `shouldMatchEach`, the stacktraces only refer to this function, not to a particular failure. This change fixes that, so that:
- if there is a single failure, it merges the stacktraces
- if there are multiple failures, it explicitly puts the stacktraces into the string failures representations
It is a draft for now, because:
- it changes the public api (didn't find a way how to do that without this change)
- it doesn't change all operators consistently, only one
Any kind of feedback is welcome.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/CollectionMatchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/CollectionMatchers.kt
@@ -2,10 +2,12 @@ package io.kotest.matchers.collections
 
 import io.kotest.assertions.ErrorCollectionMode
 import io.kotest.assertions.errorCollector
+import io.kotest.assertions.failure
 import io.kotest.assertions.print.print
 import io.kotest.assertions.runWithMode
 import io.kotest.matchers.Matcher
 import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.MatcherResultWithError
 import io.kotest.matchers.neverNullMatcher
 
 fun <T> existInOrder(vararg ps: (T) -> Boolean): Matcher<Collection<T>?> = existInOrder(ps.asList())
@@ -167,17 +169,37 @@ fun <T> matchEach(expected: List<T>, asserter: (T, T) -> Unit): Matcher<Collecti
       }
    })
 
+private sealed interface MatchEachProblem {
+   val atIndex: Int
+   /** A short message without a stack trace. */
+   val shortMessage: String
+   /** A long message including a stack trace if available. */
+   val longMessage: String
+
+   data class SizeMismatch(override val atIndex: Int, val problem: String) : MatchEachProblem {
+      override val shortMessage: String
+         get() = problem
+      override val longMessage: String
+         get() = problem
+   }
+
+   data class Thrown(override val atIndex: Int, val problem: Throwable) : MatchEachProblem {
+      override val shortMessage: String
+         get() = problem.message ?: "Exception $problem thrown"
+      override val longMessage: String
+         get() = problem.stackTraceToString()
+   }
+}
+
 /**
  * Asserts that each element in the collection matches its corresponding matcher in [assertions].
  * Elements will be compared sequentially in the order given by the iterators of the collections.
  */
 fun <T> matchEach(assertions: List<(T) -> Unit>): Matcher<Collection<T>?> = neverNullMatcher { actual ->
-   data class MatchEachProblem(val atIndex: Int, val problem: String?)
-
    val problems = errorCollector.runWithMode(ErrorCollectionMode.Hard) {
       actual.mapIndexedNotNull { index, element ->
          if (index !in assertions.indices) {
-            MatchEachProblem(
+            MatchEachProblem.SizeMismatch(
                index,
                "Element has no corresponding assertion. Only ${assertions.size} assertions provided"
             )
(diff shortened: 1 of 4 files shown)


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/MatcherResultWithError.kt
```diff
@@ -1,6 +1,6 @@
 package io.kotest.matchers
 
-internal interface MatcherResultWithError : MatcherResult {
+interface MatcherResultWithError : MatcherResult {
```
Kantis, line 6:
```suggestion
@KotestInternal
interface MatcherResultWithError : MatcherResult {
```

`@KotestInternal` should be used to flag API:s that are exposed to make them available across modules within Kotest, but which shouldn't be used by Kotest users.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/CollectionMatchers.kt
```diff
@@ -176,41 +175,77 @@ fun <T> matchEach(expected: List<T>, asserter: (T, T) -> Unit): Matcher<Collecti
...
+      0 -> MatcherResult(
+         true,
+         { "" },
+         { "Expected some element to fail its assertion, but all passed." },
+      )
+
+      1 -> {
+         val problem = problems.first()
+         MatcherResultWithError(
+            (problem as? MatchEachProblem.Thrown)?.problem,
```
Kantis: Can we be sure it's not a single SizeMismatch?

e.g. 
```kotlin
emptyList<Int>().shouldMatchEach({ it shouldBe 5})
```

Would probably result in problems being a single SizeMismatch element?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/CollectionMatchers.kt
```diff
@@ -176,41 +175,77 @@ fun <T> matchEach(expected: List<T>, asserter: (T, T) -> Unit): Matcher<Collecti
...
+            false,
+            { "Expected each element to pass its assertion, but found issues at indexes: [${problem.atIndex}}]\n\n${problem.atIndex} => ${problem.shortMessage}" },
```
Kantis: Wouldn't it make more sense to use the long message when there's just the single failure, and shortMessage when there's multiple?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/CollectionMatchers.kt
```diff
@@ -176,41 +175,77 @@ fun <T> matchEach(expected: List<T>, asserter: (T, T) -> Unit): Matcher<Collecti
       }
    })
 
+private sealed interface MatchEachProblem {
+   val atIndex: Int
+   val shortMessage: String?
+   val longMessage: String?
+
```
Kantis: Can we make these not null, and construct some relevant message when we have `Thrown` and there's no message/stacktrace?

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/MatcherResultWithError.kt
sebek64: Thank, change applied.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/CollectionMatchers.kt
sebek64: If so, then the expression `(problem as? MatchEachProblem.Thrown)?.problem` is evaluated as `null`, which is ok for `MatcherResultWithError` I suppose.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/CollectionMatchers.kt
sebek64: No, that's the point of the change. I tried to add the kdocs to the interface to make it more understandable.
Long message contains stacktrace, so it is suitable for use if the stacktrace would otherwise be lost.
On the other hand, short message can be used if the stack trace is preserved via `MatcherResultWithError` - for a single failure.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/CollectionMatchers.kt
sebek64: Good point, fixed.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/CollectionMatchers.kt
Kantis: Okay, that makes sense. I think I still need to see how it changes the output of tests to fully understand the context of the changes.

sksamuel:
Is this ready to come out of draft?

sebek64:
Ok, rebased and moving out of draft. The concern "it doesn't change all operators consistently, only one" is still valid.

sksamuel (approved):

sebek64:
There are test failures, sorry about not fixing them immediately. I'll provide a fixed version in a few days.

sebek64:
Tests should be fixed.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/CollectionMatchersTest.kt
```diff
@@ -880,9 +881,7 @@ class CollectionMatchersTest : WordSpec() {
                   { it shouldBeGreaterThan 7 }
                )
             }.message shouldBe """
-               Expected each element to pass its assertion, but found issues at indexes: [2]
```
AlexCue987: the number of failed assertion seems meaningful, is there a reason it's removed?

sebek64: 2 is not a number here, but an index. The logic behind this removal is that the failure just one and it is identified by the stacktrace. It can be changed, but the API of `MatcherResult` is quite hard to work with, and I don't know the overall code so well, so I'm not sure if I wouldn't break anything.

AlexCue987: > 2 is not a number here, but an index. The logic behind this removal is that the failure just one and it is identified by the stacktrace. It can be changed, but the API of `MatcherResult` is quite hard to work with, and I don't know the overall code so well, so I'm not sure if I wouldn't break anything.

I agree that adding stacktrace is an improvement and it should definitely help in some situations, but IMO there are cases when it's just easier and faster to go by the index of the failed predicate. So can we please add back the index of the failed predicate.

AlexCue987 (changes requested):
LGTM except one small thing: please add back the index of that single failed predicate

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/CollectionMatchersTest.kt
sebek64: Ok, changed it that way.

AlexCue987 (approved):

sebek64:
Tried to fix the windows build.

sebek64:
As a result of my fix, I also broke macos build :( I'll try to find a solution for all platforms.

sksamuel:
What's left to get this merged? Anything I can help with ?

sebek64:
Actually I forgot about it, I'll try to find some time to fix the missing piece in the following days.

LeoColman:
Closing this one as it's a bit stale.
Let us know if we should reopen

marcel-sebek-dbag:
Yes, unfortunately I don't have time to finish this to pass all platform-specific tests. If I find the time, I'll reopen it.

sebek64:
Yes, I don't have time to fix all platform-specific test failures. I'll reopen it if I find some time to do it. Anyone is welcome to do it, of course.

Issue #4113, opened 2024-06-16
CI: Flaky tests on slow runners


OliverO2:
**:kotest-assertions:kotest-assertions-core:jvmTest**
* [x] com.sksamuel.kotest.assertions.timing.ContinuallyTest[jvm] (macos-latest)
  ```
  continually should[jvm]
      Test fail tests start off as passing then fail within the period[jvm] FAILED (3s)

      java.lang.AssertionError: Expected exception kotlin.Throwable but no exception was thrown.
  ```

  In progress: https://github.com/kotest/kotest/pull/4109

* [x] com.sksamuel.kotest.matchers.concurrent.suspension.ConcurrentTest[jvm] (windows-latest)
    ```
    Test should fail when given lambda pass before the given time range[jvm] FAILED (1.2s)
    
    java.lang.AssertionError: Expected exception java.lang.AssertionError but no exception was thrown.
    ```
  In progress: https://github.com/kotest/kotest/pull/4115

* [x] io.kotest.assertions.nondeterministic.UntilTest[jvm] (macos-latest)
    ```
    Test until with config[jvm] FAILED (2.2s)
  
    java.lang.AssertionError: Block failed after 2s; attempted 36 time(s)
    The first error was caused by: expected:<true> but was:<false>
    ```

  In progress: https://github.com/kotest/kotest/pull/4132

- [x] io.kotest.assertions.nondeterministic.EventuallyTest[jvm]

  ```  
    Test handle java assertion errors[jvm] FAILED
  
    java.lang.AssertionError: Block failed after 157ms; attempted 1 time(s)
    The first error was caused by: boom
    java.lang.AssertionError: boom
  ```

  In progress: #4114

**:kotest-assertions:kotest-assertions-shared:jvmTest**
* [x] com.sksamuel.kotest.assertions.until.UntilTest[jvm] (macos-latest)
    ```
    Test until with boolean predicate and interval[jvm] FAILED (2s)
    
    java.lang.AssertionError: Until block failed after 2.049s; attempted 36 time(s); FixedInterval(duration=10ms) delay between attempts
    Test until with patience config[jvm] FAILED (2s)
    
    java.lang.AssertionError: Until block failed after 2.016s; attempted 35 time(s); FixedInterval(duration=10ms) delay between attempts
    Test until with predicate and interval[jvm] FAILED
    
    java.lang.AssertionError: 115ms should be < 100ms
    ```

  In progress: https://github.com/kotest/kotest/pull/4132

[...]

PR #4124, opened 2024-06-18
Stabilize GlobalTimeoutTest and EngineTimeoutTest

aSemy:
Wait until the engine has completed before asserting the results.

Part of #4113

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
@@ -22,7 +22,7 @@ class CollectingTestEngineListener : AbstractTestEngineListener(), Mutex by Mute
    /**
     * Counts the number of times [engineStarted] and [engineFinished] have been invoked.
     *
-    * Used by [waitForEngineFinished], so that internal Kotest tests can wait until engines
+    * Used by [waitForEnginesFinished], so that internal Kotest tests can wait until engines
     * are finished before performing assertions.
     */
    private val activeEngineCount = MutableStateFlow<Int?>(null)
@@ -74,7 +74,7 @@ class CollectingTestEngineListener : AbstractTestEngineListener(), Mutex by Mute
    }
 
    /** Suspends until at least one engine has started, and all engines have finished. */
-   internal suspend fun waitForEngineFinished() {
+   internal suspend fun waitForEnginesFinished() {
       // wait until `engineFinished()` has been invoked the same number of times as `engineStarted()`
       activeEngineCount.first { it == 0 }
    }
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
@@ -5,6 +5,9 @@ import io.kotest.core.names.TestName
 import io.kotest.core.spec.Spec
 import io.kotest.core.test.TestCase
 import io.kotest.core.test.TestResult
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlin.reflect.KClass
@@ -17,14 +20,14 @@ class CollectingTestEngineListener : AbstractTestEngineListener(), Mutex by Mute
    var errors = false
 
    /**
-    * A [Mutex] that will remain locked until [engineFinished] is invoked.
+    * Counts the number of times [engineStarted] and [engineFinished] have been invoked.
     *
     * (Using a read-write mutex would be more appropriate, but it's not available yet.
     * See https://github.com/Kotlin/kotlinx.coroutines/issues/94)
     *
     * @see waitForEngineFinished
     */
-   private val engineFinishedLock = Mutex(locked = true)
+   private val activeEngines = MutableStateFlow<Int?>(null)
 
    fun result(descriptor: Descriptor.TestDescriptor): TestResult? = tests.mapKeys { it.key.descriptor }[descriptor]
    fun result(testname: String): TestResult? = tests.mapKeys { it.key.name.testName }[testname]
@@ -49,9 +52,13 @@ class CollectingTestEngineListener : AbstractTestEngineListener(), Mutex by Mute
       names += testCase.name.testName
    }
 
+   override suspend fun engineStarted() {
+      activeEngines.update { (it ?: 0) + 1 }
+   }
+
    override suspend fun engineFinished(t: List<Throwable>): Unit = withLock {
       if (t.isNotEmpty()) errors = true
-      engineFinishedLock.unlock()
+      activeEngines.update { (it ?: 0) - 1 }
    }
 
    /**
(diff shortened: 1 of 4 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
```diff
@@ -16,31 +17,38 @@ class CollectingTestEngineListener : AbstractTestEngineListener(), Mutex by Mute
    val names = mutableListOf<String>()
    var errors = false
 
+   /**
+    * An active [Job] that will be completed when [engineFinished] is invoked.
+    * @see waitForEngineFinished
+    */
+   private val engineFinishedJob = Job()
```
OliverO2: This is an interesting solution, using a Job as a boolean-like state which one can wait for while suspending.

It took me some time to decipher, as I had to make sure that the Job
* does not perform actual work, and
* does not have children (in which case `complete()` would not make `join()` return immediately).

Could you add some comment to that effect?

([The same could be achieved with a `Mutex` (playground)](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMi4wLjAiLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIiwibm9uZU1hcmtlcnMiOnRydWUsInRoZW1lIjoiaWRlYSIsImNvZGUiOiJpbXBvcnQga290bGlueC5jb3JvdXRpbmVzLmRlbGF5XG5pbXBvcnQga290bGlueC5jb3JvdXRpbmVzLmxhdW5jaFxuaW1wb3J0IGtvdGxpbnguY29yb3V0aW5lcy5ydW5CbG9ja2luZ1xuaW1wb3J0IGtvdGxpbnguY29yb3V0aW5lcy5zeW5jLk11dGV4XG5pbXBvcnQga290bGlueC5jb3JvdXRpbmVzLnN5bmMud2l0aExvY2tcbmltcG9ydCBrb3RsaW4udGltZS5EdXJhdGlvbi5Db21wYW5pb24uc2Vjb25kc1xuXG5mdW4gbWFpbigpOiBVbml0ID0gcnVuQmxvY2tpbmcge1xuICAgIHZhbCBlbmdpbmVGaW5pc2hlZCA9IE11dGV4KGxvY2tlZCA9IHRydWUpXG5cbiAgICBsYXVuY2gge1xuICAgICAgICBkZWxheSgyLnNlY29uZHMpXG4gICAgICAgIGVuZ2luZUZpbmlzaGVkLnVubG9jaygpXG4gICAgICAgIHByaW50bG4oXCJmaW5pc2hlZFwiKVxuICAgIH1cblxuICAgIGxhdW5jaCB7XG4gICAgICAgIGRlbGF5KDAuNS5zZWNvbmRzKVxuICAgICAgICBwcmludGxuKFwib2JzZXJ2ZXIgMTogd2FpdGluZ1wiKVxuICAgICAgICBlbmdpbmVGaW5pc2hlZC53aXRoTG9jayB7fVxuICAgICAgICBwcmludGxuKFwib2JzZXJ2ZXIgMTogZGV0ZWN0ZWQgZmluaXNoXCIpXG4gICAgfVxuXG4gICAgbGF1bmNoIHtcbiAgICAgICAgZGVsYXkoMS5zZWNvbmRzKVxuICAgICAgICBwcmludGxuKFwib2JzZXJ2ZXIgMjogd2FpdGluZ1wiKVxuICAgICAgICBlbmdpbmVGaW5pc2hlZC53aXRoTG9jayB7fVxuICAgICAgICBwcmludGxuKFwib2JzZXJ2ZXIgMjogZGV0ZWN0ZWQgZmluaXNoXCIpXG4gICAgfVxuXG4gICAgbGF1bmNoIHtcbiAgICAgICAgZGVsYXkoMy5zZWNvbmRzKVxuICAgICAgICBwcmludGxuKFwib2JzZXJ2ZXIgMzogd2FpdGluZ1wiKVxuICAgICAgICBlbmdpbmVGaW5pc2hlZC53aXRoTG9jayB7fVxuICAgICAgICBwcmludGxuKFwib2JzZXJ2ZXIgMzogZGV0ZWN0ZWQgZmluaXNoXCIpXG4gICAgfVxufVxuIn0=), but that would also require some comment explaining to its uncommon use. I have a similar thing called [`Gate` in this KotlinConf talk companion repo](https://github.com/OliverO2/kmp-e2e-stress-testing-example/blob/main/application/library-testing/src/commonMain/kotlin/library/testing/Gate.kt).)

OliverO2 (changes requested):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
aSemy: Ahh yes, the `mutex.withLock {}` is probably better. I'll update the code and with a comment.

OliverO2 (approved):

aSemy:
Hmmm apparently `engineFinished()` can get called multiple times?

```
> Task :kotest-framework:kotest-framework-engine:jvmTest

com.sksamuel.kotest.engine.extensions.spec.BeforeSpecListenerTest[jvm]
  Test BeforeSpecListener inline should be triggered before tests[jvm] FAILED (2.3s)
  java.lang.IllegalStateException: This mutex is not locked
  Caused by: java.lang.IllegalStateException: This mutex is not locked

FAILURE: Executed 1423 tests in 6m 5s (1 failed, 186 skipped)
```

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
```diff
@@ -16,31 +19,44 @@ class CollectingTestEngineListener : AbstractTestEngineListener(), Mutex by Mute
    val names = mutableListOf<String>()
    var errors = false
 
+   /**
+    * Counts the number of times [engineStarted] and [engineFinished] have been invoked.
+    *
+    * Used by [waitForEngineFinished], so that internal Kotest tests can wait until engines
+    * are finished before performing assertions.
+    */
+   private val activeEngines = MutableStateFlow<Int?>(null)
```
OliverO2:
```suggestion
   private val activeEngineCount = MutableStateFlow<Int?>(null)
```

`activeEngines` suggests some collection of `Engine`s. I find that in such cases `count` helps at use sites.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/CollectingTestEngineListener.kt
```diff
@@ -56,4 +72,10 @@ class CollectingTestEngineListener : AbstractTestEngineListener(), Mutex by Mute
    fun TestCase.toKey(): TestCaseKey {
       return TestCaseKey(this.descriptor, this.name, this.spec::class)
    }
+
+   /** Suspends until at least one engine has started, and all engines have finished. */
+   internal suspend fun waitForEngineFinished() {
```
OliverO2:
```suggestion
   internal suspend fun waitForEnginesFinished() {
```

OliverO2 (approved):
Looks good, given the new insights. I also wonder why the same listener would observe multiple engines. 🤔

aSemy:
> Looks good, given the new insights. I also wonder why the same listener would observe multiple engines. 🤔

Oh, I figured it out and forgot to update my message. The reason is one of the test re-uses a listener, which I think is reasonable.

https://github.com/kotest/kotest/blob/e2ebd497ee304128ae2bf7325505565b3263b73c/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/extensions/spec/BeforeSpecListenerTest.kt#L78-L90

OliverO2:
Thanks for clarifying! Helps to demystify what I suspected was one more piece of magic. Now I can relax again. :)

PR #4132, opened 2024-06-19
Stabilize tests on slow runners

OliverO2:
Stabilize tests:
* Relax timing in UntilTest (2 variants)
* Extend timeout in PerformanceTest
* Relax timing in deprecated test UntilSpec
* Relax timing in MultipleTestTimeoutTest
* Relax timing in SpecInvocationTimeoutShouldNotApplyToContainersTest
* Relax timing in PropTestConfigConstraintsTest
* Extend timeout in PerformanceTest
* Relax performance expectations in CluesTests
* Increase blocking time in engine.concurrency.TimeoutTest
* Relax timing in assertions.async.TimeoutTest
* Support separate runs of tests marked with `@Tags("Deprecated")`
* Make tag-dependent tests immune to tags set in system property

Not in scope: Run deprecated tests separately (blocked by #4135)

### DIFF
--- a/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
@@ -1,5 +1,6 @@
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+import utils.SystemPropertiesArgumentProvider
 
 plugins {
    `java-library`
@@ -13,15 +14,9 @@ testlogger {
 
 tasks.withType<Test>().configureEach {
    useJUnitPlatform()
-   // Pass "kotest.*" system properties from the Gradle invocation to the test launcher.
-   // https://junit.org/junit5/docs/current/user-guide/#running-tests-config-params
-   for ((name, value) in System.getProperties()) {
-      if (name is String && name.startsWith("kotest.")) {
-         value as String
-         inputs.property(name, value)
-         systemProperty(name, value)
-      }
-   }
+
+   val kotestSystemProps = providers.systemPropertiesPrefixedBy("kotest")
+   jvmArgumentProviders += SystemPropertiesArgumentProvider(kotestSystemProps)
    filter {
       isFailOnNoMatchingTests = false
    }
--- a/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
@@ -13,6 +13,15 @@ testlogger {
 
 tasks.withType<Test>().configureEach {
    useJUnitPlatform()
+   // Pass "kotest.*" system properties from the Gradle invocation to the test launcher.
+   // https://junit.org/junit5/docs/current/user-guide/#running-tests-config-params
+   for ((name, value) in System.getProperties()) {
+      if (name is String && name.startsWith("kotest.")) {
+         value as String
+         inputs.property(name, value)
+         systemProperty(name, value)
+      }
+   }
    filter {
       isFailOnNoMatchingTests = false
    }
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -66,7 +66,7 @@ jobs:
    validate-primary-deprecated:
       name: Deprecated tests on primary runner
       if: ${{ github.repository == 'kotest/kotest' && needs.determine-workflows-to-run.outputs.run_tests == 'true' }}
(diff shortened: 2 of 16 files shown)


OliverO2:
With these changes, the remaining flaky test failures should be covered by the other PRs mentioned in #4113.

buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
```diff
@@ -13,6 +13,15 @@ testlogger {
...
    useJUnitPlatform()
+   // Pass "kotest.*" system properties from the Gradle invocation to the test launcher.
+   // https://junit.org/junit5/docs/current/user-guide/#running-tests-config-params
+   for ((name, value) in System.getProperties()) {
+      if (name is String && name.startsWith("kotest.")) {
+         value as String
+         inputs.property(name, value)
+         systemProperty(name, value)
+      }
+   }
```
aSemy: I think that, unfortunately, iterating over all system properties will make configuration cache, and possibly build cache, be dependent on unrelated system properties. That means the caches can't be reused whenever an unrelated system property changes.

Fortunately Gradle has `providers.systemPropertiesPrefixedBy("kotest")`.

There's also a special way of passing in system properties to workaround [Gradle not fully supporting the Provider API](https://github.com/gradle/gradle/issues/12247).

I assume that you'll be happy with me fixing and pushing these changes directly?

OliverO2: I've actually thought about that, but was not aware of the above solution. So yes, go ahead by any means! Thank you!

aSemy: Thanks!

Is there a list of valid Kotest system properties?

OliverO2: Not that I'm aware of.

OliverO2: This one: https://github.com/kotest/kotest/blob/0df229949563cca89f97770eb0b6e554fe997b8b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/internal/KotestEngineProperties.kt

OliverO2: I've searched for call sites of `syspropOrEnv()`. All system properties are listed in `KotestEngineProperties.kt`, plus two isolated ones: `KOTEST_DEBUG`, `KOTEST_DEBUG_PATH`. That should be complete.

OliverO2: There are these extra places with system properties without an environment counterpart:
* https://github.com/kotest/kotest/blob/397466e17774dc98e2f479965aad91b0670b92b6/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/AssertionsConfig.kt
* https://github.com/kotest/kotest/blob/b72c3729e6a4142d448bfe016e4cb582ea44a7b9/kotest-common/src/commonMain/kotlin/io/kotest/common/intellij.kt

aSemy: Thanks!

I wanted to have a look because if any of the properties were constantly changing (e.g. a timestamp), or some other system-dependent property (like a file path that changes per machine), then we'd have to take some extra care to avoid registering them as task inputs and breaking Gradle up-to-date checks.

For now I'll not register the values as task inputs. That means that even if they change then Gradle might not re-run the test tasks, but that's the case already.

It would be nice to have these defined in a single place, and in such a way that the Kotest Gradle Plugin could provide a nice DSL to set them in the correct way.

aSemy (approved):
Nice!

I think that ideally some of these tests should be converted to use virtual time - but I only had a cursory look. Just thinking out loud, and pondering whether to revisit them...

OliverO2:
Regarding virtual time: Surely a good idea for anything coroutine-only. Would not work for stuff that actually verifies real-time performance, blocking, thread count.

For now, I'd immediately merge this one and encourage you to do the same with the other stabilization PRs. That way we'll again have master merges, snapshot builds and updates of build cache files in GitHub caches, so that future tests run faster. 

Feel free to revisit places touched by this PR for further optimization later.

buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
OliverO2: When we switch to tag-based testing, registering values as input might become essential: We'll oscillate between
`-Dkotest.tags=!Deprecated` and `-Dkotest.tags=Deprecated` in consecutive test runs. But we can postpone covering that until we really need it.

PR #4136, opened 2024-06-20
Publish Kotest KMP Gradle Plugin to Maven Central

### LINKED ISSUE #4012: Kotest Gradle plugin is missing in snapshot publications

aSemy:
Resolves #4012


OliverO2 (approved):
I assume this "just works". Could this be tested somehow outside GH CI?

aSemy:
> I assume this "just works". Could this be tested somehow outside GH CI?

Hmmmm I guess the only way to really test it is to try publishing to Maven Central snapshots!

Publishing in general can be tested by running `./gradlew publishAllPublicationsToRootBuildDirRepository`, and check `$rootDir/build/maven-repo/`. But that won't test misc publishing config, like signing (unless you have some valid signing config set up.)

OliverO2:
Wouldn't that change also push the KMP Gradle plugin to Sonatype snapshots? Either way, I'd say let's just try and fail forward if anything breaks.

PR #4138, opened 2024-06-21
Enable concurrency cancellation in PR test GitHub Action

aSemy:
Enable concurrency to decrease CI pressure. (E.g. If a PR receives several commits in quick succession, only the latest commit will be run and the earlier ones will be cancelled.)


OliverO2 (approved):
Good idea to use that apparently frequently used practice!

PR #4140, opened 2024-06-21
Making helper classes internal instead, so we don't accidentally miss…

Kantis:
… something using @KotestInternal


AlexCue987 (approved):
LGTM

PR #4143, opened 2024-06-22
CI: run Gradle check on all OSes in parallel

aSemy:
Currently CI has to complete tests on Ubuntu before starting macOS and Windows test. This adds ~20 minutes to the CI run.

This PR runs tests in all PRs in parallel, immediately. This will reduce the duration of testing PRs and the merge queue.

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -57,24 +57,15 @@ jobs:
          ref: ${{ inputs.ref }}
          task: apiCheck --continue
 
-   validate-primary:
-      name: Validate on primary runner
+   gradle-check:
+      name: Run gradle check
       if: ${{ github.repository == 'kotest/kotest' && needs.determine-workflows-to-run.outputs.run_tests == 'true' }}
       needs: [ determine-workflows-to-run, validate-api ]
-      uses: ./.github/workflows/run-gradle.yml
-      with:
-         runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


OliverO2 (approved):
For now, if it helps to advance, a good thing. In the future, if we're seeing mostly positive results, which I'm hoping, I'll remain in favor. Otherwise we might consider to change back and avoid wasting resources.

aSemy:
> For now, if it helps to advance, a good thing. In the future, if we're seeing mostly positive results, which I'm hoping, I'll remain in favor. Otherwise we might consider to change back and avoid wasting resources.

Sounds good, we can always revisit it.

Once the tests are most stable, another option is to set `fail-fast: true`

sksamuel:
> Currently CI has to complete tests on Ubuntu before starting macOS and Windows test. This adds ~20 minutes to the CI run.
> 
> This PR runs tests in all PRs in parallel, immediately. This will reduce the duration of testing PRs and the merge queue.

This was done on purpose because macos runners are limited to I think 2 concurrent runs.

Since most failures are in the common layer, tying up the mac runners when your linux build fails, is  a waste of the mac resources.

aSemy:
> This was done on purpose because macos runners are limited to I think 2 concurrent runs.

I haven't heard of that restriction before, and I couldn't find anything in the docs. Perhaps it's an old restriction, and it's been lifted?

PR #4146, opened 2024-06-22
CI: Enable Configuration Cache re-use

aSemy:
Configure Gradle Configuration Cache re-use on CI by providing an encryption key.

This will speed up Gradle builds on CI.

See: https://github.com/gradle/actions/blob/v3.4.2/docs/setup-gradle.md#saving-configuration-cache-data


OliverO2 (approved):
Fine!

Kantis (approved):

PR #4149, opened 2024-06-22
Use available virtual time automatically, harmonize time measurement

OliverO2:
Provide testTimeSource() which automatically returns virtual time in
(test) coroutines configured to use it, TimeSource.Monotonic otherwise.

Contains stabilization fixes for
* GlobalTimeoutTest
* EngineTimeoutTest
* LongHttpRequestTest
* engine.concurrency.TimeoutTest
* SpecInvocationTimeoutShouldNotApplyToContainersTest
* TestTimeoutOverridesSpecFunctionTest
* EventuallyTest
* EventuallySpec


kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/timeout/GlobalTimeoutTest.kt
```diff
@@ -1,40 +1,39 @@
...
 import io.kotest.matchers.shouldBe
 import kotlinx.coroutines.delay
-import kotlin.time.Duration.Companion.days
 
 class GlobalTimeoutTest : FunSpec() {
    init {
-      test("global timeouts should apply if no other timeout is set") {
-         val c = ProjectConfiguration().apply { timeout = 3 }
+      test("global timeouts should apply if no other timeout is set")
+      {
```
aSemy: could you reformat this file please?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/timeout/GlobalTimeoutTest.kt
```diff
@@ -1,40 +1,39 @@
...
-
-         collector.waitForEnginesFinished()
-
-         collector.names.shouldContainExactly("blocked", "suspend")
-
-         collector.result("blocked").asClue { result -> result?.isError shouldBe true }
-         collector.result("suspend").asClue { result -> result?.isError shouldBe true }
+         collector.tests.mapKeys { it.key.name.testName }["blocked"]?.isError shouldBe true
+         collector.tests.mapKeys { it.key.name.testName }["suspend"]?.isError shouldBe true
       }
```
aSemy: Can the assertions be kept? They're more useful than the ones in this PR.

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/timeout/GlobalTimeoutTest.kt
OliverO2: Sure. Just waiting for the re-run to complete before pushing again.

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/timeout/GlobalTimeoutTest.kt
OliverO2: You mean these?
```kotlin
         collector.names.shouldContainExactly("blocked", "suspend")

         collector.result("blocked").asClue { result -> result?.isError shouldBe true }
         collector.result("suspend").asClue { result -> result?.isError shouldBe true }
```

No problem.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/timing/ContinuallyTest.kt
```diff
@@ -1,51 +1,63 @@
...
-        var n = 0
-        val e = shouldThrow<Throwable> {
+   init {
+      coroutineTestScope = true
+
+      "continually" should {
+         "pass working tests" {
+            continually(500.milliseconds) {
+               nonConstantTrue() shouldBe true
+            }
```
aSemy, line 30: Just to double check, is this PR supposed to make the deprecated `continually` use virtual time? I'm asking because test takes 500ms, but I'm not sure about whether it's supposed to use virtual time or not.

kotest-common/src/jvmMain/kotlin/io/kotest/common/nonConstantResults.kt
aSemy, line 1: Could and/or should `nonConstantTrue()` and `nonConstantFalse()` be hidden with `@KotestInternal`?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/timing/ContinuallyTest.kt
OliverO2: Good catch! In fact, continually is still using `TimeSource.Monotonic`, so the above is wrong. I'll try to change continually to `testTimeSource()` in order to avoid bothering us with flaky tests while continually is still around.

kotest-common/src/jvmMain/kotlin/io/kotest/common/nonConstantResults.kt
OliverO2: I've thought about that, but these seem quite stable and might be useful for Kotest users. WDYT?

aSemy: Can you think of an example usage why a Kotest user might use it? It'd help me understand the context better.

I would prefer avoiding adding top level functions, to avoid cluttering auto-complete suggestions as much as possible.

aSemy: I think they're really useful, and make Kotest's tests easier to understand. But are these functions necessary for this PR? If not, I'd really appreciate it if they were split out into a separate PR.

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/timeout/EngineTimeoutTest.kt
```diff
@@ -18,50 +17,41 @@ class EngineTimeoutTest : FunSpec() {
...
          val collector = CollectingTestEngineListener()
          TestEngineLauncher(collector)
             .withClasses(BillyBlocked::class)
             .launch()
-         collector.waitForEnginesFinished()
          collector.names shouldBe listOf("a")
-         collector.result("a").asClue { result ->
-            result?.errorOrNull?.message shouldBe "sleep interrupted"
-         }
+         collector.result("a")!!.errorOrNull!!.message!! shouldBe "sleep interrupted"
```
aSemy: Could the assertions here be kept? I think that `collector.result("a").asClue { result -> }` will be more useful if the test fails than `collector.result("a")!!`

kotest-common/src/commonMain/kotlin/io/kotest/common/TestTimeSourceContextElement.kt
aSemy, line 1: Nice work de-duping the test time source!

kotest-common/src/jvmMain/kotlin/io/kotest/common/nonConstantResults.kt
OliverO2: Dropping them completely for this PR would be tedious: There was `System.currentTimeMillis()` all over the place for dual use:
* to actually obtain wall clock time, and
* to obtain some value preventing the compiler from optimizing it away.

I wanted a clear separation between these use cases and also use Kotlin TimeMark throughout.

What I could do is `@KotestInternal`ize them for now and then expose them later. To me they seem like the equivalent of JMH's sink holes.

kotest-common/src/commonMain/kotlin/io/kotest/common/TestTimeSourceContextElement.kt
OliverO2: Was more like synchronizing it with the scheduler used. It would be error-prone to continue doing that individually and lead to accidentally using a real-time source with a virtual-time scheduler and vice versa.

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/timeout/EngineTimeoutTest.kt
OliverO2: Sure.

aSemy:
Just a thought: is automatically using virtual time whenever `coroutineTestScope` is enabled a breaking change? Might it cause issues in user's code?

kotest-common/src/jvmMain/kotlin/io/kotest/common/nonConstantResults.kt
aSemy: > To me they seem like the equivalent of JMH's sink holes.

I agree, which is why I would be careful about including them 😁. Writing performance tests is hard, and should probably be done with the proper tooling (i.e. kotlinx-benchmark or Lincheck).

OliverO2:
Isn't a fix of broken functionality always a breaking change? So in this case, I'd think this leads to the question: Could it make sense to use real time in coroutines that operate on virtual time? What would such real time measurements tell me? Of course, we could mention prominently in the release notes that this fix might have effects (i.e. expose bugs) in user tests.

But that makes me think of another question: Normally, I would not use the test scheduler in conjunction with other dispatchers. But it is possible. Maybe I could revisit `testTimeSource()` to make sure we are still on the test dispatcher, and not some child coroutine launched on a different dispatcher.

kotest-common/src/jvmMain/kotlin/io/kotest/common/nonConstantResults.kt
```diff
@@ -0,0 +1,9 @@
+package io.kotest.common
+
+/** returns `true` while preventing the compiler from optimizing it away. */
+@KotestInternal
+fun nonConstantTrue() = System.currentTimeMillis() > 0L
```
AlexCue987, line 5: the name `nonConstantTrue` does not seem intuitive to me. how about `alwaysTrueNotOptimizedAway`?

OliverO2: `NotOptimizedAway` is not guaranteed by the code. So while that's the intention, it is also a bit misleading. In any case, this stuff is internal only and I my view probably not worth another iteration.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/timing/ContinuallyTest.kt
```diff
@@ -1,51 +1,63 @@
...
+         }
+
+         "fail broken tests immediately" {
+            shouldThrowAny {
+               continually(12.hours) {
+                  nonConstantFalse() shouldBe true
+               }
+            }
```
AlexCue987, line 38: this test verifies "fail broken tests", but does it verify "immediately"?

AlexCue987 (approved):
After merging the latest commits from master, one test has failed:
```
9 tests completed, 1 failed
> Task :kotest-tests:kotest-tests-spec-parallelism:jvmTest
After Project Error[jvm]

  Test initializationError[jvm] FAILED

  java.lang.IllegalStateException: Parallel execution failure: Execution time was 3.649012400s
```

should we merge this PR anyway, because it's a huge improvement already?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/timing/ContinuallyTest.kt
OliverO2: It sees an immediate failure, but does not verify that it fails immediately. This should probably be addressed in another PR since it is unrelated to this PR's changes and motivation.

OliverO2:
There will be a couple of "hard real-time" tests, which would still fail occasionally with slow runners after merging this PR. I have another idea to address such cases: A global adjustment factor for real-time timeouts could be set in a runner-dependent way, so that fast runners fail fast, while slow runners get some extra time. But I'd favor a separate PR for this.

WRT merging, we had a long discussion on Slack (https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1719243849207139). As I see it, this one is safe to merge, but there may be still be other ideas out there, so I'm not sure if there is sufficient buy-in.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/timing/ContinuallyTest.kt
AlexCue987: > addressed in another PR

this

AlexCue987:
> As I see it, this one is safe to merge, but there may be still be other ideas out there, so I'm not sure if there is sufficient buy-in.

I'd suggest merging it: it should make my life much easier, even if it does not solve all the problems just yet. Thanks for all these improvements!

OliverO2:
@aSemy @sksamuel What do you think?

aSemy:
I think there are loads of good improvements, and this PR will absolutely improve the Kotest tests, but I'm not convinced that the non-deterministic test utils (`eventually`/`continually`/et al) should use virtual time by default, or even be configurable (outside of Kotest - we need to be able to control it of course).

I'm not convinced because it's a change in behaviour, which could be surprising. Also, it implies some support for something that conceptually doesn't make sense to me. The non-deterministic utils should be used to configure how tests themselves are run, and shouldn't be affected by what is being tested. When virtual time is enabled, nothing should be non-deterministic, so using `eventually` doesn't make sense - the result should be the same every time.

The non-deterministic utils should only respect virtual time within the context of testing Kotest. We need to be able to fake non-deterministic code, and verify that the non-deterministic utils behave appropriately.

I would prefer instead if the times (basically [the same approach that I implemented in this PR](https://github.com/kotest/kotest/pull/4114/files#diff-968ee99b2a2d357d516a492d3d3a547cca18bad8a499c7e2651f986e24bfcdccR438-R454), but updated so that the same CoroutineContext TimeSource could be injected & re-used for all non-deterministic utils.

So overall, it's a good PR, I would just like to make sure we're on aware of and on board with the change. Either way, if others agree or disagree, then I'm happy to move on!

OliverO2:
I agree that we have a change in behavior, but _only_ "for something that doesn't make sense". So Kotest would not be backwards bug-compatible.

As I understand it, the suggested solution is to make the non-deterministic functions respond to some Kotest-internal virtual time. This deviates from the default virtual time available with the well-known test dispatcher. Doing so would mean an extra layer for us to learn and maintain, increasing complexity.

So for me, that brings up the question: Are the possible issues relevant enough to warrant that extra layer?

I'm just not convinced that we're addressing relevant use cases here (and we have tried really hard to find some). So I'm reluctant to commit code that has no clearly defined purpose justifying the added complexity.

But I also wouldn't mind if someone would jump in to change things that way and maintain it (maybe by just another PR so that commit records are clear). Would that be a way to proceed?

OliverO2:
Maybe I have something that addresses the above concerns, yet is not too heavy for me to accommodate:

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/Spec.kt
```diff
@@ -305,6 +306,13 @@ abstract class Spec : TestConfiguration() {
 
    var coroutineTestScope: Boolean? = null
 
+   /**
+    * Flag controlling whether virtual time is enabled for non-deterministic test functions (`eventually`, ...).
+    * This is for Kotest-internal testing only and should not be used outside Kotest.
+    */
+   @KotestInternal
+   var nonDeterministicTestVirtualTimeEnabled = false
+
```
aSemy: 👍

kotest-common/src/commonMain/kotlin/io/kotest/common/testTimeSource.kt
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * This is [TimeSource.Monotonic] or – if separately enabled – virtual time, depending on the scheduler in use.
+ */
+@OptIn(KotestInternal::class)
+suspend fun nonDeterministicTestTimeSource(): TimeSource =
+   coroutineContext.nonDeterministicTestCoroutineSchedulerOrNull?.timeSource ?: TimeSource.Monotonic
+
+@OptIn(KotestInternal::class)
+private val CoroutineContext.nonDeterministicTestCoroutineSchedulerOrNull: TestCoroutineScheduler?
+   get() = if (this[NonDeterministicTestVirtualTimeEnabled] != null) testCoroutineSchedulerOrNull else null
```
aSemy: I think it would be better if these functions were also marked as `@KotestInternal`.

- Enabling them requires using the ``, which is also marked as `@KotestInternal`, so it makes sense to me to also mark the utils as internals.
- They're specifically for Kotest internal testing, and so shouldn't be confused with the regular `testTimeSource()` that most users should be using.

```suggestion
@KotestInternal
suspend fun nonDeterministicTestTimeSource(): TimeSource =
   coroutineContext.nonDeterministicTestCoroutineSchedulerOrNull?.timeSource ?: TimeSource.Monotonic

@KotestInternal
private val CoroutineContext.nonDeterministicTestCoroutineSchedulerOrNull: TestCoroutineScheduler?
   get() = if (this[NonDeterministicTestVirtualTimeEnabled] != null) testCoroutineSchedulerOrNull else null
```

kotest-common/api/kotest-common.api
```diff
@@ -14,6 +14,21 @@ public abstract interface annotation class io/kotest/common/JVMOnly : java/lang/
 public abstract interface annotation class io/kotest/common/KotestInternal : java/lang/annotation/Annotation {
 }
 
+public final class io/kotest/common/NonConstantResultsKt {
+	public static final fun nonConstantFalse ()Z
+	public static final fun nonConstantTrue ()Z
+}
+
+public final class io/kotest/common/NonDeterministicTestVirtualTimeEnabled : kotlin/coroutines/CoroutineContext$Element, kotlin/coroutines/CoroutineContext$Key {
```
aSemy, line 22: `NonDeterministicTestVirtualTimeEnabled` is marked as `@KotestInternal`, so I'm confused that it's here. Shouldn't it be excluded from the API dump? 🤔

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/Spec.kt
aSemy:
```suggestion
   /**
    * Flag controlling whether virtual time is enabled for non-deterministic test functions (`eventually`, ...).
    * This is for Kotest-internal testing only and should not be used outside Kotest.
    */
   @KotestInternal
   var nonDeterministicTestVirtualTimeEnabled: Boolean = false

```

A minor nitpick of mine: I think it's best to explicitly state the type, especially in non-private functions. imo it's easier to scan and to read, and indicates a stricter contract. WDYT?

kotest-common/src/commonMain/kotlin/io/kotest/common/testTimeSource.kt
aSemy: I have a KDoc suggestion too: add a reference to the flag that enables these functions.

```suggestion
/**
 * Returns the [TimeSource] used for non-deterministic test functions.
 *
 * This is [TimeSource.Monotonic] or – if separately enabled – virtual time, depending on the scheduler in use.
 *
 * @see io.kotest.core.spec.Spec.nonDeterministicTestVirtualTimeEnabled
 */
@OptIn(KotestInternal::class)
suspend fun nonDeterministicTestTimeSource(): TimeSource =
   coroutineContext.nonDeterministicTestCoroutineSchedulerOrNull?.timeSource ?: TimeSource.Monotonic

/**
 * Returns the [TimeSource] used for non-deterministic test functions, or `null` if it is not present.
 *
 * @see nonDeterministicTestTimeSource
 * @see io.kotest.core.spec.Spec.nonDeterministicTestVirtualTimeEnabled
 */
@OptIn(KotestInternal::class)
private val CoroutineContext.nonDeterministicTestCoroutineSchedulerOrNull: TestCoroutineScheduler?
   get() = if (this[NonDeterministicTestVirtualTimeEnabled] != null) testCoroutineSchedulerOrNull else null
```

kotest-common/api/kotest-common.api
OliverO2: The Kotlin BCV (`apiDump`/`apiCheck`) does not know about `@KotestInternal`. That could be changed by [configuring the BCV](https://github.com/Kotlin/binary-compatibility-validator?tab=readme-ov-file#optional-parameters), but would be beyond the scope of this PR.

kotest-common/src/commonMain/kotlin/io/kotest/common/testTimeSource.kt
OliverO2: (Overlapping suggestions don't work. Addressing this separately.)

kotest-common/api/kotest-common.api
aSemy: Got it, thanks!

kotest-common/src/commonMain/kotlin/io/kotest/common/testTimeSource.kt
OliverO2: The cross-reference to `io.kotest.core.spec.Spec.nonDeterministicTestVirtualTimeEnabled` won't work (different module). Better leave it out since it's all internal anyway and is better detected by using IDE lookup?

Adding `@see nonDeterministicTestTimeSource` to the private `nonDeterministicTestCoroutineSchedulerOrNull` pointing to its adjacent call site seems to distract more than it would help.

OliverO2:
Merging this right away to make the entire CI train run again.

aSemy (approved):

Issue #4151, opened 2024-06-26
shouldBeEqualToComparingFields not working with generic fields


luisttk:
Using Kotest 5.8.1

The following test fails: 

```Kotlin

data class ParameterValue<T : Any>(val name: String, val value: T)

class TestCase : ShouldSpec({

    should("consider them different") {
        ParameterValue("Test Param", "value") shouldNotBeEqualToComparingFields
                ParameterValue("Test Param", "well well")
    }

})
```

`shouldBeEqualToComparingFields` shows the same behaviour. If `shouldBe` is used or value is replaced with a regular String in the class definition the test works.

AlexCue987:
I'll take a look.

AlexCue987:
fixed

PR #4157, opened 2024-07-02
Should be equal to comparing fields should support generics 

AlexCue987:
fix for https://github.com/kotest/kotest/issues/4151
also some preparation for matching fields that are lists/maps/sets, which will bre really nice to have


nvmanh95:
hi @AlexCue987 any idea when this can be merged?

AlexCue987:
> hi @AlexCue987 any idea when this can be merged?

unfortunately, no idea - the PR needs to bew reviewed, then we need to cut a release

sksamuel (approved):

PR #4158, opened 2024-07-02
shouldContainJsonKey-partial-match

AlexCue987:
make `shouldContainJsonKey` search for partial key match


Kantis (approved):

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/keys.kt
```diff
@@ -28,17 +32,27 @@ fun containJsonKey(path: String) = object : Matcher<String?> {
...
-         true
-      } catch (t: PathNotFoundException) {
-         false
-      } catch (t: InvalidPathException) {
-         throw AssertionError("$path is not a valid JSON path")
-      }
-
-      return MatcherResult(
-         passed,
+      return value?.let {
```
Kantis: Perhaps doing an early return for null would make this code simpler?

AlexCue987: > Perhaps doing an early return for null would make this code simpler?

yeah. short circuited earlier. good call.

PR #4164, opened 2024-07-09
Opt in for expect-actual classes

aSemy:
> Expected and actual classes are in [Beta](https://kotlinlang.org/docs/components-stability.html). They are almost stable, but migration steps may be required in the future. We'll do our best to minimize any further changes for you to make.
> 
> https://kotlinlang.org/docs/multiplatform-expect-actual.html#expected-and-actual-classes

Part of #4085


OliverO2 (approved):
Looks good!

Issue #4163, opened 2024-07-08
withData fails on data class that has a function named after field


rafagapa:
Kotest version: 5.9.1

The following:
```
import io.kotest.core.spec.style.FreeSpec
import io.kotest.datatest.withData

class WithDataTest : FreeSpec ({

    data class DataClass(val a: Int, val b: String) {
        fun b(value: String) = copy(b = value)
    }

    "with data test" - {
        withData(
            DataClass(1, "A"),
            DataClass(2, "B"),
        ) {  }
    }
})
```
will crash with:
```
Callable expects 2 arguments, but 1 were provided.
java.lang.IllegalArgumentException: Callable expects 2 arguments, but 1 were provided.
	at kotlin.reflect.jvm.internal.calls.Caller$DefaultImpls.checkArguments(Caller.kt:20)
	at kotlin.reflect.jvm.internal.calls.CallerImpl.checkArguments(CallerImpl.kt:15)
	at kotlin.reflect.jvm.internal.calls.CallerImpl$Method$Instance.call(CallerImpl.kt:112)
	at kotlin.reflect.jvm.internal.KCallableImpl.call(KCallableImpl.kt:108)
	at io.kotest.mpp.JvmReflection$primaryConstructorMembers$1$1$1.invoke(reflection.kt:81)
	at io.kotest.mpp.StableKt.hasStableMembers(stable.kt:63)
	at io.kotest.mpp.StableKt.isStable(stable.kt:41)
	at io.kotest.datatest.StableIdentifiers.stableIdentifier(StableIdentifiers.kt:48)
	at io.kotest.datatest.IdentifierKt.getStableIdentifier(Identifier.kt:12)
	at io.kotest.datatest.ContainerKt$withData$4.invoke(container.kt:38)
	at io.kotest.datatest.ContainerKt$withData$4.invoke(container.kt:38)
	at io.kotest.datatest.ContainerKt.withData(container.kt:73)
	at io.kotest.datatest.ContainerKt.withData(container.kt:38)
	at io.kotest.datatest.ContainerKt.withData(container.kt:19)
	at tests.WithDataTest$1$1.invokeSuspend(WithDataTest.kt:14)
```
Looks like the code get's confused and tries to call `b(String)` function as a getter for `b` field.

[...]

PR #4173, opened 2024-07-11
fix edge case for isStable - when property and function have same name

AlexCue987:
fix edge case for isStable - when property and function have same name:
https://github.com/kotest/kotest/issues/4163


sksamuel (approved):

sksamuel:
nice

sksamuel:
Failing tests but logic looks good

sksamuel:
I've fixed the tests

PR #4174, opened 2024-07-13
use Dev Publish plugin to test Kotest plugin

aSemy:
Instead the test-project using `includedBuild()` to depend on Kotest, instead, publish Kotest to a project-local directory.

This will improve test performance, because it will allow running the Gradle Plugin tests in parallel.

It also improves the Kotest build's stability:

- Avoids discouraged Gradle practices (cross-project dependencies and configuration).
- Improve Build Cache and Config Cache compatibility.

Additional changes:

- Tidy up test-project build config, use Kotlin DSL utils.
- Tidy up GradleInvocation.


sksamuel:
What does this do?

aSemy:
> What does this do?

The Dev Publish plugin publishes subprojects to a project-local directory, sort of like Maven Local. This avoids the downsides of the current approach (cross-project configuration is a discouraged Gradle practice, because it negatively impacts parallelisation and caching), and also the negatives of Maven Local itself (which can be 'polluted' with other artifacts, and it's not practical to fingerprint and cache it).

My expectation is that using Dev Publish will improve the speed of running tests in general, but specifically on CI.

I'm looking generally improving the Kotest plugin test because [it takes significantly longer than other modules](https://scans.gradle.com/s/ewwxg4lzvsgl6/timeline?page=1&sort=longest#xrjmcakehxjxy), despite only having a handful of tests. That will probably always be the case, but I think some of the time can be cut-down.

buildSrc/src/main/kotlin/kotest-publishing-conventions.gradle.kts
```diff
@@ -23,21 +24,41 @@ val ossrhPassword: String by project
 val signingKey: String? by project
 val signingPassword: String? by project
 
+val mavenCentralRepoName = "Deploy"
```
OliverO2, line 27: Is the case change from "deploy" to "Deploy" intentional?

buildSrc/src/main/kotlin/kotest-publishing-conventions.gradle.kts
```diff
@@ -103,9 +124,11 @@ val mavenPublishLimiter =
...
    }
 
-tasks.withType<PublishToMavenRepository>().configureEach {
-   usesService(mavenPublishLimiter)
-}
+tasks.withType<PublishToMavenRepository>()
+   .matching { it.name.endsWith("PublicationTo${mavenCentralRepoName}Repository") }
+   .configureEach {
+      usesService(mavenPublishLimiter)
+   }
```
OliverO2, line 131: Isn't `tasks.withType` an eager API? If so:

```suggestion
tasks
   .named { it.name.endsWith("PublicationTo${mavenCentralRepoName}Repository") }
   .configureEach {
      usesService(mavenPublishLimiter)
   }
```

OliverO2 (approved):
Looks all very plausible and well written! Full disclosure: I only had a cursory look at `dev-publish-plugin` and I do not consider myself an expert on publishing.

buildSrc/src/main/kotlin/kotest-publishing-conventions.gradle.kts
aSemy: iirc `tasks.withType<> { ... }` is eager, but `tasks.withType<>().configureEach { ... }` is lazy.

OliverO2: Maybe that's the case. Their API surface is, well, amazingly complex. [Docs say](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html)
> [TaskCollection.withType()](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskCollection.html#withType-java.lang.Class-)
Ok to use. If chained withType().getByName(), use TaskCollection.named() instead.

If that's the case, the above would just be shorter as the task subtype is not required for `usesService()`.

aSemy: And from what I understand the new `named {}` filter is bugged and can sometimes be eager, so I try to avoid it.

https://github.com/gradle/gradle/issues/28347

buildSrc/src/main/kotlin/kotest-publishing-conventions.gradle.kts
aSemy: Good spot, yes, it's intentional. 

Gradle uses the repository names to create tasks, and it will capitalize the repo name. So it doesn't matter if the repository is named `deploy` or `Deploy`, the task names will still be `publishFooToDeployRepository`.

Defining the name as `Deploy` consistently saves a few characters below, when filtering based on the task name

```
it.name.endsWith("PublicationTo${mavenCentralRepoName}Repository", ignoreCase = true)
```

buildSrc/src/main/kotlin/kotest-publishing-conventions.gradle.kts
OliverO2: Makes sense, excellent!

aSemy: yes, that's true, `withType<>()` isn't needed here. But I think it makes it clearer that the code should only be configuring publishing tasks. Also, I'm not a big fan of string-based checking, so I feel a bit more comfortable with a type check :)

OliverO2: Yeah, I can relate. Although here, string matching is the decisive part, and the extra type check is a double-edged sword, providing an extra guard as well as an extra failure mode.

PR #4183, opened 2024-07-20
Optimise PBT seed files

aSemy:
Improve performance of property-based tests by avoiding file IO operations.

- Delete seed files once PBT iterations have completed sucessfully, instead of after every individual sucessful PBT iteration.
- General tidying.
- Now that performance is better, re-enable tests for `forAll` with 19, 20, 22 inputs.
- Make `proptest()` a bit simpler - create new `onSuccess()` helper function.
- Lazily initialise directories.
- Remove unnecessary `@file:Suppress("NOTHING_TO_INLINE")` (helps with #4085)

### Related

- #4113 This PR will improve performance.
- [Slack discussion with flamegraphs](https://slack-chats.kotlinlang.org/t/22733478/i-ve-been-doing-some-digging-into-some-of-the-https-scans-gr#29192b52-b79d-4f2d-83c0-1650329b269a).

### DIFF
--- a/kotest-property/api/kotest-property.api
+++ b/kotest-property/api/kotest-property.api
@@ -1574,13 +1574,6 @@ public final class io/kotest/property/seed/SeedKt {
 	public static final fun writeFailedSeedIfEnabled (JLkotlin/coroutines/Continuation;)Ljava/lang/Object;
 }
 
-public final class io/kotest/property/seed/SeedioKt {
-	public static final fun clearSeed (Lio/kotest/common/TestPath;)V
-	public static final fun readSeed (Lio/kotest/common/TestPath;)Ljava/lang/Long;
-	public static final fun seedPath (Lio/kotest/common/TestPath;)Ljava/nio/file/Path;
-	public static final fun writeSeed (Lio/kotest/common/TestPath;J)V
-}
-
 public final class io/kotest/property/statistics/CoverageKt {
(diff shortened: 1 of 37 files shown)


kotest-property/api/kotest-property.api
```diff
@@ -1574,13 +1574,6 @@ public final class io/kotest/property/seed/SeedKt {
 	public static final fun writeFailedSeedIfEnabled (JLkotlin/coroutines/Continuation;)Ljava/lang/Object;
 }
 
-public final class io/kotest/property/seed/SeedioKt {
-	public static final fun clearSeed (Lio/kotest/common/TestPath;)V
-	public static final fun readSeed (Lio/kotest/common/TestPath;)Ljava/lang/Long;
-	public static final fun seedPath (Lio/kotest/common/TestPath;)Ljava/nio/file/Path;
-	public static final fun writeSeed (Lio/kotest/common/TestPath;J)V
-}
-
```
aSemy, line 1583: #### ⚠️ Breaking change

I changed the visibility of `fun clearSeed()`, `fun readSeed()`, and `fun writeSeed()` to `internal`. This is a breaking change, but I don't see any reason why someone would be using these.

In principle, these functions should go through a deprecation cycle, where they are first marked as `@KotestInternal` and then made `internal`. But I don't think it's important for such small utils, and given that other seed-file code is marked as `@ExperimentalKotest`.

Thoughts?

sksamuel: I'm good with this.

sksamuel (approved):

PR #4184, opened 2024-07-20
Fix Kotest Gradle Plugin publishing

### LINKED ISSUE #4012: Kotest Gradle plugin is missing in snapshot publications

aSemy:
- Use `withJavadocJar()` for Gradle Plugin publishing.
- Enable `gradlePlugin.isAutomatedPublishing = true`
- Only require signing if releasing (helps with testing locally)

Also, tidy up Gradle plugin metadata.

Fix #4012 (hopefully!)


aSemy:
I tested this locally by running `gradle publishAllPublicationsToRootBuildDirRepository` and checking `./build/maven-repo/`. It seems to be correct, but I don't know exactly what Sonatype's [400 error message](https://github.com/kotest/kotest/issues/4012#issuecomment-2241035622) means.

sksamuel (approved):

PR #4189, opened 2024-07-20
Move some framework only commons into engine module

sksamuel:
My aim is to keep reducing how much is shared in commons for when we split the projects.


LeoColman (approved):

OliverO2 (approved):
Makes sense.

PR #4190, opened 2024-07-20
Remove deprecated/relocated Arb reflection utils

aSemy:
Remove Arb reflection utilities that are deprecated as "this function will be removed in 5.6". Kotest is currently v5.9.

Part of #4085


sksamuel:
I think we should not remove anything until we're ready to start on 6.0
At this point there's little to be gained from removing something if we're going to release 5.9.2
However if we're ready for 6.0 now, then lets split the repo's and start removing all the deprecated stuff ?

aSemy:
For things that deprecated as "will be removed in 6.0" I agree, but apparently these should have been removed already.

They can be kept, but in this instance I think it makes sense to remove them.

Otherwise, we can update the deprecation notice to say "will be removed in 6.0"?

sksamuel:
Yeah I think since it says 5.6 we could definitely remove them and stick to our contract but there's not really much benefit to us if we're about to start publishing 6.0 snapshots? So maybe update warning.

sksamuel (approved):
Master is now 6.0 snapshot so this can be merged.

PR #4193, opened 2024-07-21
Remove GitHub Run ID from SNAPSHOT version

### LINKED ISSUE #4012: Kotest Gradle plugin is missing in snapshot publications

aSemy:
The intention behind adding `GITHUB_RUN_NUMBER` to the version was to allow for users to select a specific, unchanging version. 

[Slack discussion](https://slack-chats.kotlinlang.org/t/22736345/what-s-the-reason-for-including-the-github-build-number-in-t#8dcb2ce5-6e1b-493d-a16d-4cea5d5b3154)

However, a consistent version is automatically available when publishing a regular `-SNAPSHOT` version. Maven will automatically timestamp each release, meaning a specific version will be available.

By removing the build ID, users have a choice of versions:

- the latest snapshot, e.g. `5.9.0-SNAPSHOT` (dynamic)
- or a specific snapshot `5.9.0-20240511.091617-1451` (unchanging)


buildSrc/src/main/kotlin/Ci.kt
```diff
@@ -1,19 +1,24 @@
...
 
-   // this is the version used for building snapshots
-   // .buildnumber-snapshot will be appended
-   private const val snapshotBase = "6.0.0"
+   /**
+    * The base version used for the release version.
+    *
+    * `-SNAPSHOT` or `-LOCAL` will be appended.
+    */
+   private const val snapshotBase = "5.10.0"
```
OliverO2:
```suggestion
   private const val snapshotBase = "6.0.0"
```

Or is there a reason for keeping the old version?

OliverO2 (approved):

buildSrc/src/main/kotlin/Ci.kt
aSemy: Oops, that's a mistake. Thanks for checking. I made this PR before the version was updated to 6.0.0.

PR #4194, opened 2024-07-21
Suffix Kotest Gradle Plugin snapshot versions with `-SNAPSHOT`

### LINKED ISSUE #4012: Kotest Gradle plugin is missing in snapshot publications

aSemy:
Use the same versioning scheme for the Kotest Gradle Plugin, so that it also has `-SNAPSHOT` versions.

This change is required by Sonatype to permit publishing Kotest Gradle Plugin snapshots.

Fixes (hopefully!) #4012 (again!)

### DIFF
--- a/kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
+++ b/kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
@@ -9,9 +9,6 @@ plugins {
    alias(libs.plugins.gradle.plugin.publish)
 }
 
-group = "io.kotest"
-version = Ci.gradleVersion
-
 dependencies {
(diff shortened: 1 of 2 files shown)


kotest-framework/kotest-framework-multiplatform-plugin-gradle/build.gradle.kts
```diff
@@ -9,9 +9,6 @@ plugins {
    alias(libs.plugins.gradle.plugin.publish)
 }
 
-group = "io.kotest"
-version = Ci.gradleVersion
```
aSemy, line 13: Removed, because they are set in `kotest-publishing-conventions.gradle.kts`

OliverO2 (approved):

PR #4196, opened 2024-07-21
Kotest Gradle Plugin - use remote build cache in tests

aSemy:
Depends on #4174


LeoColman:
https://github.com/kotest/kotest/pull/4174 was merged. Should we move forward with this guy?

aSemy:
I've thought more about this, and I think that the tests shouldn't use remote build cache. The project being tested is small, and introducing remote build cache is an unnecessary complication.

PR #4197, opened 2024-07-21
Move more commons out to the projects that need it

sksamuel:
As part of 6.0 all deprecated methods are being removed.


OliverO2 (approved):
A bit hard to review, but at least I could not find any obvious problems. Just seems that the lexicographic order of import statements is now broken.

sksamuel:
> A bit hard to review, but at least I could not find any obvious problems. Just seems that the lexicographic order of import statements is now broken.

I've just moved things about, no code changes.

Is there a setting in intellij for the order ?

OliverO2:
> Is there a setting in intellij for the order ?

I'm not sure, I'm using the Ktlint plugin with auto-formatting on save for my own stuff. That one does ordering. I suppose IntelliJ by default does that, too, as it has also settings on whether to accept star imports and such.

sksamuel:
> > Is there a setting in intellij for the order ?
> 
> I'm not sure, I'm using the Ktlint plugin with auto-formatting on save for my own stuff. That one does ordering. I suppose IntelliJ by default does that, too, as it has also settings on whether to accept star imports and such.

I have star imports disabled.

OliverO2:
Yeah, me too. Just mentioned it to support my suspicion that IJ can be pretty opinionated wrt. imports.

PR #4201, opened 2024-07-21
Kotest Gradle Plugin: remove memoryModel tests

aSemy:
The memory model flags have been deprecated, and will be removed.

```
w: -memory-model and memoryModel switches are deprecated and will be removed in a future release.
```

I propose removing these tests, since the flag is deprecated and the Gradle plugin tests are slow.


kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePluginSpec.kt
```diff
@@ -73,42 +73,6 @@ class KotestMultiplatformCompilerGradlePluginSpec : ShouldSpec({
...
-
-                  val result = invocation.run()
-
-                  withClue(result.clue) {
-                     taskNames.forAtLeastOne { taskName ->
-                        // Depending on the host machine these tests are running on,
-                        // only one of the test targets will be built and executed.
-                        shouldHaveExpectedTestResultsFor(taskName)
-                     }
-                  }
```
OliverO2, line 108: You could remove the old memory model tests, but would you really want to remove all native tests?

OliverO2 (changes requested):

kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePluginSpec.kt
aSemy: Good point! I mistakenly thought that the other `should {}` block tested the native targets.

OliverO2 (approved):
Looks good!

PR #4204, opened 2024-07-22
Include private specs by default #3915

### LINKED ISSUE #3879: Private Test Classes

sksamuel:
This is a breaking change. In 5.x private classes were excluded.
The system property and config option allows users to revert to this behavior.

Closes #3879

### DIFF
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
@@ -298,6 +298,12 @@ abstract class AbstractProjectConfig {
    open var testCoroutineDispatcher: Boolean = Defaults.testCoroutineDispatcher
 
    /**
+    * If set to false then private spec classes will be ignored by the test engine.
+    * Defaults to true.
+    */
+   open var includePrivateClasses: Boolean? = null
+
+   /**
     * Executed before the first test of the project, but after the
     * [ProjectListener.beforeProject] methods.
     */
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
@@ -301,7 +301,7 @@ abstract class AbstractProjectConfig {
     * If set to false then private spec classes will be ignored by the test engine.
     * Defaults to true.
     */
-   open var includePrivateClasses: Boolean? = null
+   open var ignorePrivateClasses: Boolean? = null
 
    /**
     * Executed before the first test of the project, but after the
--- a/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
+++ b/kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
@@ -299,7 +299,6 @@ abstract class AbstractProjectConfig {
 
    /**
     * If set to false then private spec classes will be ignored by the test engine.
-    * Defaults to true.
     */
    open var ignorePrivateClasses: Boolean? = null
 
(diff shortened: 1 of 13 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/applyConfigFromAbstractProjectConfig.kt
```diff
@@ -42,6 +42,7 @@ internal fun applyConfigFromProjectConfig(config: AbstractProjectConfig, configu
    // discovery
    config.discoveryClasspathFallbackEnabled?.let { configuration.discoveryClasspathFallbackEnabled = it }
    config.disableTestNestedJarScanning?.let { configuration.disableTestNestedJarScanning = it }
+   config.includePrivateClasses?.let { configuration.ignorePrivateClasses = it }
```
LeoColman: Include vs ignore

I think at some point it got inverted

sksamuel: Thanks.

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/spec/ClassVisibilitySpecRefInterceptor.kt
```diff
@@ -0,0 +1,34 @@
...
+         ref.kclass.visibility == KVisibility.PRIVATE && !allowPrivate() -> Result.success(emptyMap())
+         else -> fn(ref)
+      }
+   }
+
+   /**
+    * We allow private classes if the configuration flag or system property to ignore is set to true.
+    */
+   private fun allowPrivate(): Boolean {
+      return context.configuration.ignorePrivateClasses ||
```
LeoColman, line 30: Ignore seems inverted here too

sksamuel: I've gone with ignorePrivateClasses in this revision.
Since include is now the default.
Just pushed some changes + test.

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
```diff
@@ -297,6 +297,12 @@ abstract class AbstractProjectConfig {
    @ExperimentalKotest
    open var testCoroutineDispatcher: Boolean = Defaults.testCoroutineDispatcher
 
+   /**
+    * If set to false then private spec classes will be ignored by the test engine.
+    * Defaults to true.
+    */
+   open var ignorePrivateClasses: Boolean? = null
```
LeoColman, line 303: Can it really be null if we are saying the default is true?

LeoColman: I also assume the default is false now

LeoColman (approved):

kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/config/AbstractProjectConfig.kt
sksamuel: yep fixed

OliverO2:
Will private classes be discovered on native targets?

sksamuel:
> Will private classes be discovered on native targets?

No, that's determined by the compiler plugin, not the runtime lookup that the JVM uses.

OliverO2:
A cursory look shows that the plugin actually does detect private specs so with this PR we should have consistency between targets.

I wonder if we should switch JVM discovery to the compiler plugin as well and have it all in one place, and fast.

sksamuel:
I'm going to see how we do this with the new kotest gradle plugin. At the moment, we rely on gradle to supply the list of test classes. Still experimenting. In master now, there is no more classgraph though which is nice.

Issue #4208, opened 2024-07-22
Gradle build: use caching mirror instead of Maven Central


cloudshiftchris:
Maven Central has had to [introduce throttling](https://www.sonatype.com/blog/maven-central-and-the-tragedy-of-the-commons) to manage the influx of traffic directly to the authoritative repositories.

Instead of using Maven Central directly, and contributing the load, use a caching mirror, e.g.:
```
      // use a caching mirror for maven central. https://www.sonatype.com/blog/maven-central-and-the-tragedy-of-the-commons
      maven("https://cache-redirector.jetbrains.com/repo1.maven.org/maven2") {
         name = "JetBrains Maven Central mirror"
      }
```

In most cases these repositories won't be accessed due to Gradle caching for local and CI builds; it will help when there are routine dependency updates, and moreso when there are major upgrades that flush the local and/or CI cache (new workstations, new developers, Gradle upgrade, etc).

This also applies to the Gradle Plugin Portal repository which redirects to Maven Central.

I'll create a PR for this...

aSemy:
Hey, thanks for raising this. I am very interested in improving performance and avoiding unnecessary downloading.

However, I don't think using JetBrain's cache redirectors is a good idea. They aren't documented, there's no reliability or stability guarantees, and they aren't intended for public use.

I also suspect that GitHub already provides some level of caching for downloading dependency (although I can't find any docs to support this). This is most evident when [the Kotlin Native distributions are downloaded](https://scans.gradle.com/s/b5bpixt4fioyy/console-log?page=1#L19). 200MB in 10ms is quite impressive!

Kotest uses the [setup-gradle](https://github.com/gradle/actions/tree/main/setup-gradle) action, which will use the [GitHub Action cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows). This will cache downloaded dependencies. (Although it's bugged at the moment, but that will hopefully be fixed by #4207.)

cloudshiftchris:
Thanks for the feedback.  Had selected the JetBrains mirror as a) am using it elsewhere and b) it's used farther down in that settings.gradle.kts for npm & yarn deps.

Fair point re: unsupported/undocumented, unable to find any official references - seems like this has leaked out into the wild...

[...]

PR #4209, opened 2024-07-22
Use caching mirror repositories instead of Maven Central / Gradle Plugin Portal directly

cloudshiftchris:
Details in #4208 - using caching mirrors instead of Maven Central / Gradle Plugin Portal directly.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -2,8 +2,13 @@ rootProject.name = "kotest"
 
 pluginManagement {
    repositories {
-      mavenCentral()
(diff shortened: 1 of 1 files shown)


aSemy:
I would like to see better caching, but I don't think using JetBrains servers is the right way. [See my comments here](https://github.com/kotest/kotest/issues/4208#issuecomment-2243886297).

Maybe there's something I've missed, or there's something we can discuss more, but let's do that in the original issue you raised. To keep the discussion focused, I'll close this PR.

PR #4210, opened 2024-07-23
Added listener for disabled tests #4179

### LINKED ISSUE #4179: Intercept disabled tests
I would like to add to the report a list of tests that were disabled.

Is it possible to write an extension that intercepts disabled tests?

**BeforeEachListener** or **BeforeTestListener** callback won't be invoked if test is disabled.

kozlov-aa:
Closes #4179


AlexCue987:
Very interesting. Can you elaborate on the problem(s) you are trying to solve, such as in which context this feature is going to help?

kozlov-aa:
@AlexCue987 I am working on some automating testing project. We are using [Allure](https://allurereport.org/) as reporting tool. For sure we are integrate with allure by [kotest-extensions-allure](https://github.com/kotest/kotest-extensions-allure), but the report that this extension generates does not include ignored tests. This happen because [AllureTestReporter.kt](https://github.com/kotest/kotest-extensions-allure/blob/master/src/main/kotlin/io/kotest/extensions/allure/AllureTestReporter.kt) using **BeforeTestListener** and **AfterTestListener** hooks which does not invoke for ignored tests. However, if I am generate report by Junit native integration with Allure, report contains ignored tests, because Kotest tells Junit that test was ignored. This enhancement can help me to solve this problem. I just using new hook to inrich report by disabled tests.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/TestEnabledCheckInterceptor.kt
```diff
@@ -32,6 +35,7 @@ internal class TestEnabledCheckInterceptor(private val configuration: ProjectCon
          }
          false -> {
             logger.log { Pair(testCase.name.testName, "Test is disabled: ${enabled.reason}") }
+            testExtensions.disabledTestListenersInvocation(testCase, enabled.reason)
```
sksamuel: We prob want to fire this in the same place we fire the test engine listener, so we capture all places ?

flanker-c: I think yes, we want this.

flanker-c: So I think I should rename DisabledTestListener to IgnoredTestListener besause disable flags are not the only reason why a test might be ignored.

sksamuel (approved):

kozlov-aa:
@sksamuel Forgot about :apiDump task.

sksamuel (approved):

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/interceptors/EnabledCheckTestExecutionInterceptorTest.kt
```diff
@@ -3,15 +3,19 @@ package com.sksamuel.kotest.engine.test.interceptors
...
 import io.kotest.core.test.TestCase
 import io.kotest.core.test.TestResult
 import io.kotest.core.test.TestType
-import io.kotest.engine.test.scopes.TerminalTestScope
 import io.kotest.engine.test.interceptors.TestEnabledCheckInterceptor
+import io.kotest.engine.test.scopes.TerminalTestScope
 import io.kotest.matchers.shouldBe
+import io.mockk.coVerify
+import io.mockk.every
+import io.mockk.spyk
```
OliverO2, line 18: These are not used, right?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestExtensions.kt
```diff
@@ -151,4 +152,12 @@ internal class TestExtensions(private val registry: ExtensionRegistry) {
    fun logExtensions(testCase: TestCase): List<LogExtension> {
       return extensions(testCase).filterIsInstance<LogExtension>()
    }
+
+   /**
+    * Invokes all [IgnoredTestListener] for this test.
```
OliverO2:
```suggestion
    * Invokes all [IgnoredTestListener]s for this test.
```

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/extensions/test/IgnoredTestListenerTest.kt
```diff
@@ -0,0 +1,72 @@
...
+
+   val ignoredTestListener = spyk<IgnoredTestListener>()
+
+   register(ignoredTestListener)
+
+   describe("should be disabled by enabled flag in config").config(enabled = false) {
+      error("still enabled")
+   }
+
+   describe("!should be disabled by band") {
```
OliverO2:
```suggestion
   describe("!should be disabled by bang") {
```

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/extensions/test/IgnoredTestListenerTest.kt
```diff
@@ -0,0 +1,72 @@
...
+   describe("ignoredTestListener should be invoked") {
+      coVerify(exactly = 1) {
+         ignoredTestListener.ignoredTest(
+            testCase = match { it.name.testName == "should be disabled by enabled flag in config" },
+            reason = eq("Disabled by enabled flag in config")
+         )
+      }
+      coVerify(exactly = 1) {
+         ignoredTestListener.ignoredTest(
+            testCase = match { it.name.testName == "should be disabled by band" },
```
OliverO2:
```suggestion
            testCase = match { it.name.testName == "should be disabled by bang" },
```

OliverO2 (commented):
Some nitpicks.

I wonder if we could use tests without mocking? (Makes reviews and maintenance easier for those not familiar with mockk.)

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/interceptors/EnabledCheckTestExecutionInterceptorTest.kt
sksamuel: Removed all mockk from tests

sksamuel:
I think this is breaking some behavior.

PR #4214, opened 2024-07-24
Fix mapcontain and cleanup - take2

### LINKED ISSUE #3959: `shouldContain` always matches against `null` map value
kotest 5.8.1

The following succeeds when it should fail:
```kotlin
mapOf("foo" to "bar") shouldContain ("baz" to null)
```

due to https://github.com/kotest/kotest/blob/92091b869f3a8649cb9f6eed292222a0cd899cab/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/maps/matchers.kt#L10

AlexCue987:
* fix https://github.com/kotest/kotest/issues/3959
* clean up the following function that has an exact duplicate:
```
fun <K, V> contain(key: K, v: V): Matcher<Map<K, V>> = object : Matcher<Map<K, V>> {
   override fun test(value: Map<K, V>) = MatcherResult(
      value[key] == v,
      { "Map should contain mapping $key=$v but was ${buildActualValue(value)}" },
      { "Map should not contain mapping $key=$v but was $value" }
   )

   private fun buildActualValue(map: Map<K, V>) = map[key]?.let { "$key=$it" } ?: map
}
```

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/maps/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/maps/matchers.kt
@@ -6,14 +6,34 @@ import io.kotest.matchers.should
 import io.kotest.matchers.shouldNot
 
 fun <K, V> mapcontain(key: K, v: V) = object : Matcher<Map<K, V>> {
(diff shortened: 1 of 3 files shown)


AlexCue987:
Merging this because these changes were already approved in https://github.com/kotest/kotest/pull/4213

PR #4216, opened 2024-07-25
Fix lost stack trace with assertSoftly (#4171)

### LINKED ISSUE #4171: Stack trace is lost on certain assertion errors with assertSoftly
Kotest version: 5.9.1

Happens if:

- there is only one assertion error collected by `assertSoftly`
- the type of assertion error is other than `AssertionFailedError` (e.g. `AssertionError`)

E.g.:
```
import io.kotest.matchers.should
import io.kotest.matchers.string.beEmpty

class Test : StringSpec ({
    "Two errors" {
        assertSoftly {
            1 shouldBe 2                  // (1)
            null should beEmpty()         // (2)
        }
    }
    "One error" {
        assertSoftly {                    // (3)
            null should beEmpty()         // (4)
        }
    }
})
```
results with (cleaned up the output for better readability):
```
// "Two errors" output
1) expected:<2> but was:<1>
   // stack trace points to (1) as expected
2) Expecting actual not to be null
   // stack trace points to (2) as expected

// "One error" output
Expecting actual not to be null
   // stack trace points to (3), i.e. the start of assertSoftly block, should be (4)
```

In general it's not a big deal as the "broken" stack trace points to the `assertSoftly` block, but with `kotest.framework.assertion.globalassertsoftly = true` it starts to get really messy.

rafagapa:
I've identified a few places where the stack trace of the original exception is lost. 

The verification based on the line numbers might be a bit brittle, but this is the only thing I've came up with.

Fixes #4171


sksamuel (approved):

sksamuel:
👍🏻

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/AssertSoftlyTest.kt
```diff
@@ -254,6 +269,33 @@ class AssertSoftlyTest : FreeSpec({
             error.message shouldNotContain "3) "
          }
       }
+      "doesn't loose stack traces" - {
```
OliverO2:
```suggestion
      "doesn't lose stack traces" - {
```

Same at other change locations.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/ErrorCollector.kt
```diff
@@ -176,6 +176,7 @@ inline fun <reified T> ErrorCollector.runWithMode(mode: ErrorCollectionMode, blo
 internal fun List<Throwable>.toAssertionError(depth: Int, subject: Printed?): AssertionError? {
    return when {
       isEmpty() -> null
+      // todo: here we loose the original stack trace
```
aSemy: Is TODO still relevant, or can it be removed?

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/ErrorCollector.kt
```diff
@@ -63,6 +63,7 @@ private class ErrorCollectorContextElement(private val coroutineLocalErrorCollec
 
 actual fun ErrorCollector.collectiveError(): AssertionError? {
    fun prefixWithSubjectInformation(e: AssertionFailedError, subject: Printed) =
+      // todo: here we loose the original stack trace
```
aSemy: Is TODO still relevant, or can it be removed?

PR #4218, opened 2024-07-28
Improve Gradle Java support

aSemy:
Kotest is supposed to support Java 8, but there are a couple of places in the code that use Java utils that require 9+. This can be resolved by [testing using a fixed Java Toolchain](https://docs.gradle.org/current/userguide/toolchains.html#toolchains_for_tasks) to always test using Java 8.

The JDK used to compile Kotest can also be fixed. Using newer JDKs is advantageous because of improved compilation performance and fixes. The Java compiler can be configured to emit classes compatible with Java 8 with the [`-release` flag](https://www.baeldung.com/java-compiler-release-option). (Kotlin has an analogous `-Xjdk-release` flag.)

To ensure compatibility going forward, I registered a new test run for launching the JVM tests with the maximum supported JDK. This will run automatically.

Using fixed Java Toolchains for compilation and running tests improves build performance, cacheability, and stability.

### Usage

The Java versions used to compile, run, and test Kotest can be configured in `libs.versions.toml`

https://github.com/kotest/kotest/blob/c8666f1bbf8dac5564ff4e5c3690fe3a9834438c/gradle/libs.versions.toml#L25-L31

To test Java 8, run `./gradlew jvmTest`. To test Java 21, run `./gradlew jvmMaxJdkTest`.

### Summary

- Use Java toolchains to run and compile the build.
- Add Foojay toolchain resolver, that will automatically download the required toolchains.
- Use Java 21 to run Gradle and compile the code.
- Set the minimum supported Java version to 8.
- Set the maximum supported Java version to 21.
- Create a new test task to run the tests with Java 21.
- Fix code to support Java 8

PR #4221, opened 2024-07-28
Use Java 21 to run Gradle Daemon

aSemy:
To maximise Gradle cache hits, and to use the best Gradle features, use Java 21 to run Gradle builds.

Gradle has a new feature for running the Gradle Daemon with a fixed Java version. This can be configured to run automatically. Currently Gradle does not auto-provision the necessary Java version, but hopefully this will be available in Gradle 8.10.

This PR should be merged after #4218, which will ensure that a stable JDK is used for compilation and running tests.


build.gradle.kts
```diff
@@ -12,3 +14,9 @@ apiValidation {
       )
    )
 }
+
+configureGradleDaemonJvm(
```
sksamuel, line 18: what does this block do ?

sksamuel (approved):

PR #4222, opened 2024-07-28
Add coroutines opt-ins in ChannelTest

aSemy:
Part of #4085


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/channels/ChannelTest.kt
aSemy, line 1: The file wasn't formatted with the default indentation, so I recommend ignoring whitespace.

image

sksamuel (approved):

PR #4223, opened 2024-07-28
Fix compiler warnings in kotest-assertions-json

aSemy:
Refactor kotest-assertions-json main and test code to fix compiler warnings - #4085.

Also, update KDoc.

Also, uncomment `JsonLiteralsTest` and `EqualTest`. I'm not sure why they were commented out. Maybe they can just be deleted?

### DIFF
--- a/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
+++ b/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
@@ -1,5 +1,3 @@
-@file:Suppress("unused")
-
 package io.kotest.assertions.json
 
 @Deprecated("Json comparison options is now specified with `CompareJsonOptions`", ReplaceWith("TypeCoercion"))
@@ -20,8 +18,8 @@ enum class CompareMode {
    /**
     * Compare by value, coercing if possible.
     *
-    * For example, "true" and true will match because the string value can be coerced into a valid boolean.
-    * Similarly, "100" and 100 will match as the former can be coerced into an int.
+    * For example, `true` and `true` will match because the string value can be coerced into a valid boolean.
+    * Similarly, `100` and `100` will match as the former can be coerced into an int.
     */
    @Deprecated(
       "Json comparison options is now specified with `CompareJsonOptions`",
@@ -35,7 +33,7 @@ enum class CompareOrder {
    /**
     * All object properties must be in same order as expected.
     *
-    * For example, { "x": 14.2, "y": 13.0 }` and `{ "y": 13.0, "x: 14.2 }` will NOT be considered equal.
+    * For example, `{ "x": 14.2, "y": 13.0 }` and `{ "y": 13.0, "x: 14.2 }` will NOT be considered equal.
     */
    @Deprecated(
       "Json comparison options is now specified with `CompareJsonOptions`",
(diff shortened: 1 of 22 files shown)


sksamuel (approved):

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -20,8 +18,8 @@ enum class CompareMode {
    /**
     * Compare by value, coercing if possible.
     *
-    * For example, "true" and true will match because the string value can be coerced into a valid boolean.
-    * Similarly, "100" and 100 will match as the former can be coerced into an int.
+    * For example, `true` and `true` will match because the string value can be coerced into a valid boolean.
```
AlexCue987, line 21: I do not understand the change from 
```
"true" and true will match
```
to
```
`true` and `true` will match
```
after the change, both tokens are identical. Same goes for the change to
```
"true" and true will match
```
What is the difference between these two tokens?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -35,7 +33,7 @@ enum class CompareOrder {
    /**
     * All object properties must be in same order as expected.
     *
-    * For example, { "x": 14.2, "y": 13.0 }` and `{ "y": 13.0, "x: 14.2 }` will NOT be considered equal.
+    * For example, `{ "x": 14.2, "y": 13.0 }` and `{ "y": 13.0, "x: 14.2 }` will NOT be considered equal.
```
AlexCue987, line 36: I know it'sa pre-existing problem, but the following is not valid JSON: `"y": 13.0, "x: 14.2` - one double quote is missing.Let's change it to `"y": 13.0, "x": 14.2`

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/wrappers.kt
```diff
@@ -14,7 +14,7 @@ import kotlinx.serialization.json.longOrNull
 
 fun JsonElement.toJsonNode(): JsonNode = when (this) {
    JsonNull -> JsonNode.NullNode
-   is JsonObject -> JsonNode.ObjectNode(entries.map { it.key to it.value.toJsonNode() }.toMap())
+   is JsonObject -> JsonNode.ObjectNode(entries.associate { it.key to it.value.toJsonNode() })
```
AlexCue987, line 17: +100

AlexCue987 (approved):
LGTM, some cosmetic suggestions

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
aSemy: Oops, that's a mistake. IntelliJ was too helpful.

```suggestion
    * For example, `"true"` and `true` will match because the string value can be coerced into a valid boolean.
```

aSemy: ah, the PR is closed, I'll make a new one.

aSemy: https://github.com/kotest/kotest/pull/4228

PR #4228, opened 2024-07-28
Update kdoc in kotest-assertions-json

aSemy:
Minor KDoc corrections. Follow up to #4223.

Thank you @AlexCue987 for noticing the mistakes!

### DIFF
--- a/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
+++ b/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
@@ -18,8 +18,8 @@ enum class CompareMode {
    /**
     * Compare by value, coercing if possible.
     *
-    * For example, `true` and `true` will match because the string value can be coerced into a valid boolean.
-    * Similarly, `100` and `100` will match as the former can be coerced into an int.
+    * For example, `"true"` and `true` will match because the string value can be coerced into a valid boolean.
+    * Similarly, `"100"` and `100` will match as the former can be coerced into an int.
     */
    @Deprecated(
       "Json comparison options is now specified with `CompareJsonOptions`",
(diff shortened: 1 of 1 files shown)


kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -98,12 +108,29 @@ enum class PropertyOrder {
    /**
     * Default. Property order in objects does not matter.
     *
-    * Example: `"""{ "a": 0, "b": 2 }""".shouldEqualJson("""{ "b": 2, "a": 1 }""", compareJsonOptions { propertyOrder = Lenient })` will pass
```
aSemy, line 101: I think the original has a mistake. `a=0` on the left, but `a=1` on right. I updated it so that `a=1`.

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -98,12 +108,29 @@ enum class PropertyOrder {
...
+    *    .shouldEqualJson(
+    *       """{ "b": 2, "a": 1 }""",
+    *       compareJsonOptions { propertyOrder = Lenient }
+    *    )
+    * ```
     */
    Lenient,
 
    /**
-    * Properties must be in same order. E.g. `{ "a": 0, "b": 2 }` is not considered equal to `{ "b": 2, "a": 1 }`
```
aSemy, line 106: Another mistake? `a=0` on the left, but `a=1` on the right.

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
```diff
@@ -155,8 +182,11 @@ enum class TypeCoercion {
     * Types may be coerced. Strings containing numbers will be considered equal to their numbers, and booleans in
     * strings will also be compared.
     *
-    * For example: `"\"11\"".shouldEqualJson("12", compareJsonOptions { typeCoercion = TypeCoercion.Enabled })` will
```
aSemy, line 158: `"11".shouldEqualJson("12")` looks like it would always fail, regardless of type coercion? I changed it to `"12".shouldEqualJson("12")`, but that seems like it doesn't demonstrate type coercion properly?

sksamuel (approved):

AlexCue987 (approved):

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/compare.kt
aSemy: @AlexCue987 Do you have an idea what the example should be here? Is what's written correct?

PR #4230, opened 2024-07-28
Removed non-kotlin style shouldCompleteBetween

sksamuel:


kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/concurrent/suspension/concurrent.jvm.kt
aSemy, line 1: Just to highlight: These functions were only recently deprecated, just before 6.0 was started. They haven't had a release since being deprecated, so users won't have been warned and had a chance to update yet.

sksamuel: That's fair but it's a major release and  this is a minor piece.

aSemy (approved):

PR #4232, opened 2024-07-28
Use virtual time to test DescribeSpecExample

aSemy:
Use the test coroutine scope to enable virtual time, to help make DescribeSpecExample less flaky.

Part of #4113


sksamuel:
Do you know why it's flakey atm?

aSemy:
> Do you know why it's flakey atm?

I noticed that it failed in #4218, which adds two JVM tests simultaneously. I suspect that the JVM tests running together cause delayed test launching, which triggers the timeouts.

The timeouts could be increased, but I thought it would be more flexible and robust to use virtual time. This has the added benefit that the executed tests can be verified.

sksamuel (approved):

PR #4234, opened 2024-07-29
Use Gradle TestKit to test Kotest Gradle Plugin

aSemy:
WIP

Using TestKit helps with caching & test project management.

(Despite [the warning](https://github.com/kotest/kotest/blob/89152c8cf8262410945e658e3272ab1aa7bac0fa/kotest-framework/kotest-framework-multiplatform-plugin-gradle/src/test/kotlin/io/kotest/framework/multiplatform/gradle/KotestMultiplatformCompilerGradlePluginSpec.kt#L25-L27), TestKit is mostly fine, so long as you avoid [`withPluginClasspath()`](https://docs.gradle.org/current/userguide/test_kit.html#sub:test-kit-automatic-classpath-injection). It would be nice to create a better version though...)


LeoColman:
Closing this WIP as it's stale.
Let us know if we should reopen it :)

PR #4239, opened 2024-07-30
Removed assertions api module

sksamuel:
This module is pointless.
The core module adds no external deps (other than coroutines which is fine) so we should just simplify the build.


kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/resource/byteArrayMatchers.kt
```diff
@@ -53,8 +53,8 @@ fun ByteArray.shouldNotMatchResource(
 }
 
 fun matchResource(
-   resourcePath: String,
-   matcherProvider: (ByteArray) -> Matcher<ByteArray>,
+  resourcePath: String,
+  matcherProvider: (ByteArray) -> Matcher<ByteArray>,
```
aSemy:
```suggestion
   resourcePath: String,
   matcherProvider: (ByteArray) -> Matcher<ByteArray>,
```

The indent should be 3 spaces, right?

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/paths/paths.kt
```diff
@@ -203,9 +208,9 @@ infix fun Path.shouldNotContainFileDeep(name: String) = this shouldNot containFi
 fun containFileDeep(name: String): Matcher<Path> = object : Matcher<Path> {
 
    private fun fileExists(dir: Path): Boolean {
-      val contents = Files.list(dir).toList()
-      val (dirs, files) = contents.partition { Files.isDirectory(it) }
-      return files.map { it.fileName.toString() }.contains(name) || dirs.any(::fileExists)
+      val contents = dir.toFile().listFiles()
+      val (dirs, files) = contents.partition { it.isDirectory }
+      return files.map { it.name.toString() }.contains(name) || dirs.any { fileExists(it.toPath()) }
```
aSemy: Rather than converting Paths to Files, I would prefer either using the Java NIO functions (so revert this change), or use Kotlin stdlib utils so using Paths is easier

```suggestion
      val contents = dir.listDirectoryEntries()
      val (dirs, files) = contents.partition { it.isDirectory() }
      return files.map { it.name }.contains(name) || dirs.any { fileExists(it) }
```

(This change also requires `import kotlin.io.path.*`)

aSemy (changes requested):

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/paths/paths.kt
```diff
@@ -12,9 +12,8 @@ import io.kotest.matchers.shouldNotBe
 import java.io.File
 import java.nio.file.Files
 import java.nio.file.Path
-import kotlin.io.path.Path
+import java.util.stream.Stream
```
aSemy:
```suggestion
```

Unused import

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/paths/paths.kt
sksamuel: I don't know why but I can't get the original to compile anymore.

LeoColman:
I think the original reason for this module was

"I want to build a custom kotest assertion library but don't want to include the assertions or the extended core"

So it contained just the simplest bits to build assertions.


Do we have something similar to this currently in place? If not, maybe this module is not that useless and we should consider the use case (and perhaps even discard it if we want)

AlexCue987 (approved):
LGTM. Should we consolidate assertions-sql too - it's very very small as well?

sksamuel:
> LGTM. Should we consolidate assertions-sql too - it's very very small as well?

Absolutely, next PR

sksamuel:
> I think the original reason for this module was
> 
> "I want to build a custom kotest assertion library but don't want to include the assertions or the extended core"
> 
> So it contained just the simplest bits to build assertions.
> 
> Do we have something similar to this currently in place? If not, maybe this module is not that useless and we should consider the use case (and perhaps even discard it if we want)

My original intent was a tiny dependency, but the core dependency is also small and only has coroutines. It doesn't bring in any third party libs and is barely any bigger

sksamuel:
@aSemy mind reapproving

aSemy:
> @aSemy mind reapproving

I'd like to take a look at the paths stuff, I just want to check that the recent Java Toolchains changes haven't broken anything.

I'll check out the branch and take a look now.

aSemy (approved):
Looks like it compiles fine with NIO, but let's merge this and I'll make another PR to update the code.

aSemy:
Hmm weird, I started seeing some errors now 

> [MISSING_DEPENDENCY_CLASS] Cannot access class 'java.nio.file.Path'. Check your module classpath for missing or conflicting dependencies

image


I think it's a bug in IJ though, the code works fine.

aSemy:
I can see [there's a test failure on Windows](https://github.com/kotest/kotest/actions/runs/10227456251/job/28298949126#step:8:880).

```
"
[...]
expected: Blonde(a=foo, b=true, c=23423, p=abc),
[...]
" should include substring "expected: Blonde(a=foo, b=true, c=23423, p=a/b/c),"
```


`p=abc` != `p=a/b/c`. Have the changes in this PR affected path separators?

EDIT: I saw this test failed again in another PR, so made an issue https://github.com/kotest/kotest/issues/4257

Issue #4177, opened 2024-07-18
IncompatibleClassChangeError with Kotlin 2.0.20-Beta2 for native (iOS) targets


JesusMcCloud:
**Which version of Kotest are you using**
Kotest 5.9.1

Hi!

We ran into an issue testing our KMP codebase against Kotlin 2.0.20-Beta2 (Beta1 works, 2.0.10-RC also works).
Kotest fails to start the iOS tests due to an IncompatibleClassChangeError. [Here's](https://github.com/a-sit-plus/kmm-vc-library/actions/runs/9986168884/job/27598202067#step:5:440) the output of out CI pipeline triggering it.

OliverO2:
Thanks! Looks like a [change in the Kotlin compiler](https://github.com/JetBrains/kotlin/commit/7f6562285b6b228e435550e1a0a5a12fbab9a6cb) that requires updating the Kotest compiler plugin.

```
e: java.lang.IncompatibleClassChangeError: Found class org.jetbrains.kotlin.ir.declarations.IrFactory, but interface was expected
	at io.kotest.framework.multiplatform.embeddablecompiler.NativeTransformer.generateLauncher(NativeTransformer.kt:46)
	at io.kotest.framework.multiplatform.embeddablecompiler.Transformer.visitModuleFragment(Transformer.kt:65)
	at org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid.visitModuleFragment(IrElementTransformerVoid.kt:113)
	at org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid.visitModuleFragment(IrElementTransformerVoid.kt:19)
	at org.jetbrains.kotlin.ir.declarations.IrModuleFragment.accept(IrModuleFragment.kt:33)
	at org.jetbrains.kotlin.ir.declarations.IrModuleFragment.transform(IrModuleFragment.kt:36)
	at io.kotest.framework.multiplatform.embeddablecompiler.SpecIrGenerationExtension.generate(SpecIrGenerationExtension.kt:23)
	at org.jetbrains.kotlin.fir.pipeline.ConvertToIrKt.applyIrGenerationExtensions(convertToIr.kt:437)
	at org.jetbrains.kotlin.fir.pipeline.Fir2IrPipeline.runActualizationPipeline(convertToIr.kt:246)
	at org.jetbrains.kotlin.fir.pipeline.Fir2IrPipeline.convertToIrAndActualize(convertToIr.kt:130)
[...]

PR #4248, opened 2024-08-03
Updated to kotlin 2.0

sksamuel:
I've set it to 2.0.20 RC because by the time we release 6.0, Kotlin 2.0.20 will certainly be out


sksamuel:
@aSemy do we define the kotlin SDK version anywhere?
Like in some other projects I do
```
   tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
      kotlinOptions.apiVersion = "1.9"
      kotlinOptions.languageVersion = "1.9"
   }
```

aSemy:
I think no, the api/language levels aren't set, so they'll default to 1.9. 

I suggest we set the api/language level to 2.0. Since Kotest 6.0 is a major release we can focus on supporting newer versions.

sksamuel:
> I think no, the api/language levels aren't set, so they'll default to 1.9.
> 
> I suggest we set the api/language level to 2.0. Since Kotest 6.0 is a major release we can focus on supporting newer versions.

ok done

sksamuel:
@aSemy this is good for review

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
```diff
@@ -86,7 +86,7 @@ internal class TestCaseExecutor(
             timeMark,
             listOfNotNull(
                InvocationTimeoutInterceptor,
-               if (platform == Platform.JVM && testCase.config.testCoroutineDispatcher) TestDispatcherInterceptor() else null,
+               if (platform == Platform.JVM && testCase.config.testCoroutineDispatcher) createTestDispatcherInterceptor() else null,
```
OliverO2, line 89: This is made obsolete by #4251, right? So the entire expect/actual `createTestDispatcherInterceptor()` could either be deleted or used with...

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
```diff
@@ -86,7 +86,7 @@ internal class TestCaseExecutor(
...
                if (useCoroutineTestScope) TestCoroutineInterceptor() else null,
```
OliverO2, line 90: ... this one. But in that case, we'd need to remove `useCoroutineTestScope`.

Of course, we could throw an exception if a `coroutineTestScope` was requested but the platform does not support it. But I wonder if it would be easier to have the entire logic here in one place instead of distributing it across an expect/actual hierarchy. It seems to be a temporary state waiting for a fix, after all.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/CollectionMatchers.kt
```diff
@@ -26,8 +26,8 @@ fun <T> existInOrder(predicates: List<(T) -> Boolean>): Matcher<Collection<T>?>
 
    MatcherResult(
       subsequenceIndex == predicates.size,
-      { "${actual.print().value} did not match the predicates ${predicates.print().value} in order. Predicate at index $subsequenceIndex did not match." },
-      { "${actual.print().value} should not match the predicates ${predicates.print().value} in order" }
+      { "${actual.print().value} did not match the predicates in order. Predicate at index $subsequenceIndex did not match." },
+      { "${actual.print().value} should not match the predicates in order" }
```
OliverO2, line 30: How is this related to Kotlin 2.0?

OliverO2 (changes requested):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
sksamuel: Yes it'll be removed in that other PR.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/CollectionMatchers.kt
sksamuel: The toString on lambas changed in 2.0

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
sksamuel: Yes, it's in the other PR.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
OliverO2: Don't the two PRs now contradict each other? Doesn't the code added here collide with the removal of #4251?

sksamuel: I did this first to make it compile. Kotlin 2 didn't like the class not being abstract. This was to get it to compile. The other one I'll rebase after. It's only because I did this first

OliverO2 (approved):

JesusMcCloud:
Not trying to be a petty necromancer and resurrect closed issues, but how does this affect #4177?

PR #4249, opened 2024-08-03
Cache Kotlin Konan directory

aSemy:
Kotlin uses a persistent directory for caching dependencies and transforms. (The directory is `~/.konan` by default, or can be set manually using `KONAN_DATA_DIR`).

On CI the directory is not persisted, so Kotlin will re-do the same download and transformations. Caching the directory will improve performance (of the main build, and also in the Kotest Gradle Plugin tests), and is now a viable option because the GitHub Actions Cache is less overloaded (thanks to Remote build cache and fixes in the Gradle GitHub Action.)

### DIFF
--- a/.github/workflows/run-gradle.yml
+++ b/.github/workflows/run-gradle.yml
@@ -26,12 +26,19 @@ jobs:
       steps:
          -  name: "Redirect caches to fast D: drive for Windows runners"
             if: runner.os == 'Windows'
+            shell: cmd
             run: |
                mkdir D:\.gradle
                mklink /D C:\Users\runneradmin\.gradle D:\.gradle
+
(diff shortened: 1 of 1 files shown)


LeoColman (approved):
Is there a way to invalidate the cache? I'm afraid of getting the build stuck due to a bad cached entry 

I think it's possible via GitHub UI

aSemy:
> Is there a way to invalidate the cache? I'm afraid of getting the build stuck due to a bad cached entry
> 
> I think it's possible via GitHub UI

Yes, it's possible to do it manually in the UI. But TBH because Kotest regularly exceeds the 10GB cache limit this means the caches are deleted very often (much more often than I would like!)

PR #4251, opened 2024-08-04
Added DefaultTestConfig as replacement for TestCaseConfig

### LINKED ISSUE #4108: Test Case Config example shows deprecated symbols
https://kotest.io/docs/framework/testcaseconfig.html documents the following deprecated symbols:
* `val defaultTestConfig`
* `fun defaultTestCaseConfig()`
* `class TestCaseConfig`

Also, for `val defaultTestConfig`, the deprecation hint is not useful:
```
These settings should be specified individually to provide finer grain control
```

It remains unclear what to use instead.

sksamuel:
Fixes https://github.com/kotest/kotest/issues/4108
We might still need to make some further tweaks as we make other changes for 6.0 but I think the general gist is right.
Removed the listener methods that have been deprecated for extension methods.


sksamuel:
@OliverO2 I removed the `testCoroutineDispatcher` option. My understanding was we deprecated that in favor of coroutine test scope ?

OliverO2:
Yes, `useCoroutineTestScope` replaces `testCoroutineDispatcher`, so removing the latter is the intended move.

sksamuel:
@OliverO2 following on from the other PR this one is the one removing the testCoroutineDispatcher.

OliverO2:
Yep. But if I am not mistaken, now we're probably having an unused function `createTestDispatcherInterceptor()`, right?

sksamuel:
yeah let me tidy it all up

sksamuel:
@OliverO2 all updated

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
```diff
@@ -86,7 +85,6 @@ internal class TestCaseExecutor(
             timeMark,
             listOfNotNull(
                InvocationTimeoutInterceptor,
-               if (platform == Platform.JVM && testCase.config.testCoroutineDispatcher) createTestDispatcherInterceptor() else null,
                if (useCoroutineTestScope) TestCoroutineInterceptor() else null,
```
OliverO2, line 90: If I understood the changes in #4248 correctly, you'd prefer Kotest to throw an exception in case a configured test dispatcher meets a lack of platform support. While probably just a temporary situation, I think this makes sense. So the above code

``` kotlin
      val useCoroutineTestScope = when (platform) {
         Platform.JVM, Platform.Native -> testCase.config.coroutineTestScope
         Platform.JS, Platform.WasmJs -> false
      }
```

could be changed to
``` kotlin
      val useCoroutineTestScope = when (platform) {
         Platform.JVM, Platform.Native -> testCase.config.coroutineTestScope
         Platform.JS, Platform.WasmJs -> if (testCase.config.coroutineTestScope) {
            error("Configuration 'coroutineTestScope' is unsupported on $platform")
         } else false
      }
```

sksamuel: done

sksamuel:
@OliverO2 can you approve

sksamuel:
@OliverO2 can you approve

LeoColman (approved):
Approving to let you continue. I'm assuming the discussion is over and all fixes were applied

PR #4252, opened 2024-08-04
Do not invoke after spec if no active tests

### LINKED ISSUE #4022: afterSpec but not beforeSpec is called when all tests are filtered out
5.8.0

**Description**
Consider the following test:
```kotlin
class ListenersTest : StringSpec({
    beforeSpec { println("beforeSpec") }
    afterSpec { println("afterSpec") }
    beforeEach { println("beforeEach") }
    afterEach { println("afterEach") }

    "!test" {}
})
```

When I run `ListenersTest`, it prints only
```
afterSpec
```

**Expected behavior**
I would expect that it either prints nothing or prints
```
beforeSpec
afterSpec
```

sksamuel:
Fixes #4022

Also the isolation per leaf was not propagating test results back to the spec interceptors, which I fixed also.


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/interceptor/instance/AfterSpecListenerSpecInterceptor.kt
```diff
@@ -0,0 +1,36 @@
...
+               .map { results }
+         } else {
+            Result.success(results)
+         }
+      }
+   }
+
+   // the spec results are considered active if at least test has a non ignored status
+   private fun hasActiveTest(results: Map<TestCase, TestResult>): Boolean {
+      return results.any { it.value !is TestResult.Ignored }
```
LeoColman, line 34: I'm wondering if we should actually do this.

If one decides to ignore all tests in a class just to see what the database is doing or if one prepares the database before writing tests (empty list), this would be a non-intuitive issue

LeoColman: I know this is a fairly far-fetched what-if, but I don't see it as thaaat unlikely and can think of similar examples

LeoColman: If we are to keep this, we should probably do the same for BeforeSpec listeners although it's likely harder due to not having discovered tests at that step yet

sksamuel: BeforeSpec already does this, so this is aligning AfterSpec with the same behavior.

LeoColman (approved):

Issue #3445, opened 2023-03-08
Property testing: DSL for more ergonomic tests?


BenWoodworth:
## Problem
I've been using the property testing library for a bit now, and it's great, but often feels clunky to write especially when more and more generators are used

Specifically because of the disconnect between (lambda) variables and where their generators are declared
```kotlin
import kotlinx.coroutines.test.runTest

// ...

@Test
fun example_property_test() = runTest {
    checkAll(
        Arb.string(),
        Exhaustive.mySpecialStrings(),
        Arb.float()
    ) { someString, specialString, someFloat ->
        // Assertions
    }
}
```

## Proposal
Slightly inspired by Jetpack Compose and its [`remember` delegate](https://github.com/android/snippets/blob/714b86bc8858ad6e3a788cf20d0968fc9b60c9d9/compose/snippets/src/main/java/com/example/compose/snippets/state/StateOverviewSnippets.kt#L73-L90), I'm envisioning something like this which takes advantage of Kotlin's expressive language features in a way that wouldn't be possible in QuickCheck with Haskell:
```kotlin
@Test
fun example_property_test() = runPropertyTest {
    val someString by generator { Arb.string() }
    val specialString by generator { Exhaustive.mySpecialStrings() }

    // ...some more setup?

    val someFloat by generator { Arb.float() }

    // Assertions
}
```

I originally pictured values being pulled in like this, which looks a lot nicer, but leads to unnecessary creation of arbs and takes away control from the DSL implementation:
```kotlin
val someString by Arb.string()
```

A `generator` delegate function would also be easy to extend in the DSL syntax, e.g. with:
```kotlin
// Desugars to:  `generator { Arb.default<String>() }`
val someString by arb<String>()

// or, equivalently:
val someString: String by arb()
```

## Implementation
There might be some considerations, like disallowing a property from being instantiated multiple times (e.g. in a loop), or maintaining parity with the values that  would be generated from seemingly equivalent `checkAll`/`forAll`/`proptest` usages. That's all possible to ensure at runtime though. I don't feel qualified to make those decisions though

If there's interest I'd love to put up a PR or a prototype. I'd just want feedback/ideas first, and to know if this is even desired :)

myuwono:
Hey @BenWoodworth thank you for the recommendation. We did explore something similar prior in https://github.com/kotest/kotest/issues/2574#issue-1028289236

Back then I did manage to get it in some working state but I couldn't get shrinking to work. If there's a good idea there would love to hear it.

[...]

PR #4258, opened 2024-08-05
Added strawman proposal for next gen property test syntax

sksamuel:
Following on from inspiration by @BenWoodworth in https://github.com/kotest/kotest/issues/3445 I have been playing around with the following syntax as a potential for Kotest 6.0 (now that we're deep in the work for that major release :))

Goals:
1. To make property testing easier to read and write by taking advantage of Kotlin's language features.
2. To support new features by avoiding legacy syntax.
3. To enhance property testing support by coupling together with kotest framework.

To point 1, after many years of development in Kotest's property test framework, we've added a lot of features that can seem more intuitive and easier to understand if we took a fresh look based on what we now know. For instance, the way we configure config is a bolt on to the original syntax, and the way arbs are passed into `checkAll` and `forAll` can seem clunky.

To point 2, I am not trying to make all this work with the existing code. The main interfaces - `Gen`, `Arb`, `Exhaustive`, `Classifier`, and `Shrinker` - are untouched but I have avoided trying to make the `checkAll` and `forAll` methods backwards compatible.

To point 3, it is not a goal to make this work outside of Kotest. The original prop test library is fully compatible with JUnit and any other framework - and will always remain so - but this alternative would be tied into Kotest. This would allow us to have nicer syntax for defining tests directly in specs, as well as coupling into callbacks.

Here is what I have working in a very ugly branch. If we like this work, I would need to clean it all up, port over the rest of the features and tests. However, now that I have property testing iterations and shrinking working, I don't think there's anything left that would pose a problem to migrate.

Things to spot:
1. Property tests are defined directly inside the spec, no need for a "test" wrapper, etc. A prop test can be placed inside any container (eg describe in DescribeSpec) or simply at the top level
2. Config values are defined inline. No need to pass in PropConfig. You can reuse a config object though if you wish, similar to #4251 
3. Arbs are specified outside the property block, but we can switch that. I have no strong preference and it does not matter to the way it works.
4. The actual testing iteration takes place inside a `property` block inside the test
5. Notice the callbacks in the final test.

```kotlin
class DelegateTest : FunSpec() {
   init {
      proptest("delegate example 1") {

         iterations = 42 // run this property test 42 times
         failOnSeed = true // will fail if a seed is set
         edgecaseFactor = 0.1 // 10% of generated values will be edgecases
         shouldPrintGeneratedValues = true // output each generated value

         val a by gen { Arb.int() }
         val b by gen { Arb.int() }

         property {
            a + b shouldBe a + b
         }
      }

      proptest("delegate example 2") {

         duration = 100.milliseconds // run this property test for as many iteration as 100ms allows
         minSuccess = 5 // require 5 successful tests to ensure we get at least some in a duration based test
         maxFailure = 1 // allow 1 failure for a flakey test
         shrinkingMode = ShrinkingMode.Off // disable shrinking on this prop test

         val first by gen { Arb.string() }
         val second by gen { Arb.string() }

         property {
            val concat = first + second

            concat shouldStartWith first
            concat shouldEndWith second
            concat shouldHaveLength (first.length + second.length)
         }
      }

      proptest("shrinking") {

         seed = 838127382173 // override seed
         shouldPrintShrinkSteps = true // print shrink steps

         val a by gen { Arb.int() }
         val b by gen { Arb.int() }

         property {
            a * b shouldBe a + b
         }
      }

      proptest("assumptions") {

         val a by gen { Arb.int() }
         val b by gen { Arb.int() }
         val c by gen { Arb.int() }

         assume { a != b }
         assume { a != c }
         assume { b != c }

         property {
            a + b + c shouldBe c + b + a
         }
      }

      proptest("callbacks") {

         val a by gen { Arb.int() }
         val b by gen { Arb.int() }

         beforeProperty {
            // setup code
         }

         afterProperty {
            // some tear down code
         }

         property {
            a + b shouldBe b + a
         }
      }
   }
}
```

Would be interested in general thoughts on if this is something worth perusing, as well as feedback on what I might have missed that might prove difficult or impossible to do (and thus nix the whole idea).


sksamuel:
One additional nice benefit, is that shrinking includes the variable name in the output:

```
Will begin shrinking for the following failed values

Arg (val a: kotlin.Int) => -1014571228
Arg (val b: kotlin.Int) => 2091145759

Attempting to shrink arg (val a: kotlin.Int)
Shrink #0: 0 fail
Shrink result (after 1 shrinks) => 0


Attempting to shrink arg (val b: kotlin.Int)
Shrink #0: 0 pass
Shrink #1: 1 fail
Shrink result (after 2 shrinks) => 1
```

Because the delegates receive a reference to `KProperty` we can pull out whatever the user has put on the left hand side easily.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/GenDelegate.kt
```diff
@@ -0,0 +1,143 @@
+package io.kotest.property.core
+
+import io.kotest.property.Gen
+import io.kotest.property.RTree
+import io.kotest.property.RandomSource
+import io.kotest.property.Sample
+import kotlin.properties.ReadOnlyProperty
+import kotlin.reflect.KProperty
+
+private object UNINITIALZED_VALUE
```
LeoColman: VERY NITPICKY BUT I DON'T LIKE WHEN CODE SCREAMS AT ME

LeoColman (commented):
A bit off-topic, but we do go

- Gen -> Generator
- Arb -> Arbitrary
- Exhaustive -> Exhaustive


I think it's a good opportunity to decide if we want to go with shortned version (arb) or with the full version (Arbitrary).

For instance, the `arbs.kts` file is insinde the `arbitrary` package

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/runtime.kt
```diff
@@ -0,0 +1,131 @@
...
+import io.kotest.core.spec.style.scopes.addTest
+import io.kotest.property.AssumptionFailedException
+import io.kotest.property.PropertyTesting
+import io.kotest.property.ShrinkingMode
+import io.kotest.property.internal.Counter
+import io.kotest.property.internal.ShrinkResult
+import io.kotest.property.internal.throwPropertyTestAssertionError
+
+@ExperimentalKotest
+fun RootScope.proptest(name: String, configure: PropTestContext.() -> Unit) {
```
LeoColman: I like `proptest()` as a construct.

It makes clear what kind of test that is in the name, it fits all specs and factories. I really like it

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/runtime.kt
```diff
@@ -0,0 +1,131 @@
...
+
+   var k = 0
+   while (constraints.evaluate(Iteration(k))) {
+      try {
+         context.container.reset()
+         context.beforeProperty()
+         context.property.invoke()
+         context.afterProperty()
+      } catch (e: AssumptionFailedException) {
+         // we don't mark failed assumptions as errors
```
LeoColman: Any chance we end up generating bad cases for the assumption forever?

It seems to me that if we pass by this, we might complete k iterations without a single test execution

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
```diff
@@ -0,0 +1,97 @@
...
+import kotlin.time.Duration.Companion.milliseconds
+
+class DelegateTest : FunSpec() {
+   init {
+      proptest("delegate syntax 1") {
+
+         iterations = 42 // run this property test 42 times
+         failOnSeed = true // will fail if a seed is set
+         edgecaseFactor = 0.1 // 10% of generated values will be edgecases
+         shouldPrintGeneratedValues = true // output each generated value
```
LeoColman: I like to have a configuration inside the test itself.

It goes a bit against what we usually do with `.config(xx)`

sksamuel:
> A bit off-topic, but we do go
> 
> * Gen -> Generator
> * Arb -> Arbitrary
> * Exhaustive -> Exhaustive
> 
> I think it's a good opportunity to decide if we want to go with shortned version (arb) or with the full version (Arbitrary).
> 
> For instance, the `arbs.kts` file is insinde the `arbitrary` package

I don't want to change more than necessarily. Renaming interfaces doesn't gain us anything, and breaks code for others.

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
sksamuel: Yes, its different. I think more in line with current kotlin idioms though ?

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/runtime.kt
sksamuel: Yes, this is just PoC code though.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/GenDelegate.kt
sksamuel: That was copied from Kotlin's own Lazy implementation :)

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
```diff
@@ -0,0 +1,97 @@
...
+
+         val a by gen { Arb.int() }
```
aSemy: I think it would be a bit cleaner if the syntax was just 

```kotlin
val a by Arb.int()
```

But maybe the `gen {}` is important?

sksamuel: It means we can avoid the Arb.int() being created each time, if that's important. It also means I don't have to make any changes to the existing Arb interface.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/runtime.kt
aSemy: I have a suggestion of something to look at in the future, when the POC is more refined, is to add a `@DslMarker`, to try and prevent a property test being called within a property test.

https://kotlinlang.org/docs/type-safe-builders.html#scope-control-dslmarker

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
```diff
@@ -0,0 +1,97 @@
...
+         }
+      }
+
+      proptest("assumptions") {
+
+         val a by gen { Arb.int() }
+         val b by gen { Arb.int() }
+         val c by gen { Arb.int() }
+
+         assume { a != b }
```
aSemy: I really like everything in general, and particularly the assumes.

It might be nice to allow adding a message to these assumes, for more informative messages.

```kotlin
assume("a must not equal b") { a != b }
```

Just something that springs to mind as I read this. It's not necessary for the POC, and maybe wouldn't even be useful (I can't remember how/if assumption failures are logged currently!)

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
```diff
@@ -0,0 +1,97 @@
...
+         assume { a != c }
+         assume { b != c }
+
+         property {
+            a + b + c shouldBe c + b + a
+         }
```
aSemy: Should testing multiple properties in the same `proptest {}` scope be allowed?

```kotlin
property { a + b shouldBe b + a }
property { a + b + c shouldBe c + b + a }
```

I think it'd be convenient.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/runtime.kt
sksamuel: Yeah I planned on that (plus stopping the gen delegate being defined inside the property block)

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
aSemy: Ah yes, avoiding changes to the Arb interface makes sense.

To avoid creating the Arb every time, I think changing GenDelegate to implement  PropertyDelegateProvider would help, and make the code a bit simpler. But I always get confused by the delegate stuff so maybe I'm wrong.

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
sksamuel: assumptions just filter out invalid values basically. It's not meant to catch bugs.
assume is the common term, but really you can think of it as controlling the inputs to a subset

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
sksamuel: just put them inside the same property block ?

AlexCue987:
what does `failOnSeed` in the following mean: `failOnSeed = true // will fail if a seed is set`

LeoColman:
> what does `failOnSeed` in the following mean: `failOnSeed = true // will fail if a seed is set`


I understood it as "I'm testing a failing test with seed 5. Now I'm writing a new test, but forgot to clear the seed, and thus it's no longer a proper property test"

AlexCue987:
> > what does `failOnSeed` in the following mean: `failOnSeed = true // will fail if a seed is set`
> 
> I understood it as "I'm testing a failing test with seed 5. Now I'm writing a new test, but forgot to clear the seed, and thus it's no longer a proper property test"

If that's the case, should we name it `failWhenSeedAlreadySet`

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
AlexCue987: this is really convenient. is there a plan to provide a function, something like 'printCurrentConfig()` so that users can discover what settings are available, and which settings are in effect in this context?

OliverO2:
I like the approach.

Some initial thoughts, intentionally not looking into the implementation, trying to suppress existing knowledge about property testing, and trying with a fresh look.

I am aware that the term "property testing" is somewhat established, but does it express what we really do? We are using generated values for some tests, but whether we actually test "a property of some object" is really up to us. We can do anything with those generated values inside our test.

So I wonder:

* What does the `property` block do? What does its name suggest? Configuring/creating/returning a property? Would `withPermutations { ... }` be more indicative?
* Would this be a chance to be more precise and avoid the clash with the Kotlin property concept entirely by naming the tests `permutationTest` instead of `proptest`?

sksamuel:
> I like the approach.
> 
> Some initial thoughts, intentionally not looking into the implementation, trying to suppress existing knowledge about property testing, and trying with a fresh look.
> 
> I am aware that the term "property testing" is somewhat established, but does it express what we really do? We are using generated values for some tests, but whether we actually test "a property of some object" is really up to us. We can do anything with those generated values inside our test.
> 
> So I wonder:
> 
> * What does the `property` block do? What does its name suggest? Configuring/creating/returning a property? Would `withPermutations { ... }` be more indicative?
> * Would this be a chance to be more precise and avoid the clash with the Kotlin property concept entirely by naming the tests `permutationTest` instead of `proptest`?

the term "property" is well known in the space, even if most property tests are really just "run lots of iterations to find bugs tests". We could rename the `property` block (which is the actual test) to `iteration` or `permutation` like you suggested.

OliverO2:
OK, in that case I'd prefer `withPermutations { ... }` to indicate that it influences the behavior of the code inside (like `withContext`) and does not return any meaningful value by itself. "Permutations" seems more precise that "iterations" in this case.

An afterthought: Would it make sense to integrate the DSL into the normal `TestScope` and avoid the extra `proptest` variant entirely?

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
sksamuel: Good idea

sksamuel:
> OK, in that case I'd prefer `withPermutations { ... }` to indicate that it influences the behavior of the code inside (like `withContext`) and does not return any meaningful value by itself. "Permutations" seems more precise that "iterations" in this case.
> 
> An afterthought: Would it make sense to integrate the DSL into the normal `TestScope` and avoid the extra `proptest` variant entirely?

You mean make proptest just an extension on test scope? Would still need some kind of wrapper though to hold the config stuff.

sksamuel:
@oliverblaha any feedback?

oliverblaha:
I'd love to give feedback, but did you mean to actually tag me?

sksamuel:
> I'd love to give feedback, but did you mean to actually tag me?

Yes. You've opened some PRs on property testing so I figured you use it enough to have opinions.

OliverO2:
> You mean make proptest just an extension on test scope?

No, I was actually thinking of abandoning `proptest` and making `TestScope` absorb `proptest`'s scope.

> Would still need some kind of wrapper though to hold the config stuff.

Well, that could also be `TestScope`, maybe via an extra configuration object if property configuration is larger than we'd want in every `TestScope` object.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/runtime.kt
oliverblaha: How about also adding an extension method to allow sth like this?

```kotlin
"my test".proptest(configure = /* ... */) {
  // ...
}
```

IMHO especially WordSpec tests would benefit from that.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt
```diff
@@ -0,0 +1,59 @@
...
+   var constraints: Constraints? = null // specify manual constraits
+   var minSuccess = -1
+   var maxFailure = -1
+   var edgecaseFactor = PropertyTesting.defaultEdgecasesGenerationProbability
+   var shouldPrintShrinkSteps = false // sysprop("kotest.proptest.output.shrink-steps", true)
+   var shrinkingMode: ShrinkingMode = PropertyTesting.defaultShrinkingMode
+   var maxDiscardPercentage: Int = 20
+   var shouldPrintGeneratedValues: Boolean = PropertyTesting.shouldPrintGeneratedValues
+
+   // must call seed before rs otherwise has no effect
```
oliverblaha: That feels counterintuitive; couldn't `rs` be a container that gets updated as soon as the seed is changed?

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt
```diff
@@ -0,0 +1,59 @@
...
+   var seed: Long? = null
+   var failOnSeed: Boolean = sysprop("kotest.proptest.seed.fail-if-set", false)
```
oliverblaha: While I really like this being easily accessible, I'd encourage a more encapsulated approach (e.g., a config field), for two reasons:

- the more fields you have, the more likely are name clashes with local variables
- it makes centrally configuring certain settings rather difficult

I would expect to be able setting a global PropTestConfig, and inside the test having the config values set to that global (or maybe spec-wide) default, while being able to change them (either on spec or individual test level).

Actually I've encountered cases in the where I could have used this, especially with the number of iterations. (There's no global override, is there?)

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt
```diff
@@ -0,0 +1,59 @@
...
+import io.kotest.property.Gen
+import io.kotest.property.PropertyTesting
+import io.kotest.property.RandomSource
+import io.kotest.property.ShrinkingMode
+import kotlin.random.Random
+import kotlin.time.Duration
+
+class PropTestContext {
+
+   var iterations: Int? = null // converted to iteration based constraint
```
oliverblaha: Thought: would it be smart to somehow allow setting a factor of the default value?
Simple scenario:
- Dev machine default = 1000; a very slow test is limited to 100.
- CI/CD pipeline shall do more thorough testing, so default = 10000; the slow test should run 1000 times.

You can solve that by variables, but that requires manual overrides for every test.
This would likely be a mechanism that evolves over time - which would be another reason to somehow encapsulate config settings (see my other comment).

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt
```diff
@@ -0,0 +1,59 @@
...
+   var seed: Long? = null
+   var failOnSeed: Boolean = sysprop("kotest.proptest.seed.fail-if-set", false)
+
+   internal val rs = lazy { RandomSource.seeded(seed ?: Random.nextLong()) }
+
+   internal val container = GenDelegateContainer()
+
+   internal var property: suspend () -> Unit = {}
+   internal var beforeProperty: suspend () -> Unit = {}
+   internal var afterProperty: suspend () -> Unit = {}
```
oliverblaha: I regularly encountered situations that required me to clear mocks, also mocks used by the whole spec, after every iteration; it might be nice to have this on a spec-level, for all test & property-tests (like `beforeTestIteration`, which would be called before a "normal" test and before any property test iteration).

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
oliverblaha: Would directly using `Arb.int()` as delegate really create it each time? I'd expect it to be only created once, just the getValue being called multiple times - or maybe I'm misunderstanding something?

Also, regarding the interface: you don't need to change the interface, but you can implement `getValue(...)` as an extension method within the test scope; see comment at GenDelegate.

In addition, it would be nice to have the delegate functionality on a scope level. I encountered a few cases with a lot of tests basically requiring the same kind of generated data over and over again. I helped myself by building a generator for a data class, but that felt rather ugly. Having a whole set of top level delegates that get lazily initialized as soon as being accessed would be perfect for those cases.

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
oliverblaha: It might still be convenient to have more than one block, e.g. for having two different configs.
Probably not a good practice. However, it raises the question of when the actual test shall be executed; when calling `property {}` or when reaching the end of the test definition?

Unless there's a compelling reason to do the latter, I suggest having `property {}` actually start the testing, just as `checkAll` did it.

Note: I've seen a couple of cases where people actually got confused by the whole deferred execution concept and actually put test code outside the test. It kind of worked, but had some side-effects, IIRC. Anything non-sequential has the potential to break people's minds 😁

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+   fun beforeProperty(fn: suspend () -> Unit) {
+      beforeProperty = fn
+   }
+
+   fun afterProperty(fn: suspend () -> Unit) {
+      afterProperty = fn
+   }
+
+   fun <T> gen(initializer: () -> Gen<T>): GenDelegate<T> {
```
oliverblaha: Suggestion to allow direct delegation of generators:

```kotlin
override operator <A> fun Gen<A>.getValue(thisRef: Any?, property: KProperty<*>): A {
  // ...
}
```

Didn't study all the bits and pieces of GenDelegate though, so maybe there's a good reason to keep it around.

oliverblaha (commented):
First impression; will need some more iterations to get a better picture.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/constraints.kt
```diff
@@ -0,0 +1,36 @@
...
+      fun iterations(k: Int) = Constraints { it.iteration < k }
+
+      /**
+       * Returns a [Constraints] that executes the property test for a certain duration.
+       */
+      fun duration(duration: Duration) = object : Constraints {
+         val mark = TimeSource.Monotonic.markNow() + duration // Property tests cannot use virtual time
+         override fun evaluate(context: Iteration): Boolean {
+            return mark.hasNotPassedNow()
+         }
```
AlexCue987: I can't help but notice that we are beginning to duplcate extension functions against collections and sequences, such as `take`, `filter` etc. 

Maybe we should reimagine the whole concept, so that arbs are sequences, and then we can utilize all the existing extension functions such as `filter` and `take`?

sksamuel: I'm not sure we could make it work via sequences. There's more going on, ie with shrinks, that a simple sequence could provide.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt
sksamuel: Maybe a combined beforeIteration could do that. The problem is I didn't really want to have property testing as part of kotlin framework directly, but maybe we should just do that.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt
sksamuel: Scalatest did something like this for non deterministic testing and it was pretty neat. It would scale up timeouts on slower infra. 

I'll add something for this, its a good idea.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt
sksamuel: We do something similar with defaultTestConfig but also having all the top level fields.
I guess one could wrap all this in a config {} block so its contained. From what I can see, that seems to be how other kotlin frameworks do config, rather than creating a data class and passing it in.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt
sksamuel: Yes, I can do that.

sksamuel:
Appreciate all the comprehensive feedback so far. Here is round 2.

- Adopt the name "permutations". I think @OliverO2 had some great feedback that property testing isn't really what people use this for these days. It's about just testing lots of combinations. So the names reflect that.
- Dropped `proptest` construct, and just make it work directly inside any test scope. So now you can add property/permutation testing directly to any style.
- Wrapped config inside a `permutationConfig` block . We can extend this to allow sharing premade configs. (I might adjust the defaultTestConfig we just added to specs to follow a similar pattern if we like this).
- The `property` test block now becomes `withPermutation`.
- The gen delegates moved into the `withPermutation` block directly.

```
      test("property syntax 1") {

         permutationConfig {
            iterations = 42 // run this property test 42 times
            failOnSeed = true // will fail if a seed is set
            edgecaseFactor = 0.1 // 10% of generated values will be edgecases
            shouldPrintGeneratedValues = true // output each generated value
         }

         withPermutation {
            val a by gen { Arb.int() }
            val b by gen { Arb.int() }

            a + b shouldBe a + b
         }
      }

      test("delegate syntax 1") {

         permutationConfig {
            iterations = 42 // run this property test 42 times
            failOnSeed = true // will fail if a seed is set
            edgecaseFactor = 0.1 // 10% of generated values will be edgecases
            shouldPrintGeneratedValues = true // output each generated value
         }

         withPermutation {
            val a by gen { Arb.int() }
            val b by gen { Arb.int() }

            a + b shouldBe a + b
         }
      }

      test("delegate syntax 2") {

         permutationConfig {
            duration = 100.milliseconds // run this property test for as many iteration as 100ms allows
            minSuccess = 5 // require 5 successful tests to ensure we get at least some in a duration based test
            maxFailure = 1 // allow 1 failure for a flakey test
            shrinkingMode = ShrinkingMode.Off // disable shrinking on this prop test
         }

         withPermutation {

            val first by gen { Arb.string() }
            val second by gen { Arb.string() }

            val concat = first + second

            concat shouldStartWith first
            concat shouldEndWith second
            concat shouldHaveLength (first.length + second.length)
         }
      }

      test("shrinking") {

         permutationConfig {
            seed = 838127382173 // override seed
            shouldPrintShrinkSteps = true // print shrink steps
         }

         withPermutation {
            val a by gen { Arb.int() }
            val b by gen { Arb.int() }

            a * b shouldBe a + b
         }
      }

      test("assumptions") {

         withPermutation {

            val a by gen { Arb.int() }
            val b by gen { Arb.int() }
            val c by gen { Arb.int() }

            assume { a != b }
            assume { a != c }
            assume { b != c }

            a + b + c shouldBe c + b + a
         }
      }

      test("callbacks") {

         beforePermutation {
            // setup code
         }

         afterPermutation {
            // some tear down code
         }

         withPermutation {
            val a by gen { Arb.int() }
            val b by gen { Arb.int() }
            val c by gen { Arb.int() }

            a + b + c shouldBe c + b + a
         }
      }
```

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt
sksamuel: I think if we want to allow the delegates to be declared inside the test block (which is now the only place it can be declared, see proposal 2), then we need this extra "hop" to avoid the delegate being re-created each time, and the stream of values reset.

sksamuel: In proposal 2, withPermutuation{} starts the test immediately.

oliverblaha:
True. However, especially for easier reusability it would be nice to be
able declaring them outside; but I admit that it's cleaner to restrict it
to inside the block.

Another thought that might be related:

I often had the desire to create two or more variables that had
dependencies on each other (e.g. - simplified example - two lists that
would have too have exactly the same number of elements). This would often
lead to nested, borderline unreadable constructions.

I'm wondering if that would work now... and if so, could we get problems
with circular dependencies?

sksamuel:
> True. However, especially for easier reusability it would be nice to be able declaring them outside; but I admit that it's cleaner to restrict it to inside the block. Another thought that might be related: I often had the desire to create two or more variables that had dependencies on each other (e.g. - simplified example - two lists that would have too have exactly the same number of elements). This would often lead to nested, borderline unreadable constructions. I'm wondering if that would work now... and if so, could we get problems with circular dependencies?

I think it would be ok, because the streams are infinite.

```
withPermutation {
  val size = gen { Arb.int() }
  val list = gen { Arb.list(size, Arb.int()) }
  list.first() shouldBe list[0]
}

File: kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/constraints.kt (hunk 13 above)
Comment (oliverblaha) [reply kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/constraints.kt]: `Sequence` is just an interface and wrapping an iterator; so we could probably support all the functionality by simply implementing that interface. Even `Iterator` is just an interface, so technically, an Arb could also be an iterator instead. Or both - but that feels wrong.

Comment (oliverblaha) [reply kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/constraints.kt]: Upon further thought, I'm wondering how much benefit it would add - you can only really benefit from that if you explicitly use generators as sequences then. However, the beauty (especially of the new DSL) is to have _one_ element at a time, and not having to bother with generation process.

However, supporting sequence interfaces (or iterator) might help with special cases or with the beauty of Kotest's internal codebase.

File: kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt (hunk 12 above)
Comment (oliverblaha) [reply kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt]: > then we need this extra "hop" to avoid the delegate being re-created each time

In its current form that's true. I'm wondering though if it has to be that way - after all it's just a function call, so there might be a way to keep the actual generator object around. Might be tricky though.

But then, how does `gen { }` even help? If cross-dependencies are possible and generator B depends on A, you certainly need to recreate B every time. So unless there is some compiler magic detecting that, it's currently either always created for every iteration, or cross-dependencies are simply not working. Am I missing anything?

However, maybe the best approach is the most straight-forward one:

- Allow cross-dependencies, get rid of `gen { }`
- Any generator that does not have dependencies on generated data, you can simply define outside the block (and then assign via by inside the block)

Downside: two variables for one thing

File: kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt (hunk 10 above)
Comment (oliverblaha) [reply kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt]: If `iterations` isn't null by default, but pre-filled with the current default, it would be quite easy without the need of any special methods:

```kotlin
permutationConfig {
    iterations = 1000 // fixed
    iterations = iterations * 2 // factor
    iterations = maxOf(1000, iterations) // upper limit
    // ...
}
```

It might be worth having some extra methods though:

```kotlin
val PermutationContext.global : PermutationContext = TODO()
val PermutationContext.scope : PermutationContext = TODO()
```

This would already allow things like:

```kotlin
permutationConfig {
    iterations = 1000 // fixed
    iterations = scope.iterations * 2 // factor
    iterations = maxOf(1000, scope.iterations) // upper limit
    iterations = global.iterations // ignore scope
    // ...
}
```

Also helps with things beyond iterations; I figure it's rather interesting for any kind of configuration to have access to its defaults across layers.

OliverO2:
Looks really cool now, I'd expect the new syntax and integration to significantly ease onboarding!

Just one minor remark: Since we're potentially running _lots_ of permutations, I'd favor the plural forms (`withPermutations`, `beforePermutations`, `afterPermutation`) to express what happens. Whether that should be extended to `permutationConfig`, I'm undecided.

oliverblaha:
I'd raise that inside the block there's only one permutation present during
execution; so you're running that block *with* (one) *permutation*.
Technically correct, but that lacks the information that more than one is
generated.

For anything like checkPermutations, the plural would be correct though.

Before/after should remain singular, as those happen before/after every
single permutation.

OliverO2:
Valid arguments. Each naming variant captures part of what's going on. I wasn't even aware that `before`/`after` apply to each permutation, although that makes the most sense. Still, naming helps ;-).

Trying to express the whole picture:
```kotlin
    test("callbacks") {
        permutations {
            val a by gen { Arb.int() }
            val b by gen { Arb.int() }
            val c by gen { Arb.int() }

            beforeEach {
                // setup code
            }

            afterEach {
                // some tear down code
            }

            forEach {
                a + b + c shouldBe c + b + a
            }
        }
    }
```

Here, `permutations` is reduced to a scope, which doesn't do anything on its own. It is capable of collecting a configuration defined by generators, along with `beforeEach` and `afterEach`. And it provides an execution mechanism using its configuration via `forEach`.

kotest-property/kotest-property-core/src/commonMain/kotlin/io/kotest/property/core/proptest.kt
Kantis: Late to the party here.. Do you have some examples of other Kotlin libs/frameworks using this approach? I know we started doing it a bit for some assertions, but I'm starting to have second doubts about whether that was actually a good call.

There's a few clear upsides to having it as function params instead of mutable context.

```
      proptest("shrinking").config(
         seed = 838127382173, // override seed
         shouldPrintShrinkSteps = true, // print shrink steps
      ) {
         val a by gen { Arb.int() }
         val b by gen { Arb.int() }

         property {
            a * b shouldBe a + b
         }
      }
```

A) Makes it clear that it's configuration of the property test. 
B) Discoverable (the available config = the available parameters to the config function)
C) Eliminates possibility of name resolution ambiguity
D) Consistency - All config will always be at the start of the proptest
E) Eliminates possibility of defining and later re-defining configuration

sksamuel:
> Valid arguments. Each naming variant captures part of what's going on. I wasn't even aware that `before`/`after` apply to each permutation, although that makes the most sense. Still, naming helps ;-).
> 
> Trying to express the whole picture:
> 
> ```kotlin
>     test("callbacks") {
>         permutations {
>             val a by gen { Arb.int() }
>             val b by gen { Arb.int() }
>             val c by gen { Arb.int() }
> 
>             beforeEach {
>                 // setup code
>             }
> 
>             afterEach {
>                 // some tear down code
>             }
> 
>             forEach {
>                 a + b + c shouldBe c + b + a
>             }
>         }
>     }
> ```
> 
> Here, `permutations` is reduced to a scope, which doesn't do anything on its own. It is capable of collecting a configuration defined by generators, along with `beforeEach` and `afterEach`. And it provides an execution mechanism using its configuration via `forEach`.

This is a nice improvement. I'm very happy with that syntax.

oliverblaha:
How about using this to embrace the sequence as discussed in one of the threads by having the permutation context implement the `Sequence` interface?

Or being a bit more explicit by having a property to get it from? (The standalone "forEach" feels slightly off to me.)

sksamuel:
> How about using this to embrace the sequence as discussed in one of the threads by having the permutation context implement the `Sequence` interface?
> 
> Or being a bit more explicit by having a property to get it from? (The standalone "forEach" feels slightly off to me.)

I'm not really keen on changing Arbs if its a huge breaking change. Tons of people use prop test and I don't want them to have to do anything to upgrade.
Adding new syntax is a new module is different. It's opt in.

oliverblaha:
Not on the Arbs, but by the context of the permutations block.

kotest-property/kotest-property-core/src/jvmTest/kotlin/io/kotest/property/core/DelegateTest.kt
sksamuel: I've been trying to make this work:

            val a by Arb.int()
            val b by Exhaustive.ints(0..10)

But I'm hitting a dead end. In order to reset the values between runs we need access to the delegate itself (not just the values produced). We can do that by adding a provideDelegate extension method to Gen's. But that method doesn't provide access to the implicit receiver, so we have no place to store the delegates. I tried with a global container, which works unless you run tests in parallel because it doesn't know which delegates are associated with which test.

sksamuel:
I think this is in a good spot to get into the snapshot builds.
There's still some work to do - namely I have to finish wiring in the shrinks, and we may end up tweaking nomenclature.
But I'd rather merge this in so it's not a constantly shifting target.

sksamuel:
The test failure on macos is unrelated, seems to be a general issue with our build.

Kantis (approved):
Didn't have time to look through it, but I think it's fine to iterate further on main

PR #4264, opened 2024-08-07
Removed finalizeSpec from spec

### LINKED ISSUE #3794: [finalizeSpec, afterProject] don't work as document say for IsolationMode.InstancePerTest and IsolationMode.InstancePerLeaf
5.8.0

---

```kotlin
class MyTest : WordSpec({

    "a" should {
        "b" {}
        "c" {}
    }

    finalizeSpec {
        println("finalizeSpec")
    }

    afterProject {
        println("afterProject")
    }
})
```
The above code print
```
finalizeSpec
finalizeSpec
afterProject
afterProject
```
when `isolationMode = IsolationMode.InstancePerLeaf`, print
```
finalizeSpec
finalizeSpec
finalizeSpec
afterProject
afterProject
afterProject
```
when `isolationMode = IsolationMode.InstancePerTest`.

As the document [here](https://kotest.io/docs/framework/lifecycle-hooks.html) say `finalizeSpec`
>Called once per Spec, after all tests have completed for that spec.
>
>Regardless of how many times the spec is instantiated, for example, if InstancePerTest or InstancePerLeaf isolation modes are used, this callback will only be invoked once.
>
>The results parameter contains every TestCase, along with the result of that test, including tests that were ignored (which will have a TestResult that has TestStatus.Ignored).
>
>When running in the default SingleInstance isolation mode, then this callback and afterSpec are functionally the same since all tests will run in the same spec instance.

so I expect that the above code should only print "finalizeSpec" and "afterProject" once.

sksamuel:
Closes #3794 

Open to discussion but having a finalize spec callback inside the spec doesn't really make sense.
You can add an afterSpec callback, or register a finalize spec via @ApplyExtension or through config.


kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/spec/types/FreeSpecTestTypeTest.kt
```diff
@@ -7,8 +7,6 @@ import io.kotest.matchers.shouldBe
 class FreeSpecTestTypeTest : FreeSpec() {
    init {
 
-      finalizeSpec { it.b.size shouldBe 4 }
```
LeoColman, line 10: should we add the afterSpec counterpart?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/spec/isolation/perleaf/FreeSpecInstancePerLeafTest.kt
```diff
@@ -19,7 +19,7 @@ class FreeSpecInstancePerLeafTest : FreeSpec() {
 
    init {
 
-      finalizeSpec {
+      afterProject {
```
LeoColman, line 22: Feels a bit weird to me to translate a `afterSpec` to be a `afterProject`. Are these cases where `afterSpec` isn't enough?

sksamuel: Yes it's checking after all the specs have run what the total count is.

LeoColman (approved):

Issue #4257, opened 2024-08-04
ShouldContainExactlyTest is flaky on Windows


aSemy:
I've noticed that the tests in ShouldContainExactlyTest sometimes fail on Windows machines.

The assertion failure is quite large and hard to see, but I think it's something to do with path separators `/` vs `\`. Which is weird, because that should be constant on a single machine. 

- https://scans.gradle.com/s/weonl3kvvhtqi ([job](https://github.com/kotest/kotest/actions/runs/10237567487/job/28321070983?pr=4256#step:8:873))

- https://scans.gradle.com/s/jp6bidtsxwxf4

```
> Task :kotest-assertions:kotest-assertions-core:jvmTest

com.sksamuel.kotest.matchers.collections.ShouldContainExactlyTest[jvm]

  containExactly should[jvm]

    Test include extras when too many[jvm] FAILED

    io.kotest.assertions.MultiAssertionError: The following 2 assertions failed:
    1) "Collection should contain exactly: [Blonde(a=foo, b=true, c=23423, p=abc), Blonde(a=woo, b=true, c=97821, p=abc)] but was: [Blonde(a=foo, b=true, c=23423, p=abc)]
    Some elements were missing: [Blonde(a=woo, b=true, c=97821, p=abc)]
    
    Possible matches:
    
     expected: Blonde(a=foo, b=true, c=23423, p=abc),
      but was: Blonde(a=woo, b=true, c=97821, p=abc),
      The following fields did not match:
        "a" expected: <"foo">, but was: <"woo">
            "c" expected: <23423>, but was: <97821>
    expected:<[Blonde(a=foo, b=true, c=23423, p=abc), Blonde(a=woo, b=true, c=97821, p=abc)]> but was:<[Blonde(a=foo, b=true, c=23423, p=abc)]>" should include substring "expected: Blonde(a=foo, b=true, c=23423, p=a/b/c),"
       at com.sksamuel.kotest.matchers.collections.ShouldContainExactlyTest$1$8.invokeSuspend(ShouldContainExactlyTest.kt:205)
    2) "Collection should contain exactly: [Blonde(a=foo, b=true, c=23423, p=abc), Blonde(a=woo, b=true, c=97821, p=abc)] but was: [Blonde(a=foo, b=true, c=23423, p=abc)]
    Some elements were missing: [Blonde(a=woo, b=true, c=97821, p=abc)]
    
    Possible matches:
    
[...]

PR #4275, opened 2024-08-09
Register OS-specific variables as test task inputs

aSemy:
Prevent flaky tests (e.g. #4257) due to remote Build Cache loading successful results from other OSes.

Registering these properties as inputs will make sure that Gradle will re-run the tests on each OS, instead of loading a successful result from another OS.


AlexCue987 (approved):
very nice

PR #4283, opened 2024-08-11
Added configurable retry at the test level

### LINKED ISSUE #3967: Provide a more ergonomic way of automatically retrying failing tests
## Summary

Having the `retry(RetryConfig)` function is nice, but ideally, I'd like to set the same retry configuration for every single test case in the spec or project. All of my tests are prone to the same infrastructure issue, and should all be retried. Using the `retry` function in every single one of them adds a lot of boilerplate. Is there a recommended way to set retries globally? I played around with a `TestCaseExtension` to re-run tests multiple times if the result is failed, but the reporting is a bit odd because intermediate executions are reported as pending. 

Thank you for a great framework, and looking forward to your reply!

sksamuel:
Closes #3967

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/RetryTest.kt
+++ b/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/RetryTest.kt
@@ -1,0 +1,91 @@
+package com.sksamuel.kotest.engine.test
+
+import io.kotest.core.spec.style.DescribeSpec
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.core.test.config.DefaultTestConfig
+import io.kotest.matchers.comparables.shouldBeGreaterThan
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.TimeSource
+
+class RetryTest : FunSpec() {
+   init {
+
+      retries = 3
+      retryDelay = 25.milliseconds
+
+      val timesource = TimeSource.Monotonic
+      var count = 0
+      var mark = timesource.markNow()
+
+      beforeTest {
+         mark = timesource.markNow()
+      }
+
+      afterTest {
+         count = 0
+      }
+
+      test("basic retry").config(retries = 2) {
+         if (count < 2) {
+            count++
+            error("boom")
+         }
+      }
+
+      test("with delay").config(retries = 2, retryDelay = 20.milliseconds) {
+         if (count < 2) {
+            count++
+            error("boom")
+         } else {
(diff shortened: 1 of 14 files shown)


kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/RetryTest.kt
```diff
@@ -0,0 +1,91 @@
...
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.TimeSource
+
+class RetryTest : FunSpec() {
+   init {
+
+      retries = 3
+      retryDelay = 25.milliseconds
+
+      val timesource = TimeSource.Monotonic
```
OliverO2: Currently, `TestCaseExecutor` uses the hardwired `TimeSource.Monotonic`, exposing us to flaky tests on CI. Could we change this to `testTimeSource()`, which would provide virtual time conditionally?

Then we could make this test stable by modifying it like so:

```suggestion
      coroutineTestScope = true
      val timesource = testTimeSource()
```

sksamuel:
@OliverO2 I've merged your suggestion but the code no longer compiles and I don't really understand how the virtual time stuff works that you added. Can you give a brief explanation and I'll add comments to the code.

OliverO2:
Sure:

Generally it works like this: `testTimeSource()` provides `TimeSource.Monotonic` under normal circumstances, so we're using real time. But, if we're enabling the test dispatcher for a coroutine (hierarchy), then `testTimeSource()` will provide the test dispatcher's virtual time source inside that coroutine (hierarchy). This way, all time-dependent functions can be tested with virtual time, making tests stable on any runner.

The compilation failure was initially just a missing import. But if I add that, it still doesn't compile, since we need to be in a coroutine and the `init { ... }` block doesn't qualify. So we need to go `lateinit` and initialize exclusively inside `beforeTest`.

In this case, as I've been trying to point out, my suggestion alone is not enough: We also need to change `TestCaseExecutor` to use `testTimeSource()` instead of `TimeSource.Monotonic`. Even then, I wasn't completely sure whether this would work in this case, and it turns out it does not: `TestCaseExecutor` does not pick up our test dispatcher set via `coroutineTestScope = true` since it has already been set up outside of our spec. We'd probably need a separate `TestEngineLauncher.async` invocation for this to happen, right?

OliverO2:
We didn't need to change `TestCaseExecutor`. Using two-tier testing with `TestEngineLauncher.async` turned out to be sufficient. Added a commit to that effect. You can revert if you don't like it.

sksamuel:
@OliverO2 if you can time could you re-review

Issue #3759, opened 2023-11-11
Kotest >= 5.7.0: Jobs launched in beforeSpec can delay or hang the first test


wfhartford:
With the following test class, kotest 5.7.x will execute the test, but hang before calling `afterSpec` while the test would complete as expected with kotest 5.6.2:

```
class KotestHangSpec : FunSpec(
  {
    lateinit var job: Job
    beforeSpec {
      println("Starting background job")
      val parentJob = checkNotNull(Job(currentCoroutineContext()[Job])) { "No parent job" }
      job = GlobalScope.launch(parentJob) { delay(Duration.INFINITE) }
    }

    afterSpec {
      println("Cancelling background job")
      job.cancel()
    }

    afterTest { (case, result) ->
      println("Test '${case.name.testName}' $result")
    }

    test("always pass") {
      println("Executing test")
    }
  }
)
```

This happens when a coroutine is launched in the `GlobalScope` by the `beforeSpec` function, passing the calling coroutine's job to the `launch` function, or passing the entire `currentCoroutineContext()` to the `launch` function.

Further investigation and understanding of coroutines APIs leads me to believe that providing the parent job is not necessary, and possibly just plain wrong, but hanging indefinitely is undesirable, even with unexpected inputs.

wfhartford:
I've created a repo with a full project to reproduce the issue here: https://github.com/wfhartford/kotest-hang-3759

OliverO2:
Thanks for reporting and the reproducer! What [happens here](https://github.com/wfhartford/kotest-hang-3759/blob/651ffdcd34bc47b23a9cea84b07be404881998c9/src/test/kotlin/ca/cutterslade/kotesthang/KotestHangSpec.kt#L16-L17) is the following:

```kotlin
      val parentJob = checkNotNull(Job(currentCoroutineContext()[Job])) { "No parent job" }
      job = GlobalScope.launch(parentJob) { delay(Duration.INFINITE) }
```

This code _seems_ to launch a coroutine in `GlobalScope`, but it does not, because it immediately attaches a parent Job from the current coroutine context. The code is equivalent to
```kotlin
job = CoroutineScope(coroutineContext).launch { delay(Duration.INFINITE) }
```

[...]

Issue #3705, opened 2023-09-20
Discussion: should nested containers inherit testCoroutineScheduler?


Mexator:
Imagine we have a Repository that after some time changes its state.

```kotlin
class Repository(val scope: CoroutineScope) {
    var x = 0

    fun init() {
        scope.launch {
            delay(500)
            x += 1
        }
    }
}
```

And we want to verify this behaviour with a BehaviorSpec:
```kotlin
class RepositorySpec : BehaviorSpec({
    // We don't want test to run for 500ms, so we enable coroutineTestScope
    coroutineTestScope = true
    Given("repository is created") {
        val repository = Repository(scope = this)
        When("repository is initialized") {
            repository.init()
            Then("after 500ms x should be 1") {
                testCoroutineScheduler.advanceTimeBy(500)
                repository.x shouldBe 1
            }
        }
    }
})
```

The test above will fail. The reason is that the `Given` container will have a different `TestCoroutineScheduler` in its `CoroutineContext`. It is easy to see for yourself with this test (which is also failing):

```kotlin
class MySpec : BehaviorSpec({
    coroutineTestScope = true
    Given("invoking testCoroutineScheduler in Given") {
        val givenDispatcher = testCoroutineScheduler
        When("invoking testCoroutineScheduler in Given") {
            val whenDispatcher = testCoroutineScheduler
            Then("all three schedulers should be the same") {
                testCoroutineScheduler shouldBe givenDispatcher
                testCoroutineScheduler shouldBe whenDispatcher
            }
        }
    }
})
```

I wonder if this behaviour is intentional? As for me I expected containers to inherit `CoroutineContext` from its parents. That way they would have same `TestCoroutineScheduler`s.

If it is intentional I suggest to specify this in the documentation

OliverO2:
Related: #3577, #3759 

Suggested resolution in https://github.com/kotest/kotest/issues/3759#issuecomment-1807258525:

> And then, from a user's point of view, we should have a separate coroutine per spec, being the parent of everything at the next lower level (context or test). So the coroutine hierarchy should mirror the test hierarchy. That would be most in line with user expectations, I'd say.

Mexator:
While we are waiting for a solution, I have written a simple extension.
It allows to get the same `testCoroutineScheduler` every time it is accessed during a test:

```kotlin
class BetterTestSchedulerExtension : SpecExtension {

    override suspend fun intercept(spec: Spec, execute: suspend (Spec) -> Unit) {
        runTest {
[...]

Issue #3577, opened 2023-07-21
`Dispatchers.setMain()` in `beforeTest` doesn't set dispatcher


lukaszkalnik:
Kotest version 5.6.2
coroutines-test version 1.7.2
Mockk version 1.13.5

I'm using `BehaviorSpec`. When executing suspending functions in viewmodel under test using `testCoroutineScheduler.advanceUntilIdle()` inside a `Given` block, the `shouldBe` assertion inside the same `Given` block fails. However, when moving the `shouldBe` assertion into a `When` block, the assertion passes.

Also when there is a `shouldBe` assertion inside the `Given` block which doesn't rely on any suspending functions execution (i.e. there is no `advanceUntilIdle()` call needed before the assertion), the assertion passes.

```
import androidx.lifecycle.ViewModel
import androidx.lifecycle.viewModelScope
import io.kotest.core.spec.IsolationMode.InstancePerTest
import io.kotest.core.spec.style.BehaviorSpec
import io.kotest.core.test.testCoroutineScheduler
import io.kotest.matchers.shouldBe
import io.mockk.coEvery
import io.mockk.mockk
import kotlinx.coroutines.CoroutineDispatcher
import kotlinx.coroutines.Dispatchers
import kotlinx.coroutines.ExperimentalCoroutinesApi
import kotlinx.coroutines.delay
import kotlinx.coroutines.flow.MutableStateFlow
import kotlinx.coroutines.launch
import kotlinx.coroutines.test.setMain
import kotlin.coroutines.coroutineContext

data class ProductResponse(val id: String)

class ProductRepository {

    suspend fun getProducts(ean: String): ProductResponse {
        delay(500)
        return ProductResponse("")
    }
}

data class UiState(
    val product: ProductResponse?,
)

class MyViewModel(
    private val productRepository: ProductRepository = ProductRepository(),
) : ViewModel() {

    val uiState = MutableStateFlow(UiState(null))

    fun onProductEanScan(ean: String) {
        viewModelScope.launch {
            val product = productRepository.getProducts(ean)
            uiState.value = UiState(product)
        }
    }
}

@OptIn(ExperimentalStdlibApi::class, ExperimentalCoroutinesApi::class)
class MyViewModelTest : BehaviorSpec({
    coroutineTestScope = true
    isolationMode = InstancePerTest

    val productResponse = ProductResponse("testId")

    val productRepository = mockk<ProductRepository> {
        coEvery { getProducts(any()) } returns productResponse
    }

[...]

PR #4284, opened 2024-08-11
Runs before/after spec in the spec level coroutine

sksamuel:
Ensures that we run the spec level callbacks in the spec level coroutine.
And that each test has its own coroutine which inherits from the parent.

I'm not the worlds best coroutines expert, so appreciate any eyes on this @OliverO2 .

Possible fix for #3577 #3759 #3705


OliverO2:
On first glance there seem to be some antipatterns here (e.g. implicit use of `GlobalScope`, `async {}.await()`). Give me a day or two for a thorough review.

kotest-framework/kotest-framework-engine/build.gradle.kts
```diff
@@ -22,6 +22,7 @@ kotlin {
             implementation(libs.kotlinx.coroutines.core)
             // used for the test scheduler
             implementation(libs.kotlinx.coroutines.test)
+            implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.8.1")
```
OliverO2: Will this become a version catalog dependency at some point?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/interceptor/SpecInterceptorPipeline.kt
```diff
@@ -48,6 +50,8 @@ internal class SpecInterceptorPipeline(
 
    private fun createPipeline(): List<SpecInterceptor> {
       return listOfNotNull(
+         CoroutineNameInterceptor,
+         SpecContextElementInteceptor,
```
OliverO2, line 54: Could we combine these two into an interceptor that adds any type of `CoroutineContext` at spec-level and then use that interceptor for `CoroutineName` and `SpecContextElement`?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/BeforeSpecListenerInterceptor.kt
```diff
@@ -42,9 +45,15 @@ internal class BeforeSpecListenerInterceptor(
...
-               .fold(
+            else -> {
+
+               // beforeSpec needs to execute on the spec level coroutine, but here we are in
+               // the test executor pipline. We are doing the beforeSpec here and not in the spec pipeline
+               // because we only want to run it when we see an enabled test.
+               val context = scope.coroutineContext[SpecContextElement] ?: error("No SpecContextElement")
+               CoroutineScope(context.context).async {
+                  SpecExtensions(registry).beforeSpec(testCase.spec)
+               }.await().fold(
```
OliverO2:
```suggestion
            else -> {
               // beforeSpec needs to execute on the spec level coroutine, but here we are in
               // the test executor pipline. We are doing the beforeSpec here and not in the spec pipeline
               // because we only want to run it when we see an enabled test.
               val specContext = scope.coroutineContext[SpecContextElement]?.context ?: error("No SpecContextElement")
               withContext(specContext) {
                  SpecExtensions(registry).beforeSpec(testCase.spec)
               }.fold(
```

This is equivalent yet more idiomatic.

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/spec/runners/SingleInstanceSpecRunner.kt
```diff
@@ -45,12 +47,14 @@ internal class SingleInstanceSpecRunner(
...
-               Result.success(results)
-            }
+            async { // fresh coroutine for each spec
+               pipeline.execute(spec) {
+                  val rootTests = materializer.materialize(spec)
+                  logger.log { Pair(spec::class.bestName(), "Launching ${rootTests.size} root tests on $scheduler") }
+                  scheduler.schedule({ runTest(it, null) }, rootTests)
+                  Result.success(results)
+               }
+            }.await()
```
OliverO2:
```suggestion
            pipeline.execute(spec) {
               val rootTests = materializer.materialize(spec)
               logger.log { Pair(spec::class.bestName(), "Launching ${rootTests.size} root tests on $scheduler") }
               scheduler.schedule({ runTest(it, null) }, rootTests)
               Result.success(results)
            }
```

The enclosing `coroutineScope` already does create a new coroutine. We don't gain anything when we create another one inside, and `async { ... }.await()` is effectively identical to `coroutineScope { ... }`.

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/CallbackCoroutinesSpecificationTest.kt
```diff
@@ -0,0 +1,112 @@
...
+            .withClasses(Callbacks::class)
+            .launch()
+         beforeSpecName shouldBe "kotest-spec-Callbacks"
+      }
+
+      test("afterSpec should be executed in the spec level coroutine") {
+         TestEngineLauncher(NoopTestEngineListener)
+            .withClasses(Callbacks::class)
+            .launch()
+         beforeSpecName shouldBe "kotest-spec-Callbacks"
```
OliverO2: `afterSpecName`?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/CallbackCoroutinesSpecificationTest.kt
```diff
@@ -0,0 +1,112 @@
...
+import io.kotest.engine.listener.NoopTestEngineListener
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.string.shouldStartWith
+import kotlinx.coroutines.CoroutineName
+import kotlin.coroutines.coroutineContext
+
+/**
+ * This test is meant to be the definite source of truth for the behavior of coroutines and callbacks in kotest.
+ */
+class CallbackCoroutinesSpecificationTest : FunSpec() {
```
OliverO2, line 13: Could we launch the test engine once and then have each test just check for its respective variable?

OliverO2 (changes requested):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/interceptors/BeforeSpecListenerInterceptor.kt
OliverO2: ([IntelliJ failed to properly support a multiline suggestion.](https://youtrack.jetbrains.com/issue/IJPL-73076))

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/spec/runners/SingleInstanceSpecRunner.kt
sksamuel: Are you sure coroutineScope makes a new coroutine?
Because if I do this:

```
suspend fun main() {
   System.setProperty("kotlinx.coroutines.debug", "on")
   println("Current coroutine $coroutineContext")
   coroutineScope {
      println("Current coroutine $coroutineContext")
      launch {
         println("Current coroutine $coroutineContext")
         coroutineScope {
            println("Current coroutine $coroutineContext")
            launch {
               println("Current coroutine $coroutineContext")
            }
         }
      }
   }
}
```

Only the launch causes the coroutine id to be increased. Maybe coroutineScope makes a new one but keeps the same id ?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/CallbackCoroutinesSpecificationTest.kt
sksamuel: Done

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/CallbackCoroutinesSpecificationTest.kt
sksamuel: Fixed

kotest-framework/kotest-framework-engine/build.gradle.kts
sksamuel: It was already there, my bad. Fixed.

sksamuel:
@OliverO2 curious on a change you added.

```
val specContext = scope.coroutineContext[SpecContextElement]?.context ?: error("No SpecContextElement")
               withContext(specContext) {
                  SpecExtensions(registry).beforeSpec(testCase.spec)
               }
```

Does this mean the extension runs in a coroutine that is attached to the spec level coroutine? Or does withContext just add in the new context elements into the existing coroutine, and so the parent/child relationship is still the same as before aka, still running it on the test level coroutine).

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/spec/runners/SingleInstanceSpecRunner.kt
OliverO2: Yes `coroutineScope` does create a coroutine, namely a `ScopeCoroutine`.

We should not rely on `CoroutineId` providing any meaningful value. It is internal and not documented properly. There are lots of coroutine types, and only some of them (like from `launch`, `async`) do get a new `CoroutineId`.

The easiest way to differentiate between coroutines is via their `Job` context element. If you change the above code to print out just `${coroutineContext[Job]}`, each println will show a new object. And the job hierarchy is what's relevant in terms of waiting and cancellation.

For all practical purposes, I'd say we can safely think of `Job` and coroutine (`AbstractCoroutine` derivative) being equivalent. I'm not sure whether the complicated differentiation between those two inside `kotlinx.coroutines` is still necessary or a historical relic (you know, those in the museum basement), but we should probably not care too much.

On a side note: Use `async { ... }` only if you're interested in the block's result. Otherwise `launch { ... }` is the idiomatic abstraction, along with `join()` to wait for it.

OliverO2:
> Does this mean the extension runs in a coroutine that is attached to the spec level coroutine?

Yes, its coroutine becomes a direct child of the spec-level coroutine. You can verify this via

```kotlin
println("current job: ${coroutineContext[Job]}, parent job: ${coroutineContext[Job]?.parent}")
```

> Or does withContext just add in the new context elements into the existing coroutine, and so the parent/child relationship is still the same as before aka, still running it on the test level coroutine).

No, a coroutine context is immutable. Its elements can be inherited by a new coroutine context, but never changed in a context, once it has been created.

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/spec/runners/SingleInstanceSpecRunner.kt
sksamuel: Ok that makes sense. So I guess withContext makes a new coroutine too, unless the context isn't actually changing?

OliverO2:
The tests fail on JS platforms because of `IllegalStateException: No SpecContextElement`.

For whatever reason, the relevant coroutine context when invoking the function returned by `BeforeSpecListenerInterceptor.intercept` on these platforms looks like this:

wasmJs:
```
scope.coroutineContext=[CoroutineName(kotest-test-be able to do arithmetic), TestPathContextElement@1270228841, TestNameContextElement@383059438, SupervisorCoroutine{Active}@810667426, NodeDispatcher@1833589807]
```

js:
```
scope.coroutineContext=[CoroutineName(kotest-test-be able to do arithmetic), [object Object], [object Object], SupervisorCoroutine{Active}@1, NodeDispatcher@2]
```

In contrast, on the JVM:
```
scope.coroutineContext=[kotlinx.coroutines.UndispatchedMarker@43f1fd63, io.kotest.engine.spec.interceptor.instance.SpecContextElement@353009fd, ProjectContextElement(projectContext=ProjectContext(suite=TestSuite(specs=[Reference(kclass=class TestSpec)]), tags=TagExpression(expression=), configuration=io.kotest.core.config.ProjectConfiguration@629cf344)), ConfigurationContextElement(configuration=io.kotest.core.config.ProjectConfiguration@629cf344), CoroutineId(3), CoroutineName(kotest-test-be able to do arithmetic), io.kotest.common.TestPathContextElement@6b3e7bcb, io.kotest.common.TestNameContextElement@27e8892a, "kotest-test-be able to do arithmetic#3":DispatchedCoroutine{Active}@6c9b6c1c, java.util.concurrent.Executors$FinalizableDelegatedExecutorService@6ea5961f]
```

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/interceptor/instance/SpecContextElementInteceptor.kt
```diff
@@ -0,0 +1,32 @@
...
+import kotlinx.coroutines.withContext
+import kotlin.coroutines.AbstractCoroutineContextElement
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.coroutineContext
+
+/**
+ * A [SpecInterceptor] that adds a [SpecContextElement] to this coroutine's context so later
+ * we can get access to the spec level coroutine.
+ */
+internal object SpecContextElementInteceptor : SpecInterceptor {
```
OliverO2, line 16: Typo: Can I buy an `r`?

OliverO2:
The `SpecContextElement` is missing on non-JVM platforms because the `SpecInterceptorPipeline` is only created by `SpecRunner` derivatives. These in turn are only created by `JvmSpecExecutorDelegate`.

So the `SpecInterceptorPipeline` is not present on `DefaultSpecExecutorDelegate` and `KotlinJsTestSpecExecutorDelegate`. I'm not sure where the right place to add it might be, and it would be a lot of duplicated code.

This leads me to #4293.

sksamuel:
Closing in favor of https://github.com/kotest/kotest/pull/4568

Issue #3979, opened 2024-04-23
Can I use prepareSpec without @AutoScan?


dondod:
I currently use an approach like the following to set up a database schema before tests run.  It needs to happen before any test class is loaded. 

```kotlin
class KotestExtensions {
    @AutoScan
    class PrepareSpecTestListener : TestListener {
        override suspend fun prepareSpec(kclass: KClass<out Spec>) {
            // set up database, etc.
        }
    }
}
```

But this generates this warning: 
```
Warning: Kotest autoscan is enabled. This means Kotest will scan the classpath for extensions that are annotated with @AutoScan.
 To avoid this startup cost, set autoscan to false by setting the system property 'kotest.framework.classpath.scanning.config.disable=true'. In 6.0 this value will default to true. For further details see 
https://kotest.io/docs/next/framework/project-config.html#runtime-detection
```

I've tried disabling `autoscan` in another project that doesn't need the database setup, and it speeds up the test startup by several seconds (maybe 5 or more).

Is there a way to use `prepareSpec` without using `@AutoScan`?

Kantis:
You can define `kotest.properties` (details [here](https://kotest.io/docs/intellij/intellij-properties.html)) with the following content:
```
kotest.framework.classpath.scanning.config.disable=true
kotest.framework.classpath.scanning.autoscan.disable=true
kotest.framework.config.fqn=com.myproject.KotestConfig
```

Where `com.myproject.KotestConfig` is
```kotlin
object KotestConfig : AbstractProjectConfig() {
  override fun extensions() = listOf(
    PrepareSpecTestListener()
  )
}
```

That should get rid of runtime detection while also letting you configure Kotest.

dondod:
Thanks!  That almost worked.. I get an error though: 
```
io.kotest.engine.spec.SpecInstantiationException: Could not create instance of class com.myproject.MySpec. Specs must have a public zero-arg constructor.
	at io.kotest.engine.spec.InstantiateSpecKt.javaReflectNewInstance(instantiateSpec.kt:46)
	at io.kotest.engine.spec.InstantiateSpecKt.createAndInitializeSpec(instantiateSpec.kt:30)
	at io.kotest.engine.spec.InstantiateSpecKt.instantiate(instantiateSpec.kt:11)
	at io.kotest.engine.spec.SpecRefKt.instance(SpecRef.kt:14)
[...]

PR #4285, opened 2024-08-11
Adds after listener which does not run for nested scopes

### LINKED ISSUE #2792: Nested "before" lifecycle hook

sksamuel:
Closes #2792 #3979


kotest-framework/kotest-framework-api/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/ContainerScope.kt
```diff
@@ -98,6 +98,23 @@ interface ContainerScope : TestScope {
       afterAny(f)
    }
 
+   /**
+    * Registers a callback that is executed once for this container after all tests have completed.
+    *
+    * This differs from [afterContainer] which is executed after each container including child containers.
+    */
+   fun after(f: suspend (TestCase) -> Unit) {
```
AlexCue987: should we rename `after` to `afterAllTestsInContainerCompleted`? I'm sure there is a better name than what I've suggested, but IMO just `after` is not clear enough

sksamuel: The name was the longest part of the discussion in #2792 
I'm open to anything, there's `afterScope` perhaps. I don't think we want something too wordy, like `afterContaineratThisLevelOnly`

AlexCue987: `afterScope` sounds good

OliverO2: I'm always interested in improving naming. In this case, I don't feel like I'm able to meaningfully contribute unless I have a complete visualization of our test hierarchy along with corresponding lifecycle methods (listeners, interceptors, whatever). Would be cool if we could get that documented completely in one place.

sksamuel: afterScope it is
And will make a PR to  document all our callbacks.

AlexCue987 (approved):

dondod:
I'm confused. Why was `afterScope` added but not `beforeScope`?  I know you can just put the code directly inside the block, but it's quite odd.

sksamuel:
> I'm confused. Why was `afterScope` added but not `beforeScope`? I know you can just put the code directly inside the block, but it's quite odd.

It's literally because it wouldn't do anything that just putting the code at the same level does.
Do you think there's value to have a beforeScope block that is the same? If so, let me know why, we can consider it.

dondod:
Thanks. Yeah, I think there's great value in adding it.  It provides a consistency that people might expect (consistency with other frameworks they might be coming from), and provides consistency within the framework (other "after" hooks have a corresponding "before" hook).  It also helps organize the code and make it obvious what is happening.  If I see code inside a `beforeScope` method, I know that it runs before the scope, and the corresponding `afterScope` runs after the scope.   I've had people at work ask me why I sometimes put code directly inside the `describe` block instead of in a `beforeEach` hook.  I explain that putting the code inside the `describe` block works like a "before all" hook.  They always ask, "why isn't there a `beforeAll` hook?".   Not having to explain that would be worth it to me :).  The `afterScope` hook is a huge win though.

sksamuel:
Coming back to this and thinking about it - one of the issues with adding `beforeScope`, would be that people might assume it always runs before nested tests regardless of where it is defined, but it would be very complicated (and break existing behavior).

Eg,

```kotlin
context("foo") {
  test("bar") { }
  beforeScope { println("hello") }
}
```

Here it is reasonable to think that the `beforeScope` will run before bar, but it wouldn't as bar is executed as soon as it is discovered. So it would end up being counter intuitive I think. Whereas this looks like how it works.

```kotlin
context("foo") {
  test("bar") { }
  println("hello")
}
```

dondod:
That's a good point, but I think it would be even more counterintuitive to put a `beforeScope` after a test at all.  In other test frameworks with a similar hook, I've never seen anyone put the "before" hook after a test, even if it would work (it would in rspec or mocha, for example).

On a side note, I recently had Claude Code refactor some `DescribeSpec` tests, and I told it to use `afterScope`, etc. to consolidate setup and cleanup.  It assumed there was a `beforeScope` and kept asking to search the kotest docs :).

PR #4286, opened 2024-08-11
Define order of test level extensions

### LINKED ISSUE #3555: Unexpected execution order of lifecycle callbacks
I am using an extension to setup a database for each test. For this the `beforeEach` function of the extension must be called before the `beforeEach` function of the test classes, because the latter requires the database to be already setup.

My expectation would be that the lifecycle callbacks of the installed extension are always executed before the lifecycle callbacks of test class, but this actually depends on the way the extension is installed and the test class implements the callbacks which I find very unintuitive.

For example, these two variants work:

```kotlin
class MyTest : StringSpec({
    val dbExtension = extension(DatabaseExtension())
    
    beforeEach {
        dbExtension.doSomething()
    }
    
    ...
}
```

```kotlin
class MyTest : StringSpec() {
    private val dbExtension = DatabaseExtension()
    
    override fun extensions(): List<Extension> = listOf(dbExtension)

    override suspend fun beforeEach(testCase: TestCase) {
        dbExtension.doSomething()
    }
    
    ...
}
```

While this variant does not work:

```kotlin
class MyTest : StringSpec() {
    private val dbExtension = extension(DatabaseExtension())

    override suspend fun beforeEach(testCase: TestCase) {
        dbExtension.doSomething()
    }
    
    ...
}
```

The reason is the [following code block](https://github.com/kotest/kotest/blob/9d31ad7a46b5b706d587aad7d46d89344ea89a4e/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestExtensions.kt#L37-L44) which puts the test listeners of installed extensions after the test listener created for the overridden lifecycle functions:

```kotlin
fun extensions(testCase: TestCase): List<Extension> {
    return testCase.config.extensions +
        testCase.spec.extensions() + // overriding the extensions function in the spec
        testCase.spec.listeners() + // overriding the listeners function in the spec
        testCase.spec.functionOverrideCallbacks() + // spec level dsl eg beforeTest { }
        testCase.spec.registeredExtensions() + // added to the spec via register
        registry.all() // globals
} 
```

* In variant 1 both the extension and the `beforeEach` function are registered as extensions and are contained in `testCase.spec.registeredExtensions()`. The order is as expected because the extension is installed first.
* In variant 2 the order is as expected because the extension is contained in `testCase.spec.extensions()` and the test class callbacks are contained in `testCase.spec.functionOverrideCallbacks()`.
* In variant 3 the order is not as expected because the extension is contained in `testCase.spec.registeredExtensions()` but the test class callbacks are contained in `testCase.spec.functionOverrideCallbacks()`.

Is there a reason for this order or could `testCase.spec.functionOverrideCallbacks()` safely be moved below `testCase.spec.registeredExtensions()`? For my use case it seems that this would provide a more intuitive behavior.

sksamuel:
Closes #3555

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/extensions/spec/AfterEachOrderSpecificationTest.kt
+++ b/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/extensions/spec/AfterEachOrderSpecificationTest.kt
@@ -12,22 +12,22 @@ import io.kotest.matchers.shouldBe
  */
 class AfterEachOrderSpecificationTest : FunSpec() {
 
-   private var order = ""
+   private val order = mutableListOf<String>()
 
    override suspend fun afterEach(testCase: TestCase, result: TestResult) {
-      order += "a"
+      order.add("fn_override")
    }
 
    override fun extensions(): List<Extension> {
       return listOf(
          object : AfterEachListener {
             override suspend fun afterEach(testCase: TestCase, result: TestResult) {
-               order += "f"
+               order.add("extension1")
             }
          },
          object : AfterEachListener {
             override suspend fun afterEach(testCase: TestCase, result: TestResult) {
-               order += "g"
+               order.add("extension2")
             }
          },
       )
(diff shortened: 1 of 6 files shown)


kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/extensions/spec/AfterEachOrderSpecificationTest.kt
```diff
@@ -0,0 +1,66 @@
...
+               order += "g"
+            }
+         },
+      )
+   }
+
+   init {
+
+      afterProject {
+         order shouldBe "fgacbfgaedcbfgaedcb"
```
AlexCue987: will this test become more readable if single characters are replaced with meaningful names, so that `order` is a comma-separated list like this: `listener0, listener1,...`?

sksamuel: done, good suggestion, much clearer

AlexCue987 (approved):

Issue #4029, opened 2024-05-20
Address nullability issue with Arb.edgecase


Kantis:
An `Arb<T>`, where `T` is nullable currently has no way of expressing the null edgecase, since returning null from `Arb.edgecase` is interpreted as "no edgecase exists" by the property testing framework. 

## Suggestion
Change  `Arb<T>.edgecase(rs: RandomSource): T?` to `Arb<T>.edgecase(rs: RandomSource): Sample<T>?`. Returning null can still be interpreted as no edgecase exists. Sample lets us box the null edgecase so it can be represented as well.

## Reference

- [Ticket for nullability in Arb.lists](https://github.com/kotest/kotest/issues/3981)
- [Attempted fix for above, which caused performance degradation for nested lists](https://github.com/kotest/kotest/pull/3982) 
- [Ticket failed related to performance impact](https://github.com/kotest/kotest/issues/4016)

AlexCue987:
Is this a good use case for Kotlin's own `Result`?

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/

Kantis:
Arb interface already uses `Sample` to communicate samples being pulled from the Arb. 
https://github.com/kotest/kotest/blob/7d07732a10b18c329dddaea5b592e68ccfdff584/kotest-property/src/commonMain/kotlin/io/kotest/property/Gen.kt#L108

oliverblaha:
@Kantis I've added a draft PR to showcase how that would look like. Some thoughts are in the PR's comments.
@AlexCue987 `Result` requires an exception in case of failure, so it's not an ideal match here.

dconeybe:
I've implemented a workaround for `edgecase()` returning null in [Firebase Data Connect](https://github.com/firebase/firebase-android-sdk/tree/main/firebase-dataconnect) using this code:

```kt
import io.kotest.property.Arb
import io.kotest.property.RTree
import io.kotest.property.RandomSource
import io.kotest.property.Sample
import io.kotest.property.arbitrary.orNull
import io.kotest.property.asSample
import io.kotest.property.map

/**
 * A class that simply wraps a reference to another object, which may be null.
 * This class can be useful for use in the case where the meaning of `null` is overloaded,
 * such as [io.kotest.property.Arb.edgecase] and
 * [kotlinx.coroutines.flow.MutableStateFlow.compareAndSet]
 */
class NullableReference<out T>(val ref: T? = null) {
[...]

PR #4288, opened 2024-08-12
Add workaround for Arb.list failing with edge cases including null values

oliverblaha:
This shall pass the regression test for https://github.com/kotest/kotest/issues/4016 while avoiding the original problem from https://github.com/kotest/kotest/issues/3981.

For proper support of nullability in edge cases, see https://github.com/kotest/kotest/issues/4029.


sksamuel (approved):

sksamuel:
lgtm

PR #4289, opened 2024-08-12
Change edgecase(rs): A? to return Sample<A>?

oliverblaha:
This would address https://github.com/kotest/kotest/issues/4029 and make return types of edgecase and sample consistent.

Caveat: it requires any edgecaseFn (e.g., in arbitrary calls) to define a function that returns a Sample<A> instead of A - that's not really nice. However, it's still better than the alternative: still allowing a special meaning of "null" in those functions, as that would be inconsistent and probably confusing.

Keeping non-nullable alternatives wouldn't work nicely due to name clashes.

It might be worth considering other alternatives, but I can't think of any that hasn't any major drawbacks.
Maybe, instead using Sample<A>, introduce a new interface Edgecase<A>?


sksamuel:
I think Sample is the right word still.
I thought about this when doing the new syntax proposal and ended up thinking that both types of values are samples. Random samples and edge case samples.

sksamuel:
I tried to make this change too and couldn't get the trampoline stuff to work, so you're a better man than me :)

sksamuel (approved):

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/BuilderTest.kt
```diff
@@ -195,8 +196,8 @@ class BuilderTest : FunSpec() {
             arbitrary(classifier) { arb.bind() }.single() shouldBe 5
             arbitrary(shrinker, classifier) { arb.bind() }.single() shouldBe 5
             arbitrary(listOf(5)) { arb.bind() }.single() shouldBe 5
-            arbitrary({ 5 }) { arb.bind() }.single() shouldBe 5
-            arbitrary({ 5 }, shrinker) { arb.bind() }.single() shouldBe 5
+            arbitrary({ 5.asSample() }) { arb.bind() }.single() shouldBe 5
+            arbitrary({ 5.asSample() }, shrinker) { arb.bind() }.single() shouldBe 5
```
oliverblaha, line 203: @sksamuel the `asSample()` is the part I meant - that looks slightly ugly now. Don't really have a better idea though.

Some thought:

```kotlin
inline fun <reified A : Any> sample(block: () -> A) = block().let {
    when {
        it is Sample<A> -> it
        else -> it.asSample()
    }
}

arbitrary(sample { 5 }, /* ... */)
```

AlexCue987 (approved):

sksamuel:
@oliverblaha can you run apiDump

AlexCue987:
can you resolve merge conflicts @oliverblaha

sksamuel (approved):

PR #4296, opened 2024-08-21
Fix #4294 - reversed message in shouldNotBeBefore matcher

### LINKED ISSUE #4294: Need to reverse/change the assertion error message
https://github.com/kotest/kotest/blame/9a1e24bad9a5eadb9f4069d6912fba0020e2d34e/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/date/instant.kt#L64

when using `shouldNotBeBefore` (other methods are probably similar) the assertion error message is incorrect.
example test
```kotlin
        @Test
        fun `test error message`() {
            val a = Instant.now()
            val b = from.minusMillis(1)
            b.shouldNotBeBefore(a)
        }
```
It fails as expected but the error message reads
```2024-08-19T09:20:57.794921Z is not expected to be before 2024-08-19T09:20:57.777346Z.```
so the statement is misleading

abneroliveira-toast:
This fix the issue https://github.com/kotest/kotest/issues/4294


AlexCue987 (approved):

sksamuel:
Tests are failing

Issue #4297, opened 2024-08-25
Multiple TestRun implementations cause test report overwrites


aSemy:
Continued from https://github.com/kotest/kotest/pull/4291#issuecomment-2286872567

[Slack discussion](https://kotlinlang.slack.com/archives/CSYLDDZUY/p1724580427631109)

---

`:kotest-tests:kotest-tests-junitxml:check` is failing on CI.

I'm looking at fixing the failing `:kotest-tests:kotest-tests-junitxml` tests

https://github.com/kotest/kotest/pull/4291#issuecomment-2287047133

I think the problem is that Gradle permits creating multiple test runs. In KGP you do so like this:

```kotlin
kotlin {
   jvm {
      testRuns.named("main") {
         // the regular `./gradlew :jvmTest` task`
      }
      testRuns.create("jvmCustomTest") {
         // a custom `./gradlew :jvmCustomTest` task`
      }
   }
}
```

However, Kotest doesn't pick up on this, and so it outputs the reports for both runs to the same directory.

https://github.com/kotest/kotest/blob/408d5f0611a397df533119f23be9539721a915ca/kotest-extensions/kotest-extensions-junitxml/src/jvmMain/kotlin/io/kotest/extensions/junitxml/JunitXmlReporter.kt#L46

I'd like to be able to fix this automatically, so that the reports naturally get put into independent directories, but I'm not sure how.

Gradle launches Kotest using [KotestJunitPlatformTestEngine](https://github.com/kotest/kotest/blob/47f35f93c2fecda800e25795ad4adbbdc9e86da8/kotest-runner/kotest-runner-junit5/src/jvmMain/kotlin/io/kotest/runner/junit/platform/KotestJunitPlatformTestEngine.kt#L27). Does Gradle report the task-specific report directory to KotestJunitPlatformTestEngine?

Alternatively, could Kotest pick up the name of the executed task in another way?

sksamuel:
> Gradle launches Kotest using [KotestJunitPlatformTestEngine](https://github.com/kotest/kotest/blob/47f35f93c2fecda800e25795ad4adbbdc9e86da8/kotest-runner/kotest-runner-junit5/src/jvmMain/kotlin/io/kotest/runner/junit/platform/KotestJunitPlatformTestEngine.kt#L27). Does Gradle report the task-specific report directory to KotestJunitPlatformTestEngine?

No, because gradle just listens to the test engine output and generates its own format.
[...]

PR #4298, opened 2024-08-25
Use task-specific report directory for JUnit XML reporter test

aSemy:
Kotest has multiple JVM test runs, meaning multiple Gradle JVM test tasks. By default Kotest uses the same report directory for all tasks, which means the tests overwrite each other's reports.

This commit passes a system property with a distinct directory for each test, and wipes the directory before each test.

Additionally, I tidied up the assertions in JunitXmlReporterTest.

Partial fix for #4297 - but I think probably Kotest should have a 'test report' configuration property, or some other mechanism that is more robust and can be more easily applied in other projects.


AlexCue987 (commented):
is it still a draft?

AlexCue987 (approved):
very nice!

PR #4299, opened 2024-08-25
Tidy JunitXmlReporter

aSemy:
- Update KDoc.
- Pass in a Clock, so the time can be customised.
- Minor tidying to make the code more Kotlin-esque (although I'll concede that's a personal opinion)

I made these refactorings while looking into #4298, and while they didn't turn out to be necessary (I wanted to pass in a clock so that I could run the JUnit XML report tests programatically), I think the refactorings are still useful.


aSemy:
Marked as a draft until #4298 is ready

AlexCue987 (commented):
is it still a draft?

LeoColman:
Just for safety, could you run your branch against Kotest and share the results so we can take a look?

sksamuel (approved):

PR #4302, opened 2024-08-26
Update CollectionsTest.kt

AlexCue987:
up timeout for one flaky test


aSemy:
Is this test failing again? I'm a little concerned that this duration keeps being bumped up...

We should implement proper performance tests.

I'd like to use this: https://benchmark-action.github.io/github-action-benchmark/dev/bench/

AlexCue987:
> Is this test failing again? I'm a little concerned that this duration keeps being bumped up...
> 
> We should implement proper performance tests.
> 
> I'd like to use this: https://benchmark-action.github.io/github-action-benchmark/dev/bench/

long term I really like this suggestion. but short term, I'm in the middle of doing something else, which makes it very tempting to just kick this can down this road one more time. WDYT?

aSemy:
Ah, I just checked the git history and this test was 1.second when it was introduced, so I must be thinking of another test.

aSemy (approved):

PR #4303, opened 2024-08-27
Test `beforeProject()` is called before all tests in spec-parallelism test

aSemy:
Avoid `lateinit var` for the main time source, because if it fails (possibly due to multithreading??), then it leads to uninformative test failures.

(I would like to make `beforeProjectCalled` a `lateinit var`, but value-classes can't be `lateinit vars`.)

### DIFF
--- a/kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
+++ b/kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
@@ -1,18 +1,20 @@
 package com.sksamuel.kotest.parallelism
 
-import com.sksamuel.kotest.parallelism.ProjectConfig.projectStart
+import com.sksamuel.kotest.parallelism.ProjectConfig.startOfTest
 import com.sksamuel.kotest.parallelism.TestStatus.Status.Finished
 import com.sksamuel.kotest.parallelism.TestStatus.Status.Started
 import com.sksamuel.kotest.parallelism.TestStatus.Status.TimedOut
 import io.kotest.assertions.withClue
 import io.kotest.core.config.AbstractProjectConfig
 import io.kotest.core.config.ProjectConfiguration
+import io.kotest.core.log
 import io.kotest.core.test.TestScope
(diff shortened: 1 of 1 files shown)


kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
```diff
@@ -47,9 +49,11 @@ object ProjectConfig : AbstractProjectConfig() {
...
       CoroutineScope(Dispatchers.IO) + CoroutineName("TestStatusCollector")
 
-   /** Marks the start of the entire tests, when [beforeProject] is called, before any tests are launched. */
-   internal lateinit var projectStart: TimeMark
-      private set
+   /** Mark the start of the entire test case. */
+   internal val startOfTest: TimeMark = TimeSource.Monotonic.markNow()
+
+   /** Marks when [beforeProject] is called, which should be before any tests are launched. */
+   private var beforeProjectCalled: Duration? = null
```
AlexCue987, line 56: are any race conditions possible? Should we wrap it in `AtomicReference` or something?

AlexCue987 (commented):
LGTM, concern over thread safety

kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
aSemy: Good point, probably should be made write-only-once somehow...

Maybe make it a (threadsafe) list and ensure there's only one value.

AlexCue987 (approved):

kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
aSemy: I added this improvement in another PR https://github.com/kotest/kotest/pull/4311

Issue #4280, opened 2024-08-09
Exhaustive property testing: Support skipping N test cases


edgao:
(I might just not be using kotest-property correctly - very possible that pebkac!)

When running an exhaustive property test via forAll, it seems tricky to debug a failed test case. For example:
```kotlin
forAll(
  Exhaustive.int(0..5),
  Exhaustive.int(0..5),
) { row, column ->
  // do something that fails on row=3, column=3
}
```

Other than creating a new test for specifically the failed case, I don't think there's a great way to run it in isolation? Something like this would be nice (since kotest already prints out a nice `Property failed after 18 attempts` message):
```kotlin
forAll(
  PropTestConfig(skipExhaustiveIterations=18),
  ...
) {...}
```

(also happy to take a stab at implementing this. Seems like [proptest.kt](https://github.com/kotest/kotest/blob/master/kotest-property/src/commonMain/kotlin/io/kotest/property/internal/proptest.kt) is the right file to modify?)

sksamuel:
Have you tried using the seed ?

edgao:
hm, I thought I tried that and it still executed all the previous test cases. can try again next week

edgao:
am I using the seed wrongly? My process was:

Added `testImplementation 'io.kotest:kotest-property:5.9.1'` to my build.gradle, then ran a test like this:
```kotlin
    // not sure if it matters: this is junit's annotation org.junit.jupiter.api.Test
    @Test
    fun demo() {
        runBlocking {
            forAll(
                Exhaustive.ints(0..4),
                Exhaustive.ints(0..4),
            ) { row, column ->
                logger.info { "Running with $row, $column" }
                row != 3 || column != 3
            }
        }
    }
```
Which failed as expected:
```
    java.lang.AssertionError: Property failed after 19 attempts
    Repeat this test by using seed 41085330395708674
    Caused by AssertionFailedError: expected:<true> but was:<false>
```

So I plugged in the seed and reran:
```kotlin
    // intellij shows a warning without this annotation?
    @OptIn(ExperimentalKotest::class)
    @Test
    fun demo() {
        runBlocking {
            forAll(
                PropTestConfig(seed = 41085330395708674),
                Exhaustive.ints(0..4),
                Exhaustive.ints(0..4),
            ) { row, column ->
                logger.info { "Running with $row, $column" }
                row != 3 || column != 3
            }
        }
    }
```

[...]

PR #4307, opened 2024-08-29
Add skipTo for skipping N test cases (#4280)

baldovalievlj:
Add a `skipTo` option as requested on #4280

### DIFF
--- a/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/ForAll2Test.kt
+++ b/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/ForAll2Test.kt
@@ -121,4 +121,18 @@ Repeat this test by using seed 1234
 
 Caused by IllegalArgumentException: something unexpected happened"""
    }
+
+   test("forAll with 2 arbs should skip first 4 tests") {
+      shouldThrowAny {
+         forAll(
+            config = PropTestConfig(skipTo = 5, seed = 5847062201763421121),
+            Exhaustive.ints(0..10),
+            Exhaustive.ints(0..10)
+         ) { a, b -> a <= b }
+      }.message shouldBe """Property failed after 8 attempts
+
+Repeat this test by using seed 5847062201763421121
+
+Caused by AssertionFailedError: expected:<true> but was:<false>"""
+   }
 })
--- a/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/ForAll2Test.kt
+++ b/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/ForAll2Test.kt
@@ -1,6 +1,7 @@
 package com.sksamuel.kotest.property
 
 import io.kotest.assertions.throwables.shouldNotThrowAny
+import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.assertions.throwables.shouldThrowAny
 import io.kotest.core.spec.style.FunSpec
 import io.kotest.matchers.shouldBe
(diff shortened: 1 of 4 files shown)


AlexCue987:
@baldovalievlj can you run apiDump step and push the changes?

baldovalievlj:
Yes of course, sorry about that. Pushed.

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/ForAll2Test.kt
```diff
@@ -121,4 +121,18 @@ Repeat this test by using seed 1234
 
 Caused by IllegalArgumentException: something unexpected happened"""
    }
+
+   test("forAll with 2 arbs should skip first 4 tests") {
+      shouldThrowAny {
+         forAll(
+            config = PropTestConfig(skipTo = 5, seed = 5847062201763421121),
```
AlexCue987, line 129: In real life usage, how do we know how many tests to skip? Do we need to document that somewhere?

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/ForAll2Test.kt
```diff
@@ -121,4 +121,18 @@ Repeat this test by using seed 1234
 
 Caused by IllegalArgumentException: something unexpected happened"""
    }
+
+   test("forAll with 2 arbs should skip first 4 tests") {
+      shouldThrowAny {
```
AlexCue987: do we need to be specific, `shouldThrow<AssertionError>`?

AlexCue987 (commented):
LGTM, some concerns about documentation

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/ForAll2Test.kt
OliverO2: As I'm not that familiar with this case, just an idea: If Kotest currently prints
```
Property failed after 18 attempts
```
we could change that message to
```
Property failed after 18 attempts. Use `config = PropTestConfig(skipTo = 17)` to trigger the failed case directly.
```

That kind of message, tailored to the actual case, helps more than just burying a hint in the docs.

AlexCue987: > As I'm not that familiar with this case, just an idea: If Kotest currently prints
> 
> ```
> Property failed after 18 attempts
> ```
> 
> we could change that message to
> 
> ```
> Property failed after 18 attempts. Use `config = PropTestConfig(skipTo = 17)` to trigger the failed case directly.
> ```
> 
> That kind of message, tailored to the actual case, helps more than just burying a hint in the docs.

I like this suggestion, but IMO "Property failed after 18 attempts" means "it succeeded 17 times, then failed", and so if should be " Use `config = PropTestConfig(skipTo = 18)` to trigger the failed case directly." because generally `skipTo = 17` means "skip 16 cases, and do case 17 and onwards"

OliverO2: Don't take my number too seriously. I was just parroting [what Sam wrote](https://github.com/kotest/kotest/issues/4280#issuecomment-2308470266).

baldovalievlj: So what's the consensus here, should we change the message?

AlexCue987: IMO we are good here, "skipTo = 5" does mean "skip first 4 tests"

AlexCue987 (approved):

PR #4311, opened 2024-09-02
Fix kotest-tests-spec-parallelism

aSemy:
Currently the tests run, but the ProjectConfig is not loaded, so the tests aren't tested!

- Set `kotest.framework.config.fqn`
- Fix `statuses.shouldForAll {}` assertion.
- Verify `beforeProject()` is only called once.

### DIFF
--- a/kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
+++ b/kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
@@ -9,22 +9,25 @@ import io.kotest.core.config.AbstractProjectConfig
 import io.kotest.core.config.ProjectConfiguration
 import io.kotest.core.log
 import io.kotest.core.test.TestScope
+import io.kotest.inspectors.shouldForAll
 import io.kotest.inspectors.shouldForNone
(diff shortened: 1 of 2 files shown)


kotest-tests/kotest-tests-spec-parallelism/src/jvmTest/kotlin/com/sksamuel/kotest/parallelism/ProjectConfig.kt
```diff
@@ -52,8 +55,8 @@ object ProjectConfig : AbstractProjectConfig() {
    /** Mark the start of the entire test case. */
    internal val startOfTest: TimeMark = TimeSource.Monotonic.markNow()
 
-   /** Marks when [beforeProject] is called, which should be before any tests are launched. */
-   private var beforeProjectCalled: Duration? = null
+   /** Marks when [beforeProject] is called, which should only be once, and before any tests are launched. */
+   private val beforeProjectCalled = MutableStateFlow(listOf<Duration>())
```
AlexCue987, line 59: what would happen if it's called more than once, so the folling assumption has failed: "should only be once, and before any tests are launched."

aSemy: If `beforeProject()` is called more than once then the assertion on line 97 will fail.

AlexCue987 (approved):

PR #4308, opened 2024-08-31
Reduce the size of stack traces

angusholder:
Kotest stack traces are quite large right now. While intellij does collapse them, I still run into them when seeing test runs on a terminal, like in Github Actions, and it would be nice if they were a bit shorter to make it easier to find my own code in the trace.

I noticed that currently there's three stack trace lines between each interceptor invocation. This is down to them using lambdas via a generic lambda type like `suspend (EngineContext) -> EngineResult`, where, due to type-erasure, the compiler has to generate some extra synthetic methods in between. If we instead define dedicated interfaces for each of those lambda types, the extra synthetic methods go away, reducing the stacktraces to only one extra line between each interceptor.

See the change in the example stack traces below: each lambda call goes from being `invoke()` + `invoke()` + `invokeSuspend()`, to just `invoke()`.

<details>
<summary><b>Stack trace before</b> (53 lines)</summary>
<pre>
**snip**
	at io.kotest.engine.ConcurrentTestSuiteScheduler.schedule(ConcurrentTestSuiteScheduler.kt:41)
	at io.kotest.engine.TestEngine$execute$innerExecute$1.invokeSuspend(TestEngine.kt:65)
	at io.kotest.engine.TestEngine$execute$innerExecute$1.invoke(TestEngine.kt)
	at io.kotest.engine.TestEngine$execute$innerExecute$1.invoke(TestEngine.kt)
	at io.kotest.engine.interceptors.TestEngineInitializedInterceptor.intercept(TestEngineInitializedInterceptor.kt:16)
	at io.kotest.engine.TestEngine$execute$execute$1$1.invokeSuspend(TestEngine.kt:71)
	at io.kotest.engine.TestEngine$execute$execute$1$1.invoke(TestEngine.kt)
	at io.kotest.engine.TestEngine$execute$execute$1$1.invoke(TestEngine.kt)
	at io.kotest.engine.interceptors.DumpConfigInterceptor.intercept(DumpConfigInterceptor.kt:21)
	at io.kotest.engine.TestEngine$execute$execute$1$1.invokeSuspend(TestEngine.kt:71)
	at io.kotest.engine.TestEngine$execute$execute$1$1.invoke(TestEngine.kt)
	at io.kotest.engine.TestEngine$execute$execute$1$1.invoke(TestEngine.kt)
	at io.kotest.engine.interceptors.WriteFailuresInterceptor.intercept(WriteFailuresInterceptor.kt:34)
[...]

PR #4312, opened 2024-09-02
Reduce the size of stack traces (2)

angusholder:
Finishes what I started in #4308, doing the same for SpecInterceptor and SpecRefInterceptor too.

I had to change `ClassVisibilitySpecRefInterceptor` to internal to get it to compile. I might be missing something but it seems like a mistake that that class was made public in the first place? It only implements the internal `SpecRefInterceptor` interface, and doesn't seem to have any utility otherwise. It was introduced in this commit 22nd July https://github.com/kotest/kotest/commit/3b196b1b598aec01de26ca37a52e505391e6ece5

Alternatively I could change `NextSpecRefInterceptor` to `@KotestInternal` and leave that class public.


AlexCue987:
tests failing as follows: "java.lang.ClassCastException: kotlin.Result cannot be cast to java.util.Map"

sksamuel (approved):

sksamuel:
i tried to fix this but couldn't. No idea why the compiler is complaining, it looks fine to me.
Might need to make a new PR which does this bit by bit. If the original author is not around, I can pick it up.

sksamuel:
Got this working via #4400  and #4403 so closing this.

angusholder:
Thanks @sksamuel! I hadn't had the chance to take another look yet

PR #4314, opened 2024-09-03
Alex/should pass (#4313)

AlexCue987:
Allow non-kotest assertions to run inside `assertSoftly`, via a new assertion named `shouldPass`:

```kotlin
assertSoftly {
  (2+2) shouldBe 5
  shouldPass {
    verify(exactly = 1) { myClass.myMethod(any()) }
  }
}
```

### DIFF
--- a/documentation/docs/assertions/soft_assertions.md
+++ b/documentation/docs/assertions/soft_assertions.md
@@ -51,3 +51,26 @@ assertSoftly {
 }
 ```
 
+So if we want to invoke non-kotest assertions inside `assrtSoftly` blocks, they need to be invoked via `shouldPass`.
+In the following example both `verify` and the second assertion can fail, and we shall get both errors accumulated:
+
+```kotlin
+assertSoftly {
+  shouldPass {
+    verify(exactly = 1) { myClass.myMethod(any()) }
+  }
+  foo shouldBe bar
+}
+```
+
+Likewise, in the following example the failure of `verify` will not be ignored, it will be added along with the failure of the first assertion:
+
+
+```kotlin
+assertSoftly {
+  (2+2) shouldBe 5
+  shouldPass {
+    verify(exactly = 1) { myClass.myMethod(any()) }
+  }
+}
+```
--- a/documentation/docs/assertions/soft_assertions.md
+++ b/documentation/docs/assertions/soft_assertions.md
@@ -33,7 +33,7 @@ assertSoftly(foo) {
 
 We can configure assert softly to be implicitly added to every test via [project config](../framework/project_config.md).
 
-**Note:** only Kotest's own assertions can be asserted softly. If any other checks fail and throw an `AssertionError`, it will not respect `assertSoftly` and bubble up, erasing the results of previous assertions. This includes Kotest's own `fail()` function, so when the following code runs, we won't know if the first assertion `foo shouldBe bar` succeeded or failed:
+**Note:** only Kotest's own assertions can be asserted softly. To be compatible with `assertSoftly`, assertions from other libraries must be wrapped in `shouldPass`, which is described leter in this section. If any other checks fail and throw an `AssertionError`, it will not respect `assertSoftly` and bubble up, erasing the results of previous assertions. This includes Kotest's own `fail()` function, so when the following code runs, we won't know if the first assertion `foo shouldBe bar` succeeded or failed:
 
 ```kotlin
 assertSoftly {
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/throwablehandling/CovariantThrowableHandlingTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/throwablehandling/CovariantThrowableHandlingTest.kt
@@ -1,14 +1,20 @@
 package com.sksamuel.kotest.throwablehandling
 
+import io.kotest.assertions.MultiAssertionError
+import io.kotest.assertions.assertSoftly
 import io.kotest.assertions.throwables.shouldNotThrow
+import io.kotest.assertions.throwables.shouldNotThrowAny
 import io.kotest.assertions.throwables.shouldNotThrowUnit
+import io.kotest.assertions.throwables.shouldPass
 import io.kotest.assertions.throwables.shouldThrow
+import io.kotest.assertions.throwables.shouldThrowAny
 import io.kotest.assertions.throwables.shouldThrowUnit
 import io.kotest.assertions.throwables.shouldThrowUnitWithMessage
 import io.kotest.assertions.throwables.shouldThrowWithMessage
 import io.kotest.core.spec.style.FreeSpec
 import io.kotest.matchers.booleans.shouldBeTrue
 import io.kotest.matchers.shouldBe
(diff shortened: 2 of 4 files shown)


documentation/docs/assertions/soft_assertions.md
LeoColman, line 1: lines 34~36 should be adjusted

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/throwablehandling/CovariantThrowableHandlingTest.kt
```diff
@@ -184,6 +190,79 @@ class CovariantThrowableHandlingTest : FreeSpec() {
...
+                     }
+                  }
+               }
+            thrown.message.shouldContainInOrder(
+               """The following 2 assertions failed:""",
+               """1) expected:<5> but was:<4>""",
+               """Unexpected Exception was thrown with the following message: "Non-Assertion Failure!""",
+            )
+         }
+         "asdfa" {
```
AlexCue987: oops need to clean these two tests up. sorry about that

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/throwables/CovariantThrowableHandling.kt
```diff
@@ -260,3 +260,32 @@ inline fun <reified T : Throwable> shouldNotThrow(block: () -> Any?) {
...
+
+/**
+ * Verifies if a block of code will either throw an [AssertionError] or no exception at all. It can only be invoked inside [assertSoftly] blocks.
+ *
+ * Use this function to wrap a block of code that might throw an [AssertionError]. Use it to wrap invocations of non-kotest functions
+ * such as [verify] from mockk library, so that they can be invoked inside [assertSoftly] blocks, and
+ * errors are only thrown at the end of the block, rather than immediately. Therefore, the signature cannot return
+ * the throwable type, as in the case of a failure, there would be neither an immediate throws, nor a type to return.
+ */
+inline fun shouldPass(block: () -> Any?) {
```
LeoColman, line 262: This function is very similar to `shouldNotThrow<> { ... }`

Perhaps we should either reuse the method or have this one have a similar name as  `shouldNotThrow`. This way we don't have 2 similar block code lying around

documentation/docs/assertions/soft_assertions.md
AlexCue987: added `"To be compatible with `assertSoftly`, assertions from other libraries must be wrapped in `shouldPass`, which is described leter in this section"`

AlexCue987: > added `"To be compatible with `assertSoftly`, assertions from other libraries must be wrapped in `shouldPass`, which is described leter in this section"`

added "To be compatible with assertSoftly, assertions from other libraries must be wrapped in shouldPass, which is described leter in this section" - is that what you were suggesting @LeoColman ?

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/throwables/CovariantThrowableHandling.kt
AlexCue987: > This function is very similar to `shouldNotThrow<> { ... }`
> 
> Perhaps we should either reuse the method or have this one have a similar name as `shouldNotThrow`. This way we don't have 2 similar block code lying around

I concur it is similar, but unfortunately it won't do what I want:

```kotlin
class VerifyTest: StringSpec() {
    private val mockService = mockk<MyService>(relaxed = true)

    init {
        "try with shouldNotThrow" {
            assertSoftly {
                shouldNotThrow<AssertionError> {
                    verify(exactly = 1) { mockService.doSomething(any()) }
                }
                (2 + 2) shouldBe 5
            }
        }
    }
}

class MyService {
    fun doSomething(params: String) {}
}

--- the output

No exception expected, but a AssertionError was thrown.
java.lang.AssertionError: No exception expected, but a AssertionError was thrown.

(snip)

Caused by: java.lang.AssertionError: Verification failed: call 1 of 1: MyService(#1).doSomething(any())) was not called

```

There are two issues here:
* `(2 + 2) shouldBe 5` is not run at all
* Unnecessary noise here: "No exception expected, but a AssertionError was thrown." - ideally all we need is "Verification failed: call 1 of 1: MyService(#1).doSomething(any())) was not called"

Can you suggest a better name instead of `shouldPass`?

I concur the following code is very similar:

```kotlin
   assertionCounter.inc()
   val thrownThrowable = try {
      block()
      null
   } catch (thrown: Throwable) {
      thrown
   }
```
but it is not exactly the same. Let me try reusing code.

AlexCue987: IMO it's more similar to `shouldThrowSoftly`

AlexCue987: > This function is very similar to `shouldNotThrow<> { ... }`
> 
> Perhaps we should either reuse the method or have this one have a similar name as `shouldNotThrow`. This way we don't have 2 similar block code lying around

I've extracted some common code into a function which is reused

LeoColman (approved):

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/throwables/CovariantThrowableHandling.kt
Kantis: What Leo said was the first thing that sprung to mind for me as well. Perhaps we should take some time to discuss further before merging? 

As a user of Kotest, I feel that `shouldPass` and `shouldNotThrowAny { ... }`  are semantically indifferent and it's confusing why both would exist.

A possible workaround could be to somehow create a DSL-scope within `assertSoftly` and only expose the `shouldPass` API there?

AlexCue987: > What Leo said was the first thing that sprung to mind for me as well. Perhaps we should take some time to discuss further before merging?
> 
> As a user of Kotest, I feel that `shouldPass` and `shouldNotThrowAny { ... }` are semantically indifferent and it's confusing why both would exist.
> 
> A possible workaround could be to somehow create a DSL-scope within `assertSoftly` and only expose the `shouldPass` API there?

I concur that it would be natural to use `shouldNotThrowAny` in this use case. But it won't do what we want at this time, as it isn't compatible with `assertSoftly`. 

So how about the following change: make all the functions like `shouldNotThrowAny` , `shouldNotThrow`, `shouldThrowAny` etc compatible with `assertSoftly`. Also deprecate `shouldThrowSoftly` - right now it's the only function that works with `assertSoftly`. That done, we shall remove `shouldPass`, it won't be needed anymore. WDYT @Kantis

PR #4316, opened 2024-09-04
Fix deletion of files with unmappable characters

sschuberth:
In contrast to the File's extension function, the Path's one can delete files with unmappable characters like the one at [1].

[1]: https://github.com/fshost/node-dir/tree/a57c3b1b571dd91f464ae398090ba40f64ba38a2/test/fixtures/testdir5

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/spec/tempfile.kt
+++ b/kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/spec/tempfile.kt
@@ -2,13 +2,16 @@ package io.kotest.engine.spec
 
 import io.kotest.core.TestConfiguration
 import java.io.File
+import kotlin.io.path.deleteExisting
 
 fun TestConfiguration.tempfile(prefix: String? = null, suffix: String? = null): File {
-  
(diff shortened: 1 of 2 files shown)


kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/spec/tempfile.kt
```diff
@@ -2,13 +2,16 @@ package io.kotest.engine.spec
...
+import kotlin.io.path.deleteExisting
 
 fun TestConfiguration.tempfile(prefix: String? = null, suffix: String? = null): File {
-  
    val file = kotlin.io.path.createTempFile(prefix ?: javaClass.name, suffix).toFile()
    afterSpec {
-      if (!file.delete())
+      runCatching {
+         file.toPath().deleteExisting()
+      }.onFailure {
```
AlexCue987, line 12: do we need any tests changed/added to go with this code changes?

sschuberth: It's difficult to do, as at least the JVM platform does not support creating such files with unmappable characters to test their deletion...

AlexCue987 (approved):

PR #4327, opened 2024-09-10
Alex/contain null more details draft2 (#4326)

AlexCue987:
make the output of `shouldNotContainNull` more specific:

` "Sequence should not contain any nulls, but contained at least one at index ${indexOfFirstNull!!.first}"`

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
@@ -12,6 +12,7 @@ import io.kotest.matchers.neverNullMatcher
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldHave
 import io.kotest.matchers.shouldNot
+import kotlin.collections.indexOfFirst
 
 /*
 How should infinite sequences be detected, and how should they be dealt with?
@@ -38,7 +39,7 @@ fun <T> Sequence<T>.shouldContainNull() = this should containNull()
 fun <T> Sequence<T>.shouldNotContainNull() = this shouldNot containNull()
 fun <T> containNull() = object : Matcher<Sequence<T>> {
    override fun test(value: Sequence<T>): MatcherResult {
-      val indexOfFirstNull = indexOfFirst { it == null }
+      val indexOfFirstNull = value.indexOfFirst { it == null }
       return MatcherResult(
          indexOfFirstNull != -1,
          { "Sequence should contain at least one null" },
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
```diff
@@ -37,12 +37,14 @@ fun <T> containOnlyNulls() = object : Matcher<Sequence<T>> {
 fun <T> Sequence<T>.shouldContainNull() = this should containNull()
 fun <T> Sequence<T>.shouldNotContainNull() = this shouldNot containNull()
 fun <T> containNull() = object : Matcher<Sequence<T>> {
-   override fun test(value: Sequence<T>) =
-      MatcherResult(
-         value.any { it == null },
+   override fun test(value: Sequence<T>): MatcherResult {
+      val indexOfFirstNull = value.mapIndexed { index, t -> index to t }.firstOrNull { it.second == null }
```
LeoColman: 🟡 We can cleanup this guy with `indexOfFirst`

```suggestion
      val indexOfFirstNull = value.indexOfFirst { it == null }
```

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
```diff
@@ -37,12 +37,14 @@ fun <T> containOnlyNulls() = object : Matcher<Sequence<T>> {
...
+      return MatcherResult(
+         indexOfFirstNull != null,
```
LeoColman: 🟡 Then this would need to be adapted to
```suggestion
         indexOfFirstNull != -1,
```

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
```diff
@@ -37,12 +37,14 @@ fun <T> containOnlyNulls() = object : Matcher<Sequence<T>> {
...
          { "Sequence should contain at least one null" },
-         { "Sequence should not contain any nulls" }
+         { "Sequence should not contain any nulls, but contained at least one at index ${indexOfFirstNull!!.first}" }
```
LeoColman: 🟡 And finally,
```suggestion
         { "Sequence should not contain any nulls, but contained at least one at index $indexOfFirstNull" }
```

LeoColman (changes requested):
We can cleanup the logic a bit :)

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
AlexCue987: oops forgot about that thing, TYM

LeoColman (approved):
I think some imports are missing on the files

PR #4331, opened 2024-09-11
haveElementAt-more-details (#4330)

AlexCue987:
when `haveElementAt` fails for a `Sequence`, output more details on what exactly failed. Also make a breaking change on how empty/short sequences are handled, to make it consistent with the behavior of similar matchers for Lists, Arrays etc.

```kotlin
         "print that the sequence is empty" {
            shouldThrow<AssertionError> {
               sequenceOf<String>().shouldHaveElementAt(3, "banana")
            }.message shouldBe """Sequence should contain "banana" at index 3, but the sequence was empty."""
         }

         "print that the sequence is shorter" {
            shouldThrow<AssertionError> {
               sequenceOf("apple", "orange", "lemon").shouldHaveElementAt(3, "banana")
            }.message shouldBe """Sequence should contain "banana" at index 3, but the sequence only had 3 elements"""
         }

         "print that the actual element did not match" {
            shouldThrow<AssertionError> {
               sequenceOf("apple", "orange", "lemon").shouldHaveElementAt(2, "banana")
            }.message shouldBe """Sequence should contain "banana" at index 2, but the value was different: "lemon"."""
         }
```

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
@@ -49,12 +49,19 @@ fun <T> Sequence<T>.shouldHaveElementAt(index: Int, element: T) = this should ha
 fun <T> Sequence<T>.shouldNotHaveElementAt(index: Int, element: T) = this shouldNot haveElementAt(index, element)
 
 fun <T, S : Sequence<T>> haveElementAt(index: Int, element: T) = object : Matcher<S> {
-   override fun test(value: S) =
-      MatcherResult(
-         value.elementAt(index) == element,
-         { "Sequence should contain $element at index $index" },
-         { "Sequence should not contain $element at index $index" }
+   override fun test(value: S): MatcherResult {
+      val elementAtIndex = value.take(index + 1).mapIndexed { index, t -> index to t }.lastOrNull()
+      val passed = elementAtIndex?.let { it.first == index && it.second == element } ?: false
+      val description = elementAtIndex?.let {
+         if(it.first == index) ", but the value was different: ${it.second.print().value}."
+         else ", but the sequence only had ${it.first + 1} elements"
+      } ?: ", but the sequence was empty."
+      return MatcherResult(
+         passed,
+         { "Sequence should contain ${element.print().value} at index $index$description" },
+         { "Sequence should not contain ${element.print().value} at index $index" }
       )
+   }
 }
 
 fun <T> Sequence<T>.shouldContainNoNulls() = this should containNoNulls()
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
@@ -50,12 +50,14 @@ fun <T> Sequence<T>.shouldNotHaveElementAt(index: Int, element: T) = this should
 
 fun <T, S : Sequence<T>> haveElementAt(index: Int, element: T) = object : Matcher<S> {
    override fun test(value: S): MatcherResult {
-      val elementAtIndex = value.take(index + 1).mapIndexed { index, t -> index to t }.lastOrNull()
-      val passed = elementAtIndex?.let { it.first == index && it.second == element } ?: false
-      val description = elementAtIndex?.let {
-         if(it.first == index) ", but the value was different: ${it.second.print().value}."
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
```diff
@@ -49,12 +49,19 @@ fun <T> Sequence<T>.shouldHaveElementAt(index: Int, element: T) = this should ha
 fun <T> Sequence<T>.shouldNotHaveElementAt(index: Int, element: T) = this shouldNot haveElementAt(index, element)
 
 fun <T, S : Sequence<T>> haveElementAt(index: Int, element: T) = object : Matcher<S> {
-   override fun test(value: S) =
-      MatcherResult(
-         value.elementAt(index) == element,
-         { "Sequence should contain $element at index $index" },
-         { "Sequence should not contain $element at index $index" }
+   override fun test(value: S): MatcherResult {
+      val elementAtIndex = value.take(index + 1).mapIndexed { index, t -> index to t }.lastOrNull()
```
LeoColman: 🔵 Why can't this be `value.elementAt(index)`?

You have the index from `index` if you need to use it downstream

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
```diff
@@ -49,12 +49,19 @@ fun <T> Sequence<T>.shouldHaveElementAt(index: Int, element: T) = this should ha
...
+      val passed = elementAtIndex?.let { it.first == index && it.second == element } ?: false
```
LeoColman: 🔵 Following what I said above, this could be simplified to `element == elementAtIndex`

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
```diff
@@ -49,12 +49,19 @@ fun <T> Sequence<T>.shouldHaveElementAt(index: Int, element: T) = this should ha
...
+      val description = elementAtIndex?.let {
+         if(it.first == index) ", but the value was different: ${it.second.print().value}."
```
LeoColman: 🟡 I think this pair being passed around should at least be desconstructed with proper names.

I want to suggest an alternative for this description with so many branches and magiic numbers, but it became hard to navigate it. I think that needs change

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/SequenceMatchersTest.kt
```diff
@@ -431,12 +433,16 @@ class SequenceMatchersTest : WordSpec() {
       }
 
       "have element at" should {
-         abort<IndexOutOfBoundsException>("for empty") {
-            sampleData.empty.shouldHaveElementAt(sampleData.empty.count(), 0)
+         "handle empty sequence" {
+            shouldThrow<AssertionError> {
+               sampleData.empty.shouldHaveElementAt(sampleData.empty.count(), 0)
+            }.message shouldContain "but the sequence was empty"
```
LeoColman: 🔵 I see this more as a bugfix than a breaking change. I think we are fine on that front

LeoColman (changes requested):
I think we can simplify the logic a little bit

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
AlexCue987: > elementAt



> 🔵 Why can't this be `value.elementAt(index)`?
> 
> You have the index from `index` if you need to use it downstream

If we go for `elementAt`, we do not know the actual length of the sequence, we only know that it is too short.

```kotlin
class MyTest: StringSpec() {
    init {
        "blows up" {
            val element = sequenceOf(0, 1, 2).elementAt(4)
            println(element)
        }
    }
}

Sequence doesn't contain element at index 4.
java.lang.IndexOutOfBoundsException: Sequence doesn't contain element at index 4.

```

Because we want to say things like "but the sequence is empty" or "but the sequence only has 3 elements", we need to do some extra work. Does it make sense to you?

LeoColman: In that case, can't we use a `.size` or `.count()` and still simplify the rest?

LeoColman (approved):

PR #4336, opened 2024-09-12
Leo-s-PR-comments

AlexCue987:


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sequences/matchers.kt
```diff
@@ -50,12 +50,14 @@ fun <T> Sequence<T>.shouldNotHaveElementAt(index: Int, element: T) = this should
...
-      val passed = elementAtIndex?.let { it.first == index && it.second == element } ?: false
-      val description = elementAtIndex?.let {
-         if(it.first == index) ", but the value was different: ${it.second.print().value}."
-         else ", but the sequence only had ${it.first + 1} elements"
-      } ?: ", but the sequence was empty."
+      val sequenceHead = value.take(index + 1).toList()
+      val elementAtIndex = sequenceHead.elementAtOrNull(index)
+      val passed = elementAtIndex == element
+      val description = when{
+         passed == element -> ""
```
LeoColman: 🔴 I think this is wrong, `passed` is a boolean and is being compared to T

AlexCue987: > 🔴 I think this is wrong, `passed` is a boolean and is being compared to T

yep. fixed

PR #4340, opened 2024-09-15
Add system properties for default shrinking mode

JordanLongstaff:
I thought it'd be a good idea to have system properties to globally configure shrinking. The concept should be relatively straightforward:

- `kotest.property.default.shrinking.mode`: can be set to `off`, `bounded` or `unbounded`, and throws an error with any other value
- `kotest.property.default.shrinking.bound`: if the mode is `bounded`, sets the bound

### DIFF
--- a/documentation/docs/proptest/shrinking.md
+++ b/documentation/docs/proptest/shrinking.md
@@ -70,6 +70,20 @@ Arb.positiveInt().checkAll(PropTestConfig(shrinkingMode = ShrinkingMode.Unbounde
 }
 ```
 
+If you want to configure the behaviour globally, then either modify the value of `PropertyTesting.defaultShrinkingMode` or use the following system properties as in this example:
+
+```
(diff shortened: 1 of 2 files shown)


JordanLongstaff:
Any chance I can poke the owners to review this soon? I'd treat this as something to include ASAP, like in the next release.

My use case is that I am using Pitest with a lot of property-testing-enabled tests, and Pitest's mutations mean that the unit tests will fail and then shrinking will occur. I suspect the shrinking is making so many of my Pitest runs time out instead of being killed properly. I need to turn shrinking off globally to make sure the Pitest results are accurate, because I'm not willing to set a super-high timeout factor and wait eons for it to finish.

LeoColman:
This is definitely non-critical for next release as you can already control it via global configuration.

That said, LGTM :)

LeoColman (approved):

PR #4341, opened 2024-09-15
fix: Improve LocalDate generation

### LINKED ISSUE #3877: Arb.localDate(LocalDate.MIN, LocalDate.MAX) fails

LeoColman:
Completes #3878
Fixes #3877

### DIFF
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
@@ -50,7 +50,6 @@ fun Arb.Companion.localDate() = Arb.Companion.localDate(LocalDate.of(1970, 1, 1)
  * It includes special edge cases for testing purposes:
  * - If the range includes any leap years, the date **February 29th** of the first such leap year is included.
  * - If the range includes any century years (years divisible by 100), the date **January 1st** of the first such century year is included.
- * - If the range includes any millennium years (years divisible by 1000), the date **January 1st** of the first such millennium year is included.
  *
  * These edge cases are added to increase the likelihood of covering date-related boundary conditions in tests.
  *
@@ -75,9 +74,6 @@ fun Arb.Companion.localDate(
 
    val centuryYear = (minDate.year..maxDate.year).firstOrNull { it % 100 == 0 }
    if (centuryYear != null) { edgeCases += LocalDate.of(centuryYear, 1, 1) }
-
-   val millenniumYear = (minDate.year..maxDate.year).firstOrNull { it % 1000 == 0 }
-   if (millenniumYear != null) { edgeCases += LocalDate.of(millenniumYear, 1, 1) }
 
    edgeCases.removeAll { it !in minDate..maxDate }
 
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
@@ -69,13 +69,11 @@ fun Arb.Companion.localDate(
 
    val edgeCases = mutableListOf(minDate, maxDate)
 
-   val leapYear = (minDate.year..maxDate.year).firstOrNull { isLeap(it.toLong()) }
+   val leapYear = (minDate.year..maxDate.year).firstOrNull { isLeap(it.toLong()) && LocalDate.of(it, 2, 29) in minDate..maxDate }
    if (leapYear != null) { edgeCases += LocalDate.of(leapYear, 2, 29) }
 
-   val centuryYear = (minDate.year..maxDate.year).firstOrNull { it % 100 == 0 }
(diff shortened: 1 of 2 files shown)


kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
```diff
@@ -42,31 +43,47 @@ fun Arb.Companion.period(maxYear: Int = 10): Arb<Period> = arbitrary(listOf(Peri
...
-   minDate == maxDate -> Arb.constant(minDate)
-   else -> {
-      val leapYears = (minDate.year..maxDate.year).filter { isLeap(it.toLong()) }
+): Arb<LocalDate> {
+   require(minDate <= maxDate) { "minDate must be before or equal to maxDate" }
+   if (minDate == maxDate) return Arb.constant(minDate)
+
+   val edgeCases = mutableListOf(minDate, maxDate)
+
+   val leapYear = (minDate.year..maxDate.year).firstOrNull { isLeap(it.toLong()) }
```
AlexCue987: Let's modify it to 
`val leapYear = (minDate.year..maxDate.year).firstOrNull { isLeap(it.toLong()) && LocalDate.of(it, 2, 29) >= minDate }`
so that we don't skip this edge case if, for instance, `minDate=LocalDate.of(2024, 3, 1)`

Do we need a unit test for this scenario?

The same goes for two other edge cases - we need to make sure edge cases are inside the range, right in the filter

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
```diff
@@ -42,31 +43,47 @@ fun Arb.Companion.period(maxYear: Int = 10): Arb<Period> = arbitrary(listOf(Peri
...
+   if (leapYear != null) { edgeCases += LocalDate.of(leapYear, 2, 29) }
+
+   val centuryYear = (minDate.year..maxDate.year).firstOrNull { it % 100 == 0 }
```
AlexCue987: Do we need unit tests to verify that these edge cases are included:
* centuryYear
* millenniumYear
* leapYear

AlexCue987 (changes requested):

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
LeoColman: This specific scenario is covered out by having the `removeAll` clause.

>  so that we don't skip this edge case if, for instance, minDate=LocalDate.of(2024, 3, 1)

I don't get this example, as 2024-02-29 is indeed less than 2024-03-01. Could you clarify?

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
LeoColman: Good idea.

We have one for leap year, but none for century and millennium. I'll include them

LeoColman: Done. I've also removed the millennium edge cases altogether.

They don't add much value considering we already created the century edge case which will most of the times overlap with the millennium unless under very specific scenarios the first millenium is different from the first century in the date range.

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/dates.kt
AlexCue987: Suppose the range is as follows:
```kotlin
 val minDate: LocalDate = LocalDate.of(2024, 3, 1),
  val maxDate: LocalDate = LocalDate.of(2030, 12, 31)
```
the way I've grokked the PR, this means that with the code as is `leapYear == 2024`, and later `2024-02-29` is filtered out by `edgeCases.removeAll { it !in minDate..maxDate }`.

This is why I'm suggesting  a change that will compute `leapYear == 2028` and include `2028-02-29` as the edge case. Does it make sense to you?

AlexCue987: Likewise, if the range is "2000-01-02".."2100-01-02", the centureYear edge case of "2100-01-01" is in the range., but "2000-01-01" is not. Should we provide "2100-01-01"? Please correct me if I am wrong.

LeoColman: Ah, I see.
That's crystal clear now! I'll adjust the code

LeoColman: Done

AlexCue987 (approved):

Issue #4306, opened 2024-08-29
`shouldContainAll` -> is the output of `expected` and `but was`  incorrectly swapped?


mrclrchtr:
Kotest version `5.9.1`

Is the output of `expected` and `but was` in the `shouldContainAll` check incorrectly swapped?

```kotlin
priceSettings.priceModifiersWoodType shouldContainAll priceSettingsDto.priceModifiersWoodType.map {
        it.toModel()
      }
```

```
Collection should contain all of [PriceModifierWoodType(value=10.0099999999999997868371792719699442386627197265625, woodType=ASH), PriceModifierWoodType(value=20.00, woodType=OAK)] but was missing [PriceModifierWoodType(value=10.0099999999999997868371792719699442386627197265625, woodType=ASH)]Possible matches:
java.lang.AssertionError: Collection should contain all of [PriceModifierWoodType(value=10.0099999999999997868371792719699442386627197265625, woodType=ASH), PriceModifierWoodType(value=20.00, woodType=OAK)] but was missing [PriceModifierWoodType(value=10.0099999999999997868371792719699442386627197265625, woodType=ASH)]Possible matches:
 expected: PriceModifierWoodType(value=10.01, woodType=ASH),
  but was: PriceModifierWoodType(value=10.0099999999999997868371792719699442386627197265625, woodType=ASH),
```

the `value=10.0099999999..` is in `priceSettingsDto`. The `priceSettings` is the one, which is tested.

If it is a mistake I am willing to look for the bug and create a PR with a fix.

jvalduvieco:
Same happens with `shouldContain`

AlexCue987:
@mrclrchtr I concur, the order is wrong. Thank you for calling it out. The following test reproduces the issue using standard test data:

```kotest
class ReproduceIssue: StringSpec() {
   init {
       "reproduce wrong order" {
          listOf(sweetGreenPear, sweetGreenApple) shouldContainAll listOf(sweetRedApple)
       }
   }
}

Collection should contain all of [Fruit(name=apple, color=red, taste=sweet)] but was missing [Fruit(name=apple, color=red, taste=sweet)]Possible matches:
 expected: Fruit(name=apple, color=green, taste=sweet),
  but was: Fruit(name=apple, color=red, taste=sweet),
  The following fields did not match:
    "color" expected: <"green">, but was: <"red">
```

Of course it should be the other way around. Good luck fixing this thing.

AlexCue987:
[...]

PR #4344, opened 2024-09-17
Alex/fix should contain all  

AlexCue987:
Why: For partial matches of similar elements, in lots of places `actual` and `expected` are swapped. This was called out in https://github.com/kotest/kotest/issues/4306 but the issue is in multiple other places. Fixing this incorrect order across the board.


AlexCue987:
The failed test is unrelated: Test maintain performance fixed by https://github.com/kotest/kotest/issues/4016[jvm] FAILED

sksamuel (approved):

PR #4359, opened 2024-09-21
Add a few test cases to ArrayEqualsTest for completeness

okarmazin:
I randomly opened this test and saw a few missing primitive arrays from the test cases.

Also renamed the class to better reflect its contents.


LeoColman (approved):
Nice catch :D

Issue #4354, opened 2024-09-20
Collection matchers should work on primitive arrays


okarmazin:
A picture is worth a 1000 words they say.

<img width="480" alt="image" src="https://github.com/user-attachments/assets/ba899b14-7ee5-4414-9100-b0ba7b080fe2">

There should be no problem implementing it the same way the current matchers on Array work - through `.asList()`

- [x] `bein`
- [x] `bounds` #4378
- [x] `contain`
- [x] `containAll`
- [x] `containExactly`
- [x] `decreasing`
- [x] `duplicates` #4369 
- [x] `empty` #4363
- [x] `increasing`
- [x] `inorder`
- [x] `larger`
- [x] `smaller`
- [x] `sorted`
- [x] `startwith`
- [x] `unique` #4375 

In general a collection assertion has to have these overloads to cover all "reasonable" receivers:

1) primitive arrays
2) `Array<T>`
3) `Iterable<T>` (this covers all Collections and most Ranges)
3) `Sequence<T>` (Sequence does not implement `Iterable`, but can give you an iterator - cheaply representable as `Iterable`. Big opportunity for reusing the Iterable Matcher implementation under the hood.)
4) If appropriate, Map<K, T> - if the assertion makes sense to treat Map as a collection of Entries. Otherwise, Map doesn't implement Iterable on its own.

All of these can share a single `Matcher<Iterable<T>>` implementation since they are either Iterables or can be cheaply represented as an Iterable without big allocations (`Array.asList`, `Sequence.asIterable`).

Where appropriate and whenever the `Matcher<Iterable<T>>` implementation works with container size, it will check the underlying type to see if it is a `Collection` and use the known `Collection.size` property to prevent O(N) `Iterable.count()`.

It will also guard against infinite Sequences and Iterables where appropriate. (example: https://github.com/kotest/kotest/pull/4365)

AlexCue987:
This is an interesting suggestion - would you like to take care of it?

okarmazin:
Sure, I can take a stab at it.

sksamuel:
All completed and incuded in 6.1.8

PR #4363, opened 2024-09-22
Modernize shouldBeEmpty and shouldNotBeEmpty collection assertions

okarmazin:
- Add assertions for all primitive arrays (#4354)
- Matcher now prints a more precise collection type.
- shouldBeEmpty failures now print size of the failed collection, if size is well known.
- Out-of-place Sequence tests were moved to the appropriate file. While they were testing the same operation (isEmpty), Sequence matchers live in their own separate package. The tests were relocated to that same package.
- Iterable.shouldBeEmpty no longer deadlocks on infinite iterables.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/empty.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/empty.kt
@@ -117,29 +117,29 @@ fun <T> Array<T>?.shouldNotBeEmpty(): Array<T> {
 
 fun <T, C : Collection<T>> C?.shouldBeEmpty(): C {
    if (this == null) fail("Collection")
-   this should beEmpty("Collection")
+   this should beEmpty(this.containerName())
    return this
 }
 
 fun <T, C : Collection<T>> C?.shouldNotBeEmpty(): C {
    if (this == null) fail("Collection")
-   this shouldNot beEmpty("Collection")
+   this shouldNot beEmpty(this.containerName())
    return this
 }
 
 fun <T, I : Iterable<T>> I?.shouldBeEmpty(): I {
    if (this == null) fail("Iterable")
-   toList() should beEmpty(this.matcherName())
+   toList() should beEmpty(this.containerName())
    return this
 }
 
 fun <T, I : Iterable<T>> I?.shouldNotBeEmpty(): I {
    if (this == null) fail("Iterable")
-   toList() shouldNot beEmpty(this.matcherName())
+   toList() shouldNot beEmpty(this.containerName())
    return this
 }
 
-fun <T> beEmpty(): Matcher<Collection<T>> = beEmpty("Collection")
+fun <T> beEmpty(): Matcher<Iterable<T>> = beEmpty("Iterable")
 
 fun <T> beEmptyArray(): Matcher<Array<T>> = object : Matcher<Array<T>> {
    override fun test(value: Array<T>): MatcherResult = MatcherResult(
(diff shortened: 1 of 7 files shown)


okarmazin:
Hey, guys, as part of the proposed addition of primitive array assertions (#4354) I took the liberty of reorganizing the assertions, adding more tests, deprecating obsolete matcher factory functions, changing the matcher messages.

Ultimately I would like to remove the old (now deprecated at level ERROR) matcher factories, but unfortunately they are a part of the public API.

How do you feel about removing them?

I realize these are breaking changes. However, this is the type of comprehensive cleanup I would like to do for the collection assertions. Not just adding new stuff.

The Kotest codebase is showing signs of bitrot which I'd like to help manage with these more comprehensive collection assertion changes. I don't want to just add more tech debt 🙂.

I would implement a similar modernization for the other assertions as well. But I will only continue if I receive a permission to make the breaking changes exemplified by this PR.

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/shrinking/FloatShrinkerTest.kt
```diff
@@ -21,7 +20,7 @@ class FloatShrinkerTest : FunSpec() {
          ).map(::row).toTypedArray()
 
          forAll(*values) { value ->
-            FloatShrinker.shrink(value) should beEmpty()
+            FloatShrinker.shrink(value).shouldBeEmpty()
```
AlexCue987: `should beEmpty()` and `shouldNot beEmpty()` should be supported, it's part of commonly used style, and most assertions do support both `shouldXxx()` and `should Xxx()`. Is there a reason this PR makes `beEmpty` private? I'd advocate for reverting that change, so that this test, and presumably many others in other code bases, can be kept as is.

AlexCue987 (changes requested):
Overall a very nice addition of features, but should revert some breaking changes.

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/shrinking/FloatShrinkerTest.kt
okarmazin: Okay. In that case I'll revert this as unrelated I'll add `beEmpty` matcher factories for all the newly added arrays so that matcher-style and assertion-style are on par.

okarmazin:
@AlexCue987 @LeoColman 

Hello, here are further changes I'd like to make as part of this PR:

~1) [Breaking] `Map.shouldBeEmpty` is needlessly implemented in its own special package. `collections.maps`, but Map is already an `Iterable`. No need for this duplication. I would like to remove the specialization from the public API, relying instead on `Iterable.shouldBeEmpty`. I see this as a positive change that declutters Kotest's public API and improves maintainability. Indeed, I neve even knew this specialization existed before I added some test cases for Map, and was surprised what assertion was resolved.
**Users will have to replace import `io.kotest.matchers.maps.shouldBeEmpty` with `io.kotest.matchers.collections.shouldBeEmpty`**~
Edit: I was mistaken, Map doesn't actually implement Iterable.

2) I would like to remove the `Collection.shouldBeEmpty` overload from public API since we already have `Iterable.shouldBeEmpty` and Collections are Iterables. As far as matcher messages go, my new implementation of `Iterable.shouldBeEmpty` inspects the type and customizes error messages to make them more precise. That makes `Collection.shouldBeEmpty` obsolete.

What do you think?

AlexCue987:
> @AlexCue987 @LeoColman
> 
> Hello, here are further changes I'd like to make as part of this PR:
> 
> 1. [Breaking] `Map.shouldBeEmpty` is needlessly implemented in its own special package. `collections.maps`, but Map is already an `Iterable`. No need for this duplication. I would like to remove the specialization from the public API, relying instead on `Iterable.shouldBeEmpty`. I see this as a positive change that declutters Kotest's public API and improves maintainability. Indeed, I neve even knew this specialization existed before I added some test cases for Map, and was surprised what assertion was resolved.
>    **Users will have to replace import `io.kotest.matchers.maps.shouldBeEmpty` with `io.kotest.matchers.collections.shouldBeEmpty`**
> 2. I would like to remove the `Collection.shouldBeEmpty` overload from public API since we already have `Iterable.shouldBeEmpty` and Collections are Iterables. As far as matcher messages go, my new implementation of `Iterable.shouldBeEmpty` inspects the type and customizes error messages to make them more precise. That makes `Collection.shouldBeEmpty` obsolete.
> 
> What do you think?

I concur on both suggestions, but let's not do that in this PR. Instead, let's implement each one separately, in its own small atomic PR.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/empty.kt
```diff
@@ -7,50 +7,140 @@ import io.kotest.matchers.invokeMatcher
...
+   if (this == null) fail("LongArray")
+   asList() shouldNot beEmpty("LongArray")
+   return this
+}
+
+fun FloatArray?.shouldBeEmpty(): FloatArray {
+   if (this == null) fail("FloatArray")
+   asList() should beEmpty("FloatArray")
+   return this
+}
```
AlexCue987, line 86: these functions look very similar. have you explored the possibility of implementing the logic once via a generic with reified type?

okarmazin: I don't think there is any reasonable way to change this since `FloatArray != Array<Float>`. There is no common supertype for primitive arrays. If you wanted a single generic function `shouldBeEmpty` that works across all primitive arrays, you would have to make it an extension on any `T` - and that is not desirable IMO.

okarmazin: Right now I believe a reasonable balance has been struck - all significant logic is extracted to a single matcher and the assertions which use it are organized in a way that can easily be checked by human eye for correctness.

okarmazin:
Following up on https://github.com/kotest/kotest/pull/4363#issuecomment-2368174353, I am marking this as ready. Those changes will be added in separate PRs.

AlexCue987 (approved):

PR #4365, opened 2024-09-23
Generate at most one element in sequence isEmpty matcher

### LINKED ISSUE #4364: Sequence.shouldBeEmpty needlessly generates the entire sequence
`Sequence.isEmpty` is implemented as follows: `value.count() == 0`. This will deadlock on infinite sequences (eventually throwing integer overflow) or even on sequences with heavy generator computations.

It should be implemented as `!iterator().hasNext()`.

okarmazin:
Closes #4364 

The previous implementation used `Sequence.count()` which generates the entire sequence. Checking if a Sequence is empty only needs the first element to determine whether it is in fact empty.

Also ran autoformatter, so there are several whitespace adjustments

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/sequences/SequenceMatchersTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/sequences/SequenceMatchersTest.kt
@@ -4,11 +4,17 @@ import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.StringSpec
 import io.kotest.matchers.sequences.shouldContainExactly
 import io.kotest.matchers.sequences.shouldHaveSingleElement
+import io.kotest.matchers.sequences.shouldNotBeEmpty
 import io.kotest.matchers.sequences.shouldNotContainExactly
 import io.kotest.matchers.sequences.shouldNotHaveSingleElement
 import io.kotest.matchers.shouldBe
 
 class SequenceMatchersTest : StringSpec({
+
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/sequences/SequenceMatchersTest.kt
```diff
@@ -4,12 +4,20 @@ import io.kotest.assertions.throwables.shouldThrow
...
 import io.kotest.matchers.sequences.shouldNotContainExactly
 import io.kotest.matchers.sequences.shouldNotHaveSingleElement
 import io.kotest.matchers.shouldBe
+import kotlinx.coroutines.isActive
+import kotlin.time.Duration.Companion.milliseconds
 
 class SequenceMatchersTest : StringSpec({
 
+   "beEmpty does not generate the entire sequence".config(timeout = 500.milliseconds) {
+      val infinite = sequence { while (isActive) yield(1) }
```
AlexCue987: Should this test ever fail, it would take 500 ms to do so. The following fails faster:
```kotlin
val canConsumeOneElement = sequence { yield(1); throw Exception("should not get second element") }
```
WDYT

AlexCue987 (approved):
Good idea, suggested minor improvement

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/sequences/SequenceMatchersTest.kt
okarmazin: Changed as suggested

Issue #4362, opened 2024-09-22
Adopt uniform wording and textual representation of ranges/intervals


okarmazin:
Matcher messages and KDocs containing ranges are all over the place. Round parens are used for representing closed intervals - should be square brackets. The word "inclusive" is not used consistently - either let it be everywhere, or nowhere. Sometimes intervals are spelled out, other times they use mathematical notation.

Almost everywhere the intervals are closed - except for JVM datetime comparisons, which are open. But the matcher functions have identical names.

Here are a few examples for `shouldBeBetween`:

- Verifies that the given integer is between a and b inclusive.
- "$value should be between ($a, $b)"
- "$value should be between ($lower, $upper) inclusive"
- "$value is between ($from, $to)"
- Asserts that this is between [a] and [b]

`shouldBeIn(range)` is similar.

--------------------

The wording of both Kdoc and matcher messages should be uniform and predictable.
With the understanding that changing matcher messages is a potentially breaking change, here's one possible variant to start the discussion:

Use the proper mathematical notation everywhere, reinforced with words "inclusive" and "exclusive" after the interval. Some people may not know what the different brackets mean.
- [Kdoc] Verifies that the given integer is between `[a, b]` (inclusive, inclusive)
- [Matcher] "$value should be between [$a, $b] (inclusive, inclusive)"

LeoColman:
Was this implemented in #4369 ?

PR #4369, opened 2024-09-24
Modernize shouldContainDuplicates collection assertions

okarmazin:
This PR is a continuation of work on #4354.

- Adds should(Not)ContainDuplicates assertions for primitive arrays
- Matcher messages report a more precise container type
- Array.shouldContainDuplicates now returns the array for chaining. Not returning it was an oversight.

The diff looks large, but the bulk of it are test cases.


kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
```diff
@@ -463,10 +463,26 @@ public final class io/kotest/matchers/collections/DuplicatesKt {
 	public static final fun containDuplicates ()Lio/kotest/matchers/Matcher;
 	public static final fun shouldContainDuplicates (Ljava/lang/Iterable;)Ljava/lang/Iterable;
 	public static final fun shouldContainDuplicates (Ljava/util/Collection;)Ljava/util/Collection;
-	public static final fun shouldContainDuplicates ([Ljava/lang/Object;)V
```
okarmazin, line 480: `Array<T>.shouldContainDuplicates`

This was changed to return the Array.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/empty.kt
```diff
@@ -172,17 +172,6 @@ private fun <T> beEmpty(name: String?): Matcher<Iterable<T>> = object : Matcher<
          { "$name should not be empty" }
       )
    }
-
-   private fun Iterable<*>.containerName(): String {
```
okarmazin, line 176: Was extracted as a shared helper

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/CollectionMatchersTest.kt
```diff
@@ -260,21 +257,6 @@ class CollectionMatchersTest : WordSpec() {
...
-            listOf(1, 2, 3, 3) should containDuplicates()
-            listOf(1, 2, 3, 4) shouldNot containDuplicates()
-            listOf(1, 2, 3, 3).shouldContainDuplicates()
-            listOf(1, 2, 3, 4).shouldNotContainDuplicates()
-         }
-
-         "print duplicates in message" {
-            shouldThrowAny {
-               listOf(1, 2, 3, 4, 2, 1) shouldNot containDuplicates()
-            }.shouldHaveMessage("Collection should not contain duplicates, but has some: [1, 2]")
```
okarmazin, line 274: These tests were removed because they are duplicates (See DuplicatesTest.kt)

okarmazin: Duplication tests were themselves duplicated. I don't know why that's so funny to me :D

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/duplicates.kt
```diff
@@ -6,48 +6,130 @@ import io.kotest.matchers.MatcherResult
...
    return this
 }
 
 fun <T> Array<T>.shouldNotContainDuplicates(): Array<T> {
-   asList().shouldNotContainDuplicates()
+   asList() shouldNot containDuplicates("Array")
+   return this
+}
+
+fun <T, C : Collection<T>> C.shouldContainDuplicates(): C {
```
LeoColman, line 99: If we have Iterable, do we need the collection overflow still?

okarmazin: No, it is scheduled for removal in a followup PR. Alex pointed out that these removals should be added as separate PRs, which I'll do. In fact, the removal of this Collection overload is already waiting for this PR to be merged - #4371

okarmazin:
If there are no more questions about the changes, I'm kindly asking for a merge. Other PRs are waiting for this.

sksamuel:
I'll review all outstanding prs this weekend.

okarmazin:
To provide more context to these recent "Modernize XYZ" PRs:

I originally just wanted to add some assertions for primitive arrays, because I wanted them for my own projects where I use Kotest and couldn't perform assertions on `ByteArrays` (original suggestion: #4354).

After opening Kotest in IDEA to add these, I started noticing a few other cracks:
- Mismatched assertion receivers: some assertions are only defined for a subset of the types for which they should be "logically" defined. Example: #4377
- A lot of missing KDoc on public API
- Misplaced assertions. Some `Sequence` assertions are incorrectly placed in `collections` package, but Sequences have their own package `sequence` #4374 
- For Collection assertions specifically, there sometimes are both `Collection.shouldXXX` and `Iterable.shouldXXX` versions, which only serve to pollute IDE autocompletion. `Collection` interface implements `Iterable`, so the `Collection` version shouldn't exist.
- Duplicated implementations between `Sequence` and `Iterable` versions of the same assertion. It is understandable, since Sequences don't implement Iterable. But they can be cheaply interpreted as an Iterable via their `.asIterable` method, which just wraps the Sequence. This presents a good opportunity to delegate the implementation to the Iterable matcher. #4372 
- Needless List allocations - Iterable and Sequence assertions tend to convert the receiver to List and work with that List. This is not necessary, most assertions can work with a single pass over the Iterable, sometimes not even over all elements (#4365). It is understood that some assertions may need to convert to a List. But those single-iteration ones don't. 
- Some assertions are susceptible to infinite Sequences and Iterables - causing either OOM (if they try to convert an infinite stream to List, see previous point) or index overflow (if they try to compute `.count()` of an infinite stream #4365.
- Inconsistent wording of certain concepts. Example: #4362
- Some assertions could be removed due to just being inversions of existing ones (up for debate) #4373 

For these reasons I decided to make a more substantial contribution to Kotest than just adding a few array assertions. I intend to review assertion implementations and fix the aforementioned problems in order to reduce the technical debt accrued over the years. Obviously, that cannot be done in a single PR or a single sitting. And some design choices (https://github.com/kotest/kotest/pull/4378#discussion_r1779928977) will have to be discussed and decided separately. And written as rules for future implementers. These "Modernize XYZ" PRs are just the beginning of what I intend to contribute.

All that said, I would kindly ask the following:
- Do not block a merge because you see a duplicated `Collection.shouldXXX` and `Iterable.shouldXXX`. Make a comment, but rest assured these duplications will be removed shortly. The chances are that the followup PR is already waiting in line! #4378 #4379 
- Keep in mind that these PRs are small parts of a bigger strategy. Be very thorough with your review, point out every little thing, but an "obviously missing" change is most likely being left for a followup PR, a merge shouldn't be blocked over it.

LeoColman (approved):

sksamuel:
Need to re-run api dump

PR #4367, opened 2024-09-23
Remove Collection.shouldBeEmpty overloads

okarmazin:
Closes #4366 

Followup for #4363 - the `Collection.shouldBeEmpty` overload is not necessary since there's already `Iterable.shouldBeEmpty` version, and Collection implements Iterable.

This cleans up Kotest's public API.

See also: Kotlin standard library function `forEach` - it is declared once for Iterable without the need for additional overloads for concrete Iterable implementors.

PR #4370, opened 2024-09-24
Report indexes of duplicated elements in failed shouldNotContainDuplicates messages

okarmazin:
This PR builds upon #4369 and adds the indexes of violating elements to the assertion messages.

`listOf(1, 2, 3, null, null, 3, 2).shouldNotContainDuplicates()`

Before:
```
List should not contain duplicates, but has some: [2, 3, <null>]
```

After:
```
List should not contain duplicates, but has:
2 at indexes: [1, 6]
3 at indexes: [2, 5]
<null> at indexes: [3, 4]
```

Affected matchers/assertions:
1) All `shouldNotContainDuplicates`
2) `Sequence.shouldBeUnique` - this assertion was and remains to be implemented via "contains duplicates"

PR #4371, opened 2024-09-24
Remove Collection.shouldContainDuplicates

okarmazin:
Cleans up Kotest's public API. All Collections are Iterables, so having both overloads in place only pollutes IDE autocompletion.

See also https://github.com/kotest/kotest/pull/4367

Reused example:

Before:
isempty-before

After:
isempty-after


okarmazin:
More context: https://github.com/kotest/kotest/pull/4369#issuecomment-2381280124

okarmazin:
Waiting for https://github.com/kotest/kotest/pull/4370, will rebase after that gets merged

sksamuel (approved):

PR #4372, opened 2024-09-24
Reuse Iterable isEmpty implementation for Sequences

okarmazin:
This PR deduplicates logic by delegating the Sequence check to the Iterable matcher.


okarmazin:
More context: https://github.com/kotest/kotest/pull/4369#issuecomment-2381280124

sksamuel (approved):

sksamuel:
requires api dump

okarmazin:
Forgot to apply a change, now fixed. I even found a better way of implementing this - utilizing an import alias which allows keeping the Iterable matcher as an anonymous object instead of a class.

PR #4377, opened 2024-09-25
Deduplicate shouldBeBetween

### LINKED ISSUE #4376: Deduplicate shouldBeBetween assertions
`shouldBeBetween` is fundamentally an assertion on any `Comparable`. Currently there is a "random" assortment of `shouldBeBetween` on some primitive types, not even all of them. They are also scattered across different files as they were added over the years, with different failure messages.

I propose implementing a single `<T : Comparable<T>> T.shouldBeBetween(lower: T, upper: T)` and removing all the other varieties.

okarmazin:
Closes #4376 

This PR implements `shouldBeBetween` for any `Comparable`.

Previously this assertion was implemented individually for `Byte, UByte, Shor, UShort, Char, Int, Long, ULong`. All these versions were deprecated.


AlexCue987:
regarding "users will need to change imports to io.kotest.matchers.comparables.shouldBeBetween" - is there a reason we did not deprecate those functions instead of deleting them outright? IMO that users should have a fair amount of time to change.

okarmazin:
I didn't find any deprecation policy in CONTRIBUTING.md, so I just made the change.

I can add these functions back and mark as deprecated, if a deprecation cycle is the preferred way to go.

AlexCue987:
Please undelete and deprecate all the functions. Maybe we should add to `CONTRIBUTING.md`, as you are suggesting, in another PR of course. Let's think about it a bit.

okarmazin:

okarmazin:
PR reworked and ready for another review:
I deprecated the existing matchers instead of removing them. Replacement imports were added for convenience.
All previous standalone implementations now simply delegate to the single Comparable implementation. Also dropped the Markdown reformat commit in the interest of clarity.

LeoColman (approved):
Good job! We will eventually proceed on the deprecation cycles for the remaining dead code

Issue #3960, opened 2024-04-02
Adapt assertion extensions for Collections and Iterables to use bounded types


hanszt:
Some of the assertions on collections now look like the following:

```kt
infix fun <T> Collection<T>.shouldHaveSize(size: Int): Collection<T> {
...
return this
}
```

I suggest to change this to 

```kt
infix fun <T, C : Collection<T>> C.shouldHaveSize(size: Int): C {
   ...
   return this
}
```

The benefit of this is that you then can do the following:

```kt
val (_, _, third) = listOf(1, 2, 3) shouldHaveSize 3
third.shouldBe(3)
```

---

A second example:

When

```kt
fun <T> Iterable<T>.shouldNotContainDuplicates(): Iterable<T> {
   ...
   return this
}

```

is changed to

```kt
fun <T, I : Iterable<T>> I.shouldNotContainDuplicates(): I {
   ...
   return this
}

```

This will then allow you to write something like the following in your tests.

```kt
class Game(val name: String, players: Iterable<String>) : Iterable<String> by players

val game = Game("Risk", listOf("p1", "p2", "p3", "p4")).shouldNotContainDuplicates()
game.name shouldBe "Risk"
```

---

These are only a few examples of functions that could be improved in this way

---

hanszt:
See also [Pull request 3961](https://github.com/kotest/kotest/pull/3961)

AlexCue987:
Personally I feel the following code would be confusing to me, as well as to most developers:
`val game = Game("Risk", listOf("p1", "p2", "p3", "p4")).shouldNotContainDuplicates()`
should I ever encounter that, I'd rather refactor it like this:
```
val game = Game("Risk", listOf("p1", "p2", "p3", "p4"))
game. players.shouldNotContainDuplicates()
```
So I don't perceive this change as a big improvement.

hanszt:
I see what you mean. 

But as a more general guidline: Isn't it more desirable to have a return type that is more specific?

Something on the line of this post: https://stackoverflow.com/questions/3434367/is-it-better-to-return-the-most-specific-or-most-general-type-from-an-action-met

This makes the use of these functions more flexible don't you think? Especially in light of the first example above?
I'm curious what you think

PR #4378, opened 2024-09-28
Modernize shouldHaveUpperBound, shouldHaveLowerBound

okarmazin:
Continuation of work on #4354.

- Adds `shouldHaveUpper/LowerBound` for primitive arrays except BooleanArray. BooleanArray was left out in the interest of reducing API bloat. Booleans are better served with `shouldContain true/false`.
- Matcher messages report a more precise container type
- Adds more test cases for empty / single element containers
- Removes needless allocations: Iterables were being converted to lists, then checked iterated. Now they are just iterated once.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
@@ -138,18 +138,108 @@ private fun Iterable<*>.containerName(): String {
    }
 }
 
-infix fun <T : Comparable<T>> Array<T>.shouldHaveLowerBound(t: T): Array<T> {
-   asList() should haveLowerBound(t, "Array")
+
+// BooleanArray left out in the interest of reducing API bloat.
+// As Boolean only has 2 values, it is more natural to use
+// "shouldContain true / shouldContain false"-type assertions
+
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun ByteArray.shouldHaveLowerBound(value: Byte): ByteArray {
+   asList() should haveLowerBound(value, "ByteArray")
    return this
 }
 
-infix fun <T : Comparable<T>, I : Iterable<T>> I.shouldHaveLowerBound(t: T): I {
-   toList() should haveLowerBound(t, null)
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun ShortArray.shouldHaveLowerBound(value: Short): ShortArray {
+   asList() should haveLowerBound(value, "ShortArray")
    return this
 }
 
-infix fun <T : Comparable<T>, C : Collection<T>> C.shouldHaveLowerBound(t: T): C {
-   this should haveLowerBound(t, null)
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun CharArray.shouldHaveLowerBound(value: Char): CharArray {
+   asList() should haveLowerBound(value, "CharArray")
+   return this
+}
+
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun IntArray.shouldHaveLowerBound(value: Int): IntArray {
+   asList() should haveLowerBound(value, "IntArray")
+   return this
+}
+
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun LongArray.shouldHaveLowerBound(value: Long): LongArray {
+   asList() should haveLowerBound(value, "LongArray")
+   return this
+}
+
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun FloatArray.shouldHaveLowerBound(value: Float): FloatArray {
+   asList() should haveLowerBound(value, "FloatArray")
+   return this
+}
+
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun DoubleArray.shouldHaveLowerBound(value: Double): DoubleArray {
+   asList() should haveLowerBound(value, "DoubleArray")
+   return this
+}
+
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun <T : Comparable<T>> Array<T>.shouldHaveLowerBound(value: T): Array<T> {
+   asList() should haveLowerBound(value, "Array")
+   return this
+}
+
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun <T : Comparable<T>, C : Collection<T>> C.shouldHaveLowerBound(value: T): C {
+   this should haveLowerBound(value, null)
+   return this
+}
+
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun <T : Comparable<T>, I : Iterable<T>> I.shouldHaveLowerBound(value: T): I {
+   this should haveLowerBound(value, null)
    return this
 }
 
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
@@ -5,22 +5,116 @@ import io.kotest.matchers.Matcher
 import io.kotest.matchers.MatcherResult
 import io.kotest.matchers.should
 
-infix fun <T : Comparable<T>> Array<T>.shouldHaveUpperBound(t: T): Array<T> {
-   asList() should haveUpperBound(t, "Array")
+// BooleanArray left out in the interest of reducing API bloat.
+// As Boolean only has 2 values, it is more natural to use
+// "shouldContain true / shouldContain false"-type assertions
+
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun ByteArray.shouldHaveUpperBound(value: Byte): ByteArray {
+   asList() should haveUpperBound(value, "ByteArray")
    return this
 }
 
-infix fun <T : Comparable<T>, C : Collection<T>> C.shouldHaveUpperBound(t: T): C {
-   this should haveUpperBound(t, null)
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun ShortArray.shouldHaveUpperBound(value: Short): ShortArray {
+   asList() should haveUpperBound(value, "ShortArray")
    return this
 }
 
-infix fun <T : Comparable<T>, I : Iterable<T>> I.shouldHaveUpperBound(t: T): I {
-   this should haveUpperBound(t, null)
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun CharArray.shouldHaveUpperBound(value: Char): CharArray {
+   asList() should haveUpperBound(value, "CharArray")
    return this
 }
 
-fun <T : Comparable<T>, C : Collection<T>> haveUpperBound(t: T): Matcher<C> = haveUpperBound(t, null)
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun IntArray.shouldHaveUpperBound(value: Int): IntArray {
+   asList() should haveUpperBound(value, "IntArray")
+   return this
+}
+
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun LongArray.shouldHaveUpperBound(value: Long): LongArray {
+   asList() should haveUpperBound(value, "LongArray")
+   return this
+}
+
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun FloatArray.shouldHaveUpperBound(value: Float): FloatArray {
+   asList() should haveUpperBound(value, "FloatArray")
+   return this
+}
+
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun DoubleArray.shouldHaveUpperBound(value: Double): DoubleArray {
+   asList() should haveUpperBound(value, "DoubleArray")
+   return this
+}
+
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun <T : Comparable<T>> Array<T>.shouldHaveUpperBound(value: T): Array<T> {
+   asList() should haveUpperBound(value, "Array")
+   return this
+}
+
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun <T : Comparable<T>, C : Collection<T>> C.shouldHaveUpperBound(value: T): C {
+   this should haveUpperBound(value, null)
+   return this
+}
+
+/**
+ * Verifies that all elements are less than or equal to [value].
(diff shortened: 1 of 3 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
```diff
@@ -5,52 +5,253 @@ import io.kotest.matchers.Matcher
...
 }
 
-fun <T : Comparable<T>, C : Collection<T>> haveUpperBound(t: T) = object : Matcher<C> {
-   override fun test(value: C): MatcherResult {
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun <T : Comparable<T>> Array<T>.shouldHaveUpperBound(value: T): Array<T> {
```
LeoColman, line 87: Do we need to have this overload? I think Iterable would catch it, right?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
```diff
@@ -5,52 +5,253 @@ import io.kotest.matchers.Matcher
...
+   asList() should haveUpperBound(value, "Array")
+   return this
+}
+
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun <T : Comparable<T>, C : Collection<T>> C.shouldHaveUpperBound(value: T): C {
```
LeoColman, line 97: Same for this one, Iterable should catch all

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
```diff
@@ -5,52 +5,253 @@ import io.kotest.matchers.Matcher
...
+   this should haveUpperBound(value, null)
+   return this
+}
+
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+fun <T : Comparable<T>, C : Collection<T>> haveUpperBound(value: T): Matcher<C> = haveUpperBound(value, null)
```
LeoColman, line 117: Do we need C here?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
```diff
@@ -5,52 +5,253 @@ import io.kotest.matchers.Matcher
...
+         { "$name should not have upper bound $t" })
+   }
+}
+
+private fun Iterable<*>.containerName(): String {
+   return when (this) {
+      is List -> "List"
+      is Set -> "Set"
+      is Map<*, *> -> "Map"
+      is ClosedRange<*>, is OpenEndRange<*> -> "Range"
```
LeoColman, line 135: Maybe better to split these two for more information

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
```diff
@@ -5,52 +5,253 @@ import io.kotest.matchers.Matcher
...
+/**
+ * Verifies that all elements are less than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+fun <T : Comparable<T>, C : Collection<T>> haveUpperBound(value: T): Matcher<C> = haveUpperBound(value, null)
+
+private fun <T : Comparable<T>, I : Iterable<T>> haveUpperBound(t: T, name: String?): Matcher<I> = object : Matcher<I> {
+   override fun test(value: I): MatcherResult {
+      val name = name ?: value.containerName()
```
LeoColman, line 121: As we are messing with this code already, perhaps we should guard against empty lists.

Empty lists will throw an error if they fail and pass silently otherwise.

We should place something like 'require(value.isNotEmpty())' or something similar. I think we do this in a few other collection matchers

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
```diff
@@ -5,52 +5,253 @@ import io.kotest.matchers.Matcher
...
+   asList() should haveLowerBound(value, "DoubleArray")
+   return this
+}
+
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun <T : Comparable<T>> Array<T>.shouldHaveLowerBound(value: T): Array<T> {
```
LeoColman, line 221: Same questioning from UpperBound: Should we have this overload?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
```diff
@@ -5,52 +5,253 @@ import io.kotest.matchers.Matcher
...
+   asList() should haveLowerBound(value, "Array")
+   return this
+}
+
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun <T : Comparable<T>, C : Collection<T>> C.shouldHaveLowerBound(value: T): C {
```
LeoColman, line 231: Same here

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
```diff
@@ -5,52 +5,253 @@ import io.kotest.matchers.Matcher
...
+   this should haveLowerBound(value, null)
+   return this
+}
+
+/**
+ * Verifies that all elements are greater than or equal to [value].
+ *
+ * Passes if `this` is empty.
+ */
+infix fun <T : Comparable<T>, I : Iterable<T>> I.shouldHaveLowerBound(value: T): I {
```
LeoColman, line 241: Do we actually need to define `I` here?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
```diff
@@ -5,52 +5,253 @@ import io.kotest.matchers.Matcher
...
+   this should haveLowerBound(value, null)
+   return this
+}
+
+fun <T : Comparable<T>, C : Collection<T>> haveLowerBound(t: T): Matcher<C> = haveLowerBound(t, null)
+
+private fun <T : Comparable<T>, I : Iterable<T>> haveLowerBound(t: T, name: String?): Matcher<I> = object : Matcher<I> {
+   override fun test(value: I): MatcherResult {
+      val name = name ?: value.containerName()
```
LeoColman, line 250: same comment here, we could add a neat empty-list guard here

LeoColman (changes requested):
I believe we have some space to improve.
All in all, these are good modernizations :D

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/BoundsTest.kt
```diff
@@ -10,24 +10,225 @@ class BoundsTest : WordSpec() {
...
+            longArrayOf(3) shouldHaveLowerBound 3
+            longArrayOf(1, 2, 3) shouldHaveLowerBound 1
+            longArrayOf(1, 2, 3) shouldHaveLowerBound 0
+
+            floatArrayOf() shouldHaveLowerBound 0.3f
+            floatArrayOf(0.3f) shouldHaveLowerBound 0.3f
+            floatArrayOf(0.1f, 0.2f, 0.3f) shouldHaveLowerBound 0.1f
+            floatArrayOf(0.1f, 0.2f, 0.3f) shouldHaveLowerBound 0f
+
+            doubleArrayOf() shouldHaveLowerBound 0.3
```
LeoColman, line 155: As an example for the aforementioned guard-clauses, an empty double array will not have 0.3 as lowerbound

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
okarmazin: Yes, a followup PR is already waiting - https://github.com/kotest/kotest/pull/4379

As I said before, I was instructed by @AlexCue987 to make these removals into separate PRs.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
okarmazin: > I think Iterable would catch it, right?

No, `Array` does not implement `Iterable` - `Array<MyType>` is just `MyType[]`.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
okarmazin: https://github.com/kotest/kotest/issues/4381

Will resolve separately

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
okarmazin: This is up for debate IMO. If you word `hasUpperBound(T)` as "all elements are less than or equal to T", then an empty set does evaluate to true, because there are no elements to violate the constraint. This is known as a [_vacuous truth_](https://en.wikipedia.org/wiki/Vacuous_truth) -- statement is true, but doesn't really say anything.

I argue current behavior is correct. If you want to check for presence of elements, there is `shouldNotBeEmpty` for that.

See also: [stdlib function `all`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/all.html)

@AlexCue987 @sksamuel What do you think?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
okarmazin: In general a collection assertion has to have these overloads to cover all "reasonable" receivers:

1) primitive arrays
2) `Array<T>`
3) `Iterable<T>` (also covers all Collections and some Ranges)
3) `Sequence<T>` (Sequence does not implement `Iterable`, but can give you an iterator - cheaply representable as `Iterable`. Big opportunity for reusing the Iterable Matcher implementation under the hood.)
4) If appropriate, Map<K, T> - if the assertion makes sense to treat Map as a collection of Entries. Otherwise, Map doesn't implement Iterable on its own.


All of these can share a single `Matcher<Iterable<T>>` implementation since they are either Iterables or can be cheaply represented as an Iterable without big allocations (`Array.asList`, `Sequence.asIterable`).

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/BoundsTest.kt
okarmazin: Yep, I'll close this comment as resolved to keep that discussion in one thread - https://github.com/kotest/kotest/pull/4378#discussion_r1779820867

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
okarmazin: Yep, I'll close this comment as resolved to keep that discussion in one thread - https://github.com/kotest/kotest/pull/4378#discussion_r1779820867

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
okarmazin: Yep, scheduled for already in #4379

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
okarmazin: Someone requested the underlying type to be returned due to having custom Iterable implementations. That way he could preserve the underlying type through assertion chain.

See https://github.com/kotest/kotest/issues/3960

okarmazin: As returning the underlying type is the currently accepted behavior - and was so before this PR - I'll close this as resolved. Please reopen if you think there should be more discussion or if you think this should be changed to return Iterable again. But keep in mind that someone requested this behavior and it was accepted.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
okarmazin: Nope, scheduled for removal #4379

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
okarmazin: > Empty lists will throw an error if they fail and pass silently otherwise.

I apologize, I don't understand this sentence. Could you please reword it?

okarmazin:

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bounds.kt
LeoColman: > I argue current behavior is correct. If you want to check for presence of elements, there is shouldNotBeEmpty for that.

Exactly. If we want to check for empty collections it should happen on `shouldNotBeEmpty`, and not accidentally on `shouldHaveUpperBound`. Therefore, if you're making a check against a collection but it's actually empty, we should prevent such thing to happen

> statement is true, but doesn't really say anything.

So the test would basically do nothing? That's not a very useful test, is it?



https://github.com/kotest/kotest/blob/1360aa5a85e5e44bdf85ef04718ad24019ca47bf/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/ranges/bein.kt#L54

https://github.com/kotest/kotest/blob/1360aa5a85e5e44bdf85ef04718ad24019ca47bf/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/matchers.kt#L133

These are examples in our codebase when we do this kind of guard clause.


> See also: [stdlib function all](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/all.html)

The stdlib function is for general purpose, and not testing/assertion purposes. I think this comparison doesn't apply

okarmazin: I don't have a horse in this race. There's just a decision to be made about which behavior is deemed more in line with the principle of least surprise. 
I argued that I'd like to follow mathematical logic - but I'll happily throw an error on empty container. It just must be consistent everywhere. That's the important bit.

> So the test would basically do nothing? That's not a very useful test, is it?

No, it doesn't "do nothing". It verifies that the statement is true, which it is. An assertion is just a logical statement. Now, is it desirable for `hasLowerBound` to be just `hasLowerBound`, or should it be a conjunction of `isNotEmpty && hasLowerBound`?

okarmazin: It is also important to note that this PR doesn't modify previous behavior. It would be a behavioral change if this proposed `isNotEmpty` condition was added.

If all collection assertions should suddenly start rejecting empty containers, that's a separate issue for a separate PR. I'll open a ticket for it.

LeoColman: > It is also important to note that this PR doesn't modify previous behavior. It would be a behavioral change if this proposed isNotEmpty condition was added.

Fair point, I'll resolve this as we can track it in the new ticket you created

LeoColman (approved):

PR #4385, opened 2024-09-30
Extract shouldBeSortedDescending into a separate file

okarmazin:
Due to the large API surface of the various `shouldBeSorted` assertions I decided to perform their modernization in several sequential PRs.

This PR:
- Splits off `shouldBeSortedDescending` into its own file `sortedDesc.kt`
- Moves several forgotten `sortedDescending` matchers from `CollectionMatchers.kt` into `sortedDesc.kt` (that's the API dump change). All `sortedBy` assertions are now in `sorted.kt` and `sortedDescendingBy` in `sortedDesc.kt`.
- Introduces a standalone test class for sorting matchers, moves the existing tests into it
- Reorders the functions in `sorted.kt` and `sortedDesc.kt`

No logic changes were made in this PR. Code was just moved around in preparation of further changes.


okarmazin:

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/sortedDesc.kt
```diff
@@ -0,0 +1,58 @@
+package io.kotest.matchers.collections
```
AlexCue987, line 1: this all but destroys git history, which is very valuable, so there should be a very strong reason for doing so. what is it?

okarmazin: The reason I moved this into its own file is that there are generally files whose name corresponds to the assertion it contains. With the notable exception of the old `CollectionMatchers.kt` and `matchers.kt` files that contain a mishmash of matchers. 

What I observed so far is that each named file contains the following:
1) A set of assertions - extensions on a container - `shouldXXX`
2) Negations of those assertions - `shouldNotXXX`
3) Matchers implementing the assertion algorithm, utilized by the extensions - `XXX`

It seems to me that the file as a unit of organization contains a _single set_ of these assertions/matchers i.e. a single algorithm.
`Sorted` and `SortedDesc` are currently two separate algorithms with two sets of matchers, not one matcher.

Two sets of matchers to me indicates (according to the observed pattern) that these assertions should live in standalone files, such that each file contains its own respective algorithm and assertion set.

What do you think? Is it a reasonable observation? There have been similar breakouts in the past where an assertion was given its own file.

okarmazin: There are files `containExactly.kt` and `containExactlyInAnyOrder.kt` - two similar-yet-different assertions. `sorted.kt` and `sortedDesc.kt` seems to be of similar kind to me.

okarmazin: Please note I'm also a big fan of good history - if I didn't know that all PRs get squashed, this very change would be a set of smaller changes. First colocation of the matchers, then reordering, then breakout. One small step at a time. But as it gets squashed, there wouldn't be much point to it.

sksamuel (approved):

PR #4391, opened 2024-09-30
submatching-for-endWith (#4390)

AlexCue987:
if `shouldEndWith` fails for a String, try to find a partial match, like this:

```kotlin
               "The quick brown fox jumps over the lazy dog" should endWith("fox jumps over the lazy cat")
            }.message
            message.shouldContainInOrder(
               "Match[0]: expected[0..23] matched actual[16..39]",
               """Match[0]= ----------------++++++++++++++++++++++++---"""
            )
```


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/string/EndWithTest.kt
```diff
@@ -36,6 +38,16 @@ class EndWithTest : FreeSpec() {
...
             }
          }
+         "find submatch in the middle of value" {
+            val message = shouldThrow<AssertionError> {
+               "The quick brown fox jumps over the lazy dog" should endWith("fox jumps over the lazy cat")
+            }.message
+            message.shouldContainInOrder(
+               "Match[0]: expected[0..23] matched actual[16..39]",
+               """Line[0] ="The quick brown fox jumps over the lazy dog"""",
+               """Match[0]= ----------------++++++++++++++++++++++++---"""
```
LeoColman, line 48: I don't understand how this works.
Shouldn't the expected part be something like [16..43] ?

LeoColman: The way it reads right now is
"The string should have position [0..23] matching, but only matched [16..39]"

I feel like it should read
"The string should have position [16..42] matching, but only matched [16..39]"

AlexCue987: > The way it reads right now is "The string should have position [0..23] matching, but only matched [16..39]"
> 
> I feel like it should read "The string should have position [16..42] matching, but only matched [16..39]"

makes sense, I'll change it, thx!

sksamuel (approved):

PR #4396, opened 2024-10-01
Increase flaky performance test timeout on Mac OS

### LINKED ISSUE #4380: Change timeout value for the constantly timeouting test on macOS runner
There's a test called `"maintain performance fixed by https://github.com/kotest/kotest/issues/4016".config(timeout = 2.seconds)`.

This test keeps timeouting on the macOS runner all the time, keeping unrelated PRs from getting merged. Whether the test passes is a matter of luck - it probably depends on the current load of GitHub's macOS runner at the time.

The macOS runner is just slower in general TBH.

I propose that the timeout value be changed to 2.5 - 3 seconds such that the test maintains its function (keeping an eye out on performance regressions) without impacting unrelated PRs.

Kotest's CI pipeline runs for about 1 hour, which makes this false positive failure especially egregious. It wastes hours of time.

okarmazin:
Pursuant to #4380, increases the timeout from 2s to 5s.

Closes #4380


okarmazin:
In case this doesn't yield much effect even after a larger increase, it might be prudent to disable the test on macOS altogether. The other runners will still be able to catch any regressions.

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/CollectionsTest.kt
```diff
@@ -134,7 +133,8 @@ class CollectionsTest : DescribeSpec({
          }
       }
 
-      it("maintain performance fixed by https://github.com/kotest/kotest/issues/4016").config(timeout = 2.seconds) {
+      val timeout = if (System.getProperty("os.name").contains("mac", true)) 3.seconds else 2.seconds
```
AlexCue987: let's make it `5.seconds else 2.seconds`

sksamuel:
> In case this doesn't yield much effect even after a larger increase, it might be prudent to disable the test on macOS altogether. The other runners will still be able to catch any regressions.

Yes we could just disable it on mac if it still doesn't complete.

sksamuel (approved):

AlexCue987 (approved):

PR #4403, opened 2024-10-01
Adjust spec ref interceptor to reduce the size of stack traces

sksamuel:
Trying to split up https://github.com/kotest/kotest/pull/4312 to get it over the line (part 2)


sksamuel:
This is the one that refuses to work properly, even after I remove operator / fun from the interface

sksamuel:
Think I got this now @AlexCue987

PR #4407, opened 2024-10-01
alex/shouldNotContainADigit-details 

AlexCue987:
have `shouldNotContainADigit` print the first found digit
```kotlin
"shouldNotContainADigit should print first digit" {
            shouldThrow<AssertionError> {
               "aww1b2c3" shouldNot containADigit()
            }.message shouldBe "\"aww1b2c3\" should not contain any digits, but contained '1' at index 3"
         }
```


sksamuel (approved):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/string/matchers.kt
```diff
@@ -40,10 +40,11 @@ fun String?.shouldNotContainADigit(): String? {
 }
 
 fun containADigit() = neverNullMatcher<String> { value ->
+   val firstDigit = value.asSequence().withIndex().firstOrNull { it.value in '0'..'9' }
```
LeoColman: I think you can maybe go `indexOfFirst { }` and get the index only?

If you want to use it on the description we could maybe destructure them instead
```suggestion
   val (firstDigitIndex, firstDigit) = value.asSequence().withIndex().firstOrNull { it.value in '0'..'9' }
```

LeoColman: I also think this could benefit from being a "FirstDigitIndex" to being a "FirstNumberIndex"

Let's say

The string "I paid 200 already" != "I paid already"


The error message could be "Contained 200 at index 7", which would have more meaning than "Contained 2 at index 7". I don't know how important this would be, and we would also need to guard against having incredibly big numbers printed to the console, but if you haven't thought of this option yet, here it is :D

AlexCue987: > I also think this could benefit from being a "FirstDigitIndex" to being a "FirstNumberIndex"
> 
> Let's say
> 
> The string "I paid 200 already" != "I paid already"
> 
> The error message could be "Contained 200 at index 7", which would have more meaning than "Contained 2 at index 7". I don't know how important this would be, and we would also need to guard against having incredibly big numbers printed to the console, but if you haven't thought of this option yet, here it is :D

makes sense. I'll improve tomorrow, TY for calling it out!

AlexCue987: > I think you can maybe go `indexOfFirst { }` and get the index only?

done, and the code is cleaner. TY

PR #4425, opened 2024-10-07
Alex/search for partial matches in string should be

AlexCue987:
This follows a Slack discussion with @sksamuel and @Kantis .

When matching strings, search for partial matches, such as:

```kotlin
     val expected = "One quick brown fox jumps over the lazy cat"
      val value ="The quick brown fox jumps over the lazy dog"
      val actual = shouldThrow<AssertionError> { value shouldBe expected }.message
      actual.shouldContainInOrder(
         "Match[0]: part of slice with indexes [2..39] matched actual[2..39]",
         """Match[0]= --++++++++++++++++++++++++++++++++++++++---""",
         "expected:<One quick brown fox jumps over the lazy cat> but was:<The quick brown fox jumps over the lazy dog>"
      )
```

To make this happen, I needed to move around a few functions, and to make some of them public.


AlexCue987:
too many conflicts, not worth resolving

PR #4431, opened 2024-10-13
Modernize shouldBeSortedDescending

okarmazin:
- Adds shouldBeSortedDescending for primitive arrays
- Generalizes the matchers from List to Iterable

BooleanArray omitted - it is not very natural to check if a collection of booleans is "sorted".

The List overload will be removed in a separate PR.


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/sortedDesc.kt
```diff
@@ -26,33 +61,76 @@ infix fun <T, E : Comparable<E>> List<T>.shouldBeSortedDescendingBy(transform: (
...
+         return positiveResult
       }
-      return MatcherResult(
-         failure == null,
-         { "List ${value.print().value} should be sorted$elementMessage" },
-         { "List ${value.print().value} should not be sorted" }
-      )
+   }
+
+internal inline fun <T> Iterable<T>.zippedWithNext(
```
AlexCue987, line 113: Is there a reason the following short idiomatic one-liner was expanded into a 20-line  non-idiomatic method:

`". Element ${failure.value} at index ${failure.index} was less than element ${value[failure.index + 1]}"`

Let's revert back to that simple one-liner.

okarmazin: Iterable isn't a `RandomAccess` container, so it has neither the `get` operator nor `size ` or `lastIndex`.

This algorithm performs a single pass over the container without extraneous allocations and will be reused for both ascending `beSorted` matchers and sequence `beSorted` matchers.

AlexCue987: > Iterable isn't a `RandomAccess` container, so it has neither the `get` operator nor `size ` or `lastIndex`.
> 
> This algorithm performs a single pass over the container without extraneous allocations and will be reused for both ascending `beSorted` matchers and sequence `beSorted` matchers.

Kotlin's own  `zipWithNext` and `firstOrNull` do all that as well, so please use them instead or reimplementing.

AlexCue987 (changes requested):
let's use `firstOrNull` from Kotlin standard library instead of reimplementing it.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/sortedDesc.kt
okarmazin: > > Iterable isn't a `RandomAccess` container, so it has neither the `get` operator nor `size ` or `lastIndex`.
> > This algorithm performs a single pass over the container without extraneous allocations and will be reused for both ascending `beSorted` matchers and sequence `beSorted` matchers.
> 
> Kotlin's own `zipWithNext` and `firstOrNull` do all that as well, so please use them instead or reimplementing.

No, it doesn't - `zipWithNext` is a terminal operation which buffers the entire Iterable in memory by allocating a new list of the resulting pairs. If it did do the same as the introduced function, I would have used it. Do you want to forgo the single-pass algorithm in favor of the extraneous allocation?

AlexCue987: can you use this one:

https://github.com/JetBrains/kotlin/blob/eb6e6b7b526f06ff72b56de3c7949d91f84d50e6/libraries/stdlib/common/src/generated/_Sequences.kt#L2851

sksamuel:
@okarmazin are you able to finish this one off ?

okarmazin:
Yeah, I think I'll get back to Kotest during the Christmas holidays.

PR #4432, opened 2024-10-13
fix: Stacktrace is lost with one assertion in AssertSoftly, #4171

pientaa:
It looks like `MultiAssertionError` factory method already handles properly following cases:
 - `size == 1 && subject != null`
 - `size == 1 && subject == null `
 
So the easiest fix is to use `MultiAssertionError` even though there is just one assertion error - this way we don't loose the stacktrace. I provided appropriate test which covers the case presented in issue #4171

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/AssertSoftlyTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/AssertSoftlyTest.kt
@@ -156,7 +156,10 @@ class AssertSoftlyTest : FreeSpec({
       }
 
 
-      "Should not loose stacktrace with only one assertion" {
+      "Should not lose stacktrace with only one assertion" {
+         val expectedLineNumber = Exception().stackTrace
+            .first { it.className.contains("AssertSoftlyTest") }.lineNumber + 5
+
          shouldThrow<AssertionError> {
             assertSoftly {
                null should beEmpty()
@@ -165,7 +168,7 @@ class AssertSoftlyTest : FreeSpec({
             message.shouldContainInOrder(
                "The following assertion failed:",
                "1) Expecting actual not to be null",
-               "   at com.sksamuel.kotest.matchers.AssertSoftlyTest${'$'}1${'$'}1${'$'}9.invokeSuspend(AssertSoftlyTest.kt:162)"
+               "   at com.sksamuel.kotest.matchers.AssertSoftlyTest${'$'}1${'$'}1${'$'}9.invokeSuspend(AssertSoftlyTest.kt:$expectedLineNumber)"
             )
          }
       }
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/AssertSoftlyTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/AssertSoftlyTest.kt
@@ -19,6 +19,7 @@ import io.kotest.matchers.should
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.shouldNot
 import io.kotest.matchers.shouldNotBe
+import io.kotest.matchers.string.beEmpty
 import io.kotest.matchers.string.contain
 import io.kotest.matchers.string.endWith
 import io.kotest.matchers.string.shouldContain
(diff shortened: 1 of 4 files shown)


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/AssertSoftlyTest.kt
```diff
@@ -154,6 +155,21 @@ class AssertSoftlyTest : FreeSpec({
          }
       }
 
+
+      "Should not loose stacktrace with only one assertion" {
```
AlexCue987: Nit: please replace "loose" with "lose".

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/AssertSoftlyTest.kt
```diff
@@ -154,6 +155,21 @@ class AssertSoftlyTest : FreeSpec({
...
+         shouldThrow<AssertionError> {
+            assertSoftly {
+               null should beEmpty()
+            }
+         }.run {
+            message.shouldContainInOrder(
+               "The following assertion failed:",
+               "1) Expecting actual not to be null",
+               "   at com.sksamuel.kotest.matchers.AssertSoftlyTest${'$'}1${'$'}1${'$'}9.invokeSuspend(AssertSoftlyTest.kt:162)"
```
AlexCue987: Minor but blocking: let's remove line number 162 from expected string number 3, to make the test less fragile. Otherwise any refactoring/reformatting will break this test.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/maps/MapMatchersTest.kt
```diff
@@ -596,7 +596,11 @@ private fun matchMapTests(contextName: String) = wordSpec {
                mapOf("key" to "hi") should matcher("key" to { it shouldHaveLength 4 })
             }
          }.also {
-            it.message shouldBe """Expected map to match all assertions. Missing keys were=[], Mismatched values were=[(key, "hi" should have length 4, but instead was 2)], Unexpected keys were []."""
+            it.message.shouldContainInOrder(
+               "The following assertion failed:",
+               "1) Expected map to match all assertions. Missing keys were=[], Mismatched values were=[(key, \"hi\" should have length 4, but instead was 2)], Unexpected keys were [].",
+               "   at com.sksamuel.kotest.matchers.maps.MapMatchersTestKt${'$'}matchMapTests${'$'}1${'$'}1${'$'}3.invokeSuspend(MapMatchersTest.kt:596)",
```
AlexCue987: Same suggestion: can you remove `596` to make the test less fragile?

AlexCue987 (changes requested):
LGTM, a few minor suggestions

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/AssertSoftlyTest.kt
pientaa: That makes sense. However, this was one of the problems reported in the issue: an incorrect pointer in the stack trace. I'll move that test to a separate class to prevent it from affecting the other tests. What do you think?

AlexCue987: > That makes sense. However, this was one of the problems reported in the issue: an incorrect pointer in the stack trace. I'll move that test to a separate class to prevent it from affecting the other tests. What do you think?

I concur - let's extract this test into its own file.

pientaa: Please take a look now, I changed the hardcoded line number to dynamically calculated variable.

AlexCue987: > Please take a look now, I changed the hardcoded line number to dynamically calculated variable.

LGTM, started the build

AlexCue987: > Please take a look now, I changed the hardcoded line number to dynamically calculated variable.

the following test is failing, and it might be related: "adds an Exception to an empty collection of assertion failures"

pientaa: Yes, I see that there is a test class marked to run on Linux only, which is why I didn't catch it failing in my local run.

I've already fixed it. The issue was that the expected exception class had changed, since any assertion failure in the `assertSoftly` block should now be classified as a `MultiAssertionError`, even if only one assertion fails.

AlexCue987 (approved):
nice!

PR #4434, opened 2024-10-14
feat(extensions): support resetting environment variables via SystemEnvironmentListener

tbcrawford:
This adds null value support for resetting environment variables to `SystemEnvironmentTestListener` and `SystemEnvironmentProjectListener` so that they function the same as `withEnvironment`


kotest-extensions/src/jvmTest/kotlin/com/sksamuel/kt/extensions/system/SystemEnvironmentExtensionTest.kt
```diff
@@ -63,14 +63,21 @@ class SystemEnvironmentExtensionTest : FreeSpec() {
...
+      mode = OverrideMode.SetOrOverride
+   )
 
    override fun extensions() = listOf(setl)
 
    init {
       "sys environment extension" should {
          "set environment variable" {
             System.getenv("mop") shouldBe "dop"
+            System.getenv("dop") shouldBe null
```
LeoColman, line 90: 🔴 I think this test needs to check that the variable was removed from the environment, I.E. having it configured before the test runs

Otherwise we might catch a false positive here where our implementation does nothing with the variable and thus it stays null as before

LeoColman (dismissed):
There's only one piece I think needs correction on the tests.

Otherwise this is a very good PR!

kotest-extensions/src/jvmTest/kotlin/com/sksamuel/kt/extensions/system/SystemEnvironmentExtensionTest.kt
tbcrawford: @LeoColman Apologies for the long delay! The tests are ready for a re-run and I believe I addressed your test concerns here so please re-review when you have a moment!

AlexCue987 (approved):

sksamuel (approved):

sksamuel:
@LeoColman can you approve, it won't merge as you have the block on it

PR #4436, opened 2024-10-14
fix: Print clues on any Exception thrown, #4346

### LINKED ISSUE #4346: `asClue` and `withClue` do not print clues when the block throws an exception that is not of type `AssertionFailedError`
Kotest version: 5.9.0

Consider test:
```kotlin
test("foo") {
  "some clue".asClue {
    listOf("a", "b")
      .single { it.length == 2 }
      // chain some other assertion/operation on returned element
  }
}
```

On failure a clue is not printed:
```
List is empty.
java.util.NoSuchElementException: List is empty.
	at kotlin.collections.CollectionsKt___CollectionsKt.single(_Collections.kt:616)
```

Expected something like:
```
some clue
java.util.NoSuchElementException: List is empty.
	at kotlin.collections.CollectionsKt___CollectionsKt.single(_Collections.kt:616)
```

I suspect this is because `asClue` and `withClue` only work when the block throws kotest's `AssertionFailedError`. And because `single` throws kotlin's `NPE` or `IllegalArgumentException`, it doesn't work nicely with clues.

In many cases, our tests and testing infrastructure performs assertions without using kotest APIs (using methods like `single`, `first`, etc that throw non `AssertionFailedError` exceptions). Because of this, majority of our test failures are missing clues.

It would be very helpful if `asClue`/`withClue` worked still when any exception is throws.

Slack `kotlinlang` discussion: https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1726702539652919

pientaa:
Fixes #4346

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/ExceptionWithClue.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/ExceptionWithClue.kt
@@ -12,9 +12,9 @@ package io.kotest.assertions
  */
 class ExceptionWithClue(
    val clue: String,
-   override val message: String?,
-   override val cause: Throwable?,
-) : RuntimeException() {
+   message: String?,
+   cause: Throwable?,
+) : RuntimeException(message, cause) {
    companion object {
       @Suppress("ThrowableNotThrown")
       fun from(clue: String, e: Exception): ExceptionWithClue =
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/ExceptionWithClue.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/ExceptionWithClue.kt
@@ -1,0 +1,27 @@
+package io.kotest.assertions
+
+/**
+ * Exception class that enhances [withClue] and [asClue] functions by capturing additional context (clue)
+ * and attaching it to the thrown exception.
+ *
+ * @property clue Additional context to enrich the assertion error message if an assertion fails.
+ * @property message The error message, combining the clue and the original exception message.
+ * @property cause The original exception that triggered this one.
+ *
+ * This class simplifies debugging by bundling relevant clues into the exception when using [withClue] and [asClue].
+ */
+class ExceptionWithClue(
+   val clue: String,
+   override val message: String?,
+   override val cause: Throwable?,
+) : RuntimeException() {
+   companion object {
+      @Suppress("ThrowableNotThrown")
+      fun from(clue: String, e: Exception): ExceptionWithClue =
+         ExceptionWithClue(
+            clue = clue,
+            message = clueContextAsString() + (e.message ?: ""),
+            cause = e,
+         )
+   }
+}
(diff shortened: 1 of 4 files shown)


OliverO2:
Wouldn't this duplicate clues in case of nested `withClue` invocations?

```kotlin
withClue("outer") {
    withClue("inner") {
        "a" shouldBe "b"
    }
}
```

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/clues.kt
```diff
@@ -28,6 +28,8 @@ inline fun <R> withClue(crossinline clue: () -> Any?, thunk: () -> R): R {
       // this is a special control exception used by coroutines
    } catch (t: TimeoutCancellationException) {
       throw Exceptions.createAssertionError(clueContextAsString() + (t.message ?: ""), t)
+   } catch (e: Exception) {
+      throw Exception(clueContextAsString() + (e.message ?: ""), e)
```
AlexCue987: Let's throw a very specific Exception here, such as `ExceptionWithClue`, not a generic one. And if we catch it, let's just rethrow it:
```kotlin
catch(c: ExceptionWithClue) { throw c }
```

pientaa:
@OliverO2, good catch! You're right. The best solution for the current design would probably be reusing the `Exceptions.createAssertionError` factory function. That way, we won’t be throwing a generic `Exception` anymore. @AlexCue987, let me know what you think.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/clues.kt
```diff
@@ -25,11 +23,8 @@ inline fun <R> withClue(crossinline clue: () -> Any?, thunk: () -> R): R {
    try {
       collector.pushClue { clue.invoke().toString() }
       return thunk()
-      // this is a special control exception used by coroutines
-   } catch (t: TimeoutCancellationException) {
-      throw Exceptions.createAssertionError(clueContextAsString() + (t.message ?: ""), t)
    } catch (e: Exception) {
-      throw Exception(clueContextAsString() + (e.message ?: ""), e)
+      throw Exceptions.createAssertionError(clueContextAsString() + (e.message ?: ""), e)
```
AlexCue987: I don't think AssertionError is appropriate here - no Assertion has explicitly failed here, it's a different situation. I'd suggest creating another Exception specifically for this purpose, not reusing AssertionError.

pientaa: Sure, I'll work on this later

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/ExceptionWithClue.kt
```diff
@@ -0,0 +1,27 @@
...
+ * @property cause The original exception that triggered this one.
+ *
+ * This class simplifies debugging by bundling relevant clues into the exception when using [withClue] and [asClue].
+ */
+class ExceptionWithClue(
+   val clue: String,
+   override val message: String?,
+   override val cause: Throwable?,
+) : RuntimeException() {
+   companion object {
```
AlexCue987, line 18: LGTM, but can I suggest a simplification, because we don't need to override Throwable's fields, we can just use them:
```kotlin
class ExceptionWithClue(
   val clue: String,
   message: String?,
   cause: Throwable?,
) : RuntimeException(message, cause)
```

pientaa: sure

pientaa:
@AlexCue987 I noticed the build is failing, but I don’t believe it's related to my changes.

AlexCue987:
> @AlexCue987 I noticed the build is failing, but I don’t believe it's related to my changes.

agreed

AlexCue987 (approved):
LGTM

PR #4439, opened 2024-10-15
Alex/custom matching of fields take0 (#4438)

AlexCue987:
extend `shouldBeEqualUsingFields`, allow custom matching of fields, such as
* case-insensitive String
* numbers/temporals with tolernace
* Lists ignoring order

Example:

```kotlin
      "basic example: match name ignoring case" {
         val expected = SimpleDataClass("apple", 1.0, LocalDateTime.now())
         val actual = expected.copy(name = "Apple")
         shouldThrow<AssertionError> {
            actual shouldBeEqualUsingFields expected
         }.message.shouldContainInOrder(
            "Fields that differ:",
            """- name  =>  expected:<"apple"> but was:<"Apple">""",
         )
         actual shouldBeEqualUsingFields {
            overrideMatchers = mapOf(
               SimpleDataClass::name to matchStringsIgnoringCase
            )
            expected
         }
      }
```


LeoColman (approved):
Should we add these to the docs?

AlexCue987:
> Should we add these to the docs?

of course!

PR #4440, opened 2024-10-16
fix: Nested scopes do not share a test scheduler, #4318

### LINKED ISSUE #4318: Nested scopes do not share a test scheduler
Following test:

```kotlin
class DispatchersTest : WordSpec(
    {
        coroutineTestScope = true

        "a" should {
            println("Scheduler in a: $testCoroutineScheduler")
            "b" {
                println("Scheduler in b: $testCoroutineScheduler")
            }
        }
    },
)

```

Will print out:

```
Scheduler in a: kotlinx.coroutines.test.TestCoroutineScheduler@22d0639b
Scheduler in b: kotlinx.coroutines.test.TestCoroutineScheduler@4229d48d
```

Note that every block has a different coroutine scheduler.

In a practical sense, this means that nested scopes do not share virtual time, which could weak havoc, if you want to control the time:

```kotlin
class DispatchersTest : WordSpec(
    {
        coroutineTestScope = true

        "a" should {
            println("time before delay: ${testCoroutineScheduler.currentTime}")

            delay(1_000)

            println("time after delay but before b: ${testCoroutineScheduler.currentTime}")

            "b" {
                println("time in b: ${testCoroutineScheduler.currentTime}")
            }
        }
    },
)
```

will print

```
time before delay: 0
time after delay but before b: 1000
time in b: 0
```

Note that the time has reset to 0, even though code should just execute sequentially from a to the nested b.

This is essentially a reopen of the https://github.com/kotest/kotest/issues/3238, it looks like the fix was not complete.

**Which version of Kotest are you using** 5.6.1

pientaa:
Fixes #4318

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/engine/DispatchersSharedCoroutineSchedulerTest.kt
+++ b/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/engine/DispatchersSharedCoroutineSchedulerTest.kt
@@ -1,0 +1,26 @@
+package io.kotest.engine
+
+import io.kotest.core.spec.IsolationMode
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.core.test.testCoroutineScheduler
+import io.kotest.matchers.shouldBe
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.delay
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class DispatchersSharedCoroutineSchedulerTest : WordSpec({
+   coroutineTestScope = true
+   isolationMode = IsolationMode.InstancePerLeaf
+
+   "a" should {
+      testCoroutineScheduler.currentTime shouldBe 0
+
+      delay(1_000)
+
+      testCoroutineScheduler.currentTime shouldBe 1_000
+
+      "b" {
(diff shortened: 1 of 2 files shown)


kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/engine/DispatchersSharedCoroutineSchedulerTest.kt
```diff
@@ -0,0 +1,26 @@
...
+import io.kotest.matchers.shouldBe
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.delay
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class DispatchersSharedCoroutineSchedulerTest : WordSpec({
+   coroutineTestScope = true
+   isolationMode = IsolationMode.InstancePerLeaf
+
+   "a" should {
```
AlexCue987: should we replace `"a"` and `"b"`  with  more descriptive test names?

pientaa: sure

AlexCue987 (approved):

matejdro:
Any chance of a release with this fix included, @sksamuel? Thanks.

PR #4442, opened 2024-10-18
Increase the number of arguments in forAll to 10

KakeruHirokami:
To migrate from kotlintest to kotest, I have changed the number of arguments in forAll to 10.

I am doing a kotest migration from kotlintest v3.3.2
I am replacing as shown below, an error occurred due to an insufficient number of arguments in forAll.
- `io.kotlintest.data.forall` -> `io.kotest.data.blocking.forAll`
- `io.kotlintest.data.suspend.forall` -> `io.kotest.data.forAll`

In kotlintest it accepts up to 10 arguments
`io.kotlintest.data.forall`
https://github.com/kotest/kotest/blob/3.3.2/kotlintest-assertions/src/main/kotlin/io/kotlintest/data/DataDrivenTesting.kt
`io.kotlintest.data.suspend.forall`
https://github.com/kotest/kotest/blob/3.3.2/kotlintest-assertions/src/main/kotlin/io/kotlintest/data/suspend/DataDrivenTesting.kt

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/api/kotest-assertions-shared.api
+++ b/kotest-assertions/kotest-assertions-shared/api/kotest-assertions-shared.api
@@ -722,6 +722,15 @@ public final class io/kotest/data/ErrorsKt {
 	public static final fun forNoneError (Ljava/util/List;Ljava/util/List;)Ljava/lang/Throwable;
 }
 
+public final class io/kotest/data/ForAll10Kt {
+	public static final fun forAll (Lio/kotest/data/Table10;Lkotlin/jvm/functions/Function10;)V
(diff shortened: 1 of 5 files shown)


AlexCue987 (commented):
LGTM, but should we add tests that execute all new code? I know that similar preexisting functions do not have tests either, but should we improve that as well, in a separate PR(s) maybe?

KakeruHirokami:
Thank you for your confirmation.
I agree that add tests that execute all new code, but other functions also need to have tests added.
I think it would be better to handle this in a separate PR(s).

AlexCue987:
> Thank you for your confirmation. I agree that add tests that execute all new code, but other functions also need to have tests added. I think it would be better to handle this in a separate PR(s).

can you run apiDump

KakeruHirokami:
Sorry, I hadn't run apiDump. I've now run apiDump and pushed it.

Thank you.

AlexCue987 (approved):
TY

Issue #4152, opened 2024-06-26
When property testing strings using codepoints with unprintable characters, the output is not helpful


Zordid:
Currently, when using a simple `Arb.string` with codepoints like `Codepoint.ascii()` there will at some point be unprintable characters contained in the strings generated.

If such a test value fails the assertions, shrinking kicks in and in the end, you might end up with a message like this:

```
Property failed after 1 attempts
java.lang.AssertionError: Property failed after 1 attempts

	Arg 0: "" (shrunk from -&X74]"?@�mcayJL:,T`g{}ycwpC.Dux5"c1;F?@;So^)

Repeat this test by using seed -703120417207228322
```

Even worse. In IntelliJ, this shrunk string displays as `""` (like an empty string)! You simply do not know what exactly the problem might be.

So, my feature request is: in case of unprintable chars in a test string, please output the hex codes instead. Especially when the shrunk result is that single character that is causing the problem.

Fleshgrinder:
With hex codes, do you mean Unicode code point notation? Non-standardized hex coding can be confusing because it’s not always clear what they represent. For instance, the JVM uses UTF-16 by default, so the hex codes of the individual bytes wouldn’t correspond directly to UTF-8 or UTF-32 code points.

Normalizing the output to Unicode notation (`U+xxxx` to `U+xxxxxxxx`) would likely be best, as it provides a standardized reference. However, this still depends on the underlying encoding, as it might not lead to the correct code point for encodings like KOI-8.

An alternative could be using [Unicode control pictures](https://www.unicodepedia.com/groups/control-pictures/), which can help represent control characters visually in a more standardized way.

sksamuel:
Fixed in https://github.com/kotest/kotest/pull/4447

PR #4447, opened 2024-10-18
feat: Output Hex for Unprintable characters, #4152

pientaa:
Implements #4152. 

This PR introduces the feature to handle unprintable characters in failure messages by escaping them to a standardized Unicode notation (U+xxxx).

Once the overall approach is approved, I will update the documentation accordingly.

### DIFF
--- a/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/OutputHexForUnprintableCharsTest.kt
+++ b/kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/OutputHexForUnprintableCharsTest.kt
@@ -1,0 +1,45 @@
+package com.sksamuel.kotest.property.arbitrary
+
+import io.kotest.assertions.AssertionFailedError
+import io.kotest.assertions.throwables.shouldThrow
+import io.kotest.common.ExperimentalKotest
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.string.shouldContain
+import io.kotest.property.Arb
+import io.kotest.property.PropTestConfig
+import io.kotest.property.arbitrary.Codepoint
+import io.kotest.property.arbitrary.ascii
+import io.kotest.property.arbitrary.printableAscii
+import io.kotest.property.arbitrary.string
+import io.kotest.property.checkAll
+import io.kotest.property.internal.escapeUnprintable
+
+@OptIn(ExperimentalKotest::class)
+class OutputHexForUnprintableCharsTest : StringSpec({
+
+   "should handle unprintable characters in failure messages" {
+      checkAll(
+         PropTestConfig(outputHexForUnprintableChars = true, seed = -5556043119863981334),
+         Arb.string(1..127, codepoints = Codepoint.ascii())
+      ) { str ->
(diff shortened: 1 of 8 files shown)


pientaa:
@AlexCue987 could you take a look and share your thoughts?

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/OutputHexForUnprintableCharsTest.kt
```diff
@@ -0,0 +1,45 @@
...
+class OutputHexForUnprintableCharsTest : StringSpec({
+
+   "should handle unprintable characters in failure messages" {
+      checkAll(
+         PropTestConfig(outputHexForUnprintableChars = true, seed = -5556043119863981334),
+         Arb.string(1..127, codepoints = Codepoint.ascii())
+      ) { str ->
+         val message = shouldThrow<AssertionFailedError> { str shouldBe "expectedString" }.message
+
+         val escapedMessage = message?.escapeUnprintable()
```
AlexCue987, line 30: presumably for some iterations if this test `message != escapedMessage`, correct? do we need to validate that assumption?

AlexCue987 (commented):
I think these code changes are good to go. I'd suggest a small change in one test, but that's minor. I concur the docs can be updated later in another PR.

But I don't work with non-ASCII characters, so I don't fully understand why we need this feature. Can you elaborate? And hopefully @Kantis @OliverO2 and @LeoColman can chip in with their thoughts.

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/OutputHexForUnprintableCharsTest.kt
pientaa: I'm not sure if I understand what you mean, but let's take a look at my last commit and you can share your thoughts.

pientaa:
>  I don't work with non-ASCII characters, so I don't fully understand why we need this feature. Can you elaborate? 

@AlexCue987 sure. I also don't work with non-ASCII characters, but based on the issue description and a little bit of research I found out what the problem is about.

When working exclusively with ASCII characters using `Codepoint.ascii()`, the generated strings can include control characters and other unprintable ASCII characters. These characters are part of the ASCII table but are not printable, meaning they don't correspond to visible symbols when rendered in most contexts.

In property-based testing, it becomes difficult to understand what the actual input was that caused the failure. For example, the output might display as an empty string or as the Unicode replacement character �, which doesn't provide useful information for debugging.

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/OutputHexForUnprintableCharsTest.kt
AlexCue987: > I'm not sure if I understand what you mean, but let's take a look at my last commit and you can share your thoughts.

that';s exactly what I meant, TY

pientaa:
@AlexCue987 are we good to go with these changes?

AlexCue987 (approved):
nice!

AlexCue987:
> @AlexCue987 are we good to go with these changes?

sure, I'll merge them. Thx!

PR #4468, opened 2024-10-27
Add inline keyword and support for suspend functions in InspectorAliases

### LINKED ISSUE #4462: Enhance InspectorAliases functions to support suspend functions
Using `shouldForAll` is not possible to call suspend functions inside the lambda parameter:

```kotlin
@Test
fun `shouldForAll suspend function test`() = runTest {
   list(0, 1, 2, 3).shouldForAll {
      networkCall() // Compiler error: Suspend function 'networkCall' should be called only from a coroutine or another suspend function
   }
}

private suspend fun networkCall() {
}
```

This occurs because correctly the function is not inline:

https://github.com/kotest/kotest/blob/d965b751334f364fdafeeaa711dd691ff72c223d/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/inspectors/InspectorAliases.kt#L4-L8

This behavior is not only for `shouldForAll`, but for all functions inside of `InspectorAliases.kt`.

By adding `inline` modifier to these functions will enhance the current behavior to add the possibility to call suspended functions, but I haven't fully test to see if something will beak with this change.

LeoColman:
Introduced the inline keyword to all alias functions in InspectorAliases . Added tests to support suspend functions in InspectorAliasTest for better flexibility in async scenarios.

Fixes #4462

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/inspectors/InspectorAliases.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/inspectors/InspectorAliases.kt
@@ -1,79 +1,100 @@
 package io.kotest.inspectors
 
 /** Alias for [Sequence.forAll] */
-fun <T> Sequence<T>.shouldForAll(fn: (T) -> Unit) = forAll(fn)
+inline fun <T> Sequence<T>.shouldForAll(fn: (T) -> Unit) = forAll(fn)
+
 /** Alias for [Array.forAll] */
-fun <T> Array<T>.shouldForAll(fn: (T) -> Unit) = forAll(fn)
+inline fun <T> Array<T>.shouldForAll(fn: (T) -> Unit) = forAll(fn)
+
 /** Alias for [Collection.forAll] */
-fun <T> Collection<T>.shouldForAll(fn: (T) -> Unit) = forAll(fn)
+inline fun <T> Collection<T>.shouldForAll(fn: (T) -> Unit) = forAll(fn)
 
 
 /** Alias for [Sequence.forOne] */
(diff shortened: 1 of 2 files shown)


LeoColman:
Change is virtually the same for all the inspectors:

1. Add test using a suspend function
2. Test fails due to compilation issues
3. Add inline to all aliases
4. Test passes

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/inspectors/InspectorAliases.kt
```diff
@@ -1,79 +1,100 @@
...
+inline fun <T> Array<T>.shouldForSome(fn: (T) -> Unit) = forSome(fn)
+
 /** Alias for [Collection.forSome] */
-fun <T> Collection<T>.shouldForSome(fn: (T) -> Unit) = forSome(fn)
+inline fun <T> Collection<T>.shouldForSome(fn: (T) -> Unit) = forSome(fn)
 
 
 /** Alias for [Sequence.forAny] */
-fun <T> Sequence<T>.shouldForAny(fn: (T) -> Unit) = forAny(fn)
+inline fun <T> Sequence<T>.shouldForAny(fn: (T) -> Unit) = forAny(fn)
```
AlexCue987, line 44: It seems redundant to have both `forSome` and `forAny`. What am I missing?

AlexCue987 (approved):

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/inspectors/InspectorAliases.kt
LeoColman: I think `forSome` means strictly "At least one but not all" and `forAny` means "For 1, 2 or N inclusive`

PR #4469, opened 2024-10-27
Add unit tests for containerName function in utils.kt

### LINKED ISSUE #4381: Review Iterable.containerName whether it has enough cases
TODO: After https://github.com/kotest/kotest/pull/4369
Also, use the shared impl going forward. Right now it is duplicated in order to make PRs independent.

_Originally posted by @LeoColman in https://github.com/kotest/kotest/pull/4378#discussion_r1779820481_

LeoColman:
Introduced comprehensive tests to validate that containerName accurately
 identifies different types of collections such as List, Set,
 Collection, and Iterable. Also removed classes that are not Iterable
 from being tested.

 Closes #4381


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/utils.kt
```diff
@@ -3,15 +3,14 @@ package io.kotest.matchers.collections
...
  *
- * "List" for lists, "Set" for sets, "Range" for ranges etc.
+ * "List" for lists, "Set" for sets, "Collection" for collections etc.
  */
 internal fun Iterable<*>.containerName(): String {
    return when (this) {
-      is List -> "List"
-      is Set -> "Set"
-      is Map<*, *> -> "Map"
-      is ClosedRange<*>, is OpenEndRange<*> -> "Range"
```
LeoColman, line 13: Map and Range are not iterables

AlexCue987:
```
fail for single element range[jvm] FAILED

    io.kotest.assertions.AssertionFailedError: expected:<"Range should be empty but has at least one element, first being: 1"> but was:<"Iterable should be empty but has at least one element, first being: 1">


    Test fail for single element open range[jvm] FAILED

    io.kotest.assertions.AssertionFailedError: expected:<"Range should be empty but has at least one element, first being: 1"> but was:<"Iterable should be empty but has at least one element, first being: 1">
```
it seems these changes broke some tests

LeoColman:
Decision time then!

```
"fail for single element range" {
            shouldThrowAny {
               (1..1).shouldBeEmpty()
            }.message shouldBe "Range should be empty but has at least one element, first being: 1"
         }
```

This is the kind of test that is failing.
It's a bit tricky because Ranges don't implement `Iterable`

However an `IntProgression` is an `Int`, but it's also an Iterable. It was being matched against the `Iterable.XX` matchers by pure luck; it wouldn't work for a range of BigDecimals or LocalDates, for example.


I think technically the new message "Iterable should be ..." is more correct, but losing the `Range` keyword in the message still sucks.


I don't like to add `Range` to the `when(iterable) is Range`, but the current setup is not good. What's your opinion @AlexCue987

AlexCue987:
> Decision time then!
> 
> ```
> "fail for single element range" {
>             shouldThrowAny {
>                (1..1).shouldBeEmpty()
>             }.message shouldBe "Range should be empty but has at least one element, first being: 1"
>          }
> ```
> 
> This is the kind of test that is failing. It's a bit tricky because Ranges don't implement `Iterable`
> 
> However an `IntProgression` is an `Int`, but it's also an Iterable. It was being matched against the `Iterable.XX` matchers by pure luck; it wouldn't work for a range of BigDecimals or LocalDates, for example.
> 
> I think technically the new message "Iterable should be ..." is more correct, but losing the `Range` keyword in the message still sucks.
> 
> I don't like to add `Range` to the `when(iterable) is Range`, but the current setup is not good. What's your opinion @AlexCue987

I think that in this context there is no ambiguity about what exactly is empty - we know it's `(1..1)`. So calling it `Iterable` in the message is enough. And should we rename the test `"fail for single element range"` accordingly?

LeoColman:
I think the call is what I've done on the latest commit and drop these test altogether.

We are testing Iterables, not Ranges, those tests shouldn't be there as they are just extra `Iterable` tests

sksamuel (approved):

sksamuel:
What's this waiting on to be complete?

LeoColman:
This was not thought through by me.
I'll close this and if we ever have troubles we can work on it again

PR #4484, opened 2024-10-30
Alex/search for partial matches in string should be  (#4483)

AlexCue987:
This follows a Slack discussion with @sksamuel and @Kantis .

When matching strings, search for partial matches, such as:

```kotlin
     val expected = "One quick brown fox jumps over the lazy cat"
      val value ="The quick brown fox jumps over the lazy dog"
      val actual = shouldThrow<AssertionError> { value shouldBe expected }.message
      actual.shouldContainInOrder(
         "Match[0]: part of slice with indexes [2..39] matched actual[2..39]",
         """Match[0]= --++++++++++++++++++++++++++++++++++++++---""",
         "expected:<One quick brown fox jumps over the lazy cat> but was:<The quick brown fox jumps over the lazy dog>"
      )
```

To make this happen, I needed to move around a few functions, and to make some of them public.


kotest-assertions/kotest-assertions-shared/src/jvmTest/kotlin/io/kotest/assertions/AssertSoftlyTests.kt
```diff
@@ -12,7 +12,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.withContext
 
-@EnabledIf(LinuxCondition::class)
+//@EnabledIf(LinuxCondition::class)
```
LeoColman, line 15: !

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/submatching/StringPartialMatches.kt
```diff
@@ -2,16 +2,25 @@ package io.kotest.submatching
 
 import io.kotest.assertions.AssertionsConfig
 
-internal fun describePartialMatchesInStringForSlice(expectedSlice: String, value: String) =
+class StringPartialMatch(val expected: String, val value: String) {
```
LeoColman, line 5: I don't understand why we need all these guys as part of the public API.
Where is access to them being public required?

LeoColman (approved):
To me the only confusing point so far is why some of the functions need to be public.

If there is a valid reason I think this is ready :)

AlexCue987:
> To me the only confusing point so far is why some of the functions need to be public.

the reason is:
* I needed to invoke these functions from `shouldBe` which is in `kotest-assertions-shared` 
* These functions were `internal` in `kotest-assertions-core`
* So I moved them to `kotest-assertions-shared` 
* But I still needed to invoke them from  `kotest-assertions-core`
* So I had to make them `public`

Can you suggest a way to keep them `internal`? The only thing I can think of is to merge `shared` and `core` modules.

LeoColman:
I'm ok with your reasoning. I didn't notice they were both in different source sets.

PR #4490, opened 2024-11-01
fix: YAML package and invalid equal

rbraeunlich:


kotest-assertions/kotest-assertions-yaml/src/commonMain/kotlin/io/kotest/assertions/yaml/YamlMatchers.kt
```diff
@@ -58,8 +60,25 @@ fun equalYaml(
...
          )
       } else {
-         val (expectedTree, actualTree) = parse(expected, actual)
-         equalYamlNode(expectedTree).test(actualTree)
+         val actualYaml = try {
+            yaml.parseToYamlNode(actual)
+         } catch (ex: Exception) {
+            return@Matcher MatcherResult(
+               false,
+               { "expected: actual YAML to be valid YAML: $actual" },
```
AlexCue987: might `ex.message` or the type of Exception contain any useful information to include in the message?

rbraeunlich: Good point. I added the exceptions's message in case of an invalid YAML. I hope it matches the usual text format @AlexCue987

kotest-assertions/kotest-assertions-yaml/src/commonMain/kotlin/io/kotest/assertions/yaml/YamlMatchers.kt
```diff
@@ -58,8 +60,25 @@ fun equalYaml(
...
+         val expectedYaml = try {
+            yaml.parseToYamlNode(expected)
+         } catch (ex: Exception) {
+            return@Matcher MatcherResult(
+               false,
+               { "expected: actual YAML to be valid YAML Reason: ${ex.message} Actual: $actual" },
+               { "expected: expected YAML to be invalid YAML: $expected" }
+            )
+         }
+         equalYamlNode(expectedYaml).test(actualYaml)
```
LeoColman, line 81: I'm wondering if these exceptions should instead be re-thrown.

If we are passing something that's not an YAML, we are not making an assertion, we are failing anyways. Having a hard failure for bad yamls would clean up a bit these lines

LeoColman (approved):

kotest-assertions/kotest-assertions-yaml/src/commonMain/kotlin/io/kotest/assertions/yaml/YamlMatchers.kt
rbraeunlich: I understand your reasoning. I was following the JSON matchers here. They have the same behaviour. In case of an invalid JSON they catch the exception and return the failed result.

0x1010110 (approved):

AlexCue987 (approved):

PR #4510, opened 2024-11-28
shouldIntersect 

AlexCue987:
`shouldIntersect` verifies if two Sets intersect

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sets/SetsIntersection.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sets/SetsIntersection.kt
@@ -1,0 +1,22 @@
+package io.kotest.matchers.sets
+
+import io.kotest.assertions.print.print
+import io.kotest.matchers.Matcher
+import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldNot
+
+infix fun <T> Set<T>.shouldIntersect(ts: Set<T>) = this should intersectWith(ts)
+infix fun <T> Set<T>.shouldNotIntersect(ts: Set<T>) = this shouldNot intersectWith(ts)
+
+fun <T> intersectWith(ts: Set<T>): Matcher<Set<T>> = object : Matcher<Set<T>> {
+   override fun test(value: Set<T>): MatcherResult {
+      val intersection = ts.filter { value.contains(it) }
+      val passed = intersection.isNotEmpty()
+
+      val failure = { "Set should intersect with ${ts.print().value} but did not." }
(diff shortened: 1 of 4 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sets/SetsIntersection.kt
```diff
@@ -0,0 +1,22 @@
...
+import io.kotest.assertions.print.print
+import io.kotest.matchers.Matcher
+import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldNot
+
+infix fun <T> Set<T>.shouldIntersect(ts: Set<T>) = this should intersectWith(ts)
+infix fun <T> Set<T>.shouldNotIntersect(ts: Set<T>) = this shouldNot intersectWith(ts)
+
+fun <T> intersectWith(ts: Set<T>): Matcher<Set<T>> = object : Matcher<Set<T>> {
```
LeoColman: 🟡 I think we should explain clearer what a Set Intersection means. I know it's mathematically well-defined, but users don't need to remember these concepts when reading through the suggestions from IDE

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sets/SetsIntersection.kt
```diff
@@ -0,0 +1,22 @@
...
+   override fun test(value: Set<T>): MatcherResult {
+      val intersection = ts.filter { value.contains(it) }
```
LeoColman: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/intersect.html

Can't actually believe this function exists in stdlib

```suggestion
      val intersection = ts intercect value
```

LeoColman (changes requested):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/sets/SetsIntersection.kt
AlexCue987: > 🟡 I think we should explain clearer what a Set Intersection means. I know it's mathematically well-defined, but users don't need to remember these concepts when reading through the suggestions from IDE

will do thx

AlexCue987: > 🟡 I think we should explain clearer what a Set Intersection means. I know it's mathematically well-defined, but users don't need to remember these concepts when reading through the suggestions from IDE

done

LeoColman (approved):

PR #4523, opened 2024-11-30
shouldEqualJson-both-missing-and-extra-draft (#4522)

AlexCue987:
when matching Json, detect the case when both some keys are extra and some are missing:
```
The top level object has extra field(s) [b] and missing field(s) [c]
```

before this change, it was this:
```
The top level object was missing expected field(s) [c]
```
which was indistinguishable from the case when there are no extra fields


kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/EqualTest.kt
```diff
@@ -22,7 +22,7 @@ import io.kotest.property.assume
 import io.kotest.property.checkAll
 import io.kotest.property.exhaustive.boolean
 
-@EnabledIf(LinuxCondition::class)
+//@EnabledIf(LinuxCondition::class)
```
LeoColman: !
```suggestion
@EnabledIf(LinuxCondition::class)
```

LeoColman (approved):

PR #4548, opened 2024-12-07
Add option to delete temporary directory only if tests pass #4460

### LINKED ISSUE #4460: Delete Temporary Directory Only if Test Passes
Currently, the method to create a temporary directory can either always delete the directory, or never delete the directory.

there is no option to only delete the temporary directories if the test passes. This would be useful, as the temporary directory is often times helpful for understanding why a test failed.

Possibly a candidate for #3918. #2792 would be incredibly useful for adding this.

sksamuel:
Fixes #4460


kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/spec/TempDirTest.kt
```diff
@@ -10,13 +10,18 @@ import io.kotest.matchers.file.shouldNotExist
...
+      dir1.shouldExist()
+   }
+
+   test("temp dir should support optional to not delete on failures") {
+      dir2.shouldExist()
    }
 
    afterProject {
-      dir.shouldNotExist()
+      dir1.shouldNotExist()
```
Kantis:
```suggestion
      dir1.shouldNotExist()
      dir2.shouldExist()
```

Kantis (approved):

PR #4550, opened 2024-12-08
Rework stable identifers to a single source with new types #4470

### LINKED ISSUE #4470: More Comprehensive Stable Identifiers
It would be nice if more classes has stable identifiers, as defined in these three places:
- https://github.com/kotest/kotest/blob/17d1b3b189fe4d804f1d3c6ce14280f724b0f168/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/Identifier.kt#L7-L17
- https://github.com/kotest/kotest/blob/17d1b3b189fe4d804f1d3c6ce14280f724b0f168/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/StableIdentifiers.kt#L47-L54
- https://github.com/kotest/kotest/blob/17d1b3b189fe4d804f1d3c6ce14280f724b0f168/kotest-common/src/commonMain/kotlin/io/kotest/mpp/stable.kt#L8-L84

I think that the following classes are good candidates for having stable identifiers:
- `Map<K, V>` where `K` and `V` have stable identifiers
- `Unit`
- `Array<E>` where `E` has a stable identifier
- `BooleanArray`, `ByteArray`, `CharArray`, `ShortArray`, `IntArray`, `LongArray`, `FloatArray`, and `DoubleArray`
- `List<E>` where `E` has a stable identifier
- `Uuid`
- `Duration`
- `UByte`, `UShort`, `UInt`, and `ULong`\*
- `Regex`
- `KClass<*>`

\*unsure if these get boxed to `Byte`, `Short`, `Int`, and `Long`, respectively. I'm like 75% sure they don't, but haven't tested this.

sksamuel:
Fixes #4470


scottdfedorov (approved):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/stable/StableIdents.kt
```diff
@@ -0,0 +1,133 @@
...
+            println("Warning, type $kclass used in data testing does not have a stable toString()")
+            false
+         }
+      }
+   }
+
+   internal fun isStableCollection(kclass: KClass<*>): Boolean {
+      return when (kclass) {
+         Map::class -> true
+         else -> false
```
AlexCue987: Is there a reason a `Map` is stable, but a `List` is not?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/stable/StableIdents.kt
```diff
@@ -0,0 +1,133 @@
...
+
+   /**
+    * Returns true if all members of the given [kclass] are "stable".
+    */
+   private fun hasStableMembers(kclass: KClass<*>, t: Any? = null) =
+      reflection.primaryConstructorMembers(kclass).let { members ->
+         members.isNotEmpty() && members.all { getter ->
+            val typeIsStable = isStable(getter.type)
+            if (t == null) {
+               typeIsStable
```
AlexCue987, line 116: my knee-jerk reaction is this: "if something is `null`, it surely is stable". What am I missing?

kotest-common/src/jvmTest/kotlin/io/kotest/engine/stable/StableIdentsTest.kt
```diff
@@ -0,0 +1,2 @@
+package io.kotest.engine.stable
+
```
AlexCue987: should we remove this empty file?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/stable/StableIdents.kt
sksamuel: T is the instance of the container, not a value inside the container. I guess you're right though. But that logic was just consolidated from elsewhere

AlexCue987:
it has conflicts

AlexCue987:
conflicts

kotest-common/src/jvmTest/kotlin/io/kotest/engine/stable/StableIdentsTest.kt
sksamuel: done

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/stable/StableIdents.kt
sksamuel: It wasn't stable, it's too hard to get it to work for small RoI. So I've removed it.

PR #4552, opened 2024-12-08
Added Arb.localeJdk17

### LINKED ISSUE #4154: Add locale Arb for jdk17
Apparently in jdk17 some locales have changed and java forcefully converts the old locales to the new ISO, which make my tests fail. As a workaround I'm doing `Arb.locale().filterNot { it.startsWith("in") || it.startsWith("ji") || it.startsWith("iw") || it.startsWith("no") }`
Maybe we could add another locales function to `Arb` that contains only jdk17 compliant locales.

sksamuel:
Closes #4154

### DIFF
--- a/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/locale.kt
+++ b/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/locale.kt
@@ -701,6 +701,9 @@ val locales = listOf(
    "zu_ZA",
 )
 
+private val jdk17SafeLocales =
+   locales.filterNot { it.startsWith("in") || it.startsWith("ji") || it.startsWith("iw") || it.startsWith("no") }
+
 /**
  * Returns an Arb that generates locales.
  *
@@ -708,3 +711,5 @@ val locales = listOf(
  * that has a variant.
  */
 fun Arb.Companion.locale() = arbitrary(listOf("en", "ca_ES_VALENCIA")) { locales.random(it.random) }
+
+fun Arb.Companion.localeJdk17() = arbitrary(listOf("en", "ca_ES_VALENCIA")) { jdk17SafeLocales.random(it.random) }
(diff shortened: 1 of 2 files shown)


scottdfedorov (approved):
Hmmm, interesting.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/locale.kt
```diff
@@ -701,10 +701,16 @@ val locales = listOf(
...
+
 /**
  * Returns an Arb that generates locales.
  *
  * Edgecases include a locale that only has a language (without country) and a locale
  * that has a variant.
  */
 fun Arb.Companion.locale() = arbitrary(listOf("en", "ca_ES_VALENCIA")) { locales.random(it.random) }
+
+fun Arb.Companion.localeJdk17() = arbitrary(listOf("en", "ca_ES_VALENCIA")) { jdk17SafeLocales.random(it.random) }
```
LeoColman, line 733:
```suggestion
/**
 * An Arb that produces locales filtering out locales that are not compatible with JDK 17.
 *
 * The excluded locales include those with language codes starting with:
 * - "in" (deprecated code for Indonesian, replaced by "id")
 * - "ji" (deprecated code for Yiddish, replaced by "yi")
 * - "iw" (deprecated code for Hebrew, replaced by "he")
 * - "no" (Norwegian, replaced by "nb" for Bokmål or "nn" for Nynorsk)
 *
 * This ensures the generated locales are safe for use in environments running JDK 17,
 * where these deprecated codes are no longer supported.
 *
 * Example edge cases include:
 * - A locale that only specifies the language (e.g., "en")
 * - A locale that includes a variant (e.g., "ca_ES_VALENCIA")
 *
 * @return an Arb instance that generates JDK 17-safe locales.
 */
fun Arb.Companion.localeJdk17() = arbitrary(listOf("en", "ca_ES_VALENCIA")) { jdk17SafeLocales.random(it.random) }
```

LeoColman: Adding some fluff

LeoColman (approved):

PR #4553, opened 2024-12-08
Workaround intellij bug on parent name prefixes #4137

### LINKED ISSUE #4137: Test Names aren't used Verbatim
I'm using the latest version of Kotest and am noticing that test names aren't used verbatim. Maybe there's a technical limitation here, but the string usage suggests that tests can be named anything, whereas in reality they cannot.

Shows how test names are being altered by Kotest.

```kotlin

class TestNaming : FunSpec({
    test("what about parenthesis ()") { true shouldBe true }
    test("other stuff 🙈 {} [] -–— works fine") { true shouldBe true }

    context("prefix") {
        test("prefixes - the context eats the prefix") { true shouldBe true }
    }
})
```

At least for me it's unexpected that parenthesis are being swallowed. Especially because they're very useful to disambiguate different overloads. Even more unexpected is that contexts eats common prefixes from the tests that are nested within.

Like the title states, I would expect that test names are taken verbatim without any alterations, if possible. If not possible the Kotest plugin should warn about problematic codepoints and include an explanation (or a link to an explanation) why they cannot be used.

sksamuel:
Fixes #4137

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/Materializer.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/Materializer.kt
@@ -68,9 +68,18 @@ class Materializer(private val configuration: ProjectConfiguration) {
     * Materializes a [NestedTest] into a runtime [TestCase] by attaching it to the given parent.
     */
    fun materialize(nested: NestedTest, parent: TestCase): TestCase {
+
+      // Note: intellij has a bug, where if a child test has a name that starts with the parent test name,
+      // then it will remove the common prefix from the child, to workaround this, we will add a dash at the
+      // start of the nested test to make the child nest have a different prefix.
+      // Also note: This only affects non-MPP tests, as MPP tests have the platform name added
+      val resolvedName = if (nested.name.testName.startsWith(parent.name.testName))
+         nested.name.copy(testName = "- " + nested.name.testName)
+      else nested.name
+
       return TestCase(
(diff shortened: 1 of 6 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/Materializer.kt
```diff
@@ -68,9 +68,18 @@ class Materializer(private val configuration: ProjectConfiguration) {
     * Materializes a [NestedTest] into a runtime [TestCase] by attaching it to the given parent.
     */
    fun materialize(nested: NestedTest, parent: TestCase): TestCase {
+
+      // Note: intellij has a bug, where if a child test has a name that starts with the parent test name,
+      // then it will remove the common prefix from the child, to workaround this, we will add a dash at the
+      // start of the nested test to make the child nest have a different prefix.
+      // Also note: This only affects non-MPP tests, as MPP tests have the platform name added
```
scottdfedorov, line 75: Can you link/ref the bug from their tracker so developers can check it later to see if it's been actioned?

scottdfedorov (approved):

LeoColman (approved):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/Materializer.kt
sksamuel: It's more an undocumented feature I guess

PR #4556, opened 2024-12-08
Add test for StringSpec data test

### LINKED ISSUE #3869: withData in StringSpec tests
It seems as if `withData` cannot be used in `StringSpec` tests, unless one accepts that each individual case gets its name assigned automatically. The following, which should work, doesn't compile, because Kotlin cannot infer the correct types.

```kotlin
class DataTest : StringSpec({
    "what this test tests" {
        withData(
            "a" to 1,
            "b" to 2,
            "c" to 3,
        ) { (first, second) ->
            // fails with runtime error
        }
    }

    withData(
        { (first, second) -> "$first <> $second" },
        "a" to 1,
        "b" to 2,
        "c" to 3,
    ) { (first, second) ->
        // doesn't compile, type inference issues
    }
})
```

The `@BuilderInference` would fix the second issue:

```kotlin

@OptIn(ExperimentalTypeInference::class)
@BuilderInference
fun <T> RootScope.withData( // ...
```

sksamuel:
Closes #3869


scottdfedorov (approved):

LeoColman (approved):

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/datatest/styles/StringSpecDataTest.kt
```diff
@@ -0,0 +1,26 @@
...
+      val results = registerRootTests()
+      var count = 0
+
+      afterTest {
+         count++
+      }
+
+      afterSpec {
+         results.assertDataTestResults()
+         count shouldBe 36
```
AlexCue987, line 23: will this `36` be fragile, change every time we change `registerRootTests`? If yes, is there a way to match to some number computed off of `registerRootTests`?

sksamuel: I don't think we change these tests often enough for it to be a worry

PR #4568, opened 2024-12-14
Launch before/after spec on spec level coroutines

### LINKED ISSUE #3759: Kotest >= 5.7.0: Jobs launched in beforeSpec can delay or hang the first test

sksamuel:
Fixes #3759

@OliverO2 Would really appreciate a review from you on this to see if this is the right approach.


AlexCue987:
conflicts

sksamuel:
Will fix conflicts. I need to propagate errors too, so this is not quite ready yet, but wanting to get validation from O2 that it's along the right lines.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/SpecExtensions.kt
```diff
@@ -40,18 +44,21 @@ internal class SpecExtensions(private val registry: ExtensionRegistry) {
...
+   fun beforeSpec(spec: Spec, specCoroutineContext: CoroutineContext): Result<Spec> {
       logger.log { Pair(spec::class.bestName(), "beforeSpec $spec") }
-
-      val errors = extensions(spec)
-         .filterIsInstance<BeforeSpecListener>()
+      val scope = CoroutineScope(specCoroutineContext + CoroutineName("before-spec-scope"))
+      val errors = extensionsOf<BeforeSpecListener>(spec)
          .mapNotNull { ext ->
-            runCatching { ext.beforeSpec(spec) }
+            runCatching { scope.launch { ext.beforeSpec(spec) } }
```
OliverO2, line 61:
```suggestion
            runCatching { withContext(scope) { ext.beforeSpec(spec) } }
```

`runCatching` waits. Launching an extra job would obscure the fact that we're sequentially executing a suspend function.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/SpecExtensions.kt
```diff
@@ -77,9 +84,9 @@ internal class SpecExtensions(private val registry: ExtensionRegistry) {
             if (it.isInitialized()) it.value.close() else Unit
          }
       }
-
-      val errors = extensions(spec).filterIsInstance<AfterSpecListener>().mapNotNull { ext ->
-         runCatching { ext.afterSpec(spec) }
+      val scope = CoroutineScope(specCoroutineContext + CoroutineName("after-spec-scope"))
+      val errors = extensionsOf<AfterSpecListener>(spec).mapNotNull { ext ->
+         runCatching { scope.launch { ext.afterSpec(spec) } }
```
OliverO2:
```suggestion
         runCatching { withContext(scope) { ext.afterSpec(spec) } }
```

As above.

OliverO2 (commented):
As I understand it, it seems like the general approach could work, except on JS:

* The `*SpecRunner` runners now store a `coroutineContext` for the Spec.
* I can trace back invocations to `SpecExecutor.execute`, then the call hierarchy fans out across numerous `intercept` calls, as well as one `SpecRefInterceptorPipeline.execute` call.
* It seems that the `coroutineContext` for some invocations stems from executing the spec, this seems right.
* In other invocations, the `coroutineContext` stems from registering a `TestCase`, which might have a life cycle different from executing a spec. However, these cases (`InstancePerLeafSpecRunner` and `InstancePerTestSpecRunner`) seem to combine registration with immediate execution, so the `coroutineContext`'s life cycle seems OK. 
* With `KotlinJsTestSpecExecutorDelegate`, I'm not convinced that it could work at all: We're now injecting a `coroutineContext` from a test suite into a test. However, the test suite is for registration only and always executes synchronously. Its life cycle ends immediately after all of its tests have been registered. Also, we're breaking structured concurrency when we assign a new parent coroutine inside a test's `GlobalScope.testFunctionPromise` invocation. The underlying JS frameworks expect to directly control each async test, which must signal its completion by resolving the `Promise`. If we make another coroutine the parent of a test, how can the Promise still resolve when the test completes?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/SpecExtensions.kt
OliverO2: Correction: `runCatching` is not a coroutine scoping function. In this case it would just check that the launch succeeds (without anything running at this time). This is not what's intended, right?

sksamuel: Correct, no.

sksamuel: withContext(scope)  doesn't seem to compile because it expects a context not a scope.

OliverO2: Right, the whole concept cannot work. There would be two "parents" expecting the same coroutine (`ext.beforeSpec(spec)`) to complete:
1. The parent coroutine `specCoroutineContext`.
2. The blocking scope (`runCatching`, `mapNotNull`, ...).

sksamuel: Makes sense. I am going to re-work this now I merged the spec runners together and simplified that part of the stack a lot.

LeoColman:
Closing this as the whole spec thing got reworked.
Feel free to reopen

PR #4585, opened 2024-12-15
Migrate test clock to main repo

sksamuel:

### DIFF
--- a/kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/clock/TestClock.kt
+++ b/kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/clock/TestClock.kt
@@ -1,0 +1,49 @@
+package io.kotest.extensions.clock
+
+import java.time.Clock
+import java.time.Instant
+import java.time.ZoneId
+import java.time.ZoneOffset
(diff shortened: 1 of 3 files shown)


kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/clock/TestClock.kt
```diff
@@ -0,0 +1,49 @@
...
+      return TestClock(instant, zone)
+   }
+
+   override fun getZone(): ZoneId = zone
+
+   /**
+    * Sets the instant in this test clock to the given value.
+    */
+   fun setInstant(instant: Instant) {
+      this.instant = instant
```
AlexCue987, line 33: do we need this thing to be thread safe, and use `AtomicReference`?

AlexCue987 (approved):

kotest-extensions/src/jvmMain/kotlin/io/kotest/extensions/clock/TestClock.kt
sksamuel: That's up to the caller. I think its best to keep this simple.

PR #4587, opened 2024-12-15
Add SpringExtension that combines existing into a single extension

sksamuel:
Now that @ApplyExtension supports test case extensions and constructor extensions we can have a single easy to use spring extension


AlexCue987 (approved):
do we need any changes in the docs?

sksamuel:
Yeah we'll need a bunch of docs prs before 6.0 is released

PR #4594, opened 2024-12-19
Update extensions.md

jlous:
add link to project config

### DIFF
--- a/documentation/versioned_docs/version-5.9/framework/extensions/extensions.md
+++ b/documentation/versioned_docs/version-5.9/framework/extensions/extensions.md
@@ -37,7 +37,7 @@ class TestSpec : WordSpec({
 Any extensions registered inside a `Spec` will be used for all tests in that spec (including [test factories](../test_factories.md) and nested tests).
 
 To run an extension for every spec in the entire project you can either mark the listener with `@AutoScan`,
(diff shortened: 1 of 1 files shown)


documentation/versioned_docs/version-5.9/framework/extensions/extensions.md
```diff
@@ -37,7 +37,7 @@ class TestSpec : WordSpec({
 Any extensions registered inside a `Spec` will be used for all tests in that spec (including [test factories](../test_factories.md) and nested tests).
 
 To run an extension for every spec in the entire project you can either mark the listener with `@AutoScan`,
-or you can register the listener via project config.
+or you can register the listener via [project config](../project_config.md).
```
AlexCue987, line 40: Looks good to me, but can you also apply the same change to the origin at documentation/framework/extensions/extensions.md, so that it propagates to all next versions

AlexCue987 (changes requested):

documentation/versioned_docs/version-5.9/framework/extensions/extensions.md
sksamuel: I'll do that change in a separate push just so we can merge this now.

PR #4599, opened 2024-12-21
Promote withData to first class support

### LINKED ISSUE #3290: beforeEach/afterEach not working with data driven testing

### LINKED ISSUE #3753: Data-driven testing (`withData`) with `BehaviourSpec` doesn't appear to work
**Please consider asking your question on our slack channel #kotest or in stack overflow (and tagging with kotest). The github issue tracker is for bugs and feature requests.**

I'm trying to combine some data driven tests with other tests within a single `Given` context under a `BehaviourSpec`. However, I can't seem to make it work.

The docs say: "Data tests can only be defined at the root or in container scopes. They cannot be defined inside leaf scopes." (https://kotest.io/docs/framework/datatesting/data-driven-testing.html). However, when trying to define a `withData` block inside a `Given` block (isn't this a root or container scope?), I can't get it to compile.

```kotlin
class MyTests :
    BehaviorSpec({
      Given("Some Numbers") {
        withData(1, 2, 3) { (a, b, c) ->
          { When("Adding") { Then("Correct!") { a + b + c shouldBe 6 } } }
          { When("Multiplying") { Then("Yes!") { a * b * c shouldBe 6 } } }
        }
        When("doing something else") { Then("hurray!") { 4 + 5 shouldBe 9 } }        
      }
}
```

Gives compile errors of:
```
Destructuring declaration initializer of type Int must have a 'component1()' function
Destructuring declaration initializer of type Int must have a 'component2()' function
Destructuring declaration initializer of type Int must have a 'component3()' function
Suspension functions can be called only within coroutine body
```

Obviously, the simple `FunSpec` example works fine:

```kotlin
class MyTests :
    FunSpec({
      context("Given some numbers, they're all positive") {
         withData(1, 2, 3) { i -> run { i shouldBeGreaterThan 0 } } 
      }
    })
```

What are the rules on where `withData` can be used? Is there a way to do data-driven testing with `BehaviourSpec`?

### LINKED ISSUE #4003: Revert WithData to testContainer level
Before 5 version kotest have ability writing steps in withData section. For example:

```kotlin
"Test withData" - {
  withData(
    listOf(
      PythagTriple(3, 4, 5),
      PythagTriple(6, 8, 10),
    )
  ){ (a, b, c) ->
   "Do something" {
       ...
    }
   "Check isPythagTriple" {
       isPythagTriple(a, b, c) shouldBe true
    }
  }
}
```

But in 5.8 version this code drop error `suspend infix operator fun String.invoke(test: suspend FreeSpecTerminalScope.() -> Unit): Unit' can't be called in this context by implicit receiver. Use the explicit one if necessary`

Please revert to old behaviour. It's more usable.

sksamuel:
* Allows using the DSL structure inside withData, eg further tests can be defined inside the withData blocks using the same elements as the regular DSL.
* Removes TestType.Dynamic enitrely, so callback semantics are aligned with test/container
* The syntax is backwards compatible.

Closes #4003
Closes #3753
Closes #3290


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/BehaviorSpecGivenContainerScope.kt
```diff
@@ -94,4 +106,71 @@ class BehaviorSpecGivenContainerScope(
...
+    * Registers tests inside the given [FunSpecContainerScope] for each element of [ts].
+    * The test name will be generated from the given [nameFn] function.
+    */
+   suspend fun <T> withData(
+      nameFn: (T) -> String,
+      @BuilderInference ts: Iterable<T>,
+      @BuilderInference test: suspend BehaviorSpecGivenContainerScope.(T) -> Unit
+   ) {
+      ts.forEach { t -> addAnd(nameFn(t), false) { test(t) } }
+   }
```
Kantis: Why is `@BuilderInference` required here, but not on the [similar signature](https://github.com/kotest/kotest/pull/4599/files?diff=split&w=0#diff-0891129a199de5858e0d80f2f27ec463938bb8ab096088592560846b7756e15aR97) in `BehaviorSpecContextContainerScope`?

Kantis (commented):
Awesome. I think this will be a huge improvement to data-driven tests.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/BehaviorSpecGivenContainerScope.kt
sksamuel: Just a copy n paste thing from the existing withData.
I've removed them.

Kantis (approved):

PR #4602, opened 2024-12-22
Sanitize test path's using TestPathBuilder

### LINKED ISSUE #2981: Excluded when the escape character is used in the `DescribeSpec`
```kotlin
plugins {
    kotlin("jvm") version "1.6.21"
    id("org.jlleitschuh.gradle.ktlint") version "10.3.0"
}

dependencies {
    testImplementation("org.junit.jupiter", "junit-jupiter", "5.8.2")
    testImplementation("org.assertj", "assertj-core", "3.22.0")
    testImplementation("io.kotest", "kotest-runner-junit5", "5.2.3")
}
```

When I used the escape character in the test, I had the experience of being excluded from execution. The example code below is a sample. Circle Symbols are at each execution point.
🟢  : run 
⚪  : excluded

```kotlin

class EscapeStringFilterTest : DescribeSpec({ // 🟢


    describe("include escape string \\n"){ // ⚪
        context("something..."){ // ⚪
            it("something..."){ // ⚪
                true shouldBe true
            }
        }
    }

    describe("something..."){ // 🟢
        context("include escape string \\n"){ // ⚪
            it("something true"){ // ⚪
                true shouldBe true
            }
        }

        context("something ..."){  // 🟢
            it("include escape string \\n"){  // 🟢
                true shouldBe true
            }
        }
    }
})

```

### LINKED ISSUE #3966: Test names containing line breaks are ignored when run individually
5.0.0 and greater.

I believe this is both a Kotest and a Kotest Intellij Plugin issue, so I have also created an issue there: https://github.com/kotest/kotest/issues/5365

The problem appears to have been introduced in Kotest 5.0.0 from my testing.

I have seen the issue with Kotest versions 5.0.0, 5.5.5 and 5.81, Kotlin versions 1.6, 1.7 and 1.9 as well as JVM 11 and 17. The issue does not occur with Kotest version 4.6.4.

An example file:

```
import io.kotest.core.spec.style.BehaviorSpec

class MainTests : BehaviorSpec({
    Given("If you run from this Given, the test will run") {
        println("Given test 1")
        When(
            """
            But if you run from this When
            with line breaks,
            the test will be ignored
            """
        ) {
            println("When test 1")
            Then("It will also be ignored if you try to run it from here") {
                println("Then test 1")
            }
        }
    }
    Given("""A test case
        with a line break
        at a Given will also be ignored
        """
    ) {
        println("Given test 2")
        When("And same if you try to run it from here") {
            println("When test 2")
            Then("Same here") {
                println("Then test 2")
            }
        }
    }
    Given("""Test names with line breaks
        that trim any part of the string
        don't even get the run button
    """.trimMargin()) {
        println("Given test 3")
        When("And this test will be ignored even if you run it from here") {
            println("When test 3")
            Then("Same at this spot") {
                println("Then test 3")
            }
        }
    }
    Given("""Line
        breaks
        """
    ) {
        println("Given test 4")
        When("Shorter test name") {
            println("When test 4")
            Then("Shorter test name also") {
                println("Then test 4")
            }
        }
    }
})
```

Expected behaviour (as seen in 4.6.4) (the individual test runs):
image

Behaviour with kotest version 5.X (all tests are ignored):
image
the output:
```
C:\Users\sdawson\.jdks\corretto-17.0.10\bin\java.exe -Didea.test.cyclic.buffer.size=1048576 "-javaagent:C:\Program Files\JetBrains\IntelliJ IDEA Community Edition 2023.1.2\lib\idea_rt.jar=60675:C:\Program Files\JetBrains\IntelliJ IDEA Community Edition 2023.1.2\bin" -Dfile.encoding=UTF-8 @C:\Users\sdawson\AppData\Local\Temp\idea_arg_file1953867495 io.kotest.engine.launcher.MainKt --spec MainTests --testpath "A test case
        with a line break
        at a Given will also be ignored
        " --reporter teamcity
Testing started at 7:30 PM ...

If you run from this Given, the test will run is excluded by test filter(s): Excluded by test path filter: 'A test case
        with a line break
        at a Given will also be ignored
        '

A test case with a line break at a Given will also be ignored is excluded by test filter(s): Excluded by test path filter: 'A test case
        with a line break
        at a Given will also be ignored
        '

Test names with line breaks that trim any part of the string don't even get the run button is excluded by test filter(s): Excluded by test path filter: 'A test case
        with a line break
        at a Given will also be ignored
        '

Line breaks is excluded by test filter(s): Excluded by test path filter: 'A test case
        with a line break
        at a Given will also be ignored
        '

```

sksamuel:
Fixes #3966
Fixes #2981

### DIFF
--- a/kotest-common/src/commonMain/kotlin/io/kotest/common/testPath.kt
+++ b/kotest-common/src/commonMain/kotlin/io/kotest/common/testPath.kt
@@ -5,7 +5,11 @@ import kotlin.coroutines.CoroutineContext
 
 /**
  * A [TestPath] is a unique flattened string identifier for a test.
- * Note: A test path can include an optional spec name.
+ *
+ * They begin with the spec fully qualified name and then include the hierarchy of test names
+ * separated by a delimiter.
+ *
+ * For example, a test path might be "io.kotest.MyTest/test case parent -- test case"
  */
 data class TestPath(val value: String) {
    init {
--- a/kotest-common/src/commonMain/kotlin/io/kotest/common/testPath.kt
+++ b/kotest-common/src/commonMain/kotlin/io/kotest/common/testPath.kt
@@ -3,7 +3,15 @@ package io.kotest.common
 import kotlin.coroutines.AbstractCoroutineContextElement
 import kotlin.coroutines.CoroutineContext
 
-data class TestPath(val value: String)
+/**
+ * A [TestPath] is a unique flattened string identifier for a test.
+ * Note: A test path can include an optional spec name.
+ */
+data class TestPath(val value: String) {
+   init {
+      require(value.trim() == value) { "TestPath cannot have leading or trailing whitespace" }
+   }
+}
 
 @ExperimentalKotest
 class TestPathContextElement(
--- a/kotest-extensions/kotest-extensions-allure/src/jvmMain/kotlin/io/kotest/extensions/allure/AllureWriter.kt
+++ b/kotest-extensions/kotest-extensions-allure/src/jvmMain/kotlin/io/kotest/extensions/allure/AllureWriter.kt
@@ -1,8 +1,8 @@
 package io.kotest.extensions.allure
 
+import io.kotest.common.TestPath
 import io.kotest.core.config.ProjectConfiguration
 import io.kotest.core.descriptors.Descriptor
-import io.kotest.core.descriptors.TestPath
 import io.kotest.core.test.TestCase
 import io.kotest.core.test.TestResult
 import io.kotest.engine.test.names.DefaultDisplayNameFormatter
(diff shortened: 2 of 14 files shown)


Kantis (approved):
Perhaps this goes outside the scope of this PR, but it feels a bit odd that the `TestPath` _can_ include a Spec, but the `TestPathTestCase` filter takes an explicit Spec reference. Why not always bake the Spec into the TestPath?

sksamuel:
> Perhaps this goes outside the scope of this PR, but it feels a bit odd that the `TestPath` _can_ include a Spec, but the `TestPathTestCase` filter takes an explicit Spec reference. Why not always bake the Spec into the TestPath?

Yeah I might tweak that. What I'm trying to do here is centralize the logic to sanitize the test paths so multiline and escaped paths have their code in one place.

sksamuel:
replaced by https://github.com/kotest/kotest/pull/4714

PR #4607, opened 2024-12-23
Alex/playback 

AlexCue987:
add two simple functions that make building test-double functions easier. it's a very lightweight alternative to mocking that I'd like to use inside kotest itself. but also let's expose it - its' a simple and useful tool for functional programming


AlexCue987:
@LeoColman @sksamuel do you have any comments/concerns?

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/playback/PlaybackElements.kt
```diff
@@ -0,0 +1,45 @@
...
+class PlaybackElements<T>(
+   source: Sequence<T>
+) {
+   private val iterator = source.iterator()
+
+   fun next(): T = iterator.next()
+}
+
+/**
+ * Provides a fake function that returns elements from a sequence.
```
LeoColman, line 13: This smells a bit like an `Arb`, doesn't it?

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/playback/PlaybackOrThrow.kt
```diff
@@ -0,0 +1,53 @@
...
+
+   fun next(): T = iterator.next().getOrThrow()
+}
+
+/**
+ * Provides a fake function that returns elements or throw exceptions from a sequence.
+ *
+ * Example:
+ *
+ *          val fakeFunction = sequenceOf(
```
LeoColman, line 16: Nit: Formatting is weird on this part

LeoColman (approved):

LeoColman:
My biggest concern about this is that it kinda mimics what we do in Arbitraries.

I'd rather have users using Arbs than converting their lists to functions.

----

Another problem I see is that this is more on the Mocking/Stubbing side than on Kotest's, and we might open a pandora's box.

I think this is one of the oldest discussions in Kotest! https://github.com/kotest/kotest/pull/229 

I'm a bit afraid that this is the first step to a full-fledged mocking library

AlexCue987:
> I'm a bit afraid that this is the first step to a full-fledged mocking library

The thing is: after ditching a dependency injection framework, that is SpringBoot in our case, we no longer need a mocking library with all its bells and whistles. When all the dependencies are functions, these two simple methods are literally all we need. So we got rid of `mockk` along with SpringBoot and never looked back.

AlexCue987:
> I'd rather have users using Arbs than converting their lists to functions.

TBH I do not understand how to accomplish that. For instance, I need to make sure my wall clock returns `2025-01-01-17-29.12.098` on first call, and `2025-01-01-17-29.15.347` on the second one. How can I implement that with Arbs?

sksamuel:
This has broken master so I'm going to revert it as I am busy in my own PRs. Please make another PR when it's fixed and we can merge it back in.

Also, try to avoid titles like this, it doesn't mean anything in the git history to see "alex/playback", just an FYI !
Thanks.

sksamuel:
Never mind, later builds went green.

AlexCue987:
> This has broken master so I'm going to revert it as I am busy in my own PRs. Please make another PR when it's fixed and we can merge it back in.
> 
> Also, try to avoid titles like this, it doesn't mean anything in the git history to see "alex/playback", just an FYI ! Thanks.

The error seems unrelated to my changes: "tests that timeout during a blocking operation should still run the 'after test' listeners".

Yeah, I concur,  the commit message could be better.

PR #4610, opened 2024-12-24
Fix typos

JuHyun419:
- Fix typos in discoveryRequest.kt


OliverO2 (approved):
Good catch, thank you!

PR #4612, opened 2024-12-28
Document and add tests to nulls.kt

LeoColman:
This commit adds javadocs to all matchers in nulls.kt; it also adds test
 cases for all the matchers in the file.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/nulls.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/nulls.kt
@@ -5,114 +5,446 @@ import io.kotest.matchers.MatcherResult
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldNot
 
+/**
+ * Asserts that this iterable contains only null elements.
+ *
+ * Verifies that all elements in the iterable are null.
+ *
+ * Opposite of [shouldNotContainOnlyNulls].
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/nulls.kt
```diff
@@ -5,114 +5,446 @@ import io.kotest.matchers.MatcherResult
...
+ *
+ * Opposite of [shouldContainOnlyNulls].
+ *
+ * Example:
+ * ```
+ * listOf(null, "a").shouldNotContainOnlyNulls()  // Assertion passes
+ * listOf(null, null).shouldNotContainOnlyNulls() // Assertion fails
+ * ```
+ *
+ * @return the same iterable for further assertions
```
AlexCue987, line 81: I know this is preexisting, but because Collection inherits from Iterable, why do we need both methods? Should we deprecate the extension method on Collection?

AlexCue987 (approved):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/nulls.kt
LeoColman: I think I asked you this a few months ago 😂 

I have no problem with plain removing it. It won't change user's code

PR #4613, opened 2024-12-31
Drop Nulls Collection Matchers

LeoColman:
This commit removes the redundant functions for Collections. These
functions are already covered by Iterable, as Collection is a subtype of
 Iterable.

This change does binary changes, but we don't need to deprecate. Users
won't need to make any changes to their tests, as the compiler will
simply pick up the new function signatures.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/nulls.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/nulls.kt
@@ -20,10 +20,7 @@ import io.kotest.matchers.shouldNot
  *
  * @return the same iterable for further assertions
  */
-fun <T, I : Iterable<T>> I.shouldContainOnlyNulls(): I {
-   toList().shouldContainOnlyNulls()
-   return this
-}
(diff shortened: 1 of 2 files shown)


LeoColman:
@Kantis @sksamuel Could you give your opinion on this one, please?

i'm a bit scared of my conclusions about the change being fine

LeoColman:
Sorry @Kantis , meant to ping @AlexCue987

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/nulls.kt
```diff
@@ -401,30 +267,7 @@ fun <T, I : Iterable<T>> I.shouldNotContainNoNulls(): I {
...
- * Opposite of [shouldContainNoNulls].
- *
- * Example:
- * ```
- * listOf("a", null).shouldNotContainNoNulls() // Assertion passes
- * listOf("a", "b").shouldNotContainNoNulls()  // Assertion fails
- * ```
- *
- * @return the same collection for further assertions
- */
```
AlexCue987, line 423: As we are keeping the public functions as is, and only removing duplicate implementations, maybe we should keep the javadoc examples?

AlexCue987 (approved):

sksamuel (approved):
I  concur, should be  no-op to users

PR #4614, opened 2024-12-31
Add Javadocs to methods in unique.kt

LeoColman:
Also did some code-golfing to reduce functions to one-liners

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/unique.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/unique.kt
@@ -5,139 +5,251 @@ import io.kotest.matchers.MatcherResult
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldNot
 
-fun BooleanArray.shouldBeUnique(): BooleanArray {
-   asList() should beUniqueByEquals("BooleanArray")
-   return this
-}
-
-fun BooleanArray.shouldNotBeUnique(): BooleanArray {
-   asList() shouldNot beUniqueByEquals("BooleanArray")
-   return this
-}
-
-fun ByteArray.shouldBeUnique(): ByteArray {
-   asList() should beUniqueByEquals("ByteArray")
-   return this
-}
-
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/unique.kt
```diff
@@ -5,139 +5,251 @@ import io.kotest.matchers.MatcherResult
...
+ * Asserts that the given [Array] contains no duplicate elements using the provided
  * [comparator] for equality.
  *
- * @return the input instance is returned for chaining
+ * This assertion uses the specified [comparator] to compare elements for uniqueness.
+ *
+ * Example:
+ * ```
+ * val array = arrayOf("apple", "banana", "APPLE")
+ * array.shouldBeUnique(String.CASE_INSENSITIVE_ORDER)   // Assertion passes
```
AlexCue987: should it be `CASE_SENSITIVE_ORDER`? because it's not unique when `"APPLE" == "apple"`

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/unique.kt
```diff
@@ -5,139 +5,251 @@ import io.kotest.matchers.MatcherResult
...
  * Asserts that the given [Iterable] contains no duplicate elements using the given
  * [comparator] for equality.
  *
- * @return the input instance is returned for chaining, maintaining the input type
+ * This assertion uses the provided [comparator] to compare elements for uniqueness.
+ *
+ * Example:
+ * ```
+ * val list = listOf("apple", "banana", "APPLE")
+ * list.shouldBeUnique(String.CASE_INSENSITIVE_ORDER)   // Assertion passes
```
AlexCue987: same suggestion here: let's make it case sensitive

AlexCue987 (changes requested):
Overall LGTM, but two examples look wrong to me

AlexCue987 (approved):

LeoColman:
@AlexCue987 both examples fixed and also added tests to validate them.

There is no "ConsiderCase", so I had to improvise

PR #4624, opened 2025-01-01
Removed @KotestInternal from funcs/classes that can simply be internal

sksamuel:


AlexCue987 (approved):

sksamuel:
Will  come back to this once other changes are merged

PR #4626, opened 2025-01-02
Rework isolation modes and concurrency

### LINKED ISSUE #3311: Add fine-grained and unambiguous control over parallelism in nested tests
I just asked [ChatGPT](https://chat.openai.com/chat) if there is a way to parallelize nested tests. It came up with an interesting feature which could be useful for giving fine-grained control over which tests and containers should be run in parallel in a Spec with multiple layers, e.g. a `FreeSpec`. 
The idea, in short, is to add a `parallel {}` block to the DSL that makes all specs and tests inside run in parallel.
Maybe `parallel {}` and `sequential {}` could be used to override the intherited/global behaviour one way or another. Maybe it's easier to just add a config flag (`"test".config(parallel=true) - { /*...*/}`). However it would be implemented, I'm missing this level of fine-grained control over parallelism in Kotest. The current situation could be improved, in my opinion.
`AbstractProjectConfig.parallelism`, `AbstractProjectConfig.concurrentSpecs`, `AbstractProjectConfig.concurrentTests`, `Spec.concurrency`, `Spec.threads` all exist, but none seem to do what I want and the documentation is not very clear about all of them. Specifically, after a certain level, containers don't seem to run in parallel anymore, not matter what I do.

ChatGPT seems to think the feature already exists, which I could not find any evidence for.
Here is the full exchange between me and the AI (text version below):

image
image
image
image

> **Me:**
> 
> 
> 
>  How to run nested tests in Kotest in parallel?
> 
> **ChatGPT:**
> 
>  To run nested tests in Kotest in parallel, you can use the ``parallel`` method, which is available on all test classes and test containers. This method allows you to specify that the tests within a given test class or container should be run in parallel.
> 
> Here is an example of how to use the ``parallel`` method to run nested tests in parallel:
> 
> ```kotlin
> class MyTest : StringSpec() {
> init {
> "test case 1" {
> // some code here
> }
> "test case 2" {
> // some code here
> }
> "test case 3" {
> // some code here
> }
> 
>         parallel {
>             "test case 4" {
>                 // some code here
>             }
>             "test case 5" {
>                 // some code here
>             }
>         }
>     }
> }
> ```
> In the above example, the ``test case 1``, ``test case 2``, and ``test case 3`` tests will be run sequentially, in the order they are defined. However, the ``test case 4`` and ``test case 5`` tests will be run in parallel, meaning that they will be executed concurrently.
> 
> Note that when running tests in parallel, you need to take care to avoid race conditions and other synchronization issues. For example, if multiple tests are accessing the same shared resources, you may need to use synchronization mechanisms, such as locks or semaphores, to ensure that the tests do not interfere with each other.
> 
> Additionally, when running tests in parallel, you should consider the potential performance implications. Running tests in parallel can significantly increase the overall execution time of your test suite, as the tests will be competing for system resources. You should carefully evaluate the benefits and drawbacks of running tests in parallel before using this feature in your test suite.
> 
> 
> 
> 
> **Me:**
> 
> 
> Which version of Kotest was that feature added?
> 
> **ChatGPT:**
> 
>  
> The ``parallel`` method was added in Kotest 4.1.0. This method allows you to run nested tests in parallel, as described in the previous answer.
> 
> Kotest is an open-source testing framework for Kotlin, and it is actively developed and maintained by the Kotlin community. New features and enhancements are regularly added to Kotest, and the latest version is always available on the project's GitHub page: https://github.com/kotest/kotest
> 
> To use the ``parallel`` method, you will need to use Kotest 4.1.0 or later. You can check the version of Kotest that you are currently using by checking the ``kotest-core`` dependency in your project's ``build.gradle`` file, or by running the ``kotest --version`` command. If you are using an older version of Kotest, you can update to the latest version by following the instructions on the project's GitHub page.
> 
> 
> 
> 
> **Me:**
> 
> 
> Is it still available in the latest version of Kotest?
> 
> **ChatGPT:**
> 
>  Yes, the ``parallel`` method is still available in the latest version of Kotest. This method was added in Kotest 4.1.0, and it has been retained in all subsequent versions of the framework.
> 
> As of the time of writing, the latest version of Kotest is 5.1.0. You can use this version of Kotest to run nested tests in parallel, using the ``parallel`` method, as described in the previous answer.
> 
> To use the latest version of Kotest, you will need to update the ``kotest-core`` dependency in your project's ``build.gradle`` file. The exact steps for doing this will depend on the build tool and version management system you are using, but in general, you should be able to update the dependency by running a command similar to the following:
> 
> ```shell
> ./gradlew updateDependencies
> ```
> 
> This will update all of the dependencies in your project to their latest versions, including the ``kotest-core`` dependency. Once the dependencies have been updated, you can use the ``parallel`` method in your tests to run nested tests in parallel.
> 
> Alternatively, you can check the Kotest project's GitHub page (https://github.com/kotest/kotest) for instructions on how to download and install the latest version of the framework. This will allow you to use the latest features and enhancements in Kotest, including the ``parallel`` method.
> 
> 
> 
> 
> **Me:**
> 
> 
> Can you give me the link to the documentation for "parallel"?
> 
> **ChatGPT:**
> 
>  The ``parallel`` method is documented on the Kotest website, in the "Concurrency" section of the documentation. You can find the documentation for this method at the following link:
> 
> https://kotest.io/4.1/core/core/#parallel
> 
> The ``parallel`` method is part of the core Kotest framework, and it is available on all test classes and test containers. This method allows you to specify that the tests within a given test class or container should be run in parallel, rather than being run sequentially.
> 
> The ``parallel`` method takes an optional ``threads`` parameter, which specifies the number of threads to use when running the tests in parallel. If this parameter is not specified, Kotest will use a default number of threads, based on the number of available CPU cores.
> 
> Here is an example of how to use the ``parallel`` method to run tests in parallel:
> 
> ```kotlin
> class MyTest : StringSpec() {
> init {
> "test case 1" {
> // some code here
> }
> "test case 2" {
> // some code here
> }
> "test case 3" {
> // some code here
> }
> 
>         parallel {
>             "test case 4" {
>                 // some code here
>             }
>             "test case 5" {
>                 // some code here
>             }
>         }
>     }
> }
> ```
> In the above example, the ``test case 1``, ``test case 2``, and ``test case 3`` tests will be run sequentially, in the order they are defined. However, the ``test case 4`` and ``test case 5`` tests will be run in parallel, meaning that they will be executed concurrently.
> 
> You can find more detailed information about the ``parallel`` method, as well as examples and usage instructions, on the Kotest website.
> 
> 
> 
> 
> **Me:**
> 
> 
> Did you just make that up? I cannot find and documentation about the parallel method anywhere...
> 
> **ChatGPT:**
> 
>  I apologize if the information I provided was not accurate. As a large language model trained by OpenAI, I do not have access to the internet and I am unable to browse websites or search for information online. I am only able to provide information based on my pre-existing knowledge, which has a fixed cutoff point.
> 
> In this case, it appears that the information I provided about the ``parallel`` method in Kotest may not be accurate, as I cannot find any documentation about this method on the Kotest website or elsewhere. It is possible that I confused the ``parallel`` method with a similar feature in another testing framework, or that the information I provided was based on outdated or incorrect information.
> 
> I apologize for any confusion or inconvenience this may have caused. Please let me know if you have any other questions, and I will do my best to provide accurate and helpful information based on my pre-existing knowledge.
> 
> 
> 
> 
> **Me:**
> 
> 
> Can you tell me the hash of the git commit in which the feature was added?
> 
> **ChatGPT:**
> 
>  As a large language model trained by OpenAI, I do not have access to the internet and I am unable to browse websites or search for information online. I am also unable to access specific files or commits in a Git repository. I am only able to provide general information based on my pre-existing knowledge, which has a fixed cutoff point.
> 
> In this case, I am not able to provide the hash of the Git commit in which the ``parallel`` method was added to Kotest, as I do not have access to the Kotest repository or the commit history. Additionally, the information I provided earlier about the ``parallel`` method in Kotest may not be accurate, as I am unable to confirm its existence or availability.
> 
> I apologize for any confusion or inconvenience this may have caused. Please let me know if you have any other questions, and I will do my best to provide accurate and helpful information based on my pre-existing knowledge.
> 
> 
> 
> 
> **Me:**
> 
> 
> It's fine, if anything it is incredible that you can make stuff up like this. Maybe you should pursue a creative career?
> 
> **ChatGPT:**
> 
>  Thank you for your kind words. As a large language model trained by OpenAI, I am not capable of making things up or creating new information. My responses are based on my pre-existing knowledge, which has a fixed cutoff point. I do not have the ability to browse the internet or access new information, and I am not capable of creative thought or expression.
> 
> My purpose is to provide accurate and helpful information to the best of my ability, based on the knowledge I have been trained on. I am not capable of pursuing a career or engaging in any other activities beyond answering questions. I am a purely digital entity, and I exist only to assist users with their questions and inquiries.
> 
> I am glad that I was able to help you, even if the information I provided was not accurate. Please let me know if you have any other questions, and I will do my best to provide helpful and accurate answers based on my pre-existing knowledge.

### LINKED ISSUE #3962: Lifecycle hooks with mockk do not work properly on `object` and java classes with static functions
`5.8.1`
**Which version of mockk are you using**
`1.13.10`

Steps to reproduce:
Creating an object which will be mocked for some tests:
```kotlin
object TestToMock {
    fun intSum(a: Int, b: Int): Int = a + b
}
```
or a java file with a static function:
```java
public class TestStaticOnly {
    private TestStaticOnly() {
    }

    public static int intSum(int a, int b) {
        return a + b;
    }
}
```

Having 2 specs which use this object or java static function:
- mocked with the `beforeSpec` and unmocked via `afterSpec`
- without mocking

```kotlin
class MockedTest : DescribeSpec({
    beforeSpec {
        mockkObject(TestToMock)
        mockkStatic(TestStaticOnly::class)
    }
    afterSpec {
        unmockkObject(TestToMock)
        unmockkStatic(TestStaticOnly::class)
    }

    describe("using a mocked object") {
        it("executes the body") {
            assertEquals(4, TestToMock.intSum(3, 1))
        }
        describe("when an answer is provided") {
            it("uses the mock response") {
                every { TestToMock.intSum(any(), any()) } returns 100
                assertEquals(100, TestToMock.intSum(3, 1))
            }
        }
    }

    describe("using a java class with static functions") {
        it("executes the body") {
            assertEquals(4, TestStaticOnly.intSum(3, 1))
        }
        describe("when an answer is provided") {
            it("uses the mock response") {
                every { TestStaticOnly.intSum(any(), any()) } returns 100
                assertEquals(100, TestStaticOnly.intSum(3, 1))
            }
        }
    }
})
```

```kotlin
class RealTest : DescribeSpec({
    describe("using an already unmocked object") {
        it("executes the body") {
            assertEquals(4, TestToMock.intSum(3, 1))
        }
    }

    describe("using an already unmocked static function") {
        it("executes the body") {
            assertEquals(4, TestStaticOnly.intSum(3, 1))
        }
    }
})
```

Executing each spec on its own does work as expected.

Having both in the same package and executing the package (or whole test suite) will crash the second test (`RealTest`) which should use an unmocked state with this assertion:
`java.lang.AssertionError: expected:<4> but was:<100>`.
Note: It is important, that the `RealTest` is executed **after** the mocked + unmocked test for this crash to happen!

It seems there is a problem with mocking/unmocking objects and static functions in combination with lifecycle callbacks.

sksamuel:
Work in progress. As we move closer to 6.0, this PR attempts to bring in a permanent solution for concurrency in spec and tests, as well as simplify the test engine.

* Removes isolation per leaf and isolation per test. These isolation modes were added before coroutines even existed in Kotlin! This will be a large breaking change, but one that hugely simplifies the test engine, and removes lots of awkward edge cases around instantiation of specs that then require tests to be replayed. This will make the unified JS/JVM/Native test runner easier to accomplish. And it will allow us to make coroutine logic more sane in how tests inherit coroutine scopes etc.
* Removes the `SpecRunner` instances and squash these into a single SpecExecutor2 (odd name, but that's because in the next follow up PR, the two layers of spec exectors will be merged).
* Adds a new `IsolationPerRoot` option to replace the above that gives you a fresh spec per top level test, which avoids replaying as root tests are registered but not executed at spec creation.
* Removes all previous parallelism settings in specs and project config such as `threads`, `parallelism`, `affinity`, `testConcurrency` and `specConcurency`. Other than `threads`, these settings have been marked as experimental since they were added.
* Adds new `TestExecutionMode` which sets concurrency inside a spec and works for both single instance and instance per root isolation modes. When set to `Concurrent` all root tests are launched concurrency. When left to the default `Sequential` each root test is launched when the previous completes. Root tests are always launched into their own coroutine regardless and this setting just determines how many are launched in parallel.
* Adds new `SpecExecutionMode` which sets concurrency across the engine for specs. When set to `Concurrent` all specs tests are launched concurrency. When left to the default `Sequential` each spec is launched when the previous spec completes.
* The `CoroutineDispatcherFactory` usage has been simplified. Instead of passing around a default (which was a no-op) the default has been removed, and the engine just doesn't apply one if there isn't one set on the spec / project configuration. This means we don't have to weave a factory through the interceptors and scopes for no reason.
* The affinity option has been removed. It's simple enough to use a custom factory (and we provide a new `ThreadPerSpecCoroutineContextFactory` to do this) if the option is needed, but it's probably not even something you want to do.

Fixes #3311
Closes #3962


kotest-extensions/kotest-extensions-blockhound/src/jvmTest/kotlin/io/kotest/extensions/blockhound/BlockHoundTest.kt
```diff
@@ -66,7 +66,7 @@ class BlockHoundSpecTest : FunSpec({
       shouldNotThrow<BlockingOperationError> { blockInNonBlockingContext() }
    }
 
-   test("parallelism").config(invocations = 2, threads = 2) {
+   test("with blocking trhead").config(invocations = 2, blockingTest = true) {
```
LeoColman: Nit
```suggestion
   test("with blocking thread").config(invocations = 2, blockingTest = true) {
```

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/annotation/concurrency.kt
```diff
@@ -0,0 +1,25 @@
+package io.kotest.core.annotation
+
+/**
+ * When added to a spec, will mark that spec to not run in parallel, regardless
+ * of concurrency settings.
+ */
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class Isolate
```
LeoColman, line 9: Very nitty but important for the public API:

I know the previous version was called Isolate, but if this is a breaking change I suggest changing this guy to `Isolated`. As `Parallel` in this case is a noun and `Isolate` is more of a verb.

I think it is more sense to have 
- My test is running parallel
- My test is running isolated

Than having
- My test is running parallel
- My test is running isolate

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/annotation/tags.kt
```diff
@@ -10,3 +10,10 @@ package io.kotest.core.annotation
 @Retention(AnnotationRetention.RUNTIME)
 annotation class Tags(vararg val values: String)
 
+/**
+ * Attach to a [io.kotest.core.spec.Spec], and that spec won't be instantiated if the tag(s) are not
+ * provided at runtime. If more than one tag is specified then all tags must be provided at runtime.
```
LeoColman, line 15: Nit:

Add to these docs on how to add a tag to a spec (even if just a link)

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/config/Defaults.kt
```diff
@@ -35,6 +39,7 @@ object Defaults {
    val severity: TestCaseSeverityLevel = TestCaseSeverityLevel.NORMAL
 
    const val coroutineDebugProbes: Boolean = false
+//   const val testCoroutineDispatcher: Boolean = false
```
LeoColman, line 42: Intentional?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/config/Defaults.kt
```diff
@@ -59,8 +62,7 @@ object Defaults {
    val defaultTestNameCase: TestNameCase = TestNameCase.AsIs
    val specExecutionOrder = SpecExecutionOrder.Lexicographic
 
-   const val concurrentTests = ProjectConfiguration.Sequential
-   const val dispatcherAffinity = true
+//   const val dispatcherAffinity = true
```
LeoColman: Intentional?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/config/ProjectConfiguration.kt
```diff
@@ -88,95 +78,20 @@ class ProjectConfiguration {
...
+//    *
+//    * When this value is false, the framework is free to assign different dispatchers to different
+//    * root tests (nested tests always run in the same thread as their parent test).
+//    *
+//    * Note: This setting has no effect unless the number of threads is increasd; see [parallelism].
+//    *
+//    * Defaults to [Defaults.dispatcherAffinity].
+//    */
+//   @ExperimentalKotest
+//   var dispatcherAffinity: Boolean = Defaults.dispatcherAffinity
```
LeoColman: Poor guy

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/IsolationMode.kt
```diff
@@ -4,71 +4,13 @@ enum class IsolationMode {
...
-    *
-    * ```text
-    * a
-    * a
-    * b
-    * a
-    * c
-    * ```
-    *
+    * A new instance of the [Spec] class is instantiated for every top level (root)
```
LeoColman, line 12: I think we should keep the example code showing what is a root and what is not

For instance, I don't know if


Funspec
- Context1
    - -Test1
    - -Test2
- Context2
    - -Test3
    - -Test4


Will create 2 instances (Context1, Context2) or 3 instances (Funspec, Context1, Context2)

Or even a different thing altogether

LeoColman, line 12: I think we should keep the example code showing what is a root and what is not

For instance, I don't know if


Funspec
- Context1
    - -Test1
    - -Test2
- Context2
    - -Test3
    - -Test4


Will create 2 instances (Context1, Context2) or 3 instances (Funspec, Context1, Context2)

Or even a different thing altogether

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/annotation/concurrency.kt
sksamuel: It's not a breaking change, the package has stayed the same.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/annotation/tags.kt
sksamuel: This was just a file move.
Once we have this all merged in I will write up considerably detailed docs on all these changes (and the rest of 6.0

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/config/Defaults.kt
sksamuel: Yes, note the PR comments that this is still tbd while I think about it. I won't merge until it's in a final state.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/config/Defaults.kt
sksamuel: Yes, note the PR comments that this is still tbd while I think about it. I won't merge until it's in a final state.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/IsolationMode.kt
sksamuel: Comprehensive docs will follow, once it's in the final state.

AlexCue987 (commented):
I might have overlooked something, but I did not see a clear definition of what `InstancePerRoot` is. Do we have it documented?

sksamuel:
> I might have overlooked something, but I did not see a clear definition of what `InstancePerRoot` is. Do we have it documented?

No not yet, I'll be doing comprehensive docs if /when merged

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestEngine.kt
```diff
@@ -48,22 +46,11 @@ class TestEngine(private val config: TestEngineConfig) {
     * It is recommended that this method is not invoked, but instead the engine
     * is launched via the [TestEngineLauncher].
     */
-   @OptIn(KotestInternal::class, ExperimentalKotest::class)
    internal suspend fun execute(suite: TestSuite): EngineResult {
-      logger.log { Pair(null, "Executing test suite with ${suite.specs.size} specs") }
+      logger.log { Pair(null, "Initiating test suite with ${suite.specs.size} specs") }
```
LeoColman, line 50: much much better 😂

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/SpecExecutor.kt
```diff
@@ -47,6 +48,7 @@ internal class SpecExecutor(
 
    private suspend fun executeInDelegate(spec: Spec): Result<Map<TestCase, TestResult>> {
       return try {
+         @Suppress("DEPRECATION")
```
LeoColman, line 48: Nit:

You did this at the beggining of the file on other files. Now we have 3 of these in the same file

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestEngine.kt
sksamuel: lol

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/SpecExecutor.kt
sksamuel: I didn't want to just blanket use it here.

sksamuel:
Once the tests are updated, I think this is now good to go.

LeoColman (approved):

AlexCue987 (approved):

matejdro:
This sounds like a big downgrade of the kotest to me. This now means that every `When`/`should`/`scenario` is affected by all the tests before it. Doesn't that break the fundamental aspect of the tests - that all tests should be isolated from each other?

sgrell:
For me it is a very important feature that you can run every test in isolation. Otherwise you quickly have test bleeding and that is simply an unnecessary source of errors that could easily be avoided with the deprecated options. Please continue to support this feature or something similar.

PR #4634, opened 2025-01-04
add Exhaustive.powerSet() generator

AlexCue987:
add Exhaustive.powerSet() generator:

```kotlin
  Exhaustive.Companion.powerSet(listOf("a", "b", "c")).values shouldContainExactlyInAnyOrder
            listOf(
               listOf("a"),
               listOf("b"),
               listOf("c"),
               listOf("a", "b"),
               listOf("a", "c"),
               listOf("b", "c"),
               listOf("a", "b", "c"),
            )
```


sksamuel:
There's already .collection, .of, list.exhaustive and just `exhaustive(list)`
I don't think we need another ?

AlexCue987:
> There's already .collection, .of, list.exhaustive and just `exhaustive(list)` I don't think we need another ?

I did not find an exhaustive list of all samples aka subsets, such as:

```kotlin
 Exhaustive.Companion.samples(listOf("a", "b", "c")).values shouldContainExactlyInAnyOrder
            listOf(
               listOf("a"),
               listOf("b"),
               listOf("c"),
               listOf("a", "b"),
               listOf("a", "c"),
               listOf("b", "c"),
               listOf("a", "b", "c"),
            )
```

What am I missing?

sksamuel:
I wouldn't call it samples, that indicates randomness, which is the opposite of exhaustive.
I think what you're looking for is the powerset.

AlexCue987:
> I wouldn't call it samples, that indicates randomness, which is the opposite of exhaustive. I think what you're looking for is the powerset.

renamed to `powerSet`, added `emptyList` to results to match the definition

kotest-property/api/kotest-property.api
```diff
@@ -1389,6 +1389,7 @@ public final class io/kotest/property/exhaustive/CollectionsKt {
 	public static final fun collection (Lio/kotest/property/Exhaustive$Companion;Ljava/util/Collection;)Lio/kotest/property/Exhaustive;
 	public static final fun permutations (Lio/kotest/property/Exhaustive$Companion;Ljava/util/List;I)Lio/kotest/property/Exhaustive;
 	public static synthetic fun permutations$default (Lio/kotest/property/Exhaustive$Companion;Ljava/util/List;IILjava/lang/Object;)Lio/kotest/property/Exhaustive;
+	public static final fun samples (Lio/kotest/property/Exhaustive$Companion;Ljava/util/List;)Lio/kotest/property/Exhaustive;
```
sksamuel: samples shouldn't be here anymore

kotest-property/src/commonMain/kotlin/io/kotest/property/exhaustive/collections.kt
```diff
@@ -21,3 +21,39 @@ fun <A> Exhaustive.Companion.permutations(list: List<A>, length: Int = list.size
...
+*  Returns an [Exhaustive] which provides all the samples of elements from the given list.
+* For instance:
+* Exhaustive.Companion.samples(listOf("a", "b")).values shouldContainExactlyInAnyOrder
+* listOf(
+*   listOf("a"),
+*   listOf("b"),
+*   listOf("a", "b"),
+* )
+ */
+fun <A> Exhaustive.Companion.powerSet(list: List<A>): Exhaustive<List<A>> {
```
sksamuel, line 35: I would add an option to include the empty list ?

sksamuel, line 36: I would update the docs to be clear that the empty set is included.

sksamuel (approved):

kotest-property/api/kotest-property.api
AlexCue987: > samples shouldn't be here anymore

fixed

kotest-property/src/commonMain/kotlin/io/kotest/property/exhaustive/collections.kt
AlexCue987: > I would update the docs to be clear that the empty set is included.

fixed

PR #4643, opened 2025-01-06
Add system property key for displayFullTestPath

JordanLongstaff:
Added another system property key, `kotest.framework.testname.display.full.path`, used to configure the option to display the full test paths in reports. This is a nice-to-have for my use case where I want to upload my test results for CI and have to use this setting, but currently I can only use an `AbstractProjectConfig`.


sksamuel (approved):
awesome thanks !

sksamuel:
I think you need to run apiDump

PR #4655, opened 2025-01-10
when values have different types but same printed value, describe mismatch of types

AlexCue987:
there are edge cases when values are of different types, but they print exactly the same. and the message is not super helpful: 

`Map should contain mapping apple=1.5 but was apple=1.5`

let's add more info to help, `Expected type java.math.BigDecimal, but was kotlin.Double`:
```kotlin
               mapOf("apple" to 1.5) shouldContain ("apple" to BigDecimal("1.5"))
            }.message
            message.shouldContainInOrder(
               "Map should contain mapping apple=1.5 but was apple=1.5",
               "Expected type java.math.BigDecimal, but was kotlin.Double",
            )
```

I've made `describeTypedMismatch` `public` because there are some other cases in other modules when we want to use it, in another PR, later on


AlexCue987:
will fix and reopen

PR #4658, opened 2025-01-10
adding filterMatching

Kantis:


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/inspectors/FilterMatchingTest.kt
```diff
@@ -0,0 +1,57 @@
...
+import io.kotest.matchers.string.shouldNotContain
+import io.kotest.matchers.types.shouldBeInstanceOf
+import kotlin.system.measureTimeMillis
+import kotlin.time.Duration.Companion.seconds
+
+class FilterMatchingTest : FunSpec({
+   test("Filtering empty list should return empty list") {
+      emptyList<String>().filterMatching {
+         it shouldHaveLength 5
+      } shouldBe emptyList()
```
AlexCue987: does this test verify extension functions for Sequence and Array?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/inspectors/FilterMatchingTest.kt
```diff
@@ -0,0 +1,57 @@
...
+   }
+
+   test("Filtering should only return matching elements") {
+      arrayOf("a", "bb", "ccc", "dddd").filterMatching {
+         it shouldHaveLength 1
+      } shouldBe listOf("a")
```
AlexCue987: does this test verify extension functions for Sequence and Array?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/inspectors/FilterMatchingTest.kt
```diff
@@ -0,0 +1,57 @@
...
+   }
+
+   test("Filtering applies all assertions to each element") {
+      arrayOf("a", "b", "bb", "ccc", "dddd").filterMatching {
+         it shouldHaveLength 1
+         it shouldNotContain "b"
+      } shouldBe listOf("a")
```
AlexCue987: does this test verify extension functions for Sequence and Array?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/inspectors/FilterMatchingTest.kt
```diff
@@ -0,0 +1,57 @@
...
+         sequenceOf(1, 2, 3)
+            .filterMatching { it.shouldBeEven() }
+            .shouldBeInstanceOf<Sequence<Int>>()
+            .shouldContainExactly(2)
+      }
+   }
+
+   test("Filtering sequences should be done lazily").config(invocationTimeout = 1.seconds) {
+      // Generating an infinite sequence. If the filtering is not done lazily, this test will timeout
+      generateSequence(1) { it + 1 }
```
AlexCue987: will this test fail faster if the sequence yields 100 elements and then throws an exception?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/inspectors/FilterMatchingTest.kt
```diff
@@ -0,0 +1,57 @@
...
+         1 shouldBe 1
+         arrayOf(1, 2, 3).filterMatching { it.shouldBeEven() } shouldBe listOf(2)
+         listOf(1, 2, 3).filterMatching { it.shouldBeEven() } shouldBe listOf(2)
+
+         sequenceOf(1, 2, 3)
+            .filterMatching { it.shouldBeEven() }
+            .shouldBeInstanceOf<Sequence<Int>>()
+            .shouldContainExactly(2)
+      }
+   }
```
AlexCue987: do we need a test where all elements are filtered out?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/inspectors/FilterMatchingTest.kt
Kantis: Thanks for the quick review! I think it's pretty obvious that it does not? We do not test all type variants of all inspector for all corner cases, as far as I can tell. In fact it looks like we do not have explicit tests for `Sequence` variants at all if you check [here](https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/inspectors/CollectionInspectorsTest.kt). 

I added what I felt were relevant tests for the various types. If you feel something more is meaningful to add, please elaborate more on how you feel that it would add value.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/inspectors/FilterMatchingTest.kt
Kantis: Good point, will add.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/inspectors/FilterMatchingTest.kt
Kantis: It currently runs in (rounded to) 0ms on my machine, so I would assume so.

AlexCue987: it succeeds right away, but should it ever fail, it would take a whole second - care to speed that up?

Kantis: I've had issues in the past with GH actions runners which sometimes pause threads for long enough to cause sub-second timeouts to become flaky. 

But since this code actually never suspends / yields, there would never be a point in time where the `InvocationTimeoutInterceptor`'s `withTimeout` gets to run, so it wouldn't be a problem. But that also means that the test is broken as-is. Will adjust it a bit, but I think we should keep the timeout at 1s to prevent flakiness from creeping in.

AlexCue987: > I think we should keep the timeout at 1s to prevent flakiness from creeping in.

I'd suggest designing away the very need for timeout: replace `generateSequence(1) { it + 1 }`
with something like
```kotlin
sequence {
(1..100).forEach { yield(it) }
throw Exception("Oops!")
}
```
then we won't need any timeout at all

Kantis: Clever 😃 sounds good, should cover the scenario better.

AlexCue987 (approved):

PR #4663, opened 2025-01-11
handle edge case when different values print exactly the same

AlexCue987:
handle edge case when different values print exactly the same - print their different types. for instance, the message before the fix might be: `Map should contain mapping apple=1.5 but was apple=1.5`

the complete example:

```kotlin
               mapOf("apple" to 1.5) shouldContain ("apple" to BigDecimal("1.5"))
            }.message
            message.shouldContainInOrder(
               "Map should contain mapping apple=1.5 but was apple=1.5",
               "Expected type java.math.BigDecimal, but was kotlin.Double",
            )
```

add a public function `describeTypedMismatch`, because we shall use this in multiple places in different modules - this is just the very first step


sksamuel:
What do you think about changing the rendering so that it appears how it would in code ?
So like instead of us outputting 1.5 we would output 1.5L or 1.5F or BigDecimal(1.5)

AlexCue987:
> ```kotlin
> Map should contain mapping apple=1.5 but was apple=1.5",
>                "Expected type java.math.BigDecimal, but was kotlin.Double"
> ```

there can be many cases when printed values are the same yet types are different, and we can't account for all of them. so I'd like to output qualified names. maybe like this:

```
Map should contain mapping apple=1.5 of type java.math.BigDecimal,  but was apple=1.5 of type kotlin.Double
```
or maybe this
```
Map should contain mapping apple=java.math.BigDecimal<1.5>,  but was apple=kotlin.Double<1.5> 
```

I see quite a few data classes with bespoke `toString` like the following:
```
data class PhoneNumber(val area: Int, val localNumber: Int) {
override fun toString() { // output as (218)340-7692
```
they are indistinguishable from String when printed, so we need to output qualified name 

WDYT?

sksamuel:
I like the idea.

Maybe we should do this with a new data class, PrintedType, that wraps a Printed value and returns the type as well?

AlexCue987:
> I like the idea.
> 
> Maybe we should do this with a new data class, PrintedType, that wraps a Printed value and returns the type as well?

I don't see where/how `PrintedType` should be used. Can you elaborate?

AlexCue987:
@sksamuel I don't understand your suggestion, can you clarify

Kantis:
> I like the idea.
> 
> Maybe we should do this with a new data class, PrintedType, that wraps a Printed value and returns the type as well?

We have [PrintedWithType](https://github.com/kotest/kotest/blob/a7b0fe0c0f8fda6b4ac0f5a26a07997ea6899b4a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/print/Print.kt#L37) which I think should fulfill the same need? 

@AlexCue987 you can construct a failure with type info included using `failureWithTypeInformation`. I think if `Eq` uses `failureWithTypeInformation` in appropriate places, we should see the type info being propagated to more advanced matchers as well.

AlexCue987:
> > I like the idea.
> > Maybe we should do this with a new data class, PrintedType, that wraps a Printed value and returns the type as well?
> 
> We have [PrintedWithType](https://github.com/kotest/kotest/blob/a7b0fe0c0f8fda6b4ac0f5a26a07997ea6899b4a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/print/Print.kt#L37) which I think should fulfill the same need?
> 
> @AlexCue987 you can construct a failure with type info included using `failureWithTypeInformation`. I think if `Eq` uses `failureWithTypeInformation` in appropriate places, we should see the type info being propagated to more advanced matchers as well.

TY for the explanation. Let's see how to plug it in here.

felixwiemuth:
I just came across the same issue, comparing maps with value types `Long` and a custom numerical type backed by `Long`, printing the same.
What was the resolution here? Will it be added to the default `Eq` matcher? I tested with version 5.9.1.
I think it is important that this becomes the default. Even better would be if `should/be` required the same types on both sides.

AlexCue987:
> I just came across the same issue, comparing maps with value types `Long` and a custom numerical type backed by `Long`, printing the same. What was the resolution here? Will it be added to the default `Eq` matcher? I tested with version 5.9.1. I think it is important that this becomes the default. Even better would be if `should/be` required the same types on both sides.

I fixed that for maps and merged to master: https://github.com/kotest/kotest/commit/7bd442b451c6282f792c259dff778f8be730e688

I concur we should go for this approach in multiple other places, including `eq` - it's on to-do list. What other places would you suggest to use it in?

PR #4695, opened 2025-01-19
lazily-evaluate-startWith 

AlexCue987:
lazily-evaluate-startWith


sksamuel (approved):

sksamuel:
There's not a huge benefit to the string only ones.

PR #4705, opened 2025-01-20
Also make the block version of `shouldNotBeNull` return `this`

sschuberth:
This is convenient to assign the return value.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/nulls/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/nulls/matchers.kt
@@ -80,9 +80,10 @@ fun <T> T?.shouldNotBeNull(): T {
  *     length should beEven()
  * }
  */
-infix fun <T : Any> T?.shouldNotBeNull(block: T.() -> Unit) {
+infix fun <T : Any> T?.shouldNotBeNull(block: T.() -> Unit): T {
    this.shouldNotBeNull()
    block()
+   return this
 }
 
 /**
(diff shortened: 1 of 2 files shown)


AlexCue987:
Very interesting. For my education, can you share an example where this feature comes in handy?

sschuberth:
I came across the need when refactoring [this function](https://github.com/oss-review-toolkit/ort/pull/9781/files#diff-f7dab3818dd75045bd70a24a91783fae97f8a2e726f10d70bed5d48f02619a92R46) in our project. I wanted to write something like

    val nonNullResultList = resultList.shouldNotBeNull {
        shouldBeSingleton()
    }

    val result = nonNullResultList.single().let {
        if (resolveScopes) managerResult.resolveScopes(it) else it
    }

but couldn't.

PS: I'm also thinking of another change to make `shouldBeSingleton` actually return the single value, and not the collection with one value.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/nulls/matchers.kt
```diff
@@ -80,9 +80,10 @@ fun <T> T?.shouldNotBeNull(): T {
  *     length should beEven()
  * }
  */
-infix fun <T : Any> T?.shouldNotBeNull(block: T.() -> Unit) {
+infix fun <T : Any> T?.shouldNotBeNull(block: T.() -> Unit): T {
    this.shouldNotBeNull()
    block()
+   return this
```
AlexCue987, line 86: do we need a new/changed unit test to reflect this code change?

AlexCue987 (approved):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/nulls/matchers.kt
sschuberth: TBH, I was a bit lazy adding any, as it would probably only test trivial stuff:
- the return value being non-null is already implied by the function signature
- as well as the fact that now *something* is returned (instead of `Unit`)
- that the return value matches the receiver if it's non-null seem trivial

PR #4706, opened 2025-01-21
Kotest plugin creates a kotest target per Kotlin target

Kantis:


kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
```diff
@@ -25,20 +35,37 @@ open class KotestPlugin : Plugin<Project> {
...
+               // TODO: I think wiring dependencies,
+//               dependsOn(project.tasks.withType<KotlinCompile>())
+            }
+         }
+      }
+
+      // Configure Kotlin multiplatform projects
+      project.pluginManager.withPlugin(KOTLIN_MULTIPLATFORM_PLUGIN) {
+         project.extensions.configure<KotlinMultiplatformExtension> {
+            targets.configureEach {
```
sksamuel, line 54: does targets come from the KotlinMultiplatformExtension?

Kantis: Yes :) so when I tested I created a jvm and js target, and the plugin created `kotestJvm` and `kotestJs`

sksamuel (approved):

PR #4762, opened 2025-02-11
handle edge case when different values print exactly the same 

AlexCue987:
handle edge case when different values print exactly the same:

```kotlin
 val thrown = shouldThrow<AssertionError> {
     mapOf("a" to BigDecimal("1.5")) shouldContain("a" to 1.5)
  }
  thrown.message.shouldContainInOrder(
     "Map should contain mapping a=1.5 but value was different:",
     "expected:kotlin.Double<1.5> but was:java.math.BigDecimal<1.5>"
  )
```

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/failures.kt
+++ b/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/failures.kt
@@ -56,6 +56,7 @@ fun failure(expected: Expected, actual: Actual, prependMessage: String = ""): Th
    )
 }
 
+@KotestInternal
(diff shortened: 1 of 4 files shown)


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/failures.kt
```diff
@@ -55,6 +56,16 @@ fun failure(expected: Expected, actual: Actual, prependMessage: String = ""): Th
    )
 }
 
+fun<V> getFailureWithTypeInformation(
```
Kantis, line 61:
```suggestion
@KotestInternal
fun<V> getFailureWithTypeInformation(
```

sksamuel (approved):

PR #4803, opened 2025-03-07
add Arb.plusEdgecases

AlexCue987:
example:

```kotlin
val intArb = Arb.Companion.int(-50..50)
val edgecases = intArb.edgecases().toList()
val additionalEdgecases = listOf(-42, 42)
val newEdgecases = intArb.plusEdgecases(additionalEdgecases).edgecases().toList()
newEdgecases.containsAll(edgecases) shouldBe true
newEdgecases.containsAll(additionalEdgecases) shouldBe true
```


kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/PlusEdgecasesTest.kt
```diff
@@ -0,0 +1,25 @@
...
+class PlusEdgecasesTest: StringSpec() {
+   private val intArb = Arb.Companion.int(-50..50)
+
+   init {
+       "add edgecases to intArb while keeping the original ones" {
+          val edgecases = intArb.edgecases().toList()
+          val additionalEdgecases = listOf(-42, 42)
+          val newEdgecases = intArb.plusEdgecases(additionalEdgecases).edgecases().toList()
+          newEdgecases.containsAll(edgecases) shouldBe true
+          newEdgecases.containsAll(additionalEdgecases) shouldBe true
```
LeoColman: Nit:
We should have an assertion that original `edgeCases` do not contain all edge cases. Otherwise we could be just proxying `withEdgeCases`

LeoColman (approved):

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/PlusEdgecasesTest.kt
AlexCue987: > Nit: We should have an assertion that original `edgeCases` do not contain all edge cases. Otherwise we could be just proxying `withEdgeCases`

good catch, added an assertion

Issue #4754, opened 2025-02-04
Changes in shouldNotThrowAny signature cause compilation errors


OptimumCode:
Hello,

After upgrading from kotest:6.0.0.M1 to kotest:6.0.0.M2 the signature of `shouldNotThrowAny` method was changed (in #4351):

Before:

`inline fun <T> shouldNotThrowAny(block: () -> T): T`

After:

`inline fun <T> shouldNotThrowAny(block: () -> T): T?`

As I understand, the change was made to allow collaboration of `shouldNotThrowAny` with soft assertions. However, the change is not compatible with code like that:

```kotlin
test("validation") {
  // after the change the returning type is JsonSchem? and it cannot be assigned to the variable
  val jsonSchema: JsonSchema = shouldNotThrowAny { JsonSchema.create() }
  jsonSchema.validate(createJsonElement()) shouldBe Result.Valid()
}
```

I am raising this as a question (not a bug) because I am not sure that the usage demonstrated above follows the idea of how the `shouldNotThrowAny` method should be used.

Could you please advise if the braking change was intentional? If so, could you please advise how the `shouldNotThrowAny` should be used in such cases? Right now, I need to add `shouldNotBeNull()` after `shouldNotThrowAny` to make the code compile.

Probably, @AlexCue987 or @sksamuel could help here because they took part in the original PR that introduced that change.

AlexCue987:
Correct, the motivation for this change was to make it compatible with `assertSoftly`, which comes really handy in many cases.

Can you explain why the example you've posted needs `shouldNotThrowAny` at all? In my mind the following shorter code comes to the same thing: 

```kotlin
test("validation") {
  val jsonSchema: JsonSchema =  JsonSchema.create() 
  jsonSchema.validate(createJsonElement()) shouldBe Result.Valid()
}
```

Can you elaborate why we should wrap `JsonSchema.create() ` in `shouldNotThrowAny`, what does it buy us?

OptimumCode:
In my example, I wanted to indicate that the `JsonSchema.create` **can** throw an exception but **should not**. This is just an example, but I can imagine such code in places where something like resource access or API call is wrapped into `shouldNotThrowAny` call for similar purposes.

But let's forget for a second about the example above. What would be a correct use-case for `shouldNotThrowAny`?
[...]

PR #4804, opened 2025-03-09
Update `shouldNotThrowAny` signature

devxgb:
Suggestion from https://github.com/kotest/kotest/issues/4754


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/throwables/AnyThrowableHandling.kt
```diff
@@ -94,11 +94,12 @@ inline fun shouldThrowAny(block: () -> Any?): Throwable {
  * @see [shouldNotThrow]
  *
  */
-inline fun <T> shouldNotThrowAny(block: () -> T): T? {
+inline fun shouldNotThrowAny(block: () -> Unit) {
```
AlexCue987, line 97: Personally, I think this is a step in the right direction - is simplifies the API without removing anything necessary. But we should update javadoc to explain why and when we should use this matcher - we've discussed that in the related issue.

I concur that `shouldNotThrowAnyUnit` becomes redundant, so let's deprecate it with intent to remove in 7.0.

But I don't understand why we'd ever need to return from this assertion. Can you explain @Kantis and @LeoColman ? Before approving/merging, I'll wait for your opinions.

LeoColman: It removes the possibility of chaining matchers

AlexCue987: > It removes the possibility of chaining matchers

can you share an example when it's needed?

devxgb: As I understand, the return type was made nullable to support soft-assertion.
It already lost the ability to chain nicely, as described in the issue https://github.com/kotest/kotest/issues/4754.
Now, it requires a null-check, even if the the block returns a solid type.

I see three option.
* Keep it as is. It supports soft-assertion. It does not chain nicely, and is confusing to user.
* If we want to chain nicely without null-check, we need to go back to the version where soft assertion was not supported.
* If we want to keep soft assertion support and make it less confusing for the user, we need to give up on chaining and update signature to void.

Let me know what you think.

AlexCue987: We must have an assertion that is compatible with `assertSoftly`. But it does not have to be this one. We can change `shouldNotThrowAnyUnit` instead, without changing its contract. And then we can restore this assertion to its original signature. WDYT @LeoColman

AlexCue987:
Because @LeoColman is saying we need to chain this assertion:

* I've restored the original signature here: https://github.com/kotest/kotest/pull/4809
* and moved the compatibility with assertSoftly to another assertion

Which I believe renders this PR no longer needed.

Sorry for wasting your time @devxgb

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/throwables/AnyThrowableHandling.kt
Kantis: @LeoColman I can't think of any meaningful scenarios where this needs to chain. For instance:

```kotlin
shouldNotThrowAny {
  foo()
} shouldBe bar()
```

Could just be written as 
```kotlin
foo() shouldBe bar()
```

If we look at [Alex comment](https://github.com/kotest/kotest/issues/4754#issuecomment-2635106241) regarding intended use of this assertion, maybe there's some cases where we want to chain around non-Kotest assertions?  Feels a bit far-fetched though..

devxgb:
> Because @LeoColman is saying we need to chain this assertion:
> 
> * I've restored the original signature here: [make-shouldNotThrowExactlyUnit-compatible-with-assertSoftly (#4808) #4809](https://github.com/kotest/kotest/pull/4809)
> * and moved the compatibility with assertSoftly to another assertion
> 
> Which I believe renders this PR no longer needed.
> 
> Sorry for wasting your time @devxgb

No problem at all. 🙏  It is my pleasure get involved. 🙇 
Hope to contribute more. I will look for other issues.

sksamuel:
@devxgb happy to see more people get involved! Feel free to pick up anything.

PR #4813, opened 2025-03-19
Enable coroutineTestScope on js and wasmJs

### LINKED ISSUE #4077: Support coroutineTestScope option for JS platform

fhilgers:
This is a very simple change and I am not deeply in the kotest codebase. I used this workaround for some projects internally and it seems to work.

If these changes are acceptable I'd be able to add some tests for this too, akin to the already existing tests for the coroutineTestScope on the jvm. As I am not too familiar it would be a great help if you could point me to the right direction for adding tests on js and wasmjs.

This might close #4077


sksamuel (approved):

sksamuel:
thanks!

PR #4818, opened 2025-03-21
Enabling tests on macOS, unless we are running CI

Kantis:


sksamuel (approved):
great, gonna push this over the line as its really useful

PR #4824, opened 2025-03-28
Improve documentation in TypeMatchers.kt

LeoColman:


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/types/TypeMatchers.kt
```diff
@@ -6,43 +6,123 @@ import io.kotest.matchers.MatcherResult
...
+ *
+ * Verifies that the subject and [ref] are the exact same object in memory.
+ *
+ * ```
+ * val a = "hello"
+ * val b = a
+ * val c = "hello"
+ *
+ * a should beTheSameInstanceAs(b)    // Assertion passes
+ * a shouldNot beTheSameInstanceAs(c) // Assertion passes
```
AlexCue987, line 87: this might change in the future - `a shouldNot beTheSameInstanceAs(c)` might start failing. it might be less likely with `val c = "HELLO".lowercase()` or `val c = "olleh".reversed()`

AlexCue987 (approved):

PR #4844, opened 2025-04-12
Fix possible race condition in scheduler

sksamuel:

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestSuiteScheduler.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestSuiteScheduler.kt
@@ -40,7 +40,7 @@ internal class TestSuiteScheduler(
       schedule(isolated, 1)
       logger.log { Pair(null, "Isolated specs have completed") }
 
-      // first we run the specs that have been marked as always parallel regardless of concurrency mode
+      // next we run the specs that have been marked as always parallel regardless of concurrency mode
       val parallel = suite.specs.filter { it.kclass.isParallel() }
       logger.log { Pair(null, "Parallelized spec count: ${parallel.size}") }
       schedule(parallel, Int.MAX_VALUE)
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestSuiteScheduler.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestSuiteScheduler.kt
@@ -1,27 +1,22 @@
-@file:Suppress("DEPRECATION")
-
 package io.kotest.engine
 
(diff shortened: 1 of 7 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestSuiteScheduler.kt
```diff
@@ -39,16 +34,19 @@ internal class TestSuiteScheduler(
    suspend fun schedule(suite: TestSuite): EngineResult {
       logger.log { Pair(null, "Launching ${suite.specs.size} specs") }
 
+      // first we run the specs that have been marked as always isolated
       val isolated = suite.specs.filter { it.kclass.isIsolate() }
       logger.log { Pair(null, "Isolated spec count: ${isolated.size}") }
       schedule(isolated, 1)
       logger.log { Pair(null, "Isolated specs have completed") }
 
+      // first we run the specs that have been marked as always parallel regardless of concurrency mode
```
AlexCue987: should we replace "first we run" with "next we run" - we already said "first we run" on line 37

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestSuiteScheduler.kt
```diff
@@ -73,26 +73,34 @@ internal class TestSuiteScheduler(
...
+      if (context.projectConfigResolver.projectWideFailFast() && collector.errors) {
+         logger.log { Pair(ref.kclass.bestName(), "Project wide fail fast is active, skipping spec") }
+         context.listener.specIgnored(ref.kclass, null)
+      } else {
+         try {
+            val executor = SpecExecutor(context)
+            logger.log { Pair(ref.kclass.bestName(), "Executing ref") }
+            executor.execute(ref)
+         } catch (t: Throwable) {
+            logger.log { Pair(ref.kclass.bestName(), "Unhandled error during spec execution $t") }
```
AlexCue987, line 98: do we want to include `t` in the message, to log what was the error?

AlexCue987 (approved):
LGTM, some cosmetic suggestions

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestSuiteScheduler.kt
sksamuel: t is included at the end

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/TestSuiteScheduler.kt
sksamuel: good point, fixed

PR #4848, opened 2025-04-17
Fix `before` and `after` matchers error messages

felixwiemuth:
The matchers were correct, but the inverted error messages mistakenly swapped the `value` and `anotherInstant` variables, leading to incorrect error messages.

I added simple tests asserting the correct messages for the `before` and `after` matchers.

I would also suggest to align the wording and punctuation (which I did not do as part of this bug fix - please fix or open an issue in case you agree):
- "but it's not" is not consistent with the other formulations and should probably be removed - it is clear from the context (failing assertion)
- Some error messages end with a ".", some do not

This should probably be merged both into master and 5.9 if a 5.9.2 release should be made. I based the branch on 5.9.1.


felixwiemuth:
I see that part of this (for `before`) was already fixed in #4296 , but not for `after`.
It would be nice to have the fix released soon!

AlexCue987:
are failing tests related to the changes?

sksamuel (approved):

PR #4849, opened 2025-04-18
Allow emptyList to be passed to `should[Not]ContainAnyOf`, `should[Not]BeOneOf` and `should[Not]BeIn`

### LINKED ISSUE #4847: EmptyList Is Not Supported as an Argument in `should[Not]ContainAnyOf`
**Which version of Kotest are you using**: 5.4.1

## Description:

It appears that the `shouldContainAnyOf` and `shouldNotContainAnyOf` functions do not currently support emptyList as an argument. I would like to propose a change to allow this behavior.

## Current Behavior:

When using emptyList as an argument, the behavior is as follows:

```kt
listOf(1,2,3) shouldContainAll listOf() 
// Mathematically true -> passes as expected

listOf(1,2,3) shouldNotContainAll listOf() 
// Mathematically false -> fails as expected

listOf(1,2,3) shouldContainAnyOf listOf() 
// Mathematically false -> should fail, and it does fail, but the failure is due to the suggestion to use shouldBeEmpty(), which is misleading.

listOf(1,2,3) shouldNotContainAnyOf listOf() 
// Mathematically true -> should pass, but it fails due to the suggestion to use shouldBeEmpty().
```

## Issue:

For `shouldContainAnyOf` and `shouldNotContainAnyOf`, passing an emptyList should mathematically result in false and true, respectively, regardless of the target list (e.g., `listOf(1,2,3)`). ~~The suggestion to use `shouldBeEmpty()` does not seem relevant to these cases.~~ The suggestion to use `shouldBeEmpty()` seems misplaced, as the focus of the check is typically on the left-hand side.

## Proposed Change:

- Allow emptyList as a valid argument for both `shouldContainAnyOf` and `shouldNotContainAnyOf`.
- Ensure that the behavior aligns with mathematical logic:
  - `<any list> shouldContainAnyOf listOf()` should fail, not because shouldBeEmpty() is not used, but because it logically should fail.
  - `<any list> shouldNotContainAnyOf listOf()` should pass.

## Additional Information:

Please let me know if further clarification is needed. I am happy to assist in implementing this change if required.

habara-k:
This PR resolves https://github.com/kotest/kotest/issues/4847.

## Changes

Allow emptyList to be passed as an argument to
- `should[Not]ContainAnyOf`
- `should[Not]BeOneOf`
- `should[Not]BeIn`

### Minor changes
- ~Add dependency of libs.kotlinx.coroutines.test for local testing of :kotest-assertions:kotest-assertions-core.~ Moved to #4856

## Motivation

Mathematically, the following assertions should each fail or pass as described:

```kt
// Example assertions with empty lists
listOf(...).shouldContainAnyOf(emptyList())    // Expected to fail
listOf(...).shouldNotContainAnyOf(emptyList()) // Expected to pass
(...).shouldBeOneOf(emptyList())               // Expected to fail
(...).shouldNotBeOneOf(emptyList())            // Expected to pass
(...).shouldBeIn(emptyList())                  // Expected to fail
(...).shouldNotBeIn(emptyList())               // Expected to pass
```

However, when these assertions are passed an empty collection, an error occurs stating: "Asserting content on empty collection. Use Collection.shouldBeEmpty() instead."

These assertions lack consistency with the existing specifications of the following assertions:

```kt
listOf(...).shouldContainAll(emptyList())     // Passes
listOf(...).shouldNotContainAll(emptyList())  // Fails
mapOf(...).shouldContainAll(emptyMap())       // Passes
mapOf(...).shouldNotContainAll(emptyMap())    // Fails
mapOf(...).shouldContainKeys()                // Passes
mapOf(...).shouldNotContainKeys()             // Fails
mapOf(...).shouldContainAnyKeysOf()           // Fails
mapOf(...).shouldNotContainAnyKeysOf()        // Passes
mapOf(...).shouldContainValues()              // Passes
mapOf(...).shouldNotContainValues()           // Fails
mapOf(...).shouldContainAnyValuesOf()         // Fails
mapOf(...).shouldNotContainAnyValuesOf()      // Passes
```

This PR aims to correct the inconsistent behavior and make the Kotest assertions mathematically accurate and more general.


sksamuel:
This is going to change the existing behavior of all these matchers right ?

sksamuel:
Is there something "official" that indicates what should happen for an empty collection? Like maybe something equivalent in the Kotlin SDK ?

AlexCue987:
> Is there something "official" that indicates what should happen for an empty collection? Like maybe something equivalent in the Kotlin SDK ?

Kotlin standard library states that:

`listOf(1, 2, 3).containsAll(listOf()) shouldBe true`

Mathematically this PR make sense to me: `shouldNotContainAnyOf` means "there are no elements in the collection on the right that are also in the collection on the left, which is true if the collection on the right is empty.

Likewise `listOf(...).shouldContainAll(emptyList())` means "there are no elements on the right that are not in the collection on the left" - also true.

So I agree with this PR.

sksamuel:
Do you think we should add some explicit docs for the matchers that link to why empty collections evaluate to true ?

AlexCue987:
> Do you think we should add some explicit docs for the matchers that link to why empty collections evaluate to true ?

if we pivot on this approach, then yes I agree, we should explain it in some detail. because this is a breaking change

habara-k:
Thank you for your response!

Since this behavior is mathematically valid and adopted in general libraries, I don't believe it needs to be explained in detail as an exception in the documentation. However, since it is a breaking change, it would be appropriate to increase the major version and announce it as shown in https://kotest.io/docs/changelog.html#breaking-changes-and-removed-deprecated-methods.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bein.kt
```diff
@@ -31,7 +31,7 @@ infix fun <T> T.shouldBeIn(collection: Collection<T>): T {
  * Assertion to check that this element is not any of [collection]. This assertion checks by value, and not by reference,
  * therefore any instance with same value must not be in [collection], or this will fail.
  *
- * An empty collection will always fail. If you need to check for empty collection, use [Collection.shouldBeEmpty]
+ * An empty collection will always pass.
```
AlexCue987: Can you add a short explanation why "An empty collection will always pass".

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bein.kt
```diff
@@ -64,7 +64,7 @@ fun <T> T.shouldBeIn(vararg any: T): T {
  * Assertion to check that this element is not any of [any]. This assertion checks by value, and not by reference,
  * therefore any instance with same value must not be in [any], or this will fail.
  *
- * An empty collection will always fail. If you need to check for empty collection, use [Collection.shouldBeEmpty]
+ * An empty collection will always pass.
```
AlexCue987: Can you add a short explanation why "An empty collection will always pass".

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bein.kt
```diff
@@ -99,7 +99,7 @@ infix fun <T> T.shouldBeIn(array: Array<T>): T {
  * Assertion to check that this element is not any of [array]. This assertion checks by value, and not by reference,
  * therefore any instance with same value must not be in [array], or this will fail.
  *
- * An empty array will always fail. If you need to check for empty array, use [Array.shouldBeEmpty]
+ * An empty array will always pass.
```
AlexCue987: Can you add a short explanation why "An empty collection will always pass".

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/oneof.kt
```diff
@@ -29,7 +29,7 @@ infix fun <T> T.shouldBeOneOf(collection: Collection<T>): T {
  * Assertion to check that this instance is not in [collection]. This assertion checks by reference, and not by value,
  * therefore the exact instance must not be in [collection], or this will fail.
  *
- * An empty collection will always fail. If you need to check for empty collection, use [Collection.shouldBeEmpty]
+ * An empty collection will always pass.
```
AlexCue987: Can you add a short explanation why "An empty collection will always pass".

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/oneof.kt
```diff
@@ -61,7 +61,7 @@ fun <T> T.shouldBeOneOf(vararg any: T): T {
  * Assertion to check that this instance is not any of [any]. This assertion checks by reference, and not by value,
  * therefore the exact instance must not be in [any], or this will fail.
  *
- * An empty collection will always fail. If you need to check for empty collection, use [Collection.shouldBeEmpty]
+ * An empty collection will always pass.
```
AlexCue987: Can you add a short explanation why "An empty collection will always pass".

AlexCue987 (changes requested):
I really like this PR. IMO the only one little thing it needs added are short explanations in javadocs

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/bein.kt
habara-k: I made the change in ~20f3b8e~97d2d65.

kotest-assertions/kotest-assertions-core/build.gradle.kts
```diff
@@ -34,6 +34,7 @@ kotlin {
          dependencies {
             implementation(projects.kotestProperty)
             implementation(libs.kotlinx.coroutines.core)
+            implementation(libs.kotlinx.coroutines.test)
```
AlexCue987: what did we add this dependency for?

habara-k: This is for local testing of :kotest-assertions:kotest-assertions-core.
I would like to develop in a test-driven manner, so the dependency is added first.

AlexCue987: > This is for local testing of :kotest-assertions:kotest-assertions-core. I would like to develop in a test-driven manner, so the dependency is added first.

this is an unrelated change. feel free to open a separate PR for that. to keep this PR atomic, please remove the dependency. 

other than that, I think the PR is ready to be merged, and is a nice improvement.

AlexCue987 (approved):
very nice!

PR #4850, opened 2025-04-21
Fix Exhaustive.Companion.azstring

ej-castillo:
According the [documentation](https://kotest.io/docs/proptest/property-test-generators-list.html), "`Exhaustive.azstring(1..2)` ... [returns] `a, b, c, ...., yz, zz`". The `yz` seems to imply that it generates all possible permutations; however, this doesn't seem to be the case. For example for  `Exhaustive.azstring(2..2)` should produce `aa, ab, ac, ... , zx, zy, zz` but instead produces `aa, bb, cc, ... , xx, yy, zz`. There is a mismatch between the code and the documentation. 

My changes updates the code to match the written behavior. If the current behavior is intentional, I can change this pull request to update the documentation instead (and update the test to match the current behavior)


sksamuel (approved):

sksamuel:
Great thanks!

PR #4871, opened 2025-05-08
Fix the compiler plugin to generate correct files

### LINKED ISSUE #4870: Missing `kotest-framework-multiplatform-plugin-embeddable-compiler-6.0.0.M4` dependency
6.0.0.M4

Can't run iOS tests because of the following error

```
* What went wrong:
Could not determine the dependencies of task ':myModule:compileKotlinIosSimulatorArm64'.
> Could not resolve all dependencies for configuration ':myModule:kotlinCompilerPluginClasspathIosSimulatorArm64Main'.
   > Could not find io.kotest:kotest-framework-multiplatform-plugin-embeddable-compiler:6.0.0.M4.
     Searched in the following locations:
       - file:/Users/popovych.y/.m2/repository/io/kotest/kotest-framework-multiplatform-plugin-embeddable-compiler/6.0.0.M4/kotest-framework-multiplatform-plugin-embeddable-compiler-6.0.0.M4.pom
       - https://repo.maven.apache.org/maven2/io/kotest/kotest-framework-multiplatform-plugin-embeddable-compiler/6.0.0.M4/kotest-framework-multiplatform-plugin-embeddable-compiler-6.0.0.M4.pom

```

JesusMcCloud:
Fixes #4870

### DIFF
--- a/kotest-framework/kotest-framework-multiplatform-plugin-compiler/src/jvmMain/kotlin/io/kotest/framework/multiplatform/embeddablecompiler/Transformer.kt
+++ b/kotest-framework/kotest-framework-multiplatform-plugin-compiler/src/jvmMain/kotlin/io/kotest/framework/multiplatform/embeddablecompiler/Transformer.kt
@@ -26,6 +26,7 @@ import org.jetbrains.kotlin.ir.symbols.UnsafeDuringIrConstructionAPI
 import org.jetbrains.kotlin.ir.util.constructors
 import org.jetbrains.kotlin.ir.util.getSimpleFunction
 import org.jetbrains.kotlin.ir.util.kotlinFqName
+import org.jetbrains.kotlin.ir.util.packageFqName
 import org.jetbrains.kotlin.name.ClassId
 import java.io.File
 import java.util.concurrent.CopyOnWriteArrayList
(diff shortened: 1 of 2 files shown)


JesusMcCloud:
now the generated code actually works. as for the TODO regarding how to generate code, how about something similar as we hacked together [here](https://github.com/a-sit-plus/signum/blob/main/supreme/build.gradle.kts#L190) (without the regex, but the full power of the compiler):

* setup a generated sources folder from inside the Kotest gradle plugin once and use some semi-randomly-generated name for it
* on every test run:
  1. delete all code inside
  2. use KotlinPoet to generate the code inside that generated soruces folder. This way, no mess with escaping as in this PR, etc.
  3. run tests
  4. *don't* delete the generated source as they are viable for debugging and step 1 takes care of this anyway

JesusMcCloud:
another discovery: we run into conflicting overloads due to the generated code. seems that the new compiler plugin may have been intentionally not shipped and referenced, as it is not yet ready

JesusMcCloud:
>     * setup a generated sources folder from inside the Kotest gradle plugin once and use some semi-randomly-generated name for it
> 
>     * on every test run:
>       1. delete all code inside
>       2. use KotlinPoet to generate the code inside that generated soruces folder. This way, no mess with escaping as in this PR, etc.
>       3. run tests
>       4. _don't_ delete the generated source as they are viable for debugging and step 1 takes care of this anyway

The more I think about it, the more i feel that steps 1 + 2 should be done from a dedicated gradle task that that should be specified as a dependency of all test compile tasks

sksamuel (approved):

PR #4891, opened 2025-06-11
update to ktor 3.x, remove TestApplicationResponse assertions

tKe:
The old `withTestApplication` testing API was deprecated in ktor 2.x and is now deprecated-with-error in 3.x.


sksamuel (approved):

sksamuel:
Thanks!

PR #4893, opened 2025-06-12
fix and clarify JS transformer status

JesusMcCloud:
* Don't generate bogus files if no jsTest sources exist
* Add comments about the current limitations to JsTransformer

Also: the native tests are still not run. Any pointers welcome to also get those to work (it never seemed to work with the native transformer in the first place).


JesusMcCloud:
This is only a partial band-aid. The transformer still cannot handle annotations that require op tin on tests, ecxept for those that are hardcoded

JesusMcCloud:
this is keeping us from running build scans, because files are created in places they don't belong when in fact no files should be created, so build scan runs into conflicting overloads

sksamuel:
I am hopeful this PR will remove the need for the compiler plugins at all https://github.com/kotest/kotest/pull/4913

JesusMcCloud:
That sounds awesome! Would that mean adding kotest as a `ksp(…)` dependency in the future?

sksamuel:
> That sounds awesome! Would that mean adding kotest as a `ksp(…)` dependency in the future?

Yes, it just swaps out the kotest compiler plugin for a ksp plugin. It already works for js and native. I'm aiming to wrap up a 6.0 final RC for Friday. 

If you wanna guinea pig let me know.

JesusMcCloud:
Count me in for testing. maybe publish a new snapshot. More importantly: please point me to a sample gradle project that consumes the new ksp plugin. I'd be testing with a KMP project

sksamuel:
What targets?

JesusMcCloud:
all targets, but I can selectively disable, if not everything is supported

sksamuel:
> all targets, but I can selectively disable, if not everything is supported

Shall I just write up some docs? Or you wanna pop into #kotest-contributors on kotlin slack ?

PR #4907, opened 2025-06-26
Support reflective binding of java classes

### LINKED ISSUE #4901: Arb.bind<> fails to bind Java Classes
```
kotest-property-jvm
5.9.1
```
---
I'm trying to use Reflective Arbs to build property test inputs based on complex Java Classes generated by the Maven Avro Plugin. No Kotlin Mode available there. I'm using Kotlin for my application code and would like to test it with Kotest. 

jvmbind.kt uses kclass.primaryconstructor to create class instances. This fails because Java classes don't provide a primary constructor in the Kotlin sense.

Could we fall back to the Java reflection API in these cases? Writing custom generators that delegate to default string / bool / int Arbs for over 100 properties seems like a lot of unnecessary boilerplate.

Kantis:
Fixes #4901 with minimal changes.

We can consider adding support for customizing exactly which constructor should be used if there's enough desire for it. I played around with doing so in https://github.com/kotest/kotest/commit/5e6cea232f12eb0a2fd5fab7aa76532ca248d267, but I feel the cost of adding that support is not warranted at this point in time.

If we get reasons to revisit the jvmbind APIs for some other reason, I think it would be a good thing to consider including at that time.

### DIFF
--- a/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/jvmbind.kt
+++ b/kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/jvmbind.kt
@@ -5,12 +5,14 @@ import io.kotest.property.Arb
 import io.kotest.property.resolution.CommonTypeArbResolver
 import io.kotest.property.resolution.GlobalArbResolver
 import kotlin.reflect.KClass
+import kotlin.reflect.KFunction
 import kotlin.reflect.KParameter
 import kotlin.reflect.KProperty1
 import kotlin.reflect.KType
 import kotlin.reflect.KVisibility
 import kotlin.reflect.full.primaryConstructor
 import kotlin.reflect.full.starProjectedType
+import kotlin.reflect.jvm.kotlinFunction
 import kotlin.reflect.typeOf
 
 /**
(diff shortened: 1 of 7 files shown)


kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/jvmbind.kt
```diff
@@ -104,7 +105,10 @@ internal fun <T : Any> Arb.Companion.forClassUsingConstructor(
    kclass: KClass<T>
 ): Arb<T> {
    val className = kclass.qualifiedName ?: kclass.simpleName
-   val constructor = kclass.primaryConstructor ?: error("Could not locate a primary constructor for $className")
+   val constructor = kclass.primaryConstructor
+      ?: determineFallbackConstructor(kclass)
+      ?: error("Could not locate a primary constructor for $className")
```
AlexCue987: the error message "Could not locate a primary constructor for" does not match what the code actually does.
would ""Could not locate a constructor for..." describe it better?

AlexCue987 (approved):
LGTM, one error message seems misleading

kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/jvmbind.kt
Kantis: fixed 👍🏼

PR #4913, opened 2025-07-01
Add KSP processor to replace compiler plugin

sksamuel:
If we move to KSP we can use a stable and supported API and not the experimental compiler plugins. This means we shouldn't have to keep making changes each time a new version of Kotlin is released.

This implementation brings full nested test support to JS and Native. 

Also individual tests are now working from intellij for JS platforms, so can you click the run button and run a single test. I still need to confirm that we can do the same for Kotlin Native.

image

There are some further changes required to robust this up that I'll do in follow up PRs.

### DIFF
--- a/kotest-framework/kotest-framework-engine/api/kotest-framework-engine.api
+++ b/kotest-framework/kotest-framework-engine/api/kotest-framework-engine.api
@@ -784,9 +784,6 @@ public final class io/kotest/core/TupleKt {
 }
 
 public final class io/kotest/core/WriteLogKt {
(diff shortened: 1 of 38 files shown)


sksamuel:
Native support can run individual tests too now

image

codymikol:
Hey, for my own understanding, it looks like this is basically

Finding classes to run
Parsing together a kotlin file on the fly that imports those packages and runs them with the TestEngineLauncher

What is the reason for parsing a file out like that instead of just pulling it off the classpath?

PR #4916, opened 2025-07-04
Restore instance per leaf and test; simplify spec execution scopes

sksamuel:
To avoid migration issues in the upcoming 6.0, this re-adds the removed isolation modes, albeit with instance per leaf deprecated. I've moved duplicate name handling and fail fast out of the scopes and into interceptors so in a follow up PR we can really simplify the scope stuff.


kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/extensions/spec/AfterSpecListenerTest.kt
```diff
@@ -90,14 +92,22 @@ class AfterSpecListenerTest : FunSpec() {
...
 
-      test("inline afterSpec function errors should be caught") {
-
-         val collector = CollectingTestEngineListener()
-         TestEngineLauncher(collector)
-            .withClasses(InlineAfterSpecError::class)
-            .launch()
-         collector.specs.size.shouldBe(1)
-         collector.specs[InlineAfterSpecError::class]!!.errorOrNull.shouldBeInstanceOf<ExtensionException.AfterSpecException>()
+      context("f:handle inline afterSpec exceptions") {
```
Kantis:
```suggestion
      context("handle inline afterSpec exceptions") {
```

Kantis (approved):

PR #4918, opened 2025-07-04
conditional publication enabled based on runner, to deal with sharded…

Kantis:
Cherry-picked some stuff from @aSemy's work on Ks3's build setup.

This enables us to only allow publications to be created on certain targets, so we can run macOS builds and just get Apple targets artifacts, etc.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -30,3 +30,8 @@ kotlin.incremental.js=true
 
 android.useAndroidX=true
 android.enableJetifier=true
+
+# Kotest build settings
+# enabled/disable Kotlin targets, for improving local dev & CI/CD performance
+kotest_enableKotlinJs=true
+kotest_enableKotlinNative=false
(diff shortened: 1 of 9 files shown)


aSemy (approved):

gradle.properties
```diff
@@ -30,3 +30,8 @@ kotlin.incremental.js=true
 
 android.useAndroidX=true
 android.enableJetifier=true
+
+# Kotest build settings
+# enabled/disable Kotlin targets, for improving local dev & CI/CD performance
+kotest_enableKotlinJs=true
+kotest_enableKotlinNative=false
```
aSemy, line 37: Will this config also disable the native tests on CI?

Kantis: Good catch.. we should explicitly enable all targets in CI no matter the checked in config..

PR #4925, opened 2025-07-07
Simplify test container usage for kotest 6

sksamuel:


kotest-extensions/kotest-extensions-testcontainers/src/jvmMain/kotlin/io/kotest/extensions/testcontainers/TestContainerProjectExtension.kt
```diff
@@ -0,0 +1,33 @@
...
+ * launch the container once per project, and will not reset it between specs.
+ *
+ * Note: This extension requires Kotest 6.0+
+ *
+ * @param container the specific test container type
+ */
+class TestContainerProjectExtension(
+   private val container: GenericContainer<*>,
+) : MountableExtension<GenericContainer<*>, Unit>, AfterProjectListener, TestListener {
+
```
Kantis, line 22: Would this be installed from an `AbstractProjectConfig`? Or would you just install it in every spec that uses the container anyway?

sksamuel: You install it in every one, but it's only started once, so your test suite is quick. It's like a "lazy" container basically.
We use this pattern everywhere at Grindr.

PR #4928, opened 2025-07-07
add inline to `should` and `shouldNotBeNull`

tKe:
Allow usage of suspend functions from the existing (if historic) `should` and `shouldNotBeNull`  scope functions.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/nulls/matchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/nulls/matchers.kt
@@ -80,7 +80,11 @@ fun <T> T?.shouldNotBeNull(): T {
  *     length should beEven()
  * }
  */
-infix fun <T : Any> T?.shouldNotBeNull(block: T.() -> Unit): T {
+inline infix fun <T : Any> T?.shouldNotBeNull(block: T.() -> Unit): T {
+   contract {
+      returns() implies (this@shouldNotBeNull != null)
+   }
+
    this.shouldNotBeNull()
    block()
    return this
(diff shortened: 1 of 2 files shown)


sksamuel (approved):

codymikol:
Hi, what is the usecase for

```
inline infix fun <T> T.should(matcher: (T) -> Unit) = matcher(this)
```

This makes `should` available as an extension to any type and doesn't require `T` to be a `Matcher<T>`.

I see there is already

```
infix fun <T> T.should(matcher: Matcher<T>) {
   invokeMatcher(this, matcher)
}
```

tKe:
The main usecase is/was as a more semantic version of `also`:

```kotlin
myData should {
  it.message shouldBe "hello"
  it.reason shouldBe "politeness"
}
```

sksamuel:
If we need to revert anything, we need to do it before 6.0 is released this weekend.
Is the downside just that should appears in the auto complete ?

tKe:
This hasn't been added in 6.0, merely made `inline` to support suspend functions. This shouldn't have made it more or less discoverable in autocomplete.

codymikol:
Found one of my devs was trying to use this similar to mockk's `verify`

```kotlin
verify { updateSomething() }

should { updateSomething() }
```

I can add inline documentation that explains this is a semantic alternative to `.also{}`

I'd argue that its probably better to encourage people to just use `.also{}` since

```
myData also message should be 'hello'
```
reads better than

```
myData should message should be 'hello'
```

but that would be a deprecation, not sure if that is something you'd like to do.

codymikol:
Sorry to post this on your PR @tKe , I realize that its only tangentially related :sweat_smile:

codymikol:
Made a docs PR over here https://github.com/kotest/kotest/pull/4986

PR #4930, opened 2025-07-08
Dividing the Kotest release flow into multiple runs

Kantis:
Should make the publication process faster. Also setting the publication name so it's easier to see what's in Central.


sksamuel:
what does publication name do  ?

sksamuel (approved):

Kantis:
I believe it's what we see here: 
image

But it's not really explained in the nmcp docs so gotta confirm that..

Kantis:
> what does publication name do ?

image
seems to have worked

PR #4935, opened 2025-07-10
enabling ksp publication

Kantis:


JesusMcCloud (approved):
even though my reviews don't count, I can attest to the fact that the missing artefact can now be resoved

sksamuel (approved):

PR #4938, opened 2025-07-11
sketch auto-wiring kotest

JesusMcCloud:
This is just a draft. It simply integrates what the [KMP sample](https://github.com/JesusMcCloud/KMPotest/blob/main/shared/build.gradle.kts) does to remove the need for doing it manually.

I don't fee comfortable enough around the kotest project setup to actually finish it, plus I don't have the resources to do so either.

I case there will be any manually written acknowledgement for this in the next release, please tag @a-sit-plus (not that I expect any ACKs for 20 LoC).


Kantis:
Sorry, messed your branch up pretty bad 😞 lets continue in #4943

PR #4943, opened 2025-07-12
Kotest gradle plugin add KSP dependency

Kantis:
Builds on https://github.com/kotest/kotest/pull/4938 (which I unfortunately fucked up, sorry @JesusMcCloud )

### DIFF
--- a/kotest-framework/kotest-framework-plugin-gradle/build.gradle.kts
+++ b/kotest-framework/kotest-framework-plugin-gradle/build.gradle.kts
@@ -12,6 +12,8 @@ plugins {
    id("com.github.node-gradle.node") version "7.1.0"
 }
 
+System.setProperty("kotestVersion", Ci.publishVersion)
+
 dependencies {
    compileOnly(libs.kotlin.gradle.plugin)
    implementation(libs.asm) // used to poke into classes to see if they are specs when running JVM tests
--- a/kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
+++ b/kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
@@ -178,7 +178,6 @@ abstract class KotestPlugin : Plugin<Project> {
     * 2. Wires the Kotest symbol procesor for every configured KMP target
     */
    internal fun KotlinMultiplatformExtension.wireKotestKsp() {
-
       if (!project.pluginManager.hasPlugin("com.google.devtools.ksp")) {
          throw StopExecutionException(
             "KSP neither found in root project nor ${project.name}, " +
@@ -186,26 +185,10 @@ abstract class KotestPlugin : Plugin<Project> {
          )
       }
 
-      var kspConfigsWithKotest = mutableSetOf<String>()
       val version = System.getProperty("kotestVersion")
-      targets.whenObjectAdded {
-         project.dependencies {
-            project.tasks.withType<AbstractTestTask> {
-               runCatching {
-                  project.configurations.names.filter { it.startsWith("ksp") && it.endsWith("Test") }
-                     .forEach { configurationName ->
-                        if (!kspConfigsWithKotest.contains(configurationName)) {
-                           kspConfigsWithKotest.add(configurationName)
-                           logger.lifecycle("  ${this.name}::Adding Kotest $version to $configurationName")
-                           add(
-                              configurationName,
-                              "io.kotest:kotest-framework-symbol-processor-jvm:$version"
-                           )
-                        }
-                     }
-
-               }.getOrElse { logger.warn(it.message) }
-            }
+      project.configurations.whenObjectAdded {
+         if (name.startsWith("ksp") && name.endsWith("Test")) {
+            project.dependencies.add(name, "io.kotest:kotest-framework-symbol-processor-jvm:$version")
          }
       }
    }
--- a/kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
+++ b/kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
@@ -8,14 +8,19 @@ import io.kotest.framework.gradle.tasks.KotestNativeTask
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.plugins.JavaBasePlugin
+import org.gradle.api.tasks.StopExecutionException
+import org.gradle.api.tasks.testing.AbstractTestTask
 import org.gradle.kotlin.dsl.configure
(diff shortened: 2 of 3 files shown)


kotest-framework/kotest-framework-plugin-gradle/build.gradle.kts
```diff
@@ -12,6 +12,8 @@ plugins {
    id("com.github.node-gradle.node") version "7.1.0"
 }
 
+System.setProperty("kotestVersion", Ci.publishVersion)
```
JesusMcCloud, line 15: I'm pretty sure this can be done properly in some other way, but I don't know my way around kotest well enough)

kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
```diff
@@ -155,6 +167,50 @@ abstract class KotestPlugin : Plugin<Project> {
...
+
+      if (!project.pluginManager.hasPlugin("com.google.devtools.ksp")) {
+         throw StopExecutionException(
+            "KSP neither found in root project nor ${project.name}, " +
+               "please add 'com.google.devtools.ksp' to the either project's plugins"
+         )
+      }
+
+      var kspConfigsWithKotest = mutableSetOf<String>()
+      val version = System.getProperty("kotestVersion")
```
JesusMcCloud, line 190: also here, reading this property can probably be done in a cleaner way

kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
```diff
@@ -155,6 +167,50 @@ abstract class KotestPlugin : Plugin<Project> {
...
+            project.tasks.withType<AbstractTestTask> {
+               runCatching {
+                  project.configurations.names.filter { it.startsWith("ksp") && it.endsWith("Test") }
+                     .forEach { configurationName ->
+                        if (!kspConfigsWithKotest.contains(configurationName)) {
+                           kspConfigsWithKotest.add(configurationName)
+                           logger.lifecycle("  ${this.name}::Adding Kotest $version to $configurationName")
+                           add(
+                              configurationName,
+                              "io.kotest:kotest-framework-symbol-processor-jvm:$version"
```
JesusMcCloud: this module name should probably be a constant somewhere?

JesusMcCloud:
this approach may actually be flaky. for me at least, js tests are not run at all. probably this is not happening at the right time in the lifecycle

kotest-framework/kotest-framework-plugin-gradle/build.gradle.kts
Kantis: Kotlin project seems to be using a quite similar method. They set an extra prop on the root gradle module [here](https://github.com/JetBrains/kotlin/blob/9478dd2bae67a92d6dab4c13483d1d34fc6e574c/build.gradle.kts#L76C1-L80C2), and then refer to that [here](https://github.com/JetBrains/kotlin/blob/9478dd2bae67a92d6dab4c13483d1d34fc6e574c/libraries/tools/kotlin-gradle-plugin/build.gradle.kts#L551)

I don't see any big difference in doing it this way. Do you?

kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
```diff
@@ -155,6 +167,33 @@ abstract class KotestPlugin : Plugin<Project> {
       }
    }
 
+   //TODO wire this to JVM test Sources automatically
+   private fun KotlinDependencyHandler.addKotestJvmRunner() {
+      project.logger.info("  Adding Kotest JUnit runner")
+      implementation("io.kotest:kotest-runner-junit5-jvm")
```
sksamuel: I don't think we should do this as in Kotest 6, using junit5 runner is optional and will be discouraged in favour of our more powerful gradle plugin.

JesusMcCloud: so the custom report generator will be used by default?

JesusMcCloud:
> this approach may actually be flaky. for me at least, js tests are not run at all. probably this is not happening at the right time in the lifecycle

More filtering needs to happen during auto-wiring. I'll propose changes here, once i have confidence in a working solution

kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
```diff
@@ -155,6 +167,33 @@ abstract class KotestPlugin : Plugin<Project> {
...
+         )
+      }
+
+      val version = System.getProperty("kotestVersion")
+      project.configurations.whenObjectAdded {
+         if (name.startsWith("ksp") && name.endsWith("Test")) {
+            project.dependencies.add(name, "io.kotest:kotest-framework-symbol-processor-jvm:$version")
+         }
+      }
+   }
```
JesusMcCloud: This prevents wiring to targets that cannot be built on the current host:

```suggestion
            val target = name.substring(3, name.length - 4).replaceFirstChar { it.lowercase() }
            if (project.getBuildableTargets().firstOrNull { target == it.name } != null)
              project.dependencies.add(name, "io.kotest:kotest-framework-symbol-processor-jvm:$version")
         }
      }
   }

fun Project.getBuildableTargets() =
    project.extensions.getByType<KotlinMultiplatformExtension>().targets.filter { target ->
        when {
            // Non-native targets are always buildable
            target.platformType != org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType.native -> true
            else -> {
                runCatching {
                    val konanTarget = (target as? KotlinNativeTarget)
                    konanTarget?.publishable == true
                }.getOrElse { false }
            }
        }
    }

JesusMcCloud:
I'd like to test on a linux machine but i could not get it to publish properly

JesusMcCloud:
I don't know why, but the gradle plugin is looking for io.kotest:io.kotest.gradle.plugin and this artefact does not exist. @sksamuel leaving it as draft because of this

kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
sksamuel: yep

sksamuel:
This version has slightly drifted too, so I've taken this work and made a third! attempt.
https://github.com/kotest/kotest/pull/4961

That PR is ready and I've tested it working on linux, windows and JS.

PR #4954, opened 2025-07-16
Multiplatform JUnit XML Generator

JesusMcCloud:
some TODOs left, but the code is all there


sksamuel:
Updated the PR to focus on generation of the XML file.
A follow up shortly will wire in the actual use.

sksamuel (approved):

sksamuel:
Excellent work @JesusMcCloud, this was a great foundation.

kotest-framework/kotest-framework-multiplatform-reporter/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
+plugins {
+   id("kotest-multiplatform-library-conventions")
+   id("kotest-android-native-conventions")
+   id("kotest-watchos-device-conventions")
+   id("kotest-native-conventions")
+   id("kotest-js-conventions")
+   id("kotest-publishing-conventions")
+   //TODO, once AGP is working: id("com.android.library")
```
JesusMcCloud, line 8: if we don't get this to work, then we need an ugly check in the JVM target to see whether we're running on Android or JVM (for returning the correct target name). Android native is not affected by this.

sksamuel: for initial release we could just not include the target name in the output.

PR #4964, opened 2025-07-26
Implement WASM support

sksamuel:
This setsup WASM wasi publishing for the modules required by the WASM wasi plugins.


buildSrc/src/main/kotlin/kotest-js-wasm-conventions.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+         nodejs()
+      }
+
+      @OptIn(ExperimentalKotlinGradlePluginApi::class)
+      applyHierarchyTemplate(KotlinHierarchyTemplate.default) {
+         group("common") {
+            group("jsHosted") {
+               withJs()
+               withWasmJs()
+               withWasmWasi()
```
OliverO2, line 36: Does that work? Wasm/WASI is not a JS-hosted platform in the sense that a JS API is available.

JesusMcCloud: probably depends on whether you are actually using JS APIs

sksamuel: > Does that work? Wasm/WASI is not a JS-hosted platform in the sense that a JS API is available.

I was just copying what was already there. I thought you had set it up actually. I know next to nothing about wasm.

OliverO2: Yes, it was me who introduced the "jsHosted" group for targets using the JS API. This means Kotlin/JS and Kotlin/Wasm/JS. Kotlin/Wasm/WASI does not have any JS API access, so it should not be in that group. WASI is a completely different runtime with [its own set of APIs](https://wasi.dev/interfaces).

What can be irritating is that there are runtimes like Node.js, which came from the JS world, but can now also run Wasm applications with a purely WASI API. (There are lots of other [Wasm/WASI runtimes](https://github.com/appcypher/awesome-wasm-runtimes?tab=readme-ov-file), which have have no no JS origins at all.)

So we need to take care of the situation when JS APIs are not available.

sksamuel: I've removed jsHosted now and added nonjvm as a group, which results in less boilerplate, as most of the implementations boil down to JVM vs not-JVM, with a few wasm specific things I just left in wasmJs or wasmWasi

PR #4967, opened 2025-07-27
Recursive list shrinking

### LINKED ISSUE #3813: Recursive shrinking

At the moment, shrinkers for e.g. `Arb.list` are unable to shrink the list's elements. I was wondering whether that's a limitation of the underlying `gen` used by `Arb.list`, but providing an inner `Arb` with a custom shrinker showed that the inner shrinker is never called.

I wanted to implement my own variant of `Arb.list` that shrinks recursively (preferring to try smaller lists first, then mapping the inner shrinker over values). I believe this is how other libraries work, like QuickCheck in Haskell or quickcheck in Rust. Unfortunately, kotest's architecture seems to associate shrinkers with individual samples, rather than with an entire `Arb`. A shrinker for a collection doesn't have access to the samples used to build it and therefore can't invoke inner shrinkers recursively. Since I'm used to other property-based testing libraries, this seems like a major limitation that makes shrinking almost useless. A single level of indirection is enough to turn minimal counterexamples into complex, unshrinkable, unreadable data.

Please correct my interpretation where necessary. I'm not entirely confident in my understanding of kotest's property-based testing architecture as documentation on custom generators is somewhat scarce — it only covers very basic examples and doesn't explain the underlying abstractions and algorithms.

### Possible solutions

Changing 
```kt
fun interface Shrinker<A> {
   fun shrink(value: A): List<A>
}
```
into
```kt
fun interface Shrinker<A> {
   fun shrink(value: Sample<A>): List<Sample<A>>
}
```
and making `Sample`s carry extra info is one option, if shrinker-per-sample is the right mindset. It honestly might be, depending on how are shrinkers used in the wild.

Another approach would be maintaining the shrinker hierarchy parallel to the `Arb` hierarchy, exposing shrinkers explicitly in `Arb`s, similarly to classifiers. AFAICT doing so is common in other libraries.

Sintrastes:
This PR is a prototype of the recursive list shrinking feature.

I am not sure if this is the final design we'd want to go with (as it requires `@UnsafeVariance`), and I am open to other suggestions. However, I wanted to get the ball rolling on this as I think it is a really impactful feature.

I originally attempted getting rid of the `out` variance annotations for `Gen` and subclasses, but this seemed to cause issues with `Arb.bind`. I am not sure if there is a way to work around that.

Note also that this is a breaking change, so I would assume this won't make it into kotest until the next major version.

Closes #3813


sksamuel:
Before I review, what's the breaking part of it? With 6.0 about to be released, if we want to make breaking changes, now is the time.

Sintrastes:
> Before I review, what's the breaking part of it? With 6.0 about to be released, if we want to make breaking changes, now is the time.

Actually it may not be strictly speaking breaking. I just know I had to update the `kotest-property/api/kotest-property.api` file, otherwise it wouldn't build. But I suppose that's more of a "making sure the public API doesn't inadvertantly change" thing than a "checking for breaking changes" thing.

I am currently working on fixing a few tests (with the new recursive shrinking behavior, a few of the hard-coded outputs need to be updated), but beyond that the rest of the test suite seems to be passing.

viluon:
Thanks for working on this! However, at first glance, this doesn't seem like a good solution.

Shrinking is necessarily invariant. `Gen`s are necessarily covariant. The two should probably be decoupled: rather than adding `shrinker` to `Gen`, shrinkers should be referenced from elsewhere, such as from the various `forAll` functions.

In ScalaCheck, [shrinkers are invariant](https://github.com/typelevel/scalacheck/blob/a6f79122ccab60ae5b8d77b025d8c8869193ba4b/core/shared/src/main/scala/org/scalacheck/Shrink.scala#L18), [`Gen`s are covariant](https://github.com/typelevel/scalacheck/blob/a6f79122ccab60ae5b8d77b025d8c8869193ba4b/core/shared/src/main/scala/org/scalacheck/Gen.scala#L208), and [arbitraries are invariant, but don't contain shrinkers](https://github.com/typelevel/scalacheck/blob/a6f79122ccab60ae5b8d77b025d8c8869193ba4b/core/shared/src/main/scala/org/scalacheck/Arbitrary.scala#L82-L84).

Sintrastes:
@viluon Yeah, that's why I mentioned I had to use `@UnsafeVariance`. I was following the design of [QuickCheck](https://hackage-content.haskell.org/package/QuickCheck-2.16.0.0/docs/Test-QuickCheck-Arbitrary.html#v:shrink), but I understand if that's not a desired or feasible design direction.

I know you mentioned `forAll`. Would it also be OK I'm assuming to also create something similar to `GlobalArbResolver` / `Arb.resolve`, but for `Shrinker`? I can try working it from that direction instead.

sksamuel:
At the momentt the shrink candidates are generated at the same time as the arbitraries. This gives you more power over what the shrink values should be.

Sintrastes:
> At the momentt the shrink candidates are generated at the same time as the arbitraries. This gives you more power over what the shrink values should be.

What about something like this?

```kotlin
abstract class ArbDefinition<A>: Arb<A>(), Shrinker<A>

// Update arbitrary to return a sub-class that inherits `Shrinker`
fun <A> arbitrary(
   edgecaseFn: (RandomSource) -> Sample<A>?,
   shrinker: Shrinker<A>,
   sampleFn: suspend ArbitraryBuilderContext.(RandomSource) -> A
): ArbDefinition<A> =
   object : ArbDefinition<A>() {
      override fun edgecase(rs: RandomSource): Sample<A>? = edgecaseFn(rs)
      override fun sample(rs: RandomSource): Sample<A> = delegate.sample(rs)

      private val delegate: Arb<A> = arbitraryBuilder(shrinker) { rs -> sampleFn(rs) }
      override fun shrink(value: A): List<A> {
         return shrinker.shrink(value)
      }
   }
   
 // Update Arb.list to check for sub-classes implementing Shrinker
 fun <A> Arb.Companion.list(gen: Gen<A>, range: IntRange = 0..100): Arb<List<A>> {
   check(!range.isEmpty())
   check(range.first >= 0)

   return arbitrary(
      ...
      shrinker = ListShrinker<A>(
         if (gen is ArbDefinition<*>) gen as Shrinker<A> else Shrinker { listOf() }, 
         range
      ),
      ...
   )
}
```

This trades `@UnsafeVariance` for an un-checked cast, but maybe this is at least somewhat of an improvement?

Sintrastes:
Because of the definition of ArbDefinition as well (as opposed to if we just did a `gen is Shrinker<*>` check), I think it's guaranteed that this cast will always succeed.

viluon:
@Sintrastes 
> This trades `@UnsafeVariance` for an un-checked cast

Neither should be necessary. In this case, you should be able to use `gen is ArbDefinition<A>` instead (`A`, not `*`) and thus avoid the cast altogether.

Regarding the design (using a new class to tie `Arb` and `Shrinker` together), I'd say it's a bit confusing to end-users since it's not immediately clear whether to use `Arb` or `ArbDefinition`. The name is quite unclear, though I appreciate it's a work in progress :slightly_smiling_face: 

As an end user, I'd love for it to Just Work :tm:. I'm actually not sure if covariance is important for `Arb`s -- ScalaCheck doesn't have it. It seems you can have ergonomic invariant `Arb`s if you just work with generics everywhere. The API of `forAll` functions already makes that easy. So if dropping `Arb` covariance is an option, moving `Shrinker` to `Arb` seems like a decent choice.

@sksamuel 
> At the momentt the shrink candidates are generated at the same time as the arbitraries. This gives you more power over what the shrink values should be.

Did you mean at the same time as the samples?

Sintrastes:
> Neither should be necessary. In this case, you should be able to use `gen is ArbDefinition<A>` instead (`A`, not `*`) and thus avoid the cast altogether.

I'll push my latest changes. I was in fact able to get rid of this with a bit of refactoring. I'm not sure why IntelliJ was originally not letting me do this.

> Regarding the design (using a new class to tie `Arb` and `Shrinker` together), I'd say it's a bit confusing to end-users since it's not immediately clear whether to use `Arb` or `ArbDefinition`. The name is quite unclear, though I appreciate it's a work in progress 🙂

That's fair. Note that in my latest changes I have made this class `internal` and scrubbed it from the public API. This was basically only intended originally to make that cast safe. 

> As an end user, I'd love for it to Just Work ™️. I'm actually not sure if covariance is important for `Arb`s -- ScalaCheck doesn't have it. It seems you can have ergonomic invariant `Arb`s if you just work with generics everywhere. The API of `forAll` functions already makes that easy. So if dropping `Arb` covariance is an option, moving `Shrinker` to `Arb` seems like a decent choice.

I did attempt this originally, but I was running into trouble with APIs that use an erased type, namely, callsites that use:

```
fun <T : Any> Arb.Companion.bind(
   providedArbs: Map<KClass<*>, Arb<*>>,
   kclass: KClass<T>
): Arb<T>
```

I get stuff like

```
None of the following candidates is applicable:
fun <A> Arb.Companion.choice(arb: Arb<A>, vararg arbs: Arb<A>): Arb<A>
fun <A> Arb.Companion.choice(arbs: List<Arb<A>>): Arb<A>
```

Though I guess you're proposing removing those APIs in favor of generic ones instead? I'm not sure that's possible if we want to be able to pass around `providedArbs` in various places (I've personally used this feature before myself), since Kotlin doesn't have row types. 

Trying to remove the `out` in `Arb` I also ran into a few cases where the variance seemed to be actively used e.x. to handle sub-classes of sealed classes and the like -- and I'm sure there are likely users of kotest doing the same in their code-bases.

sksamuel:
> > At the momentt the shrink candidates are generated at the same time as the arbitraries. This gives you more power over what the shrink values should be.
> 
> Did you mean at the same time as the samples?

I did. I had been drinking 😂

viluon:
@sksamuel 
> At the momentt the shrink candidates are generated at the same time as the samples. This gives you more power over what the shrink values should be.

Well, that's kind of the problem here from my PoV. To quote the original issue (that is, erm, myself):
> Unfortunately, kotest's architecture seems to associate shrinkers with individual samples, rather than with an entire `Arb`. **A shrinker for a collection doesn't have access to the samples used to build it** and therefore can't invoke inner shrinkers recursively.

I think either the shrinker interface and sample structure need to change or shrinkers need to be reassociated with `Arb`s (instead of with individual samples) to fix this. The current situation actually gives you less power than either of the proposed approaches.

I appreciate that a shrinker can currently be tailored to a particular sample. But perhaps this functionality, if it is considered important, could be implemented differently. One option is to make shrinkers take `Sample<A>` instead of `A` and hide extra information in the sample, so that the shrinker can pattern match on it and invoke tailored implementations as necessary. Making `Sample<A>` an `interface`  would allow end users to send arbitrary metadata from the `Gen` to the `Shrinker`.

Sintrastes:
> I think either the shrinker interface and sample structure need to change or shrinkers need to be reassociated with `Arb`s (instead of with individual samples) to fix this. The current situation actually gives you less power than either of the proposed approaches.

I think I agree that these are probably both cleaner options than what I have now.
 
> I appreciate that a shrinker can currently be tailored to a particular sample. But perhaps this functionality, if it is considered important, could be implemented differently. One option is to make shrinkers take `Sample<A>` instead of `A` and hide extra information in the sample, so that the shrinker can pattern match on it and invoke tailored implementations as necessary. Making `Sample<A>` an `interface` would allow end users to send arbitrary metadata from the `Gen` to the `Shrinker`.

Yeah, so I believe you had a similar suggestion to this when you originally filed your ticket. While I'm personally more partial to associating shrinkers with `Arb`s than modifying the `Shrinker` interface itself, I did briefly attempt this.

Unless I'm missing something, I'm not sure it's really possibly to implement recursive list shrinking via that approach -- at least not in a very clean way. Consider, if we change the interface to have `fun shrink(value: Sample<A>): List<Sample<A>>`, that makes the type of the shrinking function of a `List<A>` `fun shrink(value: Sample<List<A>>): List<List<A>>`. The core problem then becomes: How do we determine how to shrink an individual `A` when we are simply given a `Sample<List<A>>` (when what we really need here is either a `Sample<A>` or a `Shrinker<A>`)?

@viluon - I'm not sure if you had anything more concrete in mind for that line of thinking, or if you had any suggestions.

Sintrastes:
Also, I'm not sure why the CI is currently failing. Works On My Machine ™️. 😅

I'm on a Mac (M1) and noticed a lot of tests were skipped. Perhaps that is the cause. I'll look into that more.

Sintrastes:
> Though I guess you're proposing removing those APIs in favor of generic ones instead? I'm not sure that's possible if we want to be able to pass around providedArbs in various places (I've personally used this feature before myself), since Kotlin doesn't have row types.

Following up on this, random thought I just had: If removing the current functionality of the `providedArbs: Map<KClass<*>, Arb<*>>` parameter is not acceptable, it is possible we might be able to implement something similar using Church-encoded existential types.

Sintrastes:
@sksamuel - Does the CI include a timeout for individual tests that is not used when testing locally, perhaps? It looks like the one test that is still failing is timing out. It looks like it is working for me locally, but it does take a long time.

I wonder if the issue is adding the recursive shrinking makes the test take longer to run, and this ends up making it time out on the CI.

Sintrastes:
OK, I found some glaring mistakes in my tests and the shrinking implementation which have now been corrected. All tests should be passing now. I tried reducing the number of iterations of the test that was timing out on the CI, so hopefully that will be working now if my hypothesis for why it was failing is correct.

I'm still waiting on a consensus for how we want to go forward with this. I'm tracking the following potential courses of action based on our discussions so far:

  1. Go with the existing code (internal `ArbDefinition` class) if that is acceptable, since this does not change the user-facing API.
  2. Make `Arb` invariant, and move `val shrinker: Shrinker<A>` to `Arb`, then either trying to remove `providedArbs` or finding an alternative way to provide the same functionality.
  3. Try modifying the API of `Shrinker<A>` to `(Sample<A>) -> List<Sample<A>>` (though per my previous comment I am not sure how to proceed with this line of thought)
  
 Thoughts? @viluon, @sksamuel

viluon:
Sorry for the delay @Sintrastes.

> Unless I'm missing something, I'm not sure it's really possibly to implement recursive list shrinking via that approach -- at least not in a very clean way. Consider, if we change the interface to have `fun shrink(value: Sample<A>): List<Sample<A>>`, that makes the type of the shrinking function of a `List<A>` `fun shrink(value: Sample<List<A>>): List<List<A>>`. The core problem then becomes: How do we determine how to shrink an individual `A` when we are simply given a `Sample<List<A>>` (when what we really need here is either a `Sample<A>` or a `Shrinker<A>`)?
> 
> @viluon - I'm not sure if you had anything more concrete in mind for that line of thinking, or if you had any suggestions.

I think reconciling this issue with shrinker-per-sample is cumbersome but possible. Maybe I'm missing something myself but something like
```kotlin
data class ListSample<out A>(
    val samples: List<Sample<A>>,
) : Sample<List<A>>
```
would let you pass the underlying samples from the list `Arb`'s `Gen` to a `ListShrinker`. You could extend this structure to also pass the `Arb<A>` or whatever else you may need.

This approach is ugly for at least two reasons though:
- you need to write a custom sample class for every structure you want to support and take care of passing around stuff like the underlying `Arb`. Ideally, Kotest would take care of this for you.
- your shrinker still needs to handle arbitrary `Sample`s somehow, e.g. by not doing any shrinking at all. I.e. its interface still accepts `Sample<List<A>>`  rather than `ListSample<A>`.

viluon:
I am not a maintainer :slightly_smiling_face: While I might have opinions, I'm not sure about their relevance here.

Personally, I prefer option 2, i.e. a breaking change. My reasoning is that such a change is necessary in the long-term to support clean, first-class, recursive shrinking[^adoption]. The sooner such a change happens, the less code relies on the old behaviour.

I'll need to have a think about the alternatives and how they translate to user experience. One thing is unclear to me: your proposal and current code make `ArbDefinition` internal. That surely precludes extension (and direct use) by user code, does it not? How could I create an `Arb` that supports recursive shrinking for a custom data structure?

[^adoption]: This, in turn, I consider vital for adoption of Kotest property-based testing. All the major QuickChecks out there support this; complex property-based testing is infeasible without it, as explained in the original issue.

Sintrastes:
> Sorry for the delay

No worries!

> One thing is unclear to me: your proposal and current code make ArbDefinition internal. That surely precludes extension (and direct use) by user code, does it not? How could I create an Arb that supports recursive shrinking for a custom data structure?

You're right -- in the current state of the PR custom user-defined shrinking would not currently be possible. My goal with this prototype was primarily getting recursive shrinking working for lists, so I had not considered this, but I think that is an important requirement.

Technically I could add that back in by adding an extension function / property to `Arb`... But at that point we might as well just move `val shrinker` to `Arb` itself, or make `ArbDefinition` public again. But I think we're both in agreement that (if we can get it to work in a way that @sksamuel approves of), that's going to be the cleanest way to tackle this going forward.

I appreciate the additional details on your `Shrinker` approach, and I will keep those in mind if for whatever reason moving shrinkers to `Arb` itself for some reason does not end up working out.

sksamuel:
Just an FYI that I am not following this thread closely yet as I am laser focused on getting 6.0 ready. As soon as the remaining bug are squashed I will return to this (before 6.0 is released in case want to include this).

sksamuel:
Lets aim to get this into 6.1

Sintrastes:
@sksamuel - Did you have any thoughts about the different approaches / tradeoffs previously discussed?

sksamuel:
I'm excited to review this and will get to it soon - and include it in 6.2 (it missed 6.1 as we released that already).
Can you fix the conflict if you have time?

sksamuel:
> Technically I could add that back in by adding an extension function / property to `Arb`... But at that point we might as well just move `val shrinker` to `Arb` itself, or make `ArbDefinition` public again. But I think we're both in agreement that (if we can get it to work in a way that @sksamuel approves of), that's going to be the cleanest way to tackle this going forward.

I agree.

One option here is to limit the shrinking changes to the new permutations syntax. Since that is net new, you could change the interfaces in anyway we want.

Sintrastes:
> > Technically I could add that back in by adding an extension function / property to `Arb`... But at that point we might as well just move `val shrinker` to `Arb` itself, or make `ArbDefinition` public again. But I think we're both in agreement that (if we can get it to work in a way that @sksamuel approves of), that's going to be the cleanest way to tackle this going forward.
> 
> I agree.
> 
> One option here is to limit the shrinking changes to the new permutations syntax. Since that is net new, you could change the interfaces in anyway we want.

Sorry for going MIA on all this. I think based on this I should be able to get the branch ready for review again. Though I am not tracking this new permutations syntax you are referencing and how that'd work.

Sintrastes:
> > Neither should be necessary. In this case, you should be able to use `gen is ArbDefinition<A>` instead (`A`, not `*`) and thus avoid the cast altogether.
> 
> I'll push my latest changes. I was in fact able to get rid of this with a bit of refactoring. I'm not sure why IntelliJ was originally not letting me do this.
> 
> > Regarding the design (using a new class to tie `Arb` and `Shrinker` together), I'd say it's a bit confusing to end-users since it's not immediately clear whether to use `Arb` or `ArbDefinition`. The name is quite unclear, though I appreciate it's a work in progress 🙂
> 
> That's fair. Note that in my latest changes I have made this class `internal` and scrubbed it from the public API. This was basically only intended originally to make that cast safe.
> 
> > As an end user, I'd love for it to Just Work ™️. I'm actually not sure if covariance is important for `Arb`s -- ScalaCheck doesn't have it. It seems you can have ergonomic invariant `Arb`s if you just work with generics everywhere. The API of `forAll` functions already makes that easy. So if dropping `Arb` covariance is an option, moving `Shrinker` to `Arb` seems like a decent choice.
> 
> I did attempt this originally, but I was running into trouble with APIs that use an erased type, namely, callsites that use:
> 
> ```
> fun <T : Any> Arb.Companion.bind(
>    providedArbs: Map<KClass<*>, Arb<*>>,
>    kclass: KClass<T>
> ): Arb<T>
> ```
> 
> I get stuff like
> 
> ```
> None of the following candidates is applicable:
> fun <A> Arb.Companion.choice(arb: Arb<A>, vararg arbs: Arb<A>): Arb<A>
> fun <A> Arb.Companion.choice(arbs: List<Arb<A>>): Arb<A>
> ```
> 
> Though I guess you're proposing removing those APIs in favor of generic ones instead? I'm not sure that's possible if we want to be able to pass around `providedArbs` in various places (I've personally used this feature before myself), since Kotlin doesn't have row types.
> 
> Trying to remove the `out` in `Arb` I also ran into a few cases where the variance seemed to be actively used e.x. to handle sub-classes of sealed classes and the like -- and I'm sure there are likely users of kotest doing the same in their code-bases.

@sksamuel - If you're proposing we just try making `Arb` invariant though, do you have any suggestions on this issue?

Issue #4884, opened 2025-06-03
Kotest IntelliJ Plugin fails to show diff for org.opentest4j.AssertionFailedError


enolive:
**Which version of Kotest are you using**

5.9.1

**Version of Plugin**

1.3.95-2025.1

**Version of IDEA**

#IU-251.25410.129

# Description

To my surprise I noticed that when using the Kotest Runner, I get no diff for AssertionFailedErrors from opentest4j. In fact, the plugin only seems to show Assertion Errors that implement kotest's own ComparisonError interface.

When running the test with both the gradle or junit runner in IntelliJ, the diff link will be present

Is this the intended behavior?

Issue might be related to this comment:

https://github.com/kotest/kotest/issues/4435#issuecomment-2558358969

# Demo Code:

```kotlin
import io.kotest.common.errors.ComparisonError
import io.kotest.core.spec.style.DescribeSpec
import io.kotest.assertions.AssertionFailedError as KError
import org.opentest4j.AssertionFailedError as O4Error

class SampleWithKoTest : DescribeSpec({
  describe("I wanna try this!") {
    it("whatever") {
     // does not show any diff
      throw O4Error("failed", "Expectation", "Actual", null)
      // will show diff
      throw KError("failed", null, "Expectation", "Actual")
     // also shows diff
      throw MyError("failed", "Expectation", "Actual")
    }
  }
})

class MyError(override val message: String, val expected: String, val actual: String) : RuntimeException(message),
  ComparisonError {
  override val actualValue = actual
  override val expectedValue = expected
}
``` 

# Actual Output:

```
failed
org.opentest4j.AssertionFailedError: failed
	at SampleWithKoTest$1$1$2.invokeSuspend(SampleWithKoTest.kt:24)
...
```

# Expected Output:

```
failed
Expected :Expectation
Actual   :Actual
<Click to see difference>

org.opentest4j.AssertionFailedError: failed
	at SampleWithKoTest$1$1$2.invokeSuspend(SampleWithKoTest.kt:24)
```

sksamuel:
I get an opposite outcome - the kotest one doesn't render a diff.

Can you try this test case:

```
package com.sksamuel.kotest

import io.kotest.common.errors.ComparisonError
import io.kotest.core.spec.style.DescribeSpec
import io.kotest.assertions.AssertionFailedError as KotestAssertionFailedError
import org.opentest4j.AssertionFailedError as Opentest4jError

class DiffTest : DescribeSpec({
[...]

Issue #4710, opened 2025-01-22
Add kotlin-reflect and opentest4j as Gradle `api` dependencies


erikhuizinga:
See my extensive comment on some research (and a fix) here:

https://github.com/kotest/kotest/issues/4089#issuecomment-2606969583

Kantis:
I still think the issue relates to how the Kotest plugin executes tests when not using Gradle. I think @sksamuel's upcoming changes would make that irrelevant, as Gradle will always be invoked.

edgarmiro:
> I still think the issue relates to how the Kotest plugin executes tests when not using Gradle. I think [@sksamuel](https://github.com/sksamuel)'s upcoming changes would make that irrelevant, as Gradle will always be invoked.

Is there any news about this change? 🍿

sksamuel:
You can use the latest intellij plugin and M2 (or the about to be released M3) for a much better experience if you don't mind running prerelease.

erikhuizinga:
Thanks, sksamuel! Just for those who aren’t on top of all things kotest: how long does it usually take for an M-release (I guess a milestone release?) to stabilise? I will wait for that. Also, since I work with both Android Studio and IntelliJ, does that affect when/what I’ll experience with this new plugin version or when I get it?

sksamuel:
Do you want to try M8 with the latest setup guide? https://kotest.io/docs/next/framework/project-setup.html
This is all fixed in the latest code.

sksamuel:
M10 is out

sksamuel:
Since 6.1 all tests run via gradle now, so this issue is fixed.

erikhuizinga:
How sad. This issue ~died~ closed a day before its first birthday. Let's take a moment of silence.

(Thanks for fixing.)

Issue #4645, opened 2025-01-07
Make `MarkAbortedExceptionsAsSkippedTestInterceptor` available for Android


JordanLongstaff:
I'm trying to use [Kotest for Android](https://github.com/LeoColman/kotest-android), but it's simply not working.

```
org.opentest4j.TestAbortedException
java.lang.ClassNotFoundException: org.opentest4j.TestAbortedException
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(Unknown Source)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(Unknown Source)
	at java.base/java.lang.ClassLoader.loadClass(Unknown Source)
	at io.kotest.engine.interceptors.MarkAbortedExceptionsAsSkippedTestInterceptor.intercept(MarkAbortedExceptionsAsSkippedTestInterceptor.kt:24)
	at io.kotest.engine.test.TestCaseExecutor$execute$3$1.invokeSuspend(TestCaseExecutor.kt:100)
	at io.kotest.engine.test.TestCaseExecutor$execute$3$1.invoke(TestCaseExecutor.kt)
	at io.kotest.engine.test.TestCaseExecutor$execute$3$1.invoke(TestCaseExecutor.kt)
	at io.kotest.engine.test.interceptors.ExpectExceptionTestInterceptor.intercept(ExpectExceptionTestInterceptor.kt:18)
	at io.kotest.engine.test.TestCaseExecutor$execute$3$1.invokeSuspend(TestCaseExecutor.kt:100)
	at io.kotest.engine.test.TestCaseExecutor$execute$3$1.invoke(TestCaseExecutor.kt)
	at io.kotest.engine.test.TestCaseExecutor$execute$3$1.invoke(TestCaseExecutor.kt)
	at io.kotest.engine.test.interceptors.TimeoutInterceptor.intercept(TimeoutInterceptor.kt:33)
	at io.kotest.engine.test.TestCaseExecutor$execute$3$1.invokeSuspend(TestCaseExecutor.kt:100)
	at io.kotest.engine.test.TestCaseExecutor$execute$3$1.invoke(TestCaseExecutor.kt)
	at io.kotest.engine.test.TestCaseExecutor$execute$3$1.invoke(TestCaseExecutor.kt)
	at io.kotest.engine.test.interceptors.BlockedThreadTimeoutInterceptor.intercept(BlockedThreadTimeoutInterceptor.kt:79)
[...]

Issue #4435, opened 2024-10-14
Make diff for failed assertions in IDEA IDE


kkocel:
I would like to have similar functionality as Junit 5 has in IDEA.

When assertion fails, there is an option to click a button to see how objects asserted are different:

<img width="285" alt="image" src="https://github.com/user-attachments/assets/8c4fb706-b2c8-4f7b-8f4b-304c9ccf468a">
 
In kotest I get detailed information about objects but no easy option to see the diff.

```
java.lang.AssertionError: Collection should have size 2 but has size 1. Values: [...]
	at com.example.SomeTest.should do sth
```

AlexCue987:
do you have kotest plugin installed ?

kkocel:
Yes

Nikita-Guchakov:
The problem does not related to the plugin, it's about throwing `io.kotest.assertions.AssertionFailedError` instead of `java.lang.AssertionError` for assertions. IDE supports [opentest4j](https://github.com/ota4j-team/opentest4j) initiative, and `AssertionFailedError` extends from `org.opentest4j.AssertionFailedError`. So fixing assertions to throw proper exception will be enough and will work without plugin either.

Same issue as in the [comment](https://github.com/kotest/kotest/issues/4500#issuecomment-2558357818)

kkocel:
@Nikita-Guchakov I've seen that kotest is using those exceptions [already](https://github.com/kotest/kotest/pull/1339/files), maybe it's a bug that it is not used here...

kkocel:
@LeoColman so it's not intellij-plugin issue, it's opentest4j regression

Kantis:
What actual diff do you wish to see here?

assuming `listOf("foo") shouldHaveSize 2` fails, what is a meaningful _diff_?

kkocel:
I would rather be able to leverage [opentest4j](https://github.com/ota4j-team/opentest4j) support in kotest, the assertion I put in the description is a sample.

sksamuel:
I don't think this is an opentest4j thing - the diff is generated by formatting the message in a certain way in the output. We must be doing that wrong somewhere.

Nikita-Guchakov:
@sksamuel opentest4j format is supported by IDEs, it works properly everywhere kotest uses it. Just switching to opentest4j there as well should work for this case too

sksamuel:
[...]

PR #4971, opened 2025-07-28
Centralize all assertion handling into AssertionErrorBuilder.

sksamuel:
Centralize all assertion handling into AssertionErrorBuilder.

* This avoids the myriad of functions we had scattered about that created error messages used by assertion errors.
* Propagates opentest4j comparison errors properly to team city output to show diffs in all contexts
* Adds KMP handling for comparison errors
* Moves all the non-shared assertions/helpers into the assertions core module, leaving the namespace unpolluted for anyone not using kotest assertions
 * Unifies types into the Printed typeclass. No custom plugin mechansim exists for Printed typeclasses, so it's safe to make breaking changes
 * Move the long defunct table testing code into its own module and deprecate it all


sksamuel:
Fixes #4884
Fixes #4710
Fixes #4645
Fixes #4435

PR #4983, opened 2025-07-31
Add test report dir to native

sksamuel:


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/reports/JunitXmlReportTestEngineListener.kt
```diff
@@ -38,11 +39,10 @@ class JunitXmlTestEngineListener(private val testReportsDir: String, hostname: S
...
-      File(testReportsDir).mkdirs()
-      val xml = writer.writeXml(ref.kclass, results)
-      val fileName = ref.kclass.qualifiedName ?: ref.kclass.simpleName
-      Files.writeString(Paths.get(testReportsDir).resolve("TEST-${fileName}.xml"), xml)
-      results.clear()
+      val specName = ref.kclass.bestName()
+      val testFile = "TEST-${specName}.xml"
+      val xml = generator.xml(ref.kclass, results)
+      writeFile(testReportsDir, testFile, xml)
    }
```
JesusMcCloud, line 46: shouldn't this be guarded by a mutex or use streams or some other coroutine-safe mechanism? what can't specs finish in parallel depending on the config?

JesusMcCloud: also: can a spec contain path separators or other special chars that can cause problems for file systems? I am not 100% sure the rudimentary kotlinx.io Path and FS implementations handle it correctly.
What I did find out though, is that there is of course no file system in the browser, so when we try this on js *browser* targets, it will [fail hard](https://kotlinlang.org/api/kotlinx-io/kotlinx-io-core/kotlinx.io.files/-system-file-system.html).  
I have no idea how kotlin.test produces XMLs for browser tests.

JesusMcCloud (changes requested):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/reports/JunitXmlReportTestEngineListener.kt
sksamuel: > shouldn't this be guarded by a mutex or use streams or some other coroutine-safe mechanism? what can't specs finish in parallel depending on the config?

the test engine itself ensures it's all thread safe and only one spec is reported at once (I call it spec pinning)

sksamuel: > also: can a spec contain path separators or other special chars that can cause problems for file systems? I am not 100% sure the rudimentary kotlinx.io Path and FS implementations handle it correctly. What I did find out though, is that there is of course no file system in the browser, so when we try this on js _browser_ targets, it will [fail hard](https://kotlinlang.org/api/kotlinx-io/kotlinx-io-core/kotlinx.io.files/-system-file-system.html). I have no idea how kotlin.test produces XMLs for browser tests.

yeah once this is working for native I will start looking into wasm and js

PR #4986, opened 2025-08-02
docs: explain proper usage of T.should

codymikol:
this clarifies that this function in particular is only a semantic alias for kotlin's `T.also` function/

Fixes N/A


sksamuel (approved):

sksamuel:
ty!

PR #4987, opened 2025-08-02
Add support for -0.0 and 0.0 equality in NumberFormat.Lenient mode

G0xilla:
# Problem
In the current implementation, NumberFormat in lenient mode treats -0.0 and 0.0 as different values, which can cause unexpected behavior when comparing numbers. This issue occurs when you serializing a double value with Kotlin serialization:
```kotlin
@Serializable
data class SomeValue(
    val num: Double,
)

val json = Json.Default
val value = SomeValue(-0.0)
val actualJson = json.encodeToString(value)
println(actualJson) // prints: {"num":-0.0}
```

# Solution
Enhanced Number.lenientEquals to handle this edge case


sksamuel (approved):
ty!

PR #4998, opened 2025-08-07
docs: Fix kotest-extensions-spring documentation for 6.0.0M12

mkienenb:
Update kotest-extensions-spring documentation to reflect new API for activating Spring extension.

I've tested both the all-test-classes ProjectConfig example and the per-test example in my own 6.0.0.M12 project.


sksamuel:
ty!

PR #5000, opened 2025-08-08
Avoid quoting strings in the diff

### LINKED ISSUE #3346: idea: avoid quoting strings to make outputs copy-paste friendly

sksamuel:
Fixes #3346


Nek-12:
Congrats on #5000

PR #5030, opened 2025-08-26
Replace typo in deprecation message of ContainerExtension

gpopides:
While making the required changes for v6 i tried to copy the suggested annotation from the deprecation message which i could not import because it was misspelled


sksamuel (approved):

sksamuel:
Good spot, thank you for creating the PR.

Issue #5022, opened 2025-08-23
InstancePerLeaf executes describe/context block unnecessarily with Kotest 6.0.0


T45K:
**Which version of Kotest are you using**
Kotest 6.0.0 and Kotest 5.9.1

reproducer: https://github.com/T45K/kotest-sandbox/blob/48f1aa4153d05b066f9dead29a91abbf555da0d2/app/src/test/kotlin/io/github/t45k/kotest/instanceperleaf/alt/Test.kt

-----

### The Issue with `InstancePerLeaf` in Kotest 6.0.0

I've noticed a change in how `InstancePerLeaf` works after upgrading to 6.0.0.
My understanding is that `InstancePerLeaf` should create a new test object for each "leaf" test, like an `it` block.
This allows each `it` block to run independently.

However, in version 6.0.0, the parent `describe` and `context` blocks are re-executed for each `it` block.

### Example and Reproducer

Here's a simple test case to demonstrate the issue. I've also provided a link to a reproducer repository.

```kotlin
class Test : DescribeSpec({
    isolationMode = IsolationMode.InstancePerLeaf

    describe("d1") {
        println("d1")

        context("c1") {
            println("c1")

            it("i1") {
                println("i1")
            }
        }

        context("c2") {
            println("c2")

            it("i2") {
                println("i2")
            }
        }
    }
})
```

**Expected behavior (Kotest 5.9.1):**
`d1` -\> `c1` -\> `i1` -\> `d1` -\> `c2` -\> `i2`

**Actual behavior (Kotest 6.0.0):**
`d1` -\> `c1` -\> `d1` -\> `c1` -\> `i1` -\> `c2` -\> `d1` -\> `c2` -\> `i2`

The `d1`, `c1`, and `c2` blocks are being called more than once.
This is problematic for tests that have side effects in these blocks, such as database insertions, because they can cause test failures (e.g., due to unique key constraints).

### Conclusion and Questions

This behavior seems like a bug to me. However,
I'm also aware that `InstancePerLeaf` is deprecated.

Is this a known bug in Kotest 6.0.0, or is this new behavior intended?
Given that `InstancePerLeaf` is deprecated, is it recommended to avoid using it altogether?
I appreciate that you've kept it for migration purposes, but I want to know if I should update my tests to use a different isolation mode.

Thank you for your help.

Jadarma:
[...]

PR #5044, opened 2025-09-01
Fix `InstancePerLeaf` execution order

T45K:
ref https://github.com/kotest/kotest/issues/5022

first, let me appreciate your quick fix for my issue.

unfortunately, your fix seemed not to fix completely
https://github.com/kotest/kotest/issues/5022#issuecomment-3219284408

i tried to fix it by myself. please kindly review my first PR
___

## Solution

the same `LeafLaunchingScope` instances are executed in siblings of test nodes (Container or Leaf).
it means that the first node is always fresh (or seed), and the second and latter nodes always should be executed in fresh spec.
in my solution, i added `hasVisitedFirstNode` property to `LeafLaunchingScope` to manage this.
if the visited nodes are not first one, the node and children nodes will be executed in a fresh spec.
it means that container nodes also can be treated as `target`. so, i extended `executeInFreshSpec` and `target` object comparison

## Concerns

- i removed `AtomicBoolean` and use just `Boolean` instead because the property is enclosed in `LeafLaunchingScope`. do you think this is ok?
- i execute Container type in `executeInFreshSpec`. to do that, i remove `require` block. is that no problem?
- i added some comments and logger. if you think there are more appropriate messages, please let me know.

thank you

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/execution/InstancePerLeafSpecExecutor.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/execution/InstancePerLeafSpecExecutor.kt
@@ -25,8 +25,6 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Semaphore
 import kotlinx.coroutines.sync.withPermit
 import kotlinx.coroutines.withContext
-import kotlin.concurrent.atomics.AtomicBoolean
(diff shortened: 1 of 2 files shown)


sksamuel:
Looks great. I'll merge this in, would be good to have a test that confirms two root level tests work too.

T45K:
@sksamuel 

>  would be good to have a test that confirms two root level tests work too

i added a test https://github.com/kotest/kotest/pull/5044/commits/1ff6a7f3fef77a7c5bee9fe7cf2a7ef787363f45

sksamuel (approved):

PR #5048, opened 2025-09-01
Add test for boolean matchers

### LINKED ISSUE #5001: Improve error message when you do true shouldNotBe true
Improve error message when you do true shouldNotBe true

Image

sksamuel:
Fixes #5001

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/ShouldBeBooleanTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/ShouldBeBooleanTest.kt
@@ -1,0 +1,17 @@
+package com.sksamuel.kotest.matchers
+
+import io.kotest.assertions.throwables.shouldThrowAny
+import io.kotest.core.spec.style.FunSpec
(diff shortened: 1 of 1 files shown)


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/ShouldBeBooleanTest.kt
```diff
@@ -0,0 +1,17 @@
...
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.throwable.shouldHaveMessage
+
+class ShouldBeBooleanTest : FunSpec() {
+   init {
+      test("a shouldBe true should have comparison") {
+         val a = false
+         shouldThrowAny {
+            a shouldBe true
+         }.shouldHaveMessage("expected:<true> but was:<false>")
```
AlexCue987, line 14: do we need a test when the assertion succeeds?

sksamuel: This improvement is about the error message so if it succeeds there would be no error message

PR #5049, opened 2025-09-01
Fix shouldCompleteWithin masking errors

### LINKED ISSUE #4892: `shouldCompleteWithin` should not swallow exceptions
5.9.1

Currently an exception in the `thunk` in `io.kotest.matchers.concurrent.shouldCompleteWithin` will hit the top of the thread and consequently the `latch.countDown()` will never fire. This conceals the underlying test failure.

Something like this would work:

```kotlin
fun <A> shouldCompleteWithin(timeout: Long, unit: TimeUnit, thunk: () -> A): A {

  val ref = AtomicReference<A>(null)
  val throwableRef = AtomicReference<Throwble>(null)

  val latch = CountDownLatch(1)
  val t = thread {
    try {
      val a = thunk()
      ref.set(a)
    } catch (t: Throwable) {
      throwableRef.set(t)
    } finally {
      latch.countDown()
    }
  }

  if (!latch.await(timeout, unit)) {
    t.interrupt()
    throw failure("Test should have completed within $timeout/$unit")
  }

  throwableRef.get()?.let { throw it }

  return ref.get()
}
```

sksamuel:
Fixes #4892


kotest-common/src/jvmTest/kotlin/io/kotest/matchers/concurrent/ShouldCompleteWithinTest.kt
```diff
@@ -0,0 +1,15 @@
...
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.matchers.shouldBe
+import kotlin.time.Duration.Companion.seconds
+
+class ShouldCompleteWithinTest : FunSpec({
+
+   test("shouldCompleteWithin should not swallow threads #4892") {
+      shouldCompleteWithin(5.seconds) {
+         1 shouldBe 2
+      }
```
AlexCue987: do we need tests for other scenarios:

* pass immediately
* time out

sksamuel: There are tests already, I have now moved this test to the same test file. Thanks for the prompt!

sksamuel: Ready for approval

Issue #5061, opened 2025-09-05
`InstancePerLeaf` doesn't call `beforeSpec` hooks for each TestCase execution


T45K:
**Which version of Kotest are you using**
6.0.2

before Kotest 6.0.0, `InstancePerLeaf` calls `beforeSpec` hooks for each TestCase execution.

the below code is an example.
`count` value should be 2 finally because there are 2 test instance (`d1->c1->i1` and `d1->c2->i2`).
this is described in the office doc (https://kotest.io/docs/framework/lifecycle-hooks.html)
but now, `beforeSpec` is called only once in the example

```
private var count = 0

class Test : DescribeSpec({
    isolationMode = IsolationMode.InstancePerLeaf

    beforeSpec {
        count++
    }

    describe("d1") {
        context("c1") {
            it("i1") {
                count shouldBe 1
            }
        }

        context("c2") {
            it("i2") {
                count shouldBe 2
            }
        }
    }
})
```

T45K:
ah, ok. now i found how to call `beforeSpec`.
maybe like this

```InstancePerLeafSpecExecutor.kt
   private suspend fun executeInFreshSpec(testCase: TestCase, ref: SpecRef, specContext: SpecContext) {
      logger.log { "Enqueuing in a fresh spec ${testCase.descriptor}" }

      val spec = inflator.inflate(ref).getOrThrow()

      // we need to find the same root test but in the newly created spec
      val root = materializer.materialize(spec).first { it.descriptor.isPrefixOf(testCase.descriptor) }
      logger.log { "Located root for target $root" }

      // we switch to a new coroutine for each spec instance
      withContext(CoroutineName("spec-scope-" + spec.hashCode())) {
         pipeline.execute(spec, specContext) {
            specContext.beforeSpecInvoked.store(false) // <-- added this
            val result = executeTest(root, testCase.descriptor, specContext, ref)
            Result.success(mapOf(testCase to result))
         }
      }
   }
```

but i'm still not sure how to handle `afterSpec` correctly. i'll investigate much more

T45K:
looks better to create `specContext` again

PR #5066, opened 2025-09-06
Enable to execute `beforeSpec` lifecycle hooks for each `InstancePerLeaf` test execution

T45K:
ref #5061 

## Change points

as-is: use the same `specContext` instance on all test execution.
to-be: create new `specContext` instance every time when the tests are executed in fresh spec or new roots

## Why
[Lifecycle document](https://kotest.io/docs/framework/lifecycle-hooks.html) says that `beforeSpec` is called for each test instances (i.e., test execution in fresh spec), but currently no.
this PR fixes the behavior

## Concerns
this PR fixes only `beforeSpec` issue.
it means `afterSpec` still doesn't work well
i'll investigate this more deeply later.

thank you


sksamuel (approved):

sksamuel:
awesome

PR #5077, opened 2025-09-09
Update usage of Isolation Mode in lifecycle docs.

Jadarma:
Since `InstancePerLeaf` and `InstancePerTest` are deprecated in 6.0+, I replaced their mentions within the spec and extension lifecycle documentation, since it might confuse or enourage users to use them instead of the new `InstancePerRoot`.

```diff
-for example, if `InstancePerLeaf` or `InstancePerTest` isolation modes are used
+for example, if `InstancePerRoot` isolation mode is used
```

I also formatted the table for the (uncut) docs, since it missed trailing pipe symbols or punctuation for some rows.


sksamuel:
awesome thanks

sksamuel (approved):

PR #5083, opened 2025-09-10
Restore io.kotest.assertions.fail

sksamuel:


Zordid (commented):
Cool, this is very helpful!

PR #5085, opened 2025-09-11
feat: add Arb for kotlin.time.Instant

tKe:
This also deprecates and renames the current `io.kotest.property.arbitrary.instant` to `io.kotest.property.arbitrary.javaInstant` to avoid confusion.


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/instant.kt
```diff
@@ -0,0 +1,44 @@
...
+fun KotlinInstantRange.random(random: Random): Instant {
+   try {
+      val seconds = (start.epochSeconds..endInclusive.epochSeconds).random(random)
+
+      val nanos = when (seconds) {
+         start.epochSeconds if seconds == endInclusive.epochSeconds -> start.nanosecondsOfSecond..endInclusive.nanosecondsOfSecond
+         start.epochSeconds -> start.nanosecondsOfSecond..999_999_999
+         endInclusive.epochSeconds -> 0..endInclusive.nanosecondsOfSecond
+         else -> 0..999_999_999
+      }.random(random)
```
AlexCue987: can we add some tests for new code?

tKe: Arguably it's not new code as it was copied over from the Java InstantRange.random function (with minor refactoring to use the new property names and switch to when-guard)

If needed I can add some test cases specifically.

tKe: @AlexCue987 added some tests to cover the edge cases

AlexCue987 (approved):
Very nice!

AlexCue987:
can you run `apiDump`

tKe:
The dump is updated - should be good to add to queue

sksamuel:
Still red

tKe:
Doesn't look like the failure was due to this change - npm network connectivity?

PR #5088, opened 2025-09-12
Positive Boolean matchers return non-null by contract

JordanLongstaff:
If `shouldBeTrue` or `shouldBeFalse` pass, we know the Boolean can't be null, so enforce it by contract.


AlexCue987 (approved):
Nice

JordanLongstaff:
Why is the merge queue so allergic to this PR? Flaky tests? Everything passed on the second attempt and yet it still got kicked out.

sksamuel:
> Why is the merge queue so allergic to this PR? Flaky tests? Everything passed on the second attempt and yet it still got kicked out.

no idea but its happening elsewhere too, might be a GHA issue

PR #5090, opened 2025-09-12
Increment the assertion counter in Arrow assertions that do not call core assertions

jenssuhr:
Some of the Arrow assertions (those with an optional failure message in particular) do not increment the assertion counter and cause false positives with `AssertionMode.Error` enabled.
This PR just adds the calls to increment the counter to the affected functions.


sksamuel (approved):

jenssuhr:
Looks like the PR-Test workflow on Windows failed because of a network issue while fetching npm dependencies. A retry should probably fix it.

PR #5097, opened 2025-09-16
Add missing Extension marker to core extension and listeners.

Jadarma:
Fixed a bug where the `InstantiationListener` was not marked as an `Extension` like the other listeners are.
It could not be registered directly, or found in the extension registry [here](https://github.com/kotest/kotest/blob/090e32115e0e82763ec1d6bbdeb81566a5530039/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/SpecExtensions.kt#L92), unless the implementing class was marked as an `Extension` separately by the user:

```kotlin
suspend fun specInstantiated(spec: Spec) = runCatching {
    logger.log { Pair(spec::class.bestName(), "specInstantiated $spec") }
    specConfigResolver.extensions(spec)
        .filterIsInstance<InstantiationListener>()
        .forEach { it.specInstantiated(spec) }
}
```

Same for `SpecExecutionOrderExtension` and `SpecRefExtension`.

### DIFF
--- a/kotest-framework/kotest-framework-engine/api/kotest-framework-engine.api
+++ b/kotest-framework/kotest-framework-engine/api/kotest-framework-engine.api
@@ -1199,7 +1199,7 @@ public final class io/kotest/core/extensions/RuntimeTagExtension : io/kotest/cor
 	public fun tags ()Lio/kotest/engine/tags/TagExpression;
 }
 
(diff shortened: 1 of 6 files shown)


AlexCue987:
is it possible to add a test that would fail without this change and pass with the change?

sksamuel:
I've added tests.

sksamuel (approved):

PR #5106, opened 2025-09-24
Fix kotest plugin in project with Kotlin 2.2.20

walterbrebels:
Calling `target.publishable` throws an exception in a Kotlin multiplatform project when using Kotlin 2.2.20. This seems to be a Kotlin internal issue because `getOrThrow()` is called on a non-completed future and I can't find an exposed API to complete it.

As a workaround instead of checking if the target is publishable I check if the original task is enabled or not.


sksamuel (approved):

sksamuel:
Nice!

MaxMichel2:
Any news on when this can be merged?

sksamuel:
Can you try 6.0.4

walterbrebels:
Our tests work with kotlin 2.2.20 and kotest 6.0.4

sksamuel:
great

PR #5114, opened 2025-09-29
Enable to call `afterSpec` hooks on `InstancePerLeaf` correctly

T45K:
ref: https://github.com/kotest/kotest/pull/5066

## Background

currently, `InstancePerLeaf` test doesn't call `afterSpec` hooks for each test case execution.
this PR tries to fix the issue

## Root cause

since `executeInFreshSpec` is called as a recursive function, `pipeline#execute` for first test case execution will be completed after executing all leaves.

e.g.,

```
d1 --- i1
    |- i2
```

about the above test case, the execution order will be like

`(start d1 for i1) -> (start i1) -> (finish i1) -> (start d1 for i2) -> (start i2) -> (finish i2) -> (finish d1 for i2) -> (finish d1 for i1)`

but `afterSpec` should be called after `(finish i1)`

## Solution

change to execute test in not recursive function but isolated queue after completing one leaf execution.
in this solution, discovered test candidates are not executed immediately in the test execution but put to queue and executed after the discovering completed (i.e., afterSpec is executed)

to notify the test results to JUnit platform correctly, test ignored/finished notifications are also put to the same queue

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/execution/InstancePerLeafSpecExecutor.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/execution/InstancePerLeafSpecExecutor.kt
@@ -39,8 +39,6 @@ internal class InstancePerLeafSpecExecutor(
    private val extensions = SpecExtensions(context.specConfigResolver, context.projectConfigResolver)
    private val materializer = Materializer(context.specConfigResolver)
    private val results = TestResults()
-   private val testQueue = ArrayDeque<Pair<TestCase, SpecRef>>()
-   private val testFinishListener = mutableListOf<Pair<TestCase, TestResult>>()
 
    private val inflator = SpecRefInflator(
       registry = context.registry,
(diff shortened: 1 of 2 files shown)


T45K:
`InstancePerLeafTest2` is failing...
when `context` has multiple `it`

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/execution/InstancePerLeafSpecExecutor.kt
```diff
@@ -47,179 +51,251 @@ internal class InstancePerLeafSpecExecutor(
...
+         newSpec: Spec,
+         executor: suspend (SpecContext) -> Result<Map<TestCase, TestResult>>
+      ) {
+         val specContext = SpecContext.create()
+
+         withContext(CoroutineName("spec-scope-" + newSpec.hashCode())) {
+            pipeline.execute(newSpec, specContext) {
+               executor(specContext)
+            }
+         }.getOrThrow()
```
T45K, line 136: to fail tests under executing when afterSpec fails, throwing the exception

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/spec/execution/InstancePerLeafSpecExecutor.kt
```diff
@@ -47,179 +51,251 @@ internal class InstancePerLeafSpecExecutor(
...
+            val nestedTestCase = materializer.materialize(nested, testCase)
+
+            if (target != null && !nestedTestCase.descriptor.isPrefixOf(target)) {
+               // Should execute the given test case described by target but traversing an irrelevant test case now
+               // Just return to abort from this logic
+               return
+            }
+            if (hasVisitedFirstNode) {
+               logger.log { Pair(testCase.name.name, "Executing in fresh spec") }
+               discoveredTests.add(TestExecutionInFreshSpec(nestedTestCase, ref))
```
T45K, line 237: originally, targets are executed in fresh spec immediately

sksamuel:
great

Issue #5112, opened 2025-09-28
Make `withData` generic


alfonsoristorato:
Creating this issue to link a PR that I made to make `withData` generic as per conversations had in this PR https://github.com/kotest/kotest/pull/5075

alfonsoristorato:
Agreed to do specific withXXX methods https://github.com/kotest/kotest/pull/5144

PR #5121, opened 2025-10-03
Feature/5112 with data generic

alfonsoristorato:
This PR aims at making `withData` generic, as per this issue https://github.com/kotest/kotest/issues/5112, by exposing an interface that lets the specific `scope` override a function, which is used to determine how a test should be registered when using `withData`. 
Interfaces introduced:

- **`WithDataRootRegistrar`** -> for root tests
- **`WithDataContainerRegistrar`** -> for container tests
- **`WithDataTerminalRegistrar`** -> to help the compiler give errors when `withData` is used in a spec where it cannot be used (for now only nested `withData` in `StringSpec` and `WordSpecShouldContainerScope`)

The reason why I went with an implementable interface and not a generic `withData` extension from `RootScope` and `ContainerScope` (as it was prior to `6.0.0`) is to:
- keep allowing each `scope` determine how to register tests
- keep functionality to give compiler errors when `withData` is used when it shouldn't, as opposed to allowing it and getting Exceptions

If this was to go ahead, there will be a breaking change, because now `withData` extends from one of those interfaces, it no longer requires only to know about `T` (the data type) but also `C` (the test scope). 
Because Kotlin does not support partial inference, when an user uses with data as follows:
```
withData<SimpleClass>(...)
```
it will now need to be changed to either of these two
```
withData<SimpleClass, `SpecificTestScope`>(...)
withData<SimpleClass, _>(...)
```
There is an argument to be made on the necessity of declaring the first type in the first place, which can be fixed by solving the ambiguity of various `withData` methods, but that is not in the scope of this PR.

**Update**: the above can be completely avoided by users, if they use `nameFn` explicitly, and I have updated the docs accordingly.

### Extra Changes
- `WordSpecShouldContainerScope` never allowed `withData`, because of its inability of nesting tests, this PR allows the compiler to give an error when `withData` is attempted within it
- `withDataMap` was used for both `RootScope` and `ContainerScope` prior to `6.0.0`, however as `kotest` moved to `6.0.0` it was reflected only in the `ContainerScope` equivalents, this PR reintroduces `@JvmName("withDataMap")` for the `RootScope` equivalent (if this was intentional and not an omission, happy to remove it)
- new method to register tests -> because now each scope implements `registerWithDataTest`, it is now possible to register tests within a scope as such `registerWithDataTest("foo"){1 + 1 shouldBe 2}` - if this is not something you'd want I am happy to find ways to remove this by somehow making it internal

### Notes
1. I am sure you guys have better knowledge and direction on how to make `withData` generic, but I wanted to contribute with this in case it inspires a final solution
2. if this was to be accepted as is, I would need to update the docs where `withData` is exampled as `withData<Int>(...)` **Update**: Already done


kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/datatest/NullableDataTest.kt
```diff
@@ -12,9 +12,9 @@ class NullableDataTest : FunSpec() {
       )
 
       context("data tests should support nulls") {
-         withData<String?>({ "Test with $it" }, listOfData) { }
+         withData<String?,_>({ "Test with $it" }, listOfData) { }
```
AlexCue987: this is a breaking change, yet the `_` is telling us that the second type is irrelevant. let's try to find a way to get rid of the second type, to make it more intuitive for the end users who should not care about kotest's implementation details

alfonsoristorato: fair point, to be fair it is not irrelevant, it simply is inferred. The issue is not so much the need to declare the `_` but the fact that the code for `withData` was ambiguous to begin with. If you switch to `master` and remove the `String` type (`_` obvs will not be there) you will see that it still fails to compile as it doesn't know which `withData` to use. This seems to have been a problem in the past with other `withData` function, and this is where people opted to use `second: T` to avoid this issue. You can find more info by looking the codebase for `// we need second to help the compiler disambiguate between this and the sequence version`.

Now this further ambiguity can be solved by adding a further param to one of the two `withData` that the compiler gets confused with, even a dummy param, and and that point there will be no need to declare even the first type, and `Kotlin` will start to infer the type for `withData`. I would be happy to propose a solution for this, but in a different PR.

AlexCue987: how about giving the new function a slightly different name and deprecating the old one? long term this would result in code that is simpler to use, right?

alfonsoristorato: I can change the name from withData to something else, sure, but I fear this will create confusion for users, that being said ultimately it is your choice,  :smile. 
But from an user perspective, considering that the majority of users are not likely to have encountered the ambiguous function, this change will be seamless, as withData will continue working as expected.
This becomes a breaking change only for those that have specified a type in their withData declaration, whether intentionally or to avoid the ambiguous issue that the compiler gave them.

As for the second question, I definitely think having withData generic is indeed going to be easier to maintain long term, but also gives more flexibility to users that create their own scopes.

alfonsoristorato: In this last commit [revise ambiguity resolution in docs and update tests](https://github.com/kotest/kotest/pull/5121/commits/fa37ecc5b2e823faf15d62cd140b5a0c6ab2b9bd) I have addressed the ambiguity error by explicitly using `nameFn` type parameter, and updated the docs. I believe providing better docs to users on how to avoid having to declare the type manually is probably the best solution, especially given how this was an issue to begin with (the docs already mentioned of the compiler type inference)

sksamuel:
I would like to think of a way to avoid the type parameters, and also make it work for people who don't specify a name function (which I imagine is most people).

What about having a new parameter, scope, which the user can set to one of two values - CONTAINER or TEST (or LEAF or TEST).

I don't think the users care about "scopes" they just need to know, tests i make here will be at the container level and allow further tests (the default) or they're at TEST level and so are final.

alfonsoristorato:
> I would like to think of a way to avoid the type parameters, and also make it work for people who don't specify a name function (which I imagine is most people).
> 
> What about having a new parameter, scope, which the user can set to one of two values - CONTAINER or TEST (or LEAF or TEST).
> 
> I don't think the users care about "scopes" they just need to know, tests i make here will be at the container level and allow further tests (the default) or they're at TEST level and so are final.

Hello hello 😄 - I think I can come up with something, but just so I make sure I understand your end goal here, your withData function would look like this in pseudo code:
```
class ATest : WhateverSpec({
    // LEAF scope, can have further withData nested within
    withData(
        scope = Scope.LEAF
        nameFn = // as before
        T // this is the normal type, i.e. sequence, list, row, etc.
    ){
        // test logic
    }
    test("a test name"){
        // TEST scope, cannot have further withData nested within
        withData(
            scope = Scope.TEST
            nameFn = // as before
            T // this is the normal type, i.e. sequence, list, row, etc.
        ){
            // test logic
        }
    }
})
```
If I completely misunderstood what you'd like to achieve, please tell me as I am having fun working on this stuff :)

sksamuel:
Exactly that and name is optional, and scope is optional and if not speciifed defaults to today's behavior which is container.

It might not be possible if you have to know at compile time if it's a container scope or not.

alfonsoristorato:
> Exactly that and name is optional, and scope is optional and if not speciifed defaults to today's behavior which is container.
> 
> It might not be possible if you have to know at compile time if it's a container scope or not.

Ok, I tried a few things (this is long, apologies 😄) and I think you were onto something when you said ` It might not be possible if you have to know at compile time if it's a container scope or not.`
**Option 1**
Have an interface without parameterised receiver to expose the `withData` and implement the methods required for each scope. 
This allowed me to avoid having to declare another type, but I do not think it's feasible to use for nested `withData`, this is because at that point the compiler has no longer a reference to to the original receiver. 
The code (broken) is in this branch:
https://github.com/kotest/kotest/compare/master...alfonsoristorato:kotest:feature/5112-withData-generic-option-1 and the errors are visible in any test that nests `withData`.
In this option I was not able to find a good reason to expose a `scope` method to allow the users to choose between `CONTAINER` and `TEST`.

**Option 2**
This is very similar to option 1, with the difference that `WithDataRegistrar` is an abstract class, and instead of extending each scope, I opted to extending the `AbstractContainerScope`. This is also not viable, as even if we manage to make this abstract class usable, all the root scopes will need to somehow adapt to this, meaning that `RootScope` itself would no longer be extended in the root scopes but a new abstract class that extends it.
The code (broken) is in this branch:
https://github.com/kotest/kotest/compare/master...alfonsoristorato:kotest:feature/5112-withData-generic-option-2 and the errors are visible in any test that use `withData`.
In this option I was not able to find a good reason to expose a `scope` method to allow the users to choose between `CONTAINER` and `TEST` - even if I'd ahve fixed all the compilation errors.

**Option 3**
Go back to how it was before 6.x.x - meaning that we would have `RootScope.withData` and `ContainerScope.withData`. I did not write a POC for this, but I think the reason why this was modified to how things are now was to allow each scope to define how the `withData` tests are registered, instead of defining this generically in `RootScope` and `ContainerScope`. Also, neither at this level I was able to find a good reason to expose a `scope` method to allow the users to choose between `CONTAINER` and `TEST`.

Obviously, I am only me, and I am sure that this could be done exactly the way you want, and I just can't figure out how 😄. But, should you also not be able to figure out the best way to handle this, I still stand by this PR, which, true enough, introduces a potential breaking change for those few users that had to resort in declaring the type, but it still gives you more control on how to apply `withData` and provides compiler errors when `withData` is used wrongly.

AlexCue987:
> > Exactly that and name is optional, and scope is optional and if not speciifed defaults to today's behavior which is container.
> > It might not be possible if you have to know at compile time if it's a container scope or not.
> 
> Ok, I tried a few things (this is long, apologies 😄) and I think you were onto something when you said ` It might not be possible if you have to know at compile time if it's a container scope or not.` **Option 1** Have an interface without parameterised receiver to expose the `withData` and implement the methods required for each scope. This allowed me to avoid having to declare another type, but I do not think it's feasible to use for nested `withData`, this is because at that point the compiler has no longer a reference to to the original receiver. The code (broken) is in this branch: [master...alfonsoristorato:kotest:feature/5112-withData-generic-option-1](https://github.com/kotest/kotest/compare/master...alfonsoristorato:kotest:feature/5112-withData-generic-option-1) and the errors are visible in any test that nests `withData`. In this option I was not able to find a good reason to expose a `scope` method to allow the users to choose between `CONTAINER` and `TEST`.
> 
> **Option 2** This is very similar to option 1, with the difference that `WithDataRegistrar` is an abstract class, and instead of extending each scope, I opted to extending the `AbstractContainerScope`. This is also not viable, as even if we manage to make this abstract class usable, all the root scopes will need to somehow adapt to this, meaning that `RootScope` itself would no longer be extended in the root scopes but a new abstract class that extends it. The code (broken) is in this branch: [master...alfonsoristorato:kotest:feature/5112-withData-generic-option-2](https://github.com/kotest/kotest/compare/master...alfonsoristorato:kotest:feature/5112-withData-generic-option-2) and the errors are visible in any test that use `withData`. In this option I was not able to find a good reason to expose a `scope` method to allow the users to choose between `CONTAINER` and `TEST` - even if I'd ahve fixed all the compilation errors.
> 
> **Option 3** Go back to how it was before 6.x.x - meaning that we would have `RootScope.withData` and `ContainerScope.withData`. I did not write a POC for this, but I think the reason why this was modified to how things are now was to allow each scope to define how the `withData` tests are registered, instead of defining this generically in `RootScope` and `ContainerScope`. Also, neither at this level I was able to find a good reason to expose a `scope` method to allow the users to choose between `CONTAINER` and `TEST`.
> 
> Obviously, I am only me, and I am sure that this could be done exactly the way you want, and I just can't figure out how 😄. But, should you also not be able to figure out the best way to handle this, I still stand by this PR, which, true enough, introduces a potential breaking change for those few users that had to resort in declaring the type, but it still gives you more control on how to apply `withData` and provides compiler errors when `withData` is used wrongly.

my vote would be to sacrifice nested `withData` - most likely it's rarely used and let's go for simpler code that is easier to maintain. Can Option 1 work without nested  `withData` ?

alfonsoristorato:
> my vote would be to sacrifice nested withData - most likely it's rarely used and let's go for simpler code that is easier to maintain. Can Option 1 work without nested withData ?

It can, but that means that nothing can nest within `withData` anymore. Meaning we won't be able to do:
```
         withData(
            "foo",
            "bar",
            "baz"
         ) {
            it("test") {}
         }

         withData(
            "foo",
            "bar"
         ) {
            "context $it" - {
               "test $it" {
                  this.testCase.descriptor.path() shouldBe DescriptorPath("io.kotest.datatest.styles.FreeSpecDataTest/inside a context -- $it -- context $it -- test $it")
               }
            }
         }

         withData("p", "q") { a ->
            withData(listOf("r", "s")) { b ->
               withData(sequenceOf("x", "y")) { c ->
                  a + b + c shouldHaveLength 3
               }
            }
         }

         withData(
            "foo",
            "bar"
         ) {
            describe("describe $it") {
               it(it) {
                  this.testCase.descriptor.path() shouldBe DescriptorPath("io.kotest.datatest.styles.DescribeSpecDataTest/inside a context -- $it -- describe $it -- $it")
               }
            }
         }
```

and so forth, you get the idea. `withData` in itself will become final, which doesn't seem ideal to me.

AlexCue987:
> > my vote would be to sacrifice nested withData - most likely it's rarely used and let's go for simpler code that is easier to maintain. Can Option 1 work without nested withData ?
> 
> It can, but that means that nothing can nest within `withData` anymore. Meaning we won't be able to do:
> 
> ```
>          withData(
>             "foo",
>             "bar",
>             "baz"
>          ) {
>             it("test") {}
>          }
> 
>          withData(
>             "foo",
>             "bar"
>          ) {
>             "context $it" - {
>                "test $it" {
>                   this.testCase.descriptor.path() shouldBe DescriptorPath("io.kotest.datatest.styles.FreeSpecDataTest/inside a context -- $it -- context $it -- test $it")
>                }
>             }
>          }
> 
>          withData("p", "q") { a ->
>             withData(listOf("r", "s")) { b ->
>                withData(sequenceOf("x", "y")) { c ->
>                   a + b + c shouldHaveLength 3
>                }
>             }
>          }
> 
>          withData(
>             "foo",
>             "bar"
>          ) {
>             describe("describe $it") {
>                it(it) {
>                   this.testCase.descriptor.path() shouldBe DescriptorPath("io.kotest.datatest.styles.DescribeSpecDataTest/inside a context -- $it -- describe $it -- $it")
>                }
>             }
>          }
> ```
> 
> and so forth, you get the idea. `withData` in itself will become final, which doesn't seem ideal to me.

In my book, simplicity is really important. So yes, it totally makes sense to make `withData` a bit simpler - it's a nice compromise. So some advanced abilities might have to go, but that's worth it. We'll end up with less code, simpler code, easier to maintain.

sksamuel:
I agree with @AlexCue987 that simplicity (for the end user, doesn't matter how complicated it is for us to implement) is very important.

Let me throw another idea into the mix.

* We deprecate withData.
* We add withXXX methods which mimic the same functionality of the keywords they represent. 
* Eg, for FunSpec we provide withContext (creates container level tests) and withTest (creates test level tests).

As a full example
```kotlin
         withContexts("p", "q") { a ->
            withContexts(listOf("r", "s")) { b ->
               withTests(sequenceOf("x", "y")) { c ->
                  a + b + c shouldHaveLength 3
               }
            }
         }
```

We _could_ leave withData alive as a synonym to withContexts (and the equiv in other specs) so people don't have to migrate who don't care.

To implement this all we have to do is add some new extension functions to each scope. It would be statically resolvable. No need for the type parameters. And the user can choose if they want container tests or not.

It also means the syntax now follows the spec style too.

alfonsoristorato:
cool, thanks both. My idea of simplicity was to allow the maintainers to maintain less code and also give users less headaches moving forward when using `withData`. 
Also, my assumption was that we wanted to end up with less code, hence this PR that delegated the `withData` registrars to each spec by allowing each spec to define only the final `withData` caller.
That being said, here is a PR based off the latest suggestion os @sksamuel.
As you can see I have only done `behaviourSpec`, as I would not want to do them all and then go back to something else 😄.

As for  `behaviourSpec` itself, I do not see how we can introduce `withTests` as the way I understand it, the base tests within this spec are `and` tests.
The `withTests` to me, would make sense in a spec like `StringSpec`. 
Let me know what you think, and if we are happy I will do the rest of the specs accordingly.
https://github.com/kotest/kotest/pull/5144

sksamuel:
> cool, thanks both. My idea of simplicity was to allow the maintainers to maintain less code and also give users less headaches moving forward when using `withData`. Also, my assumption was that we wanted to end up with less code, hence this PR that delegated the `withData` registrars to each spec by allowing each spec to define only the final `withData` caller. That being said, here is a PR based off the latest suggestion os @sksamuel. As you can see I have only done `behaviourSpec`, as I would not want to do them all and then go back to something else 😄.
> 
> As for `behaviourSpec` itself, I do not see how we can introduce `withTests` as the way I understand it, the base tests within this spec are `and` tests. The `withTests` to me, would make sense in a spec like `StringSpec`. Let me know what you think, and if we are happy I will do the rest of the specs accordingly. #5144

I think this is a step in the right direction.
I don't follow the "and" test comment though, aren't the final TEST types "thens" in BehaviorSpec ?

alfonsoristorato:
> I don't follow the "and" test comment though, aren't the final TEST types "thens" in BehaviorSpec ?

What I am trying to say is that the current implementation of `withData `for `BehaviorSpec ` is to wrap the tests in `and` when in `WhenContainerScope `or  `GivenContainerScope`,  as per https://github.com/kotest/kotest/blob/master/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/behaviorSpecWhenContainerScope.kt and https://github.com/kotest/kotest/blob/master/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/behaviorSpecGivenContainerScope.kt. 

Now we have four options:
1. stick with their implementation and give a better name, like you suggested, therefore `withAnds`
2. change their implementations and completely change to `withThens` - but this will make the underlying registered tests currently in a `given `or a `when `switch `from `and to `then`
3. implement all possible types and let users choose
4. call these guys `withTests` even though they call `and `or `then`

sksamuel:
> > I don't follow the "and" test comment though, aren't the final TEST types "thens" in BehaviorSpec ?
> 
> What I am trying to say is that the current implementation of `withData `for `BehaviorSpec ` is to wrap the tests in `and` when in `WhenContainerScope `or `GivenContainerScope`, as per https://github.com/kotest/kotest/blob/master/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/behaviorSpecWhenContainerScope.kt and https://github.com/kotest/kotest/blob/master/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/behaviorSpecGivenContainerScope.kt.
> 
> Now we have four options:
> 
> 1. stick with their implementation and give a better name, like you suggested, therefore `withAnds`
> 2. change their implementations and completely change to `withThens` - but this will make the underlying registered tests currently in a `given `or a `when `switch `from `and to `then`
> 3. implement all possible types and let users choose
> 4. call these guys `withTests` even though they call `and `or `then`

Behavior spec is the style that has the most options - whens givens thens ands contexts, so it has tons of combinations.

I don't know if the best approach is to just leave withData as an alias to whatever it does today (so for behavior spec, it now forwards to withAnd, and in fun spec it forwards to withContexts) ?

alfonsoristorato:
>Behavior spec is the style that has the most options - whens givens thens ands contexts, so it has tons of combinations.
I don't know if the best approach is to just leave withData as an alias to whatever it does today (so for behavior spec, it now forwards to withAnd, and in fun spec it forwards to withContexts) ?

I guess for `behaviorSpec` it is up to you, I can write the `withThens`, `WithGivens` and whatever else for it, so users can mix and match if they want, but I would personally leave the default `withData` alias to what it does today in all specs, so like you say:
- withContexts for `funSpec`
- withAnds for `behaviorSpec`
and so forth

sksamuel:
Ok that works for me.
We probably don't need to deprecate withData either, just remove it from the docs to encourage the new style (once its out) and leave the old in place as a permanent alias.

alfonsoristorato:
in that case @sksamuel and @AlexCue987 do you want to review the other PR - as it is now fully ready for behaviourSpec, and potentially I can make more PRs with the remaining specs and one final PR with the docs?

Issue #5123, opened 2025-10-07
allow `shouldBeEqualToUsingFields` to be given different types


alfonsoristorato:
**Please describe the feature you'd like to see including any solutions in mind if you have any**
Because its similar function `shouldBeEqualToIgnoringFields` does, maybe indirectly, allow to test with different types, I think it would be cool if `shouldBeEqualToUsingFields` also allows this, as the checks happen at property level

alfonsoristorato:
merged in and will release in 6.0.5

PR #5124, opened 2025-10-07
`shouldBeEqualToUsingFields` with different types

alfonsoristorato:
#### `shouldBeEqualToUsingFields` with different types
Addressing my own feature request in https://github.com/kotest/kotest/issues/5123

This change introduces the ability in `shouldBeEqualToUsingFields` to allow for equality assertions between objects of different types.
This is useful in scenarios where you want to compare objects that share common properties but are not of the same class, such as comparing a domain model with a DTO.

#### Changes
- `KProperty1<T, *>` has been introduced instead of `KProperty<*>` to ensure at compile time that the properties provided belong to the caller's type `T`. 
- The internal comparison logic has been updated to look up the corresponding property by name on the other object, in case `T` and `V` are of different class.


kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
```diff
@@ -32,11 +32,23 @@ import kotlin.reflect.jvm.isAccessible
...
+ * foo.shouldBeEqualToUsingFields(fuu, Foo::description) // Assertion passes
+ *
+ * foo shouldBe fuu // Assertion fails, `equals` is false!
+ * ```
  *
  * Note: Throws [IllegalArgumentException] if [properties] contains any non public property
  *
  */
-fun <T : Any> T.shouldBeEqualToUsingFields(other: T, vararg properties: KProperty<*>) {
+fun <T : Any, V: Any> T.shouldBeEqualToUsingFields(other: V, vararg properties: KProperty1<T, *>) {
```
alfonsoristorato, line 51: using `KProperty1` to ensure users are forced to provide `KProperty` of the type of the receiver which should be used to obtain the value of the property, which is `T` in this case

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/equality/ReflectionKtTest.kt
```diff
@@ -96,6 +114,13 @@ class ReflectionKtTest : FunSpec() {
          car2.shouldBeEqualToIgnoringFields(car1, Car::price)
       }
 
+      test("shouldBeEqualToIgnoringFields should compare equality for class having private fields, even when other class is different type") {
```
alfonsoristorato, line 148: tests with `AnotherCar` are to prove that some sort of different type assertion was already present, possibly indirectly, as the tests here happen on the `other` directly

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
```diff
@@ -389,13 +401,35 @@ fun <T : Any> beEqualComparingFields(
...
-      val expected = it.getter.call(other)
+private fun <T, V> checkEqualityOfFields(fields: List<KProperty<*>>, value: T, other: V): List<String> {
+   requireNotNull(value)
+   requireNotNull(other)
+   val equalityChecker: (actual: Any?, expected: Any?, propertyName: String) -> String? =
+      { actual, expected, propertyName ->
+         val isEqual = EqCompare.compare(actual, expected, false) == null
+         if (isEqual) null else "$propertyName: ${actual.print().value} != ${expected.print().value}"
+      }
+   return when (value::class == other::class) {
```
AlexCue987, line 433: do we need two separate branches here? will everything work exactly as before if we keep only the `else` branch?

AlexCue987 (commented):
what if my intent is to fail my test if `expected` and `actual` are of different types - which is different from what you want but still a legitimate need? should we merge this PR as is, will my failing tests begin to pass, producing false positives?

alfonsoristorato:
> what if my intent is to fail my test if `expected` and `actual` are of different types - which is different from what you want but still a legitimate need? should we merge this PR as is, will my failing tests begin to pass, producing false positives?

That is a valid point, and yes, in theory your failing tests will start passing and vice-versa, but I have two points to this 😄 
Assuming we have such data classes:
```
   data class Foo(val a: String, val b: Int, val c: Boolean)
   data class Fuu(val a: String, val b: Int, val c: Boolean)
```
1. Why would an user have a failing test in their code?
2. In the case they have a test such as this:
    ```
             shouldThrow<IllegalArgumentException> {
                Foo("sammy", 42, true).shouldBeEqualToUsingFields(Fuu("sammy", 42, true))
             }
    ```
   then their test will start failing as such: `Expected exception java.lang.IllegalArgumentException but no exception was thrown.` because this is now a valid test (that all the fields in `Foo` are indeed the same in `Fuu`.

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
alfonsoristorato: you are right, if we stick only with the `false` case, then there would not be any difference in the assertions - I guess the `true` case avoids instantiating `otherPropertiesByName` where not needed, but it's so minor that I would be happy to stick with the `false` case directly

AlexCue987:
I agree the feature you are proposing is nice to have. But I don't think it should break existing behaviour for those of us who like it the way it is. So how about we create a new matcher for your use case, and leave the current one as is. Of course the code to implement both assertions would be largely similar and common parts can be reused.

alfonsoristorato:
> I agree the feature you are proposing is nice to have. But I don't think it should break existing behaviour for those of us who like it the way it is. So how about we create a new matcher for your use case, and leave the current one as is. Of course the code to implement both assertions would be largely similar and common parts can be reused.

That is completely fair, I just wanted to avoid having similar functions that do the exact same thing, but if it's the preferred approach I will adapt it to this :)

alfonsoristorato:
> > I agree the feature you are proposing is nice to have. But I don't think it should break existing behaviour for those of us who like it the way it is. So how about we create a new matcher for your use case, and leave the current one as is. Of course the code to implement both assertions would be largely similar and common parts can be reused.
> 
> That is completely fair, I just wanted to avoid having similar functions that do the exact same thing, but if it's the preferred approach I will adapt it to this :)

@AlexCue987 addressed your comments regarding having a different function but not https://github.com/kotest/kotest/pull/5124#discussion_r2414185086, as I think there's value in avoiding creating the map when there is no need. lmk what you think, and if this gets merged I will do other PRs for the remainder of the `reflection` functions

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/equality/reflection.kt
```diff
@@ -40,6 +40,35 @@ fun <T : Any> T.shouldBeEqualToUsingFields(other: T, vararg properties: KPropert
...
+ *
+ * foo.shouldBeEqualToUsingFields(fuu, Foo::description) // Assertion passes
+ *
+ * foo shouldBe fuu // Assertion fails, `equals` is false!
+ * ```
+ *
+ * Note: Throws [IllegalArgumentException] if [properties] contains any non-public property
+ *
+ */
+fun <T : Any, V: Any> T.shouldBeEqualToDifferentTypeUsingFields(other: V, vararg properties: KProperty1<T, *>) {
```
alfonsoristorato, line 72: using `KProperty1` to ensure users are forced to provide `KProperty` of the type of the receiver which should be used to obtain the value of the property, which is `T` in this case

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/equality/ReflectionKtTest.kt
```diff
@@ -172,6 +172,15 @@ class ReflectionKtTest : FunSpec() {
          }
       }
 
+      test("shouldBeEqualToUsingFields should throw exception when other is of different type") {
+         val car1 = Car("Car", 12345, 23)
+         val car2 = AnotherCar("Car", 12345, 23)
+
+         assertThrows<IllegalArgumentException>("object is not an instance of declaring class") {
```
AlexCue987, line 179: nice call!

alfonsoristorato:
@AlexCue987 I had forgotten about this, if you're still happy with it, can I get an approval? Can be merged in 6.1 whenever that is, but at least I will know what to do with this pr :)

AlexCue987:
> @AlexCue987 I had forgotten about this, if you're still happy with it, can I get an approval? Can be merged in 6.1 whenever that is, but at least I will know what to do with this pr :)

why is the build failing?

alfonsoristorato:
> > @AlexCue987 I had forgotten about this, if you're still happy with it, can I get an approval? Can be merged in 6.1 whenever that is, but at least I will know what to do with this pr :)
> 
> why is the build failing?

Must have been a fluke, re-running it now

alfonsoristorato:
@AlexCue987 as @sksamuel is getting ready to release 6.0.5, can we try to get this in too?

AlexCue987:
> @AlexCue987 as @sksamuel is getting ready to release 6.0.5, can we try to get this in too?

the following tests are failing on Windows:

SystemPropertyConfigClassTest[jvm] > system property override should be used for config classname[jvm] FAILED

Execution failed for task ':kotlinNpmInstall'.
>                 Process 'Resolving NPM dependencies using yarn' returns 1
                  yarn install v1.22.17

I think these are unrelated. Let's make "system property override should be used for config classname" run on Linux only. I'll take care of that tomorrow.

For now, I'm merging this as is. If you disagree @sksamuel I'll revert later.

Issue #5125, opened 2025-10-08
Test instantiation fails if the production source set contains a class or data class `PackageConfig`


ursjoss:
With kotest-6.0.3 in a kotlin project, I ran into the following issue:

```
Class PackageConfig should have a zero-arg constructor
java.lang.IllegalArgumentException: Class PackageConfig should have a zero-arg constructor
```

Turns out that if the production source set contains a class or data class called `PackageConfig`, a test in the same package with access to that production source set will fail.

For a simple reproducer, check out https://github.com/ursjoss/kotest-packageconfig-reproducer

sksamuel:
Fixed in #5131 will be in 6.0.4 later today.

ursjoss:
@sksamuel I don't see that 6.0.4 solves this issue. See https://github.com/ursjoss/kotest-packageconfig-reproducer. Please reopen the ticket or let me know what I missed and how to fix that.

sksamuel:
I cloned your project and set kotest to 6.0.4 and I cannot duplicate the issue. Can you check ?

ursjoss:
You are absolutely right. I had missed that renovate had only updated the plugins while still leaving the dependencies on 6.0.3. Should have set up the version catalog properly 🙄.

Sorry for the noise. All good. Thanks for the fix.

sksamuel:
Glad its working!

PR #5131, opened 2025-10-12
`PackageConfig` checked bug

alfonsoristorato:
Addresses bug reported in https://github.com/kotest/kotest/issues/5125 by checking that the `kclass` returned is assignable from `AbstractPackageConfig`, meaning that the `kclass` extends it, rather than just checking on name equality.

### How to test this - now
- create a random class, or data class (as per example in issue) called `PackageConfig` with some `ctor` params in a test package
- run the test and watch it not fail
- create a class called `PackageConfig`, make it extend `AbstractPackageConfig` and watch it properly use this `PackageConfig` as expected

### How to test this - before
- create a random class, or data class (as per example in issue) called `PackageConfig` with some `ctor` params in a test package
- remove the new `takeIf` clause
- run the test and watch it fail


sksamuel (approved):

sksamuel:
nice

PR #5139, opened 2025-10-16
Fix: `InstancePerLeaf` tests fail when a test leaf has siblings

T45K:
sending notification logic was incorrect


AlexCue987 (commented):
InstancePerLeaf is deprecated. We should migrate from it, not fix it.

sksamuel:
> InstancePerLeaf is deprecated. We should migrate from it, not fix it.

We have to continue to support it though given it has been around for so long.

Issue #5138, opened 2025-10-16
Clues are not collected by arrow assertions


RafeArnold:
Kotest version: 6.0.4

The assertion helper functions provided by the arrow assertions library do not collect clues upon failure.

To reproduce, run the following:
```kotlin
import arrow.core.left
import io.kotest.assertions.arrow.core.shouldBeRight
import io.kotest.assertions.withClue
import io.kotest.matchers.ints.shouldBeGreaterThan

fun main() {
  withClue("my clue") {
//    1 shouldBeGreaterThan 1
    1.left().shouldBeRight()
  }
}
```
Observe that
```
java.lang.AssertionError: Expected Either.Right, but found Either.Left with value 1
```
is thrown, but if `1 shouldBeGreaterThan 1` in the above snippet is uncommented then
```
org.opentest4j.AssertionFailedError: my clue
1 should be > 1
```
 is thrown and contains the clue value. I'd expect the first assertion error to also include the clue value in its message.

RafeArnold:
fixed by https://github.com/kotest/kotest/pull/5140

PR #5140, opened 2025-10-16
fix(arrow): collect clues on failures

RafeArnold:
fix for https://github.com/kotest/kotest/issues/5138

- **test(arrow): assert unhappy paths for either + ior**
- **fix(arrow): collect clues on failures.**
- **test(arrow-fx): assert unhappy paths for ExitCase.**
- **fix(arrow-fx): collect clues on failures.**


sksamuel (approved):

sksamuel:
nice thank you!

RafeArnold:
thanks for merging so fast and for maintaining a great library!

PR #5141, opened 2025-10-17
Add focused methods to spec styles

### LINKED ISSUE #5087: Add focused test methods
`DescribeSpec` has functions like `xdescribe` and `xit` which are the "skip" variants of the normal test descriptor functions, as inspired by Jest. Jest also has functions called `fdescribe` and `fit` which are the focus variants. I'd like to see those functions added to `DescribeSpec` also, and other kinds of specs as appropriate (e.g. `fcontext`).

sksamuel:
Closes #5087


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/DescribeSpecRootScope.kt
```diff
@@ -21,24 +19,69 @@ import io.kotest.core.test.TestScope
...
+         disabled = false,
+         config = null,
+      ) { DescribeSpecContainerScope(this).test() }
+   }
+
+   fun fcontext(name: String, test: suspend DescribeSpecContainerScope.() -> Unit) {
+      addContainer(
+         TestNameBuilder.builder(name).withPrefix("Context: ").build(),
+         focused = true,
+         disabled = false,
```
AlexCue987: combinations like the following are repeated many times in this PR:
```
focused = true,
disabled = false,
```
do we want to add an enum with three elements: REGULAR, FOCUSED, DISABLED and two attributes: `focused` and `disabled`?

then we could have just one function with the body like this:
```
addContainer(
         TestNameBuilder.builder(name).withPrefix("Context: ").build(),
         focused = testType.focused,
         disabled = testTpe.disabled,
         config = null,
      ) { DescribeSpecContainerScope(this).test() }
```
and our three similar functions will just call it with that enum:
```
fun context(name: String, test: suspend DescribeSpecContainerScope.() -> Unit) = context(name, test, REGULAR)
```

AlexCue987 (approved):
LGTM, suggested a possible simplification

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/DescribeSpecRootScope.kt
sksamuel: Ok good idea, check latest commit.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/behaviorSpecContextContainerScope.kt
```diff
@@ -228,7 +229,7 @@ suspend fun <T> BehaviorSpecContextContainerScope.withContexts(
    test: suspend BehaviorSpecContextContainerScope.(T) -> Unit
 ) {
    ts.forEach { t ->
-      context(nameFn(t), false) { this.test(t) }
+      context(nameFn(t), xmethod = TestXMethod.NONE) { this.test(t) }
```
alfonsoristorato, line 232: this will have to be repeated a few lines below to the other `withxxx`

alfonsoristorato (commented):
looks good aside the comment + api docs. 
I think on a general note this will require updates to the `intellij-plugin` to mark these new `f` accordingly

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/behaviorSpecContextContainerScope.kt
sksamuel: good spot, done

PR #5144, opened 2025-10-18
Specific data test methods for various specs

alfonsoristorato:
introduces `withXXX` for each combination in Specs, allowing withData to continue working as before.


AlexCue987:
is it still a draft?

alfonsoristorato:
> is it still a draft?

I'd say so, we are discussing it in the other pr :)

sksamuel:
lgtm,

should we go away and add a simpler one too to iron out any incompatibilities ?

sksamuel (approved):

sksamuel:
Also can we not merge this until 6.0.5 has been released, as we should put this only in the 6.1.0 release.

alfonsoristorato:
Sounds good re waiting 6.1.0, in that case, would you rather have all the code changes in this pr, so to avoid pr overload ?

sksamuel:
Your call - happy for one big one if that saves you sequencing headaches.

Should we add some tests around beforeEach, beforeAny, beforeAll and how they now interact with withXXX ?

I wonder if there's any   code to be saved by parameterizing them like we did for test config?
So that withContexts delegates to a withContainerDataTest method that just takes a type parameter for the concrete type it delegates to (similar to your original idea, but means that anyone can extend their own spec style easily if needed). I would only consider doing this if you think it saves you some lines of code overall.

alfonsoristorato:
> should we go away and add a simpler one too to iron out any incompatibilities ?

stringSpec added to this PR with `withTests`

> Should we add some tests around beforeEach, beforeAny, beforeAll and how they now interact with withXXX ?

I have added some tests for this, it's particularly visible that `beforeEach` like before counts only before every Container level test and `beforeAny` before any test, whether Container or Test scopes.

> I wonder if there's any code to be saved by parameterizing them like we did for test config?

I don't think we can get any gainings by parameterizing this at this level, as in, I can parameterize it but the DSL will still need to be repeated for every spec, if anything it might add more code into the mix.

sksamuel:
Ok cool. I wonder if we shouldn't even bother with StringSpec, kinda seems pointless to mix up plain strings with function calls. Same for free spec? Thoughts ?

alfonsoristorato:
> Ok cool. I wonder if we shouldn't even bother with StringSpec, kinda seems pointless to mix up plain strings with function calls. Same for free spec? Thoughts ?

That's fair, I will bring it back to withData, as long as we keep withData alive tho, these are my favourite specs :)

alfonsoristorato:
@sksamuel and @AlexCue987 all possible specs are done - please have a look and if we are all happy I can do the docs

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/datatest/styles/WordSpecDataTest.kt
```diff
@@ -15,26 +16,41 @@ class WordSpecDataTest : WordSpec() {
...
-         count++
+         afterTestCounter++
       }
 
       afterSpec {
-         count shouldBe 68
+         afterTestCounter shouldBe 146
+         beforeAnyCounter shouldBe 146
+         // only counting tests within the shoulds
+         beforeEachCounter shouldBe 36
```
alfonsoristorato, line 40: this was quite interesting, `beforeEach` is only called for tests within the `should` - i.e. [this](https://github.com/kotest/kotest/blob/2070a45b74ca4f9f5ae94811b09480baf187f167/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/datatest/styles/WordSpecDataTest.kt#L181-L183), [this](https://github.com/kotest/kotest/blob/2070a45b74ca4f9f5ae94811b09480baf187f167/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/datatest/styles/WordSpecDataTest.kt#L193-L195) and [this](https://github.com/kotest/kotest/blob/2070a45b74ca4f9f5ae94811b09480baf187f167/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/datatest/styles/WordSpecDataTest.kt#L198-L200)

sksamuel:
We could release 6.0.5 with what's in master, and then merge this in ready for 6.1 ?

alfonsoristorato:
@sksamuel sounds good to me, this is ready but I'd like to write some docs for it. But I am in hols this week, so if we can delay this for next week, or thereafter, it'd be good

AlexCue987:
Merging it now as is. Pls don't forget about the docs @alfonsoristorato

AlexCue987 (approved):
LGTM

alfonsoristorato:
Hey Hey, @AlexCue987 docs are already in, but are you sure you want to merge this now? I thought the plan was for this to go in 6.1.0, but it seems we'll have to do a 6.0.6 to release some fixes for mingwx.
Removed from the queue for now (did not know I could 😄) - feel free to re-add if we are ready to release this now

AlexCue987:
> Hey Hey, @AlexCue987 docs are already in, but are you sure you want to merge this now? I thought the plan was for this to go in 6.1.0, but it seems we'll have to do a 6.0.6 to release some fixes for mingwx. Removed from the queue for now (did not know I could 😄) - feel free to re-add if we are ready to release this now

I don't see how this nice feature is throwing a wrench in the next release. It seemed ready to go to me. What am I missing here?

alfonsoristorato:
I am happy to release, this is ready imo, just basing the hold on this https://github.com/kotest/kotest/pull/5144#issuecomment-3490016181

Issue #5143, opened 2025-10-17
Non-deterministic test (retry, continually) duration config doesn't interrupt the test


cheiily:
**Kotest 6.0.4**
The problem from #5110 (non-deterministic methods not interrupting the test after provided timeout) persists for `retry` and `continually`.

While 
```kotlin
"test duration eventually" {
    eventually(3.seconds) {
        delay(6.seconds)
        true shouldBe true
    }
}

"test duration until" {
    until(3.seconds) {
        delay(6.seconds)
        true
    }
}
```
will now both be interrupted after 3 seconds (as `until` uses the `eventually` function internally), these tests
```kotlin
"test duration retry" {
    retry(1, 3.seconds) {
        delay(6.seconds)
        true shouldBe true
    }
}

"test duration continually" {
    continually(3.seconds) {
        delay(6.seconds)
        true shouldBe true
    }
}
```
will still both spin for 6 seconds.
<img width="296" height="94" alt="Image" src="https://github.com/user-attachments/assets/a049fa94-f831-4551-b786-c4493c04a7ad" />

The documentation for `retry` says "We still provide a timeout period to avoid the loop running for ever.", though perhaps an immediate interrupt might not be necessary here as the retry count takes precedence here?
The docs for `continually` do not mention timeouts but i believe not interrupting after the provided period is bound to cause issues when used to test indefinitely-lasting functions with no return points, which seems to be primary aim of `continually` in the first place.

sksamuel:
Thanks, good catch. Will be in 6.0.5

sksamuel:
6.0.5 is now released

PR #5145, opened 2025-10-18
add max timeout bounds to continually and retry

alfonsoristorato:
addresses https://github.com/kotest/kotest/issues/5143 - inspired from https://github.com/kotest/kotest/pull/5134

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/StringSpecRootScope.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/StringSpecRootScope.kt
@@ -37,6 +37,7 @@ interface StringSpecRootScope : RootScope {
       enabledOrReasonIf: EnabledOrReasonIf? = null,
       coroutineDebugProbes: Boolean? = null,
       blockingTest: Boolean? = null,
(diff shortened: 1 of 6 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/StringSpecRootScope.kt
```diff
@@ -37,6 +37,7 @@ interface StringSpecRootScope : RootScope {
       enabledOrReasonIf: EnabledOrReasonIf? = null,
       coroutineDebugProbes: Boolean? = null,
       blockingTest: Boolean? = null,
+      coroutineTestScope: Boolean? = null,
```
alfonsoristorato, line 40: hopefully it wasn't intentional to not pas `coroutineTestScope` for `StringSpecRoot` configs

sksamuel (approved):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/StringSpecRootScope.kt
sksamuel: good catch

PR #5149, opened 2025-10-20
Remove periods from test names only at TCSM stage

### LINKED ISSUE #5064: Dot in test's name causes it not to run
When running a specific test using the Kotest IntelliJ plugin (6.0.101-2025.2) this works:
```kotlin
test("parse googlecom query") {}
```

and this doesn't:
```kotlin
test("parse google.com query") {}
```

```
Testing started at 20:20 ...

ParseDnsMessageTest/parse google com query is excluded by filter(s)
```

Here's my libs.versions.toml:
```toml
[versions]
kotest = "6.0.2"
kotlin = "2.2.0"

[plugins]
kotest = { id = "io.kotest", version.ref = "kotest" }

[libraries]
kotest-assertions-core = { module = "io.kotest:kotest-assertions-core" }
kotest-bom = { module = "io.kotest:kotest-bom", version.ref = "kotest" }
kotest-framework-engine = { module = "io.kotest:kotest-framework-engine" }
```

sksamuel:
Fixes #5064

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/teamcity/names/ParentNameStripper.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/teamcity/names/ParentNameStripper.kt
@@ -7,7 +7,7 @@ import io.kotest.core.names.TestName
 // start of the nested test to make the child nest have a different prefix.
 // Also note: This only affects non-MPP tests, as MPP tests have the platform name added
 internal object ParentNameStripper {
-   fun stripe(name: TestName, parent: TestName?): String {
+   fun strip(name: TestName, parent: TestName?): String {
       return when {
          parent == null -> name.name
          name.name.startsWith(parent.name) -> "- " + name.name
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/teamcity/names/ParentNameStripper.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/teamcity/names/ParentNameStripper.kt
@@ -1,0 +1,17 @@
+package io.kotest.engine.teamcity.names
+
+import io.kotest.core.names.TestName
+
+// Note: intellij has a bug, where if a child test has a name that starts with the parent test name,
+// then it will remove the common prefix from the child, to workaround this, we will add a dash at the
+// start of the nested test to make the child nest have a different prefix.
+// Also note: This only affects non-MPP tests, as MPP tests have the platform name added
+internal object ParentNameStripper {
+   fun stripe(name: TestName, parent: TestName?): String {
+      return when {
+         parent == null -> name.name
+         name.name.startsWith(parent.name) -> "- " + name.name
+         else -> name.name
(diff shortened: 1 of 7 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/teamcity/names/ParentNameStripper.kt
```diff
@@ -0,0 +1,17 @@
+package io.kotest.engine.teamcity.names
+
+import io.kotest.core.names.TestName
+
+// Note: intellij has a bug, where if a child test has a name that starts with the parent test name,
+// then it will remove the common prefix from the child, to workaround this, we will add a dash at the
+// start of the nested test to make the child nest have a different prefix.
+// Also note: This only affects non-MPP tests, as MPP tests have the platform name added
+internal object ParentNameStripper {
+   fun stripe(name: TestName, parent: TestName?): String {
```
AlexCue987: should it be `strip` not `stripe`?

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/listener/TeamCityTestEngineListenerTest.kt
```diff
@@ -409,6 +409,34 @@ a[testSuiteStarted name='TeamCityTestEngineListenerTest' id='com.sksamuel.kotest
...
+            listener.specStarted(SpecRef.Reference(TeamCityTestEngineListenerTest::class))
+            listener.testStarted(aperiods)
+            listener.testFinished(aperiods, TestResult.Success(124.milliseconds))
+            listener.specFinished(
+               SpecRef.Reference(TeamCityTestEngineListenerTest::class),
+               TestResult.Success(0.seconds)
+            )
+            listener.engineFinished(emptyList())
+         }
+         output shouldBe """a[enteredTheMatrix durationStrategy='MANUAL']
```
AlexCue987, line 435: should this test ever fail, especially in CI, it would take some time to understand what is different. should we split this string by `a[` and match the collection of parts via `containExactly`?

AlexCue987 (approved):
LGTM, minor suggestions

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/teamcity/names/ParentNameStripper.kt
sksamuel: good spot, fixed

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/listener/TeamCityTestEngineListenerTest.kt
sksamuel: I think trying to work out what's broken might be better than writing some string parsing and hoping we don't introduce bugs there ?
I don't expect this will ever break, and if it does, it would fail locally too.

PR #5154, opened 2025-10-21
Fix StackOverflowError on cyclic maps/collections (fixes #5105)

### LINKED ISSUE #5105: `StackOverflowError` when asserting deep equality on cyclic maps in Kotest 6.0.3 (works in 5.9.1)
## Summary

After upgrading from Kotest v5.9.1 to **v6.0.3**, deep equality assertions (`shouldBe`) on objects that contain a self-referential (cyclic) map trigger a `StackOverflowError`. The same test passes on 5.9.1.

This appears to be a regression in the v6 Eq comparators (`CollectionEq` / `MapEq`) where cycle detection is missing, leading to infinite recursion.

## Affected versions
- ✅ 5.9.1 — no crash (test passes)
- ❌ 6.0.3 — `StackOverflowError`

## Environment
- Kotlin: 2.2.20
- JDK: 17
- Gradle: 8.14.3
- OS: macOS (also likely OS-agnostic)

### build.gradle.kts (test deps)
```kotlin
dependencies {
    testImplementation(platform("io.kotest:kotest-bom:6.0.3"))
    testImplementation("io.kotest:kotest-runner-junit5")
    testImplementation("io.kotest:kotest-assertions-core")
}
tasks.test { useJUnitPlatform() }
```

### ReproducerSpec.kt
```kotlin

class ReproducerSpec : StringSpec({
  "deep equality over a cyclic map should not overflow" {
    // Create a self-referential map
    val a = mutableMapOf<String, Any?>()
    a["b"] = a

    // Put it inside a container map to trigger deep equality
    val parsed = mapOf("foo" to mapOf("bar" to "baz", "baz" to a))

    // This assertion overflows on Kotest 6.0.3, passes on 5.9.1
    (parsed["foo"] as Map<*, *>)["baz"] shouldBe a
  }
})
```

## Expected behavior
Kotest should either:
- handle cyclic structures safely (detect cycles and short-circuit), or
- fail fast with a clear message indicating cyclic structures aren’t supported for deep equality.

## Actual behavior (6.0.3)

The test fails with a `StackOverflowError` due to infinite recursion in the equality comparators:
```
java.lang.StackOverflowError
    at io.kotest.assertions.eq.CollectionEq.equalsIgnoringOrder(CollectionEq.kt:179)
    at io.kotest.assertions.eq.CollectionEq.checkSetEquality(CollectionEq.kt:25)
    at io.kotest.assertions.eq.CollectionEq.equals(CollectionEq.kt:12)
    at io.kotest.assertions.eq.CollectionEq.equals(CollectionEq.kt:8)
    at io.kotest.assertions.eq.EqCompare.compare(EqCompare.kt:11)
    at io.kotest.assertions.eq.MapEq.equals(MapEq.kt:16)
    at io.kotest.assertions.eq.MapEq.equals(MapEq.kt:13)
    at io.kotest.assertions.eq.EqCompare.compare(EqCompare.kt:11)
    ... (frames repeat until SOE)
```

## Notes / Analysis
- The issue surfaces when `shouldBe` performs deep structural equality on graphs containing cycles.
- In v6, `EqCompare` routes map/collection comparisons through `MapEq`/`CollectionEq` without any visited-pair guard, so comparing two cyclic maps (even the same instance) will descend indefinitely via the self-reference.
- In 5.9.1 the same scenario did not overflow (likely due to different short-circuiting paths), so this looks like a regression.

## Workarounds
- Avoid deep equality for cyclic graphs; assert identity instead:
```kotlin
val baz = (parsed["foo"] as Map<*, *>)["baz"]
(baz === a) shouldBe true
```
- Or assert cycle shape explicitly (using reference checks) without `shouldBe`.

sksamuel:
## Summary

After upgrading from Kotest v5.9.1 to **v6.0.3**, deep equality assertions (`shouldBe`) on objects that contain a self-referential (cyclic) map trigger a `StackOverflowError`. The same test passes on 5.9.1.

This PR fixes the regression by adding identity checks to prevent infinite recursion when comparing cyclic structures.

## Changes

- **MapEq**: Added identity check (`===`) at the beginning of the `equals` method
- **CollectionEq**: Added identity check (`===`) at the beginning of the `equals` method
- **Tests**: Added comprehensive test cases for cyclic maps and collections

## Root Cause

The issue occurred because `MapEq` and `CollectionEq` lacked cycle detection. When comparing a cyclic structure with itself, the code would infinitely recurse through the self-reference, eventually causing a StackOverflowError.

## Solution

Added identity checks at the start of both equality comparators. If `actual === expected` (same instance), we immediately return `null` (no error), short-circuiting the deep comparison. This handles cyclic references safely while maintaining correct behavior for all other cases.

## Test plan

- ✅ Added test case: "should handle cyclic maps without StackOverflowError"
- ✅ Added test case: "should handle cyclic maps in nested structures"
- ✅ Added test case: "should handle cyclic collections without StackOverflowError"
- ✅ All existing tests pass
- ✅ Verified the fix resolves the reproduction case from #5105

## Fixes

Closes #5105


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/MapEqTest.kt
```diff
@@ -170,4 +170,26 @@ class MapEqTest : FunSpec({
...
    }
 
+   test("should handle cyclic maps without StackOverflowError") {
+      // Create a self-referential map
+      val cyclicMap = mutableMapOf<String, Any?>()
+      cyclicMap["self"] = cyclicMap
+
+      // Comparing a cyclic map with itself should work (same instance)
+      MapEq.equals(cyclicMap, cyclicMap, false).shouldBeNull()
+   }
```
AlexCue987, line 180: should we add a test where the cycle has more than one node?

```
val cyclicList1 = mutableListOf<Any?>()
val cyclicList2 = mutableListOf<Any?>()
cyclicList1.add(cyclicList2)
cyclicList2.add(cyclicList1)
```

AlexCue987 (approved):
LGTM, suggested one more test

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/MapEqTest.kt
sksamuel: Yeah this is a good idea, this test currently fails. It's a much more complicated fix to handle recursive references. I'm going to create a new issue to handle that.

PR #5156, opened 2025-10-21
Wire kotest task to Gradle check task for better integration (fixes #5116)

### LINKED ISSUE #5116: Better gradle integration without junit5
Since version 6 of Kotest, there is a gradle plugin providing the kotest specific tasks. However:

* those tasks (like `kotest`) are not hooked as generic gradle `check` task
* the common gradle `test` task becomes unusable but hooked as `check` dependency making it to fail

The only way for using `check` task is currently using junit5 runner - which is acceptable but would be better to use direct kotest way.

The ideal solution would be if the kotest plugin would configure `test` task to use `kotest` directly.

sksamuel:
## Summary

Fixes #5116 by wiring the `kotest` task to the standard Gradle `check` task, enabling proper integration with Gradle's verification workflow.

## Problem

The Kotest Gradle plugin created a `kotest` task in the VERIFICATION_GROUP but did not wire it to the standard `check` task. This caused several issues:

1. ❌ Running `gradle check` would NOT run Kotest tests
2. ❌ Users had to remember to run `gradle kotest` explicitly instead of using standard Gradle commands
3. ❌ CI/CD pipelines using `gradle check` would miss Kotest tests entirely
4. ❌ The standard `test` task could fail if no JUnit tests existed, breaking the `check` task

## Solution

Added a simple dependency from the `check` task to the `kotest` task immediately after the kotest task is registered:

```kotlin
// Wire the kotest task to the check task so that 'gradle check' runs Kotest tests
project.tasks.named(JavaBasePlugin.CHECK_TASK_NAME).configure {
   dependsOn(KOTEST_TASK_NAME)
}
```

This follows Gradle best practices where verification tasks should be dependencies of the `check` task.

## Benefits

- ✅ `gradle check` now runs Kotest tests automatically
- ✅ Better integration with standard Gradle workflows
- ✅ CI/CD pipelines work as expected without custom configuration
- ✅ Consistent with how other testing frameworks (JUnit, TestNG) integrate with Gradle
- ✅ Users can rely on standard Gradle commands

## Test plan

- ✅ Built the Gradle plugin successfully
- ✅ Verified compilation with new task wiring
- ✅ Confirmed check task dependency is established

### DIFF
--- a/kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
+++ b/kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
@@ -65,6 +65,11 @@ abstract class KotestPlugin : Plugin<Project> {
          description = TASK_DESCRIPTION
       }
 
+      // Wire the kotest task to the check task so that 'gradle check' runs Kotest tests
+      project.tasks.named(JavaBasePlugin.CHECK_TASK_NAME).configure {
+         dependsOn(KOTEST_TASK_NAME)
+      }
+
       // configures standalone Kotlin JVM projects
       handleKotlinJvm(project)
 


alfonsoristorato (commented):
this works in general, but on a general note, should we disable the test task itself, as I think it now runs and produces nothing? 
It would also be cool to disallow setting `useJUnitPlatform` when using this plugin.
I admit my knowledge here is very limited as I normally use kotest with junit and only now trying to give the plugin and framework a go 😄 so trying to understand and my questions might be silly

PR #5157, opened 2025-10-21
Add virtualTime parameter to eventually for real-time support (fixes #5147)

### LINKED ISSUE #5147: Ugrading Kotest Assertsions from 6.0.3 to 6.0.4 Failed My Tests
In our project, we used Github Dependentbot to manage the dependency, there is a PR used to upgrade Kotest assertions to 6.0.4 from 6.0.3, there are several tests failed with the following messages.

```bash 
org.opentest4j.AssertionFailedError: Block failed after 1.593626ms; 
attempted 2 time(s)
The first error was caused by: Timed out after 6s of
 _virtual_ (kotlinx.coroutines.test) time. 

To use the real time, wrap 'withTimeout' in 'withContext(Dispatchers.Default.limitedParallelism(1))'
kotlinx.coroutines.TimeoutCancellationException: Timed out 
after 6s of _virtual_ (kotlinx.coroutines.test) time. 

To use the real time, wrap 'withTimeout' in withContext(Dispatchers.Default.limitedParallelism(1))'
at kotlinx.coroutines.TimeoutKt.TimeoutCancellationException(Timeout.kt:189)
```

sksamuel:
## Summary

Fixes #5147 by adding a `virtualTime` configuration option to the `eventually` function, allowing users to choose between virtual time and real time execution.

## Problem

PR #5134 added `withTimeout` to `eventually` to enforce timeouts, which correctly respects virtual time from `kotlinx.coroutines.test.runTest`. However, this broke existing code that:

1. Used `runTest` (which sets up virtual time by default)
2. Integrated with blocking libraries that use `Thread.sleep` or perform real I/O
3. Expected `eventually` to wait for actual elapsed time, not virtual time

Users reported errors like:
```
Timed out after 6s of _virtual_ (kotlinx.coroutines.test) time.
To use the real time, wrap 'withTimeout' in 'withContext(Dispatchers.Default.limitedParallelism(1))'
```

## Root Cause

When `runTest` sets up virtual time, `eventually`'s `withTimeout` respects that virtual time. For blocking operations (like `Thread.sleep` or I/O), virtual time doesn't advance because these aren't suspending coroutine operations. This causes timeouts even though real time hasn't actually passed.

## Solution

Added a `virtualTime: Boolean` parameter to `EventuallyConfiguration`:
- **When `true` (default)**: Uses virtual time if available - maintains current behavior
- **When `false`**: Switches to `Dispatchers.Default.limitedParallelism(1)` to use real time

This follows the exact recommendation from the `kotlinx.coroutines.test` error message.

## Changes

- **EventuallyConfiguration**: Added `virtualTime` parameter (default: `true`)
- **EventuallyConfigurationBuilder**: Added `virtualTime` property with documentation
- **eventually()**: Added logic to wrap in `withContext(Dispatchers.Default.limitedParallelism(1))` when `virtualTime=false`
- **Convenience overload**: Added `eventually(duration, virtualTime, test)` for easy usage
- **until.kt**: Updated to pass `virtualTime=true` (maintains current behavior)

## Usage

```kotlin
// For blocking operations - use real time
eventually(5.seconds, virtualTime = false) {
    launch(Dispatchers.IO) {
        Thread.sleep(1000)  // blocking operation
        success = true
    }
    success shouldBe true
}

// Or with configuration
eventually(eventuallyConfig {
    duration = 5.seconds
    virtualTime = false
}) {
    // blocking operations here
}

// Default behavior (virtual time) unchanged
eventually(5.seconds) {
    // suspending operations work as before
}
```

## Benefits

- ✅ **Backward compatible**: Default `virtualTime=true` maintains current behavior
- ✅ **Fixes reported issue**: Users can set `virtualTime=false` for blocking operations
- ✅ **Clear API**: Parameter name makes the behavior explicit
- ✅ **Follows best practices**: Uses the exact solution recommended by kotlinx.coroutines.test
- ✅ **Flexible**: Users can choose per-call which timing mode they need

## Test plan

- ✅ All existing EventuallyTest tests pass
- ✅ Verified compilation and test execution
- ✅ Default behavior unchanged (backward compatible)

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
@@ -4,7 +4,9 @@ import io.kotest.assertions.AssertionErrorBuilder
 import io.kotest.common.nonDeterministicTestTimeSource
 import io.kotest.matchers.ErrorCollectionMode
 import io.kotest.matchers.errorCollector
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeout
 import kotlin.reflect.KClass
 import kotlin.time.Duration
@@ -37,6 +39,25 @@ suspend fun <T> eventually(
 }
 
 /**
+ * Runs a function [test] until it doesn't throw as long as the specified duration hasn't passed.
+ *
+ * @param virtualTime When false, switches to Dispatchers.Default to use real time instead of
+ * virtual time from kotlinx.coroutines.test. This is useful when integrating with blocking
+ * libraries or when you need actual time to pass.
+ */
+suspend fun <T> eventually(
+   duration: Duration,
+   virtualTime: Boolean,
+   test: suspend () -> T,
+): T {
+   val config = eventuallyConfig {
+      this.duration = duration
+      this.virtualTime = virtualTime
+   }
+   return eventually(config, test)
+}
+
+/**
  * Runs a function [test] until it doesn't throw, using the supplied [config].
  */
 suspend fun <T> eventually(
@@ -52,8 +73,16 @@ suspend fun <T> eventually(
    val start = nonDeterministicTestTimeSource().markNow()
    val control = EventuallyControl(config, start)
    try {
-      return withTimeout(config.duration) {
-         runIterations(control, test, config)
+      return if (config.virtualTime) {
+         withTimeout(config.duration) {
+            runIterations(control, test, config)
+         }
+      } else {
+         withContext(Dispatchers.Default.limitedParallelism(1)) {
+            withTimeout(config.duration) {
+               runIterations(control, test, config)
+            }
+         }
       }
    } finally {
       errorCollector.setCollectionMode(originalAssertionMode)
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
```diff
@@ -52,8 +73,16 @@ suspend fun <T> eventually(
    val start = nonDeterministicTestTimeSource().markNow()
    val control = EventuallyControl(config, start)
    try {
-      return withTimeout(config.duration) {
-         runIterations(control, test, config)
+      return if (config.virtualTime) {
```
AlexCue987, line 76: this code is duplicated:
```
withTimeout(config.duration) {
            runIterations(control, test, config)
         }
```

do we want to extract a small function?

AlexCue987 (commented):
do we need to update https://kotest.io/docs/assertions/eventually.html

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/eventually.kt
sksamuel: this is just AI slop, I'll clean it up

sksamuel: Moving this to a new PR

sksamuel:
Moving this to a new PR

PR #5165, opened 2025-10-25
Added a special --tests filter for the intellij plugin

sksamuel:
This will allow us to run all tests from intellij from the gradle runner and ditch the idea runner.


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/descriptors/descriptor.kt
```diff
@@ -90,9 +83,15 @@ sealed interface Descriptor {
...
    fun path(): DescriptorPath = DescriptorPaths.render(this)
 
-   fun parts(): List<String> = when (this) {
+   /**
+    * Returns the test parts as a list of strings, excluding the spec.
+    *
+    * For example, a test with name "my test" inside a context "my context" in a spec called "my spec"
+    * would have the parts ["my context", "my test"].
+    */
+   fun testParts(): List<String> = when (this) {
```
AlexCue987, line 92: should we add `testParts()` to the interface and split this `when` into two simple implementations in both classes?

AlexCue987 (commented):
LGTM, one minor suggestion

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/descriptors/descriptor.kt
sksamuel: yeah I like that.

Issue #5171, opened 2025-10-29
It's not possible to restrict/verify boolean property values in jsonSchema assertions


TWiStErRob:
```kotlin
    @Test fun test() {
        """ { "prop": true } """ shouldMatchSchema jsonSchema {
            obj {
                boolean("prop")
            }
        }
        """ { "prop": true } """ shouldMatchSchema jsonSchema {
            obj {
                boolean("prop") { beEqual(true) } // does not compile
            }
        }
    }
```
The doc for boolean():
https://github.com/kotest/kotest/blob/76fcf698f85e85c614ec44a6efd0a9f17b19d8a3/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/builder.kt#L229-L234

explicitly states it must be either be one of them, which is cool and all, but what if my restriction is stronger than that? It must always exist with a true value?

alfonsoristorato:
Hello, raised this [PR](https://github.com/kotest/kotest/pull/5174) to support this feature. Once it gets merged, it should be in the next release.
You can see some tests in how to use such feature, but in case you can't find it, you'd be able to:
```
        """ { "prop": true } """ shouldMatchSchema jsonSchema {
            obj {
                boolean("prop") { beTrue() }
            }
        }
```

PR #5174, opened 2025-10-30
allow to verify if Boolean property in jsonSchema is true or false (addresses #5171)

alfonsoristorato:
1. Addresses https://github.com/kotest/kotest/issues/5171 by exposing `matcherBuilder: () -> Matcher<Boolean>?` in `JsonSchema.JsonBoolean` to provide some matchers.
2. adds `beTrue()` and `beFalse()` `Matcher<Boolean>` functions 
3. uses new boolean matchers for existing boolean assertions

### DIFF
--- a/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/builder.kt
+++ b/kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/builder.kt
@@ -126,7 +126,8 @@ data class JsonSchema(
       fun boolean(
          name: String,
          optional: Boolean = false,
-      ) = withProperty(name, optional) { boolean() }
+         matcherBuilder: () -> Matcher<Boolean>? = { null }
+      ) = withProperty(name, optional) { boolean(matcherBuilder) }
 
       fun `null`(
          name: String,
@@ -181,7 +182,9 @@ data class JsonSchema(
    }
 
    @Serializable
-   object JsonBoolean : JsonSchemaElement, ValueNode<Boolean> {
+   data class JsonBoolean(
+      override val matcher: Matcher<Boolean>? = null
+   ) : JsonSchemaElement, ValueNode<Boolean> {
       override fun typeName() = "boolean"
    }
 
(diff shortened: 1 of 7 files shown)


kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/builder.kt
```diff
@@ -181,7 +182,9 @@ data class JsonSchema(
    }
 
    @Serializable
-   object JsonBoolean : JsonSchemaElement, ValueNode<Boolean> {
+   data class JsonBoolean(
```
AlexCue987, line 185: I like this new feature. But was it really necessary to add `beTrue` and `beFalse`? They seem to be really trivial and as such they seem to add very little value. WDYT?

alfonsoristorato: Happy to remove them, but as they are the only two possible candidates for a Matcher Boolean, it made more sense to me to provide them to users rather than expect everyone using this to create their own matchers that will be the same for everyone.

AlexCue987 (commented):
I like this improvement. There is just one detail - if `beTrue` and `beFalse` are public, should we add them to the docs? Alternatively, should we make them `internal`? WDYT?

alfonsoristorato:
Ah fair point about the docs, I will try to find the best place for them to be placed at, unless you can point me where already 🙂.
I'd like to keep them public, so anyone can use them.

alfonsoristorato:
@AlexCue987 I had a look at the current docs, and it does not seem we specifically state the matchers equivalents anywhere, as effectively they are named exactly the same as their assertions methods equivalents. 
I.e. [beLessThanOrEqualTo](https://github.com/kotest/kotest/blob/4af10c624be33d39b88d7241765fdffb08758999/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/longs/LongMatchers.kt#L19) is only present in the docs as `shouldBeLessThanOrEqualTo`
and many more. Considering this, do you still think it's worth to add these to the docs?

AlexCue987 (approved):
very nice!

PR #5178, opened 2025-10-31
fix-data-class-eq

AlexCue987:
respect data classes implementation of `equals` - if instances are equal, do not bother matching fields


sksamuel (approved):

AlexCue987:
bypassing because the failure is an unrelated random flaky test.

PR #5182, opened 2025-11-03
make shouldCompile/shouldNotCompile support multiple files

mirror-kt:
Title as is

The current implementation of kotest-assertions-compile can only pass one file to CodeSnippet.
To handle compile testing that requires multiple files, we have added variations of shouldCompile/shouldNotCompile that take multiple files as receivers.


kotest-assertions/kotest-assertions-compiler/src/jvmMain/kotlin/io/kotest/matchers/compilation/compileCodeSnippet.kt
```diff
@@ -192,6 +250,81 @@ fun File.shouldCompile() = codeSnippet(readText()).shouldCompile()
...
+ * If [expectedMessage] is provided, the test additionally verifies that the compilation fails
+ * with an error message containing the specified text. This helps ensure that the compilation
+ * fails for the expected reason, not due to an unrelated error.
+ *
+ * It includes the classpath of the calling process,
+ * so that dependencies available to the calling process are also available to the code snippet.
+ * @see [List.shouldCompile]
+ * */
+@JvmName("shouldNotCompileFiles")
+fun List<File>.shouldNotCompile(expectedMessage: String? = null) = codeSnippet(
```
AlexCue987, line 285: is there a reason we've added new code but no tests?

mirror-kt: I was unsure about what tests to add, so they were not added initially. I have now added some tests, but are there any other test cases I should consider?

AlexCue987:
Task :kotest-assertions:kotest-assertions-compiler:jvmApiCheck FAILED

can you run `apiDump`?

mirror-kt:
@AlexCue987 

> Task :kotest-assertions:kotest-assertions-compiler:jvmApiCheck FAILED
can you run apiDump?

Done.

sksamuel (approved):

PR #5186, opened 2025-11-06
Use deprecated overloads to highlight non-determinstic arbitrary usage

tKe:
This is an attempt to provide warnings when using `.next()`, `.single()`, `.take(n)` and `.drop(n)` within either an ArbitraryBuilder or PropertyContext.

This type of usage will not propagate the `RandomSource` and will instead use `RandomSource.default()` which will result in non-deterministic tests.

NB: It's still possible to call `Arb<A>.samples()` which would still use `RandomSource.default()` without a warning, but I suspect that's rather niche.


sksamuel (approved):
nice!

PR #5192, opened 2025-11-13
use atLeast and atMost for Double (#5191)

AlexCue987:
add synonyms for `shouldBeGreaterThanOrEqualTo` => `shouldBeAtLeast` which is commonly used verbiage in Kotlin, such as `coerceAtLeast` and `coerceAtMost` in Kolin common library

currently only for Double, will be adding for other types later.


alfonsoristorato (commented):
can we maybe make these new methods, or the old methods call one  another? That way  they will be co-dependant and not repeat the body?

Edit: on second thought, my suggestion does not seem like something we normally do, so approving it anyway and letting you decide :)

alfonsoristorato (approved):

sksamuel (approved):

AlexCue987:
> can we maybe make these new methods, or the old methods call one another? That way they will be co-dependant and not repeat the body?

I concur, there is some code duplication to remove. Next PR. Thx!

PR #5194, opened 2025-11-15
add `shouldBeEqualToDifferentTypeIgnoringFields` 

alfonsoristorato:
- Adds a new matcher called `shouldBeEqualToDifferentTypeIgnoringFields` which allows comparison of two different types and enforces `Kproperty` of receiver being used.
- Ensures existing  `shouldBeEqualToIgnoringFields` remains unchanged, but disallows to provide a different type for `other` than the receiver

Follow up of PR https://github.com/kotest/kotest/pull/5124 that introduced `shouldBeEqualToDifferentTypeUsingFields`


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/equality/ReflectionKtTest.kt
```diff
@@ -60,414 +67,418 @@ class ReflectionKtTest : FunSpec() {
    data class EnumWrapper<E : Enum<E>>(val enum: E)
 
    init {
```
alfonsoristorato, line 69: tests here are unchanged - with the exception of adding new tests for new method - I simply moved each test in its own context for clarity

AlexCue987:
presumably the failure is unrelated:
```
SystemPropertyConfigClassTest[jvm] > system property override should be used for config classname[jvm] FAILED
    io.kotest.engine.test.interceptors.TestTimeoutException at InvocationTimeoutInterceptor.kt:38
1 test completed, 1 failed
```
WDYT?

AlexCue987 (approved):

alfonsoristorato:
> presumably the failure is unrelated:
> ```
> SystemPropertyConfigClassTest[jvm] > system property override should be used for config classname[jvm] FAILED
>     io.kotest.engine.test.interceptors.TestTimeoutException at InvocationTimeoutInterceptor.kt:38
> 1 test completed, 1 failed
> ```
> WDYT?

Yeah definitely unrelated, will try re-running and hope for the best 😅. Will also try to see why this happens, it was the same in another pr

Issue #5202, opened 2025-11-19
Data classes that inherit from map or collection will use wrong comparison


walterbrebels:
Seen with kotest 6.0.4

This test fails while it should succeed:
```kotlin
data class NamedMap(
    val name: String,
    val map: Map<String, String>,
) : Map<String, String> by map

val map1 = NamedMap("map1", mapOf("foo" to "bar"))
val map2 = NamedMap("map2", mapOf("foo" to "bar"))
map1 shouldNotBe map2
```

In the output it also uses the `toString()` method from the map implementation instead of the `toString()` from the actual type.

christopheml:
Kotest uses specific classes for testing equality. `EqResolver` selects one based on the types of actual and expected arguments. There are `MapEq` and a `DataClassEq` implementations but the `MapEq` is selected with a higher priority, so a data class implementing `Map` will always be evaluated with `MapEq`.

`MapEq` is both responsible for comparing directly key sets and values from the maps and producing a String representation of the maps. Since it only views objects as `Map` instances, other fields will be ignored.

A fix could be changing priorities so that the data class comparison is picked up higher, but I don't know why it has a low priority in the first place and what the impacts would be.

walterbrebels:
Changing priorities would fix this specific issue. But the comparison logic seems incorrect in general. In the example `NamedMap` could have also been a regular class and the reported issue would still be valid IMO.

AlexCue987:
nice simple repro! I got it

AlexCue987:
Unfortunately I don't see a way to fix this thing without breaking 50 other tests. In general, `shouldBe` is a Swiss Army knife, which does everything, but does nothing perfectly. For better results, we can use more specialized assertions.

walterbrebels:
shouldBe is documented that it tests for equality. Is there another matcher that always ensures equality is verified?

AlexCue987:
> shouldBe is documented that it tests for equality. Is there another matcher that always ensures equality is verified?

`beEqual` does just that, but does not try to display differences in good detail. and `shouldBeEqualUsingFields` is specifically built for data classes, and [does a good job displaying diffs](https://github.com/kotest/kotest-cookbook/blob/master/README.md#matching-data-classes-with-shouldbeequalusingfields)

scottdfedorov:
@walterbrebels Was just looking through issues and noticed this one... Kotest added custom Eq instances, which have essentially the highest priority, so you could register a custom Eq instance and it would get selected first:
[...]

PR #5218, opened 2025-11-27
Alex/fix embedded map  (#5217)

AlexCue987:
fix for https://github.com/kotest/kotest/issues/5202


AlexCue987:
this is not gonna work. maybe I'll try another approach later

PR #5227, opened 2025-11-29
Change assertion from collection to sequence

AlexCue987:
`shouldContainAllInAnyOrder` is only available for sequences


sksamuel:
should we not just add shouldContainAllInAnyOrder for collection too ?

sksamuel:
what do you think @AlexCue987 https://github.com/kotest/kotest/pull/5228

AlexCue987:
> should we not just add shouldContainAllInAnyOrder for collection too ?

sure - but let's make sure we document if it's a synonym for `shouldContainExactlyInAnyOrder` or explain the difference between them.

AlexCue987:
closing as there will be an implementation for collections

PR #5228, opened 2025-11-30
Add `shouldContainAllInAnyOrder` matchers for collections

sksamuel:


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containAllInAnyOrder.kt
```diff
@@ -0,0 +1,37 @@
...
+infix fun <T, C : Collection<T>> C?.shouldContainAllInAnyOrder(expected: C) =
+   this should containAllInAnyOrder(expected)
+
+fun <T, C : Collection<T>> C?.shouldContainAllInAnyOrder(vararg expected: T) =
+   this should containAllInAnyOrder(*expected)
+
+fun <T> containAllInAnyOrder(vararg expected: T): Matcher<Collection<T>?> =
+   containAllInAnyOrder(expected.asList())
+
+/** Assert that a collection contains all the given values and nothing else, in any order. */
```
AlexCue987: is it identical to `shouldContainExactlyInAnyOrder`? if yes, do we need a separate implementation? if no, let's document the difference

sksamuel: containAllInAnyOrder means the collection has all the given elements in any order, but may have more.
shouldContainExactlyInAnyOrder means the collection has all the given elements in any order, but no more.

I'll update docs.

AlexCue987: Besides the docs, we need to update the javadoc here:
"Assert that a collection contains all the given values and nothing else, in any order" to 
"Assert that a collection contains all the given values in any order, but may have more elements"

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/ShouldContainAllInAnyOrderTest.kt
```diff
@@ -0,0 +1,149 @@
...
+            empty.shouldContainAllInAnyOrder(listOf(null, null, null, null))
+         }
+      }
+
+      test("should succeed when elements are same") {
+         val countdown = (10 downTo 0).toList()
+         val countup = (0..10).toList()
+         countdown.shouldContainAllInAnyOrder(countup)
+      }
+
```
AlexCue987, line 31: do we need to add one more test:

```
      test("should succeed when elements are same") {
         val biggerList = (0..15).toList()
         val smallerList = (0..10).toList()
         biggerList.shouldContainAllInAnyOrder(smallerList)
      }
```

sksamuel: done

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/containAllInAnyOrder.kt
sksamuel: done

AlexCue987 (approved):

AlexCue987:
build failing

AlexCue987:
the new test is failing:

ShouldContainAllInAnyOrderTest[jvm] > containAllInAnyOrder[jvm] > should succeed when LHS is a bigger list[jvm] FAILED
    org.opentest4j.AssertionFailedError at ShouldContainAllInAnyOrderTest.kt:35

AlexCue987:
I am working on some changes for this feature

AlexCue987 (commented):
IMO the new test is failing because the code does not do what it is supposed to do. I'm adding some code to address that.

Issue #5167, opened 2025-10-25
junit 6 runner


jgrnrt:
Hello,
I wanted to ask if someone is trying to migrate to junit6?

I did try to make some first changes: https://github.com/jgrnrt/kotest/commit/9d87741afbcb22b5f8cf1ef4112e65707e601124
Maybe someone finds it useful.

Syncing gradle projects does not run with IntelliJ 2025.2.4. 
Run with ./gradlew

Kind regards

sksamuel:
We'll add a junit6 runner in kotest 6.1
Where are you using junit6 currently? Spring?

jgrnrt:
Yes, Spring. JUnit 6 should improve DX, since nullability was added with JSpecify.

sksamuel:
Merged into main and will be in 6.1

sksamuel:
released

PR #5230, opened 2025-11-30
Add specific JUnit 6 runner

sksamuel:
Also migrated shared junit platform logic to a new shared module


sksamuel:
Fixes #5167

alfonsoristorato (approved):
nice

PR #5241, opened 2025-12-01
Makes `beBoolean` matcher public

LeoColman:
- Renames matcher
- Removes private modifier
- Move matcher to the end of the file

### DIFF
--- a/kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
+++ b/kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
@@ -1065,6 +1065,7 @@ public final class io/kotest/matchers/bigdecimal/ToleranceMatcher : io/kotest/ma
 }
 
 public final class io/kotest/matchers/booleans/BooleanMatchersKt {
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
```diff
@@ -1065,6 +1065,7 @@ public final class io/kotest/matchers/bigdecimal/ToleranceMatcher : io/kotest/ma
 }
 
 public final class io/kotest/matchers/booleans/BooleanMatchersKt {
+	public static final fun beBoolean (Z)Lio/kotest/matchers/Matcher;
```
AlexCue987, line 1068: do we need to update `core.md` as well?

alfonsoristorato: we normally do not document `Matcher` in the docs

alfonsoristorato (approved):
approving as the test failure is not related

PR #5258, opened 2025-12-08
Added backwards compatible methods for nondeterministic testing

sksamuel:
Am updating one of my other projects and noticed the 5.x era versions that accept a Long have gone away, so adding back to aid anyone upgrading.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/until.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/until.kt
@@ -7,7 +7,7 @@ import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.seconds
 
 /**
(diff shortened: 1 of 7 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/nondeterministic/until.kt
```diff
@@ -19,6 +19,19 @@ suspend fun until(
...
 }
 
+/**
+ * Runs a function [test] until it returns true, as long as the specified [durationMs] hasn't passed.
+ *
+ * To supply more options to until, use the overload that accepts an [UntilConfiguration].
+ */
+suspend fun until(
+   durationMs: Long,
+   test: suspend () -> Boolean,
```
AlexCue987, line 29: do we need any new/updated tests?

sksamuel: I've made them simple delegates to the original functions and added basic tests.

AlexCue987 (approved):

Issue #5261, opened 2025-12-15
Out of memory errors running DescribeSpec


itsmattking:
**Which version of Kotest are you using** 6.0.7

When running a `DescribeSpec` in the following conditions:

* isolationMode is InstancePerLeaf
* two tests with the same name in the same describe container

```
class IssueReproductionTest : DescribeSpec({
    isolationMode = IsolationMode.InstancePerLeaf
    describe("an issue") {
        it("can be reproduced") {
            assertTrue(1 + 1 == 2)
        }
        it("can be reproduced") {
            assertTrue(2 + 2 == 4)
        }
    }
})
```

The test runs in a loop showing errors like:

```
WARNING: TestExecutionListener [org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestExecutionListener] threw exception for method: executionFinished(TestIdentifier [uniqueId = [engine:kotest]/[spec:com.example.IssueReproductionTest]/[test:an issue]/[test:can be reproduced], parentId = [engine:kotest]/[spec:com.example.IssueReproductionTest]/[test:an issue], displayName = 'can be reproduced', legacyReportingName = 'can be reproduced', source = MethodSource [className = 'com.example.IssueReproductionTest', methodName = 'an issue/can be reproduced', methodParameterTypes = null], tags = [], type = TEST], TestExecutionResult [status = SUCCESSFUL, throwable = null])

```

And eventually an `OutOfMemoryError` exception is thrown.

alfonsoristorato:
The issue has to do with test names, fixing this made me realise that the same could be applied to the old `InstancePerTest`, so that tests with duplicate names would be counted in those instances instead of getting merged together.

sksamuel:
> The issue has to do with test names, fixing this made me realise that the same could be applied to the old `InstancePerTest`, so that tests with duplicate names would be counted in those instances instead of getting merged together.

What was the underlying cause with the test names?

alfonsoristorato:
it wasn't so much with the test names, but with the test IDs that get generated.
The way I understand it (I might be wrong) is that `JUnitPlatformTestExecutionListener` expects test 1 to start and finish, and test 2 to do the same (and so on for however many tests are being run).
If these tests have the same IDs, then the first will start and finish, but the second will start in a finished state directly, which I believe causes the `AssertionError` on their side.

The infinite loop was instead caused by kotest, as it keeps on registering tests with the same matching name (in instance per leaf), and once the second test was being registered it went back to the first.

PR #5274, opened 2025-12-22
handle duplicate test names in `InstancePerLeaf` and `InstancePerTest` (issue #5261)

alfonsoristorato:
Addresses https://github.com/kotest/kotest/issues/5261 but as I addresses it I noticed that `InstancePerTest` would merge tests with the same name, so added a de-duplicator there too


sksamuel (approved):

sksamuel:
nice

PR #5286, opened 2025-12-28
Support custom eq registration

### LINKED ISSUE #5096: Add Custom Eqs for data classes and types
Would love to be able to register globally a custom comparator for determining if two things are the same for the purposes of tests. The Eq resolver selects from a couple of classes, and adding the ability to register an additional entry in the resolver to override default behavior would be awesome.
Also, in the resolver, if an item is a data class, it uses the native `equals` method to determine if two data classes are the same, but would ideally like it to use the Eq resolver on each field as well so that we can have more control over matching without having to define custom matching by fields. 

As a concrete example, when processing Instants, the resolution is nanoseconds, but many databases (postgres) don't really support nanosecond resolution meaning two data classes with Instants are almost never matching. 
My goal would be to register a Instant Eq comparator and then have all cases check for instant equality as milliseconds. 

```kotlin
data class Wibble(
  val id: String,
  val createdAt: Instant
)
```

``` kotlin
val a = Wibble("abc", Instant.now())
val b = datastore.getExisting("abc")

a shouldBe b
```
This is almost guaranteed to fail.  The issue will be the mismatch in createdAt timestamps in the nanoseconds. 
```
org.opentest4j.AssertionFailedError: data class diff for app.getpie.plans.database.tables.User
└ createdAt: expected:<2025-09-16T19:32:21.306762Z> but was:<2025-09-16T19:32:21.306Z>
```

sksamuel:
Fixes #5096


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/DefaultEqResolver.kt
```diff
@@ -5,17 +5,31 @@ import io.kotest.assertions.AssertionsConfig
...
+ * Alternatively, the [EqResolver] can be switched to an entire custom implementation.
+ *
  * See [EqCompare].
  */
+interface EqResolver {
+
+   /**
+    * Returns the [Eq] to use for comparison for the given values.
+    */
+   fun resolve(actual: Any?, expected: Any?): Eq<out Any?>
```
AlexCue987: should we make it nullable, so that when it returns a null, we fall back on the default?

suppose we only want to override a matcher for LocalDateTime, and we are totally happy with all other decisions done by `DefaultEqResolver`. it would be super easy to do this:

```
when {
  actual is LocalDateTime && expected is LocalDateTIme -> ...
  else -> null
}
```

sksamuel: This is just early days as I play about, will open it up from draft for a PR when it's ready

sksamuel: That's an interesting point though, I'll keep that in mind as I work on the idea.

sksamuel:
@AlexCue987 ready for your eyes

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/DefaultEqResolver.kt
```diff
@@ -1,29 +1,25 @@
...
-/**
- * An [DefaultEqResolver] is used by default in Kotest unless a user specifies a custom implementation.
+ * This resolver allows registration of custom [Eq] implementations for specific types, which
+ * this resolver will then delegate to if both the LHS and RHS have the same type, and there is a custom
+ * [Eq] instance available for that type.
  */
 @Suppress("DEPRECATION")
 object DefaultEqResolver {
 
+   private val customEqInstances = mutableMapOf<KClass<*>, Eq<*>>()
```
AlexCue987, line 17: is thread safety a concern here? because `customEqInstances` is both shared and mutable

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/CustomEqTest.kt
```diff
@@ -0,0 +1,41 @@
...
+import io.kotest.assertions.eq.Eq
+import io.kotest.assertions.throwables.shouldThrowAny
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.matchers.shouldBe
+
+class CustomEqTest : FunSpec() {
+   init {
+
+      beforeSpec {
+         DefaultEqResolver.register(Foo::class, fooEq)
```
AlexCue987: once we've registered it, it remains there as a side effect, potentially breaking other tests. should we use an immutable copy for this test only? should we think about scopes affected by `register`?

sksamuel: In my mind the idea is you would register it globally for all tests otherwise creating a custom eq is overkill. What do you think?

AlexCue987: Is there a way to register all overrides just once, before any tests are run, and make sure it's immutable while tests are running? otherwise the outcome of tests may depend on the order they run, and I'd rather avoid it.

sksamuel: I was thinking you'd setup all your custom Eqs in project config (or your testing framework equivalent) so it's then set for the entirely of the test suite. Any ideas on what you'd like to see here?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/DefaultEqResolver.kt
sksamuel: I think this depends on if you are setting them up in advance or not, see other comment pending your feedback.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/CustomEqTest.kt
AlexCue987: > I was thinking you'd setup all your custom Eqs in project config (or your testing framework equivalent) so it's then set for the entirely of the test suite. Any ideas on what you'd like to see here?

it would be nice to have an example of this setup in the docs

sksamuel: docs added

PR #5288, opened 2025-12-29
Add line endings to unix option in assertions.

### LINKED ISSUE #3161: Add option to ignore differences in line endings during string comparison, so tests are not OS dependent

sksamuel:
Fixes #3161

Introduces StringPreprocessor's so we can add more string manipulation in the future if needed.
Currently only covering StringEq and startsWith / endsWith, for simplicity.


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/AssertionsConfig.kt
```diff
@@ -80,6 +80,10 @@ class EnvironmentConfigValue<T>(
       }
    }
 
+   /**
+    * Executes the given [thunk] with the value of this [io.kotest.assertions.EnvironmentConfigValue] set
```
AlexCue987: lets change `thunk` to `block` to be consistent

AlexCue987 (approved):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/AssertionsConfig.kt
sksamuel: done

PR #5302, opened 2026-01-01
Allow global configuration of default range of arrays and collections in prop tests

### LINKED ISSUE #4989: Property - Allow global configuration of default range of arrays and collections
### Currently

companion arbs for classes that can contain multiple elements currently default to either `range= 1..100` `maxSize=100`. 

Examples:
[Arb.array](https://github.com/kotest/kotest/blob/6f740ecd8632900c735e3dc42150bae6c0b05939/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/array.kt#L18)
[Arb.set](https://github.com/kotest/kotest/blob/6f740ecd8632900c735e3dc42150bae6c0b05939/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/collections.kt#L60)

This is usually a fine default, but in certain situations, a user might want to have wildly different sizes of collections or arrays in a test, which can come up often when using reflective arbs

### Workaround

The current workaround would be to bind a new arb for each of the target properties, but this can be a very verbose approach when you are working with nested objects or need to set multiple bindings. Let's say in my case, I want to default to 1000 elements in several different properties.

With binding, the approach would be something like:
```kotlin
checkAll( 
    Arb.bind<ExampleClass> {
    	bind(NestedExample::recordedValues to Arb.array(Arb.int(), 1..1000))
    	bind(ExampleClass::value2 to Arb.array<AnotherExample>(Arb.bind<AnotherExample>{  }, 1..1000))
    	bind(ExampleClass2::properties to Arb.array<String>(Arb.string(), 1..1000))
       ....
    }){
// my test
}
``` 

Where I have to go through the model and identify the arrays, then map them one by one in the binding. This approach is verbose, and requires reviewing the whole model, which is one of the things I want to avoid when I'm using reflective arbs

### Proposal

Compare the previous example with a theoretical configuration property:

```kotlin
forAll<ExampleClass>(PropTestConfig(maxElements= 1000) { ExampleClass->
         // my test
      }
```

From what I understand from looking at the code, this config property could easily default to the current 1-100 range, and then be used in the defaults of companions like [Arb.companion.array](https://github.com/kotest/kotest/blob/6f740ecd8632900c735e3dc42150bae6c0b05939/kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/array.kt#L18)

This could also help reduce the time to generate large objects in cases where they dont need as many elements per array/collection, as reducing this range globally could also drastically reduce the time to generate in some cases

### Other notes

There might be a better workaround that I am unaware of, since I do not have a deep understanding of Kotlin's type system, but I have not been able to find any way to do a binding with parameter type association like:

```kotlin
checkAll( 
    Arb.bind<ExampleClass> {
    	bind(Array<T>::class to Arb.array(GlobalArbResolver.resolve(T), 1..1000))
    }){
// my test
}
```
Obviously this wont compile as is, but I think the idea is clear
If this is sort of binding actually doable in Kotlin, in a way where we dont override the Array type parameter and generate only `Array<Any>` filled with generic `Object` , then this proposal is unnecessary, although it could still improve ease of use

sksamuel:
Fixes #4989


kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
```diff
@@ -73,6 +73,8 @@ object PropertyTesting {
 
    var defaultOutputHexForUnprintableChars: Boolean =
       sysprop("kotest.proptest.arb.string.output-hex-for-unprintable-chars", false)
+
+   var defaultCollectionsRange: IntRange = 0..100
```
alfonsoristorato, line 77: does this deserve a `sysprop` setter? DYT?

alfonsoristorato (commented):
happy to approve, not approving in case you want to address the `sysprop` and we have automerge

kotest-property/src/commonMain/kotlin/io/kotest/property/config.kt
sksamuel: I didn't know what to do really, would mean parsing it out, and we don't use sysprops for everything in this file so figured this was good enough to address the issue.

alfonsoristorato: fair fair

alfonsoristorato (approved):

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
```diff
@@ -0,0 +1,76 @@
...
+      val genIter = gen.generate(it).iterator()
+      val targetSize = it.random.nextInt(range)
+      val set = mutableSetOf<A>()
+      var iterations = 0
+      val maxMisses = targetSize * slippage
+      // We may generate duplicates, but we don't know if the underlying gen has sufficient cardinality
+      // to satisfy our range, so we can try for a while, but must not try forever.
+      // The slippage factor controls how many times we will accept a non unique element before giving up,
+      // which is the number of elements in the target set * slippage
+      while (iterations < maxMisses && set.size < targetSize && genIter.hasNext()) {
```
AlexCue987, line 55: I'd like to add some explicit tests for this `while` loop, after extracting it into a function. Later.

AlexCue987 (approved):

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
sksamuel: I agree. All I did here was move it out of a giant file into its own file.

PR #5303, opened 2026-01-01
Support Java21 SequencedSet in shouldContainsExactly

### LINKED ISSUE #3996: UnmodifiableSequencedSet not supported

sksamuel:
Fixes #3996

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/assertions/eq/isOrderedSet.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/assertions/eq/isOrderedSet.kt
@@ -5,7 +5,13 @@ import java.util.SortedSet
 /**
  * JVM-specific: Allows TreeSet to be considered ordered for iterable comparisons
  */
-actual fun isOrderedSet(item: Iterable<*>) =
-   item is LinkedHashSet ||
+actual fun isOrderedSet(item: Iterable<*>): Boolean {
+   return item is LinkedHashSet ||
       item is SortedSet ||
-      (item is Set && item.size <= 1)
+      (item is Set && item.size <= 1) ||
+      isJavaSequencedSet(item::class.java)
+}
+
+private fun isJavaSequencedSet(clazz: Class<out Iterable<*>>): Boolean {
+   return clazz.name == $$"java.util.Collections$UnmodifiableSequencedSet"
+}
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/eq/IsOrderedSetTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/eq/IsOrderedSetTest.kt
@@ -1,0 +1,22 @@
+package com.sksamuel.kotest.assertions.eq
+
+import io.kotest.assertions.eq.isOrderedSet
+import io.kotest.core.spec.style.FunSpec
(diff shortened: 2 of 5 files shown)


kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/assertions/eq/isOrderedSet.kt
```diff
@@ -5,7 +5,13 @@ import java.util.SortedSet
...
+actual fun isOrderedSet(item: Iterable<*>): Boolean {
+   return item is LinkedHashSet ||
       item is SortedSet ||
-      (item is Set && item.size <= 1)
+      (item is Set && item.size <= 1) ||
+      isJavaSequencedSet(item::class.java)
+}
+
+private fun isJavaSequencedSet(clazz: Class<out Iterable<*>>): Boolean {
+   return clazz.name == $$"java.util.Collections$UnmodifiableSequencedSet"
```
AlexCue987, line 16: is it more conventional to escape `$` via a raw string: 
`return clazz.name == """java.util.Collections$UnmodifiableSequencedSet"""`

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/assertions/eq/isOrderedSet.kt
```diff
@@ -5,7 +5,13 @@ import java.util.SortedSet
 /**
  * JVM-specific: Allows TreeSet to be considered ordered for iterable comparisons
  */
-actual fun isOrderedSet(item: Iterable<*>) =
-   item is LinkedHashSet ||
+actual fun isOrderedSet(item: Iterable<*>): Boolean {
```
AlexCue987, line 8: should `isOrderedSet` have it's own unit tests?

AlexCue987 (approved):

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/assertions/eq/isOrderedSet.kt
sksamuel: That won't work as $ still works in """
So you'd need to escape it either way.

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/assertions/eq/isOrderedSet.kt
sksamuel: done

PR #5305, opened 2026-01-01
Update docs and more tests for testcontainers v2

### LINKED ISSUE #3808: Lifecycle documentation for test containers needs updating
Currently the documentation states:

> Lifecycle
By default, the lifecycle of a container is per spec - so it will be started at the install command, and shutdown as the spec is completed. This can be changed to start/stop per test, per leaf test, or per root test.

However, this is not true for many cases. The default lifecycle for both `ContainerExtension` and the `JdbcDatabaseContainerExtension` is `ContainerLifecycleMode.Project`. Additionally, there are no modes for "per test" or "per leaf test".

Our solution right now is to set kotest isolation mode to `InstancePerTest` and the container lifecycle mode to `Spec`. This seems to work fine to avoid the same database being concurrently modified by multiple tests.

sksamuel:
Closes #3808

Removed obsolete tests, added moretests for the 6.1.0 support.


kotest-extensions/kotest-extensions-testcontainers/src/jvmTest/kotlin/io/kotest/extensions/testcontainers/TestContainerProjectExtensionTest.kt
```diff
@@ -0,0 +1,50 @@
...
+
+private val container = GenericContainer("redis:5.0.3-alpine").apply {
+   startupAttempts = 2
+   withExposedPorts(6379)
+}
+
+private val extension = TestContainerProjectExtension(container)
+
+@EnabledIf(LinuxOnlyGithubCondition::class)
+@Order(1)
```
alfonsoristorato, line 20: curious as to what was the issue without Order?

alfonsoristorato (approved):

kotest-extensions/kotest-extensions-testcontainers/src/jvmTest/kotlin/io/kotest/extensions/testcontainers/TestContainerProjectExtensionTest.kt
sksamuel: the Order(1) sets up the redis, and Order(2) confirms it's still there.
If the test container wasn't shared between 1 and 2, then 2 would find null not the  pre-existing value.

PR #5309, opened 2026-01-02
Rework Eq to generate errors lazily

### LINKED ISSUE #5036: `obj should(Not)Be null` and `obj should(Not) be(null)` alters the state of `obj`
6.0.0 (but it's also present in other versions)

# Problem

I had a Java test that used `assertNotNull(iterator)`, and Junie (and some time ago also I) ported such calls to `iterator shouldNotBe null`, which led to failing (incorrectly ported) tests. After digging deeper, it turned out that it wants to compare values of `iterator` and `null`, and correct way sof porting `assertNotNull(iterator)` are `iterator.shouldNotBeNull()` or `iterator shouldNot beNull()`.

Someone not knowing this distinction may write the non-null assertion in 4 ways:

* `obj shouldNotBe null`
* `obj shouldNot be(null)`
* `obj shouldNot beNull()`
* `obj.shouldNotBeNull()`

The two first ones aren't null-aware, and treat `null` as any other value, which results in altering the object when e.g. it's an interator, and it's casted to a list.

I'm not sure if it's higher-level design problem, or just an edge case that wasn't covered in `should(Not)Be` and `should(Not)`. To me, it feels like there should be another edge case explicitly covered for `null` here:

https://github.com/kotest/kotest/blob/09094d69bb4ca62d27cbc44c7430e1e4e008a0d8/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/should.kt#L56-L59

and in other similar matchers, since `null` is a special value, and it would be nice to make all 4 calls shown above equivalent.

# Reproducer

```kotlin
import io.kotest.matchers.be
import io.kotest.matchers.nulls.beNull
import io.kotest.matchers.nulls.shouldNotBeNull
import io.kotest.matchers.shouldNot

class MyStatefulObject : Iterable<Int> {
    var state: Int? = 5
    override fun iterator(): Iterator<Int> = iterator {
        // The goal is to make the iterator return just one `5` and then get exhausted.
        if (state != null) {
            yield(state!!)
            state = null
        }
    }
}

class ReproductionTest : FunSpec({
    test("reproduction") {
        val obj = MyStatefulObject()

        // Uncomment one of the below to make the test fail
        // since the iterator is consumed, and after this assertion
        // there are no items to consume

        // obj shouldNotBe null
        // obj shouldNot be(null)

        // Uncomment one of the below to make the test pass
        // since the iterator isn't consumed
        obj.shouldNotBeNull()
        //obj shouldNot beNull()

        val iterator = obj as Iterable<*>
        iterator.iterator().hasNext() shouldBe true
    }
})
```

### LINKED ISSUE #5245: 100x Performance Bug with inverted matchers
**Which version of Kotest are you using** 6.0.4

Inverted matchers take significantly longer to execute than their normal counterparts, up to 100x longer for large sample sizes.

Here is a minimal example:

**Scenario A (shouldNotBe 0)**
Time taken: 10000ms (10s)
```kotlin
        (1..1_000_000).forEach {
          (it % 999_999) shouldNotBe 0
        }
```

**Scenario B (!=0 shouldBe true)**
Time taken: 150ms
```kotlin
        (1..1_000_000).forEach {
          ((it % 999_999) != 0) shouldBe true
        }
```

Note: The issue seems to be a similar problem to a performance issue I reported a year ago ( https://github.com/kotest/kotest/issues/4041 ), where "shouldNotBe" specifically was ~25x slower than "shouldNotBeIn", which was narrowed down to being ~4x slower after the fixes were applied.

sksamuel:
Please take a look @AlexCue987 

The `Eq` signature has been changed to return an `EqResult` (I may make this a sealed interface rather than a data class?). Callers have been updated to handle the strongly typed reply. Fixes perf issues and the shouldNotBe mutation issue.
`ComparisonMatcherResult` which eagerly generated actual and expected values has been deprecated in favor of `DiffableMatcherResult` which is lazy.

Fixes #5036
Fixes #5245


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/eq/NullShouldBeTest.kt
```diff
@@ -0,0 +1,37 @@
...
+      if (state != null) {
+         yield(state!!)
+         state = null
+      }
+   }
+}
+
+@Issue("https://github.com/kotest/kotest/issues/5036")
+class NullShouldBeTest : FunSpec({
+   test("shouldBe null and shouldNotBe null should handle nulls without alterating state") {
```
AlexCue987: nit: "alterating" => "altering"

AlexCue987 (approved):
I could not think of a shorter way to fix that issue. Glad this massive improvement is done!

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/eq/NullShouldBeTest.kt
sksamuel: fixed

sksamuel:
@AlexCue987 I saw your comment about removing comments but can't find it to reply. I've removed them.

PR #5312, opened 2026-01-02
use ClassSource for CONTAINER type tests

### LINKED ISSUE #5034: Kotest 6.0.0 and above breaks DescribeSpec nesting in Android Studio
When I have test suite like:

```kotlin
class SimpleTest: DescribeSpec({
    describe("simple test") {
        it("should pass") {
            true shouldBe true
        }
    }
    
    describe("another simple test") {
        it("should also pass") {
            (1 + 1) shouldBe 2
        }
    }
})
```

The hierarchy is fine in v5.9.1 of Kotest like so:

Image

When using Kotest 6.0.0 or 6.0.1, it shows up like:

Image

I have tried different names and strings, but whatever they may be, they all get grouped into 1 describe which doesn't make the tests look right. 

Any ideas about this?

### LINKED ISSUE #5160: Inconsistent "afterSpec" & "when" call order when using BehaviorSpec
With **Kotest 6.0.4**

Hi, passing from 5.9 to 6.0, I wanted to try a bit `InstancePerRoot` behavior with the different before/after methods and the BehaviorSpec that I'm using in my project.

The thing is there is an issue with the execution order of the line of codes, not linked to InstancePerRoot (see all the tests hereunder), because my prints show a weird behavior in line execution.

Here is the test class, with InstancePerRoot and base tests:
```kotlin
internal class MyClassTests : BehaviorSpec({
    isolationMode = IsolationMode.InstancePerRoot

    beforeSpec {
        println("beforeSpec")
    }
    afterSpec {
        println("afterSpec")
    }


    context("a broomstick should be able to be fly and come back on it's own") {
        println("context1")
        given("a broomstick") {
            println("given1")
            and("and") {
                println("and1")
                `when`("I sit on it") {
                    println("when1")
                    then("I should be able to fly") {
                        println("then1")
                        // test code
                    }
                }
                `when`("I don't sit on it") {
                    println("when2")
                    then("I should be able to fly") {
                        println("then2")
                        // test code
                    }
                }
            }
        }
    }
})
```
That gives the `then1` to be run before the `when1`:
```
beforeSpec
context1
given1
and1
when1
then1
then2
when2 // SHOULD BE BEFORE THE then2
afterSpec
```
and also the displayed tests are not what I wrote, since both `then` are under the same `when`, so basically my second `when` just disappeared:

Image

Alternative test - With `SingleInstance`, it gives the same result:
```
beforeSpec
context1
given1
and1
when1
then1
then2
when2 // SHOULD BE BEFORE THE then2
afterSpec
```
When adding an `and` clause after the first `when` then the order is **correct**
```
beforeSpec
context1
given1
and1
when1
and1.2
then1
when2
then2
afterSpec
```

Image

But, when adding an `and` clause after the second `when` (removing the first, because with the first it's working) then the order is **not** correct
```
beforeSpec
context1
given1
and1
when1
then1
and1
then2
when2 // SHOULD BE BEFORE THE then2
afterSpec
```
I read the doc a lot but maybe I still missd something?

I don't know if it's linked, but I have inconsistent behavior as well with `afterSpec` not being called at the end of the test, before the second `beforeSpec`:
```
beforeSpec
context1
given1
when1
then1
// HERE IS EXPECTED THE afterSpec
beforeSpec
context2
given2
when2
then2
afterSpec
beforeSpec
context3
given3
when3
then3
afterSpec
afterSpec // BUT HERE IT IS INSTEAD
```

I don't really understand what I'm doing wrong, those seem like bugs to me, and testing previous fix versions of the 6.0, like 6.0.3, didn't go well either.

alfonsoristorato:
Tries to address:
Fixes #5034
Fixes #5160
https://github.com/kotest/kotest/issues/5323


kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/descriptors.kt
```diff
@@ -75,10 +75,15 @@ internal fun createTestDescriptorWithMethodSource(
...
+      // For CONTAINER types, use ClassSource (like v5.9.1) to ensure proper tree structure in Android Studio.
+      // Android Studio does not display MethodSource containers correctly, hence using ClassSource for them.
       // gradle-junit-platform hides tests if we don't send a source at all
       // surefire-junit-platform (maven) needs a MethodSource in order to separate test cases from each other
       // and produce more correct XML report with test case name.
-      source = getMethodSource(testCase.spec::class, id),
+      source = when (type) {
+         TestDescriptor.Type.CONTAINER -> ClassSource.from(testCase.spec::class.java)
+         else -> getMethodSource(testCase.spec::class, id)
+      },
```
alfonsoristorato, line 86: this was a lot of trial and error and noticed this was changed from 5.x.x onwards. Built project locally and started both in IntelliJ and Android Studio and it seems to work - I would be very interested to understand more, so am reading into that, but while I read and understand, if anyone has early feedback it is appreciated

sksamuel: Is it only for containers ?

sksamuel: Can we add a test for createTestDescriptorWithMethodSource so that future PRs don't flip it back again without realizing

sksamuel (approved):
nice find :)

kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/descriptors.kt
alfonsoristorato: yes added -> [add tests](https://github.com/kotest/kotest/pull/5312/commits/b18a622066f4ea02a37302070ca227a6aa9f3528)

alfonsoristorato: > Is it only for containers ?

the issue was presenting in containers (also with `withData` as per the issue in ij-plugin), so I think for now we handle it this way until we get more reports (if any)

PR #5557, opened 2026-01-07
Improve Boolean matchers documentation

LeoColman:
- Add detailed examples for `beTrue`, `beFalse`, and `beBoolean`
- Clarify usage scenarios with `should`/`shouldNot`

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/booleans/BooleanMatchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/booleans/BooleanMatchers.kt
@@ -105,15 +105,69 @@ fun Boolean?.shouldNotBeFalse(): Boolean? {
 }
 
 /**
- * Match that verifies a given [Boolean] is `true`.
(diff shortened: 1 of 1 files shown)


sksamuel (approved):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/booleans/BooleanMatchers.kt
```diff
@@ -105,15 +105,69 @@ fun Boolean?.shouldNotBeFalse(): Boolean? {
...
+ * Creates a matcher that verifies a [Boolean] is equal to [expected].
+ *
+ * Use this when you need to pass the expected boolean value dynamically; for
+ * convenience, prefer [beTrue] and [beFalse] for the common cases.
+ *
+ * ```
+ * val expected = computeFlag()
+ * flag.should(beBoolean(expected))
+ *
+ * beBoolean(true)  // same as beTrue()
```
AlexCue987, line 161: should we provide a more detailed example, such as:

`myVar should beBoolean(true)`

PR #5559, opened 2026-01-08
extract the logic that yields sets of required length to a separate funciton, and test all its edge cases explicitly

AlexCue987:
extract the logic that yields sets of required length to a separate funciton, and test all its edge cases explicitly


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
```diff
@@ -45,25 +45,30 @@ fun <A> Arb.Companion.set(gen: Gen<A>, range: IntRange = PropertyTesting.default
    return arbitrary(SetShrinker(range)) {
       val genIter = gen.generate(it).iterator()
       val targetSize = it.random.nextInt(range)
-      val set = mutableSetOf<A>()
       var iterations = 0
```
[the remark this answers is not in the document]
AlexCue987: done

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
```diff
@@ -45,25 +45,30 @@ fun <A> Arb.Companion.set(gen: Gen<A>, range: IntRange = PropertyTesting.default
...
       // The slippage factor controls how many times we will accept a non unique element before giving up,
       // which is the number of elements in the target set * slippage
-      while (iterations < maxMisses && set.size < targetSize && genIter.hasNext()) {
-         val size = set.size
-         set.add(genIter.next().value)
-         if (set.size == size) iterations++
-      }
+      val set = takeSet(elements, targetSize, maxMisses)
       check(set.size == targetSize) {
          "the target size requirement of $targetSize could not be satisfied after $iterations consecutive samples"
```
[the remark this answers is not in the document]
AlexCue987: fixed

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
```diff
@@ -45,25 +45,30 @@ fun <A> Arb.Companion.set(gen: Gen<A>, range: IntRange = PropertyTesting.default
...
    }
 }
 
+internal fun<A> takeSet(elements: Sequence<A>, targetSize: Int, maxMisses: Int): Set<A> {
+   var elementsTaken = 0
+   val chunk = elements
+      .takeWhile { elementsTaken++ < maxMisses }
+      .distinct()
+      .take(targetSize)
+   return chunk.toSet()
```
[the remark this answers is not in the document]
AlexCue987: replacing idiomatic Kotlin with a while loop? no thanks

kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/SetsTest.kt
```diff
@@ -0,0 +1,41 @@
...
+class SetsTest: StringSpec() {
+   init {
+      "generate sets of required length" {
+         checkAll(1_000, Arb.set(
+            Arb.double(),
+            range = 3..4,
+            slippage = 10,
+            )) {
+            val actual = it
+            actual.size.shouldBeIn(3, 4)
```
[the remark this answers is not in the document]
AlexCue987: done

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
```diff
@@ -45,25 +45,29 @@ fun <A> Arb.Companion.set(gen: Gen<A>, range: IntRange = PropertyTesting.default
...
+      val set = takeSet(elements, targetSize, maxAttempts)
       check(set.size == targetSize) {
-         "the target size requirement of $targetSize could not be satisfied after $iterations consecutive samples"
+         "the target size requirement of $targetSize could not be satisfied"
       }
       set
    }
 }
 
+internal fun<A> takeSet(elements: Sequence<A>, targetSize: Int, maxAttempts: Int): Set<A> {
```
alfonsoristorato, line 62: is this internal for testing purposes only? asking cause it will be exposed in the public api

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
```diff
@@ -45,25 +45,29 @@ fun <A> Arb.Companion.set(gen: Gen<A>, range: IntRange = PropertyTesting.default
...
+   var elementsTaken = 0
+   val chunk = elements
+      .takeWhile { elementsTaken++ < maxAttempts }
+      .distinct()
```
alfonsoristorato, line 66: before we were taking care of dupes in the loop, while now we rely on `distinct` - do you think it will cause memory problems for users that generate big enough data?

alfonsoristorato (commented):
little comments / questions

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
AlexCue987: > is this internal for testing purposes only? asking cause it will be exposed in the public api

it's internal because I did not intend to expose it. why should I?

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
AlexCue987: > before we were taking care of dupes in the loop, while now we rely on `distinct` - do you think it will cause memory problems for users that generate big enough data?

why write bespoke code when `distinct` already does it for you, is well-tested, etc

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
alfonsoristorato: I assumed that internal functions go into the public API drop, I was suggesting at making it private. Unless you need it callable for some other reason or some other classes

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
alfonsoristorato: No reason, I just wondered if this was going to be less memory efficient, cause the way I understand this is that the distinct will create another loop to do the check. But happy either way

alfonsoristorato (approved):

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
AlexCue987: > I assumed that internal functions go into the public API drop, I was suggesting at making it private. Unless you need it callable for some other reason or some other classes

if it's private, I cannot explicilty unit test it, right? if it's internal, I can, and it won't generate into API dump. Please correct me if I am wrong.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
AlexCue987: > No reason, I just wondered if this was going to be less memory efficient, cause the way I understand this is that the distinct will create another loop to do the check. But happy either way

I concur that `distinct` builds another set and then throws it away, and as such it uses more memory. But I was more concerned about maintainability, and I thought that later on code built from standard well-known blocks would be easier to grok than the bespoke while loop.

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/sets.kt
alfonsoristorato: I assumed internal functions get exposed in the API, but clearly they don't, all good :)

Issue #5267, opened 2025-12-19
StackOverflowError in Print system when reporting failure on cyclic structures


jongheon1:
**Which version of Kotest are you using**
6.0.7

## Problem
While implementing cycle detection for equality checking (#5196), I discovered that the `print()` system also suffers from infinite recursion when generating error messages for cyclic structures.

## Current Behavior
When comparing two cyclic structures that are not equal (e.g., different sizes), the equality check identifies the difference. However, when building the failure message, `actual.print()` or `expected.print()` is called, leading to a `StackOverflowError`.

## Steps to Reproduce
```kotlin
val list1 = mutableListOf<Any?>()
list1.add(list1)
list1.add("extra")

val list2 = mutableListOf<Any?>()
list2.add(list2)

// Logic handles the cycle, but Print crashes during error message generation
list1 shouldBe list2
```

alfonsoristorato:
I wanted to have a go at trying to fix this, but I am not able to get the `StackOverflowError`.
Were you doing anything specific to get it to crash?

<img width="631" height="148" alt="Image" src="https://github.com/user-attachments/assets/0fa6d618-3143-4e31-adc9-bc2625ed4f8c" />

jongheon1:
> I wanted to have a go at trying to fix this, but I am not able to get the `StackOverflowError`. Were you doing anything specific to get it to crash?
> 
[...]

PR #5564, opened 2026-01-09
add handler of indirect cycles when calling print and simplifactions of print

alfonsoristorato:
For https://github.com/kotest/kotest/issues/5267
Delegates all printing logic to `recursiveRepr`, checking for ref equality and holding a context of visited nodes to avoid `StackOverFlowError`s

the print context maintains thread safety in jvm.


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/MatchersTest.kt
```diff
@@ -147,7 +147,7 @@ class MatchersTest : FreeSpec({
             l.add(l)
             l shouldBe emptyList()
          }.message shouldBe """Unexpected elements from index 0
-                                                              |expected:<[]> but was:<[[(this Collection)]]>""".trimMargin()
+                                                              |expected:<[]> but was:<[[(this ArrayList)]]>""".trimMargin()
```
alfonsoristorato: pretty sure this was a bug as `l` is an `ArrayList` not a `Collection`

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/print/ListPrint.kt
```diff
@@ -21,18 +21,15 @@ class ListPrint<T>(
...
+         Printed(
             a.joinToString(
             separator = ", ",
             prefix = "[",
             postfix = "]",
             limit = limit,
             truncated = "...and $remainingItems more$limitHint"
          ) {
-            when {
-               it is Iterable<*> && it.toList() == a && a.size == 1 -> a[0].toString()
```
alfonsoristorato, line 33: I don't think this was giving us much value, and should be covered in the new `recursiveRepr`

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/MatchersTest.kt
```diff
@@ -147,7 +147,7 @@ class MatchersTest : FreeSpec({
             l.add(l)
             l shouldBe emptyList()
          }.message shouldBe """Unexpected elements from index 0
-                                                              |expected:<[]> but was:<[[(this Collection)]]>""".trimMargin()
+                                                              |expected:<[]> but was:<[(this ArrayList)]>""".trimMargin()
```
alfonsoristorato, line 150: also, we now do not re-construct the array structure (before `[[(whatever)]]` now `[(whatever)]`)

kotest-assertions/kotest-assertions-core/src/nonjvmMain/kotlin/io/kotest/assertions/print/PrintContext.kt
```diff
@@ -0,0 +1,20 @@
+package io.kotest.assertions.print
+
+/**
+ * Non-JVM implementation of [PrintContext] using a simple [MutableList].
+ * This is safe because JS is single-threaded and Native has its own memory model.
```
alfonsoristorato, line 5: I knew about JS, but I had to ask AI for Native - and it says `Native has its own memory model` - I am going to fact check it on it and learn about it, but iirc in non jvm we always launch the tests in sequence?

sksamuel: No they still launch concurrently

alfonsoristorato: ok fair, not having much (any) experience with native an JS - do you reckon that such a test would be possible?
https://github.com/kotest/kotest/blob/eb704e4f7aa53b878f4e76e7335185b5b6a10e71/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/assertions/print/AsyncPrintTest.kt

If not, I wonder if we should leave as is, if instead such test can indeed be replicated in non-jvm then I will try and find a solution to this.
Bearing in mind that I had to go out of Kotest default way of handling concurrent tests to make this test break with the same impl of non-jvm in jvm.

For example, the same test linked simply annotated with `testExecutionMode = TestExecutionMode.Concurrent` and `InstancePerRoot` and without the coroutines stuff (`withContext` and `async`) would not fail.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/MatchersTest.kt
sksamuel: an ArrayList is a collection via the List interface

kotest-assertions/kotest-assertions-core/src/nonjvmMain/kotlin/io/kotest/assertions/print/PrintContext.kt
sksamuel: They launch concurrently, but unless the dispatcher/platform has threading support then they won't actually run in parallel. So on JS it all ends up being sequential, not sure what happens on native.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/MatchersTest.kt
alfonsoristorato: Yes, but the simpleClassName we use to print should have picked that this was an array list, not sure why it was picking Collection before

sksamuel: Understood

sksamuel (approved):

PR #5565, opened 2026-01-12
Refactor Kotest IntelliJ plugin to run Gradle-based individual tests for all targets

sksamuel:
Last piece we need to release 6.1

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/filter/IncludePatternEnvDescriptorFilter.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/filter/IncludePatternEnvDescriptorFilter.kt
@@ -10,15 +10,15 @@ import io.kotest.engine.extensions.TestPatternParser
 /**
  * An implementation of [DescriptorFilter] that uses an include pattern from an env variable.
  */
-internal object IncludeTestPatternDescriptorFilter : DescriptorFilter {
+object IncludePatternEnvDescriptorFilter : DescriptorFilter {
 
    const val ENV_NAME = "KOTEST_INCLUDE_PATTERN"
 
-   private val logger = Logger(IncludeTestPatternDescriptorFilter::class)
+   private val logger = Logger(IncludePatternEnvDescriptorFilter::class)
 
    override fun filter(descriptor: Descriptor): DescriptorFilterResult {
       val env = env(ENV_NAME)
-      // if no include pattern then we include everything by default
+      // if there is no include pattern, then we include everything by default
       return if (env.isNullOrBlank()) DescriptorFilterResult.Include else filter(env, descriptor)
    }
 
@@ -30,7 +30,7 @@ internal object IncludeTestPatternDescriptorFilter : DescriptorFilter {
 
    private fun match(pattern: TestPattern, descriptor: Descriptor): Boolean {
       logger.log { Pair(descriptor.toString(), "Testing pattern $pattern against $descriptor") }
-      if (pattern.className == null) return packageMatch(pattern, descriptor)
+      if (pattern.className == null) return packageMatch(pattern, descriptor.spec())
       val spec: Descriptor = Descriptor.SpecDescriptor(DescriptorId(pattern.packageName + "." + pattern.className))
       val tests = pattern.contexts.fold(spec) { acc, op -> acc.append(op) }
       return descriptor.hasSharedPath(tests)
@@ -38,9 +38,9 @@ internal object IncludeTestPatternDescriptorFilter : DescriptorFilter {
 
    private fun packageMatch(
       pattern: TestPattern,
-      descriptor: Descriptor
+      descriptor: Descriptor.SpecDescriptor
    ): Boolean {
-      val pck = descriptor.id.value.substringBeforeLast(".")
+      val pck = descriptor.id.value.substringBeforeLast(".") // grabs just the package by dropping the class name
       return if (pattern.subpackages) pck.startsWith(pattern.packageName) else pck == pattern.packageName
    }
 }
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/filter/IncludePatternEnvDescriptorFilter.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/filter/IncludePatternEnvDescriptorFilter.kt
@@ -7,17 +7,17 @@ import io.kotest.core.descriptors.DescriptorId
 import io.kotest.engine.extensions.TestPattern
 import io.kotest.engine.extensions.TestPatternParser
 
+const val INCLUDE_PATTERN_ENV = "KOTEST_INCLUDE_PATTERN"
+
 /**
  * An implementation of [DescriptorFilter] that uses an include pattern from an env variable.
  */
-object IncludePatternEnvDescriptorFilter : DescriptorFilter {
-
-   const val ENV_NAME = "KOTEST_INCLUDE_PATTERN"
+internal object IncludePatternEnvDescriptorFilter : DescriptorFilter {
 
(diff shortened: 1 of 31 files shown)


kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/gradle/GradleUtils.kt
```diff
@@ -72,4 +79,49 @@ object GradleUtils {
...
+
+      val libraryTable = LibraryTablesRegistrar.getInstance().getLibraryTable(module.project)
+      val dependency = libraryTable.libraries.find { it.name?.contains("io.kotest:kotest-framework-engine") ?: false }
+      LOG.info("Kotest dependency found $dependency")
+
+      val version = dependency?.name?.substringAfterLast(":") ?: return null
+
+      // we assume x.y..... format
+      val major = version.take(1).toIntOrNull() ?: return null
+      val minor = version.drop(2).take(1).toIntOrNull() ?: return null
```
alfonsoristorato: hopefully we will never get to v10 😄, how about?
```
val (major, minor) = vers.split(".").take(2)
```

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/filter/IncludePatternEnvDescriptorFilter.kt
```diff
@@ -10,15 +10,15 @@ import io.kotest.engine.extensions.TestPatternParser
 /**
  * An implementation of [DescriptorFilter] that uses an include pattern from an env variable.
  */
-internal object IncludeTestPatternDescriptorFilter : DescriptorFilter {
+object IncludePatternEnvDescriptorFilter : DescriptorFilter {
```
LeoColman: Are you sure you want to remove the internal modifier?
It's the only thing allowing you to rename this without it being a breaking change.

If it wasn't part of the public API before, why make it now?

kotest-intellij-plugin/api/kotest-intellij-plugin.api
LeoColman, line 1: Offtopic: Why do we have to keep the plugin's public API? It's not a public library with a necessarily fixed api

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/gradle/GradleUtils.kt
```diff
@@ -42,13 +48,14 @@ object GradleUtils {
       return taskNames.any { isKotestTaskName(it) }
    }
 
+   @Deprecated("Support in 6.1 has moved to use the standard gradle test tasks")
```
LeoColman, line 45: Why deprecate and not remove it altogether?

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/gradle/GradleUtils.kt
```diff
@@ -42,13 +48,14 @@ object GradleUtils {
...
    fun isKotestTaskName(taskName: String): Boolean {
       return taskName == "kotest" // jvm only task name
          || taskName == "jvmKotest" // multiplatform jvm task name
          || taskName.matches("kotest[a-zA-Z]+UnitTest".toRegex()) // android task names eg kotestReleaseUnitTest, kotestDebugUnitTest, etc.
    }
 
-
+   @Deprecated("Support in 6.1 has moved to use --tests over a gradle property")
```
LeoColman, line 52: Why deprecate and not remove it altogether?

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/gradle/GradleUtils.kt
```diff
@@ -72,4 +79,49 @@ object GradleUtils {
    fun moduleData(module: Module): GradleModuleData? {
       return CachedModuleDataFinder.getGradleModuleData(module)
    }
+
+   /**
+    * Returns the version of Kotest defined for this module or null if Kotest is a dependency.
+    * Assumes that all artifacts in the io.kotest group are using the same version.
```
LeoColman: What about the libraries that we purposefully keep in a different version than Kotest for release cadency?

I believe they're not relevant to `getKotestVersion`, but I'd like to raise the flag

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/gradle/GradleUtils.kt
```diff
@@ -72,4 +79,49 @@ object GradleUtils {
...
+      ) != TestRunner.PLATFORM
+   }
+
+   fun isPlatformRunner(module: Module?): Boolean {
+      if (module == null) return false
+      return GradleProjectSettings.getTestRunner(
+         module.project,
+         ExternalSystemApiUtil.getExternalProjectPath(module)
+      ) == TestRunner.PLATFORM
+   }
```
LeoColman:
```suggestion
   fun isGradleTestRunner(module: Module?): Boolean {
      if (module == null) return false
      return GradleProjectSettings.getTestRunner(
         module.project,
         ExternalSystemApiUtil.getExternalProjectPath(module)
      ) != TestRunner.PLATFORM
   }

   fun isPlatformRunner(module: Module?) = !isGradleTestRunner(module)
```

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/filter/IncludePatternEnvDescriptorFilter.kt
sksamuel: Because the constant is being used in the gradle module too. I'll rejig it and add the internal back.

kotest-intellij-plugin/api/kotest-intellij-plugin.api
sksamuel: Yeah you're right, I've disabled it for this module.

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/gradle/GradleUtils.kt
sksamuel: Because not everyone who upgrades the plugin will also be on 6.1

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/gradle/GradleUtils.kt
sksamuel: Because not everyone who upgrades the plugin will also be on 6.1

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/gradle/GradleUtils.kt
sksamuel: The code is actually looking for the engine module, the comment is outdated. Good catch I'll clarify.

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/gradle/GradleUtils.kt
sksamuel: There's three enums, and I might add the third option. Good suggestion to remove duplication so I'll do this:

```kotlin
   fun runner(module: Module?): TestRunner? {
      if (module == null) return null
      return GradleProjectSettings.getTestRunner(
         module.project,
         ExternalSystemApiUtil.getExternalProjectPath(module)
      )
   }

   fun isGradleTestRunner(module: Module?): Boolean = runner(module) != TestRunner.PLATFORM
   fun isPlatformRunner(module: Module?): Boolean = runner(module) == TestRunner.PLATFORM
```

alfonsoristorato (approved):

PR #5562, opened 2026-01-09
Updated SpecRef's to contain the FQN

sksamuel:
This will be used by the new --tests filtering from the plugin coming in a follow up PR.
Also adds embedded location support for jump to source.

PR #5566, opened 2026-01-13
Only show JUnit ignore reasons if enabled

### LINKED ISSUE #5190: kotest running a single test in intelliJ also lists all other tests as "is excluded by filter(s)"
I'm using 
```toml
[versions]
kotest = "6.0.4"
kotlin = "2.2.21"

[libraries]
kotest = { module = "io.kotest:kotest-framework-engine", version.ref = "kotest" }
kotest-assertions-core = { module = "io.kotest:kotest-assertions-core", version.ref = "kotest" }

[plugins]
kotest = { id = "io.kotest", version.ref = "kotest" }
kotlin-jvm = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
```
and when using the Kotest plugin to run a single test in a `FunSpec` test class that has 17 tests, the result says "16 tests ignored" and shows
```
Testing started at 3:27 PM ...
UtilKtTest/powerSet is excluded by filter(s)
UtilKtTest/powerSetSeq is excluded by filter(s)
UtilKtTest/permutations is excluded by filter(s)
UtilKtTest/neighbors4 is excluded by filter(s)
UtilKtTest/neighbors8 is excluded by filter(s)
UtilKtTest/chunkedBy is excluded by filter(s)
UtilKtTest/countingMap is excluded by filter(s)
UtilKtTest/countingMapWithInit is excluded by filter(s)
UtilKtTest/gcd is excluded by filter(s)
UtilKtTest/numbers is excluded by filter(s)
lcm: 50.004571ms
UtilKtTest/minMax is excluded by filter(s)
UtilKtTest/minMax using multiFold is excluded by filter(s)
UtilKtTest/minMax using multiReduce is excluded by filter(s)
UtilKtTest/multiFold is excluded by filter(s)
UtilKtTest/multiReduce is excluded by filter(s)
UtilKtTest/manhattanDistance is excluded by filter(s)

```
Seeing all the excluded tests makes it hard to see the actual test result. Thus, would be really good if the output could be limited to only show the executed tests.

sksamuel:
New option in the kotest plugin to enable this if so desired

Closes #5190

### DIFF
--- a/kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/JUnitTestEngineListener.kt
+++ b/kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/JUnitTestEngineListener.kt
@@ -1,5 +1,6 @@
 package io.kotest.runner.junit.platform
 
+import io.kotest.common.env
 import io.kotest.common.reflection.bestName
 import io.kotest.core.Logger
 import io.kotest.core.descriptors.Descriptor
@@ -178,7 +179,7 @@ class JUnitTestEngineListener(
       logger.log { Pair(kclass.bestName(), "Spec is being flagged as ignored") }
       val testDescriptor = findTestDescriptorForSpec(root, Descriptor.SpecDescriptor(DescriptorId(kclass.java.name)))
       if (testDescriptor != null)
-         listener.executionSkipped(testDescriptor, reason ?: "")
+         listener.executionSkipped(testDescriptor, reasonIfEnabled(reason))
    }
 
    private fun reset() {
@@ -272,7 +273,13 @@ class JUnitTestEngineListener(
       results[testCase.descriptor] = TestResult.Ignored(reason)
 
       logger.log { Pair(testCase.name.name, "executionSkipped: $testDescriptor") }
-      listener.executionSkipped(testDescriptor, reason ?: "")
+      listener.executionSkipped(testDescriptor, reasonIfEnabled(reason))
+   }
+
+   private fun reasonIfEnabled(reason: String?): String? {
+      if (reason == null) return null
+      if (env("KOTEST_SHOW_IGNORE_REASONS") == "true") return reason
+      return null
    }
 
    private fun startParents(testCase: TestCase) {
--- a/kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/JUnitTestEngineListener.kt
+++ b/kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/JUnitTestEngineListener.kt
@@ -276,17 +276,17 @@ class JUnitTestEngineListener(
       listener.executionSkipped(testDescriptor, reasonIfEnabled(reason))
    }
 
-   private fun reasonIfEnabled(reason: String?): String? {
-      if (reason == null) return null
+   private fun reasonIfEnabled(reason: String?): String {
+      if (reason == null) return ""
       if (env("KOTEST_SHOW_IGNORE_REASONS") == "true") return reason
-      return null
+      return ""
    }
 
    private fun startParents(testCase: TestCase) {
       val parent = testCase.parent
       if (parent != null) {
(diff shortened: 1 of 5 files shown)


kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/JUnitTestEngineListener.kt
```diff
@@ -272,7 +273,13 @@ class JUnitTestEngineListener(
...
 
       logger.log { Pair(testCase.name.name, "executionSkipped: $testDescriptor") }
-      listener.executionSkipped(testDescriptor, reason ?: "")
+      listener.executionSkipped(testDescriptor, reasonIfEnabled(reason))
+   }
+
+   private fun reasonIfEnabled(reason: String?): String? {
+      if (reason == null) return null
+      if (env("KOTEST_SHOW_IGNORE_REASONS") == "true") return reason
+      return null
```
alfonsoristorato: maybe make this return `""` and the method a non nullable `String` - as now we have compilation errors on `executionSkipped`

alfonsoristorato (approved):
Nice stuff - love that we have this disabled by default
also maybe worth adding to the docs about this env variable

kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/JUnitTestEngineListener.kt
sksamuel: I don't know why it's doing that now. Do you?

alfonsoristorato: because `executionSkipped` expects a reason that is not nullable, in fact before we were doing `reason ?: ""`, whereas now we are returning `null` by default. Changing this method to return `String` and instead of return null we return `""` should fix it

alfonsoristorato: something like 
```kotlin
private fun reasonIfEnabled(reason: String?): String =
    if (reason != null && env("KOTEST_SHOW_IGNORE_REASONS") == "true") reason else ""
```

sksamuel: > because `executionSkipped` expects a reason that is not nullable, in fact before we were doing `reason ?: ""`, whereas now we are returning `null` by default. Changing this method to return `String` and instead of return null we return `""` should fix it

It wasn't doing that last week i only just changed it

alfonsoristorato: Ah! [I see what you mean](https://github.com/kotest/kotest/pull/5562/changes#diff-2316de2d175b7bd60efba59dd5460a8865531d62f6412cb0730898d29f92e899).
image
Quite odd

alfonsoristorato: I am so confused, how did this ever compile. Even the previous version of `junit-platform` required that reason to be non-nullable. And whatever history I go and check I get compilation errors.

sksamuel: I wonder if somewhere we or a version bump changed kotlin to look for Java nullable annotations?

alfonsoristorato: good shout! But the last thing that could have affected this was the merging of the plugin I tried reverting `RepositoriesMode` to see if it was that, and excluded the plugin from the build, and still get the same issue
Still very very annoying 😄

PR #5569, opened 2026-01-14
Refactor TeamCity message builders to use testName for finished methods

sksamuel:


alfonsoristorato (approved):
looks fine, without much context tho, I assume you want to use `testName` method everywhere? If so, there's a few missed places in this class

sksamuel:
> looks fine, without much context tho, I assume you want to use `testName` method everywhere? If so, there's a few missed places in this class

Just for test cases, not when we put a placeholder in, which is why some are missed. We only need it for test started / finished / ingored.

PR #5570, opened 2026-01-16
add a CRON based gh action to run kotest-examples repo tests with latest release from master

alfonsoristorato:
Based off the ground work to override Kotest version in `kotest-examples` - PRs:
1. https://github.com/kotest/kotest-examples/pull/88
2. https://github.com/kotest/kotest-examples/pull/89
3. https://github.com/kotest/kotest-examples/pull/90

This PR introduces a gh action that runs all the tests defined within the repo `kotest-examples` with all targets that that repo already tests against.
The initial idea for this action was to:
- run on master merge or every day at a specific time
- pull the latest snapshot release of Kotest
- run all the tests agains the targets

Unfortunately the above is not going to be possible for some targets, as we do not deploy snapshot versions for all https://github.com/kotest/kotest/blob/9c24fa232f77400c9c46ba2bf87a89708195fcbd/.github/workflows/master.yml#L63-L74

So for now it:
- builds a LOCAL release of kotest 
- runs all tests against the targets with that release 
  - only for libraries as even though the kotest gradle plugin can be overridden with `KOTEST_GRADLE_PLUGIN_VERSION` I have not been able to run some targets with the overridden plugin (failures [here](https://github.com/kotest/kotest/actions/runs/21082392472))
- runs at 6 AM UTC every day


alfonsoristorato:
action runs successfully https://github.com/kotest/kotest/actions/runs/21081159036 (except `Linux - Android` and `Linux - wasm` as those fail in `kotest-examples` already)
will revert the hack to make it run all native targets in non-master as when this runs in a cron it will be on master and no hacks are required

alfonsoristorato:
> action runs successfully https://github.com/kotest/kotest/actions/runs/21081159036 (except `Linux - Android` and `Linux - wasm` as those fail in `kotest-examples` already) will revert the hack to make it run all native targets in non-master as when this runs in a cron it will be on master and no hacks are required

and latest run after clean-up with same results https://github.com/kotest/kotest/actions/runs/21083129861
This is now ready to be reviewed @kotest/kotest-developers 😄

.github/workflows/test-kotest-examples.yml
```diff
@@ -0,0 +1,335 @@
...
+              --no-configuration-cache
+
+         - name: Get published version
+           id: get-version
+           run: |
+              VERSION=$(ls ~/.m2/repository/io/kotest/kotest-framework-engine/ | grep -E '^[0-9]' | head -1)
+              echo "version=$VERSION" >> $GITHUB_OUTPUT
+              echo "Published Kotest version: $VERSION"
+
+         - name: Upload Maven Local artifacts
```
sksamuel, line 64: is this needed? What's the difference here between upload artifact and publishToMavenLocal

.github/workflows/test-kotest-examples.yml
```diff
@@ -0,0 +1,335 @@
...
+              path: ~/.m2/repository/io/kotest/
+              retention-days: 1
+
+   publish-local-macos:
+      name: Publish Kotest to Maven Local (macOS)
+      runs-on: macos-latest
+      if: github.repository == 'kotest/kotest'
+      steps:
+         - name: Checkout kotest
+           uses: actions/checkout@v5
```
sksamuel: I think v6 is latest

alfonsoristorato: done [switch to checkout action v6](https://github.com/kotest/kotest/pull/5570/commits/be5e0704bddda9beb10b7604ae4580bf5c3cf59a) normally I copy what is used in other actions and let renovate handle the updates when needed - should we enable it for gh actions?

.github/workflows/test-kotest-examples.yml
alfonsoristorato: I might be wrong, but I think between the builder jobs and the runner jobs the context gets lost, so I have made it so that
- in builders we
  - run publishToMavenLocal
  - upload the local kotest via `upload-artifact` (his action)
- in runners we
  - download artifacts for each target - effectively combining them all in `~/.m2/repository/io/kotest/`
  - run the tests

I think this is still needed, cause in linux for example I need the artifacts for linux but also macOs and windows.
Even if the contexts were not lost between each job, surely the linux job should not be able to read stuff from the windows job?

sksamuel: Gotcha.

sksamuel:
Would it best to have this script in the examples repo, since that's where we want to see it working ?

sksamuel (approved):

alfonsoristorato:
> Would it best to have this script in the examples repo, since that's where we want to see it working ?

I think I am happy with it being in `kotest` - this action should make sure that whatever gets pushed to master does not make the example tests fail

PR #5573, opened 2026-01-17
matching json, find all differences, top level and nested - do not stop at first diff

AlexCue987:
matching json, find all differences, top level and nested - do not stop at first diff


kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/EqualTest.kt
```diff
@@ -743,15 +745,21 @@ expected:<{
             }
             """
          a.shouldEqualJson(b)
```
alfonsoristorato, line 747: I know this was here before, but I wonder how did this test ever pass without a `shouldFail` wrapper?

alfonsoristorato (approved):

kotest-assertions/kotest-assertions-json/src/jvmTest/kotlin/com/sksamuel/kotest/tests/json/EqualTest.kt
alfonsoristorato: nvm - it's the Property Order 😄

PR #5574, opened 2026-01-17
Support to run data tests singularly

### LINKED ISSUE #5389: Run button for withData tests
Hello,
thanks kotest and idea plugin. 

Would be possible to add run button for test wrapped by `withData` from `"io.kotest:kotest-framework-datatest-jvm"`?
On attached image - run button on line 35.
Screenshot from 2023-02-08 22-06-29(2)

alfonsoristorato:
This PR introduces tags for data tests, which will ultimately allow them to be singularly run from the intelliJ plugin.
Issues:
Fixes #5389

**At Framework level:**
- when a data test is registered, two tags get automatically assigned to it:
  - `kotest.data`
  - `kotest.data.{lineNumber}` -> for `jvm` only
  - `kotest.data.nonJvm` -> for `nonJvm` only

This allows us to know exactly what data test needs to run when being invoked by the intelliJ plugin.

**At the intelliJ plugin level:**
- when a data test is recognised, we have access to the line number and we add an `env variable` in the runner
- the `env variable` has key `KOTEST_TAGS` as it is already implemented in `kotest` in general
- the `env variable` value changes according to what type of data test we have
  - for a root data test, the `env variable` would be in its simplest form -> `(kotest.data.{lineNumber}) | kotest.data.nonJvm`
  - for a data test within another data test, the `env variable` would have this structure -> `(kotest.data.{parentDataTestLineNumber} & !kotest.data.{siblingLineNumber}) | kotest.data.nonJvm`
  - for a data test within a regular container, the `env variable` would have this structure -> `((kotest.data.{parentDataTestLineNumber} & !kotest.data.{siblingLineNumber}) | !kotest.data) | kotest.data.nonJvm`

As you may have already guessed, we have added `| kotest.data.nonJvm` at the end of each of the `env variable` in the hope (I can't test native or js sadly) that this allows the framework to recognise that the `left` side of the tags are un-executable (cause in `nonJvm` those tags will not exist - aside `kotest.data`) but it should allow to execute the `right` side, and all tests in `nonJvm` will satisfy the condition of having `kotest.data.nonJvm` tag. 

## Test path for gradle runner
As the work of having test path within each test already exists, any data test that is nested within a container inherits this and re-uses - whereas any data test that is declared at root level, or is a child of a root level data test, simply relies on the tags going from the parent down, and the test path is simply the `package.spec`

## Examples
The test files in [this package](https://github.com/kotest/kotest/tree/081125c5f6b0a90d8201b9acea2b698a388984d6/kotest-intellij-plugin/src/test/resources/data-test-tags) can do a much better job than me at explaining how the tags get generates (I just hope you have a large enough monitor)

## How to run this locally
1. release kotest to `mavenLocal`
2. build the plugin
3. use the local version of kotest in whichever project
4. isntall the local plugin
5. run some tests
6. if you are doing this, it would be good to have feedback on `nonJvm` as I have not been able to test it 😄


alfonsoristorato:
TODO:
- ~fix issue that tags are not removed when running the full spec~ [simplify plugin tag extraction logic and fix issue where spec run was…](https://github.com/kotest/kotest/pull/5574/commits/423f49bc1955a92112ef9b058e217a8deb132b7e)
- ~make this work for first level nested data tests within other data tests~ [support to run first level nested data tests within other data tests](https://github.com/kotest/kotest/pull/5574/commits/f61d749c43f4fe8ef6d7f0d9544b8af208842b80)
- ~make this work for all levels nested data tests within other data tests at all levels~ [support to run nested data tests within other data tests at any level](https://github.com/kotest/kotest/pull/5574/commits/6288f9034043f24f20154f9fc8ba594265e86771)
- ~make this work for nested data tests within normal contexts as a first level nested data test~ [handle data tests within regular containers](https://github.com/kotest/kotest/pull/5574/commits/bad60c0ce8f2ef49cbe87b0f8100dcd028084947)
- ~make this work for nested data tests within normal contexts at any nested data test level~ [support data tests nested within regular containers at any level](https://github.com/kotest/kotest/pull/5574/commits/263b5d4914aaa7d79c405df237e020870e3f47de)
- ~multiple lines withXXX~ **this works already**
- ~try to find a solution for non-jvm (might not be possible and if so it might be cool to at least change the gutter icon if no jvm found to run all data tests with the generic tag - I admit I know very little about non jvm - so I my alternative might also not be possible)~ [add nonJvm support at the plugin level](https://github.com/kotest/kotest/pull/5574/commits/081125c5f6b0a90d8201b9acea2b698a388984d6) had to settle for this, sorry `nonJvm` users
- do the other specs if agreed
- ~change plugin arbitrary test name for data tests~ [change default tooltip for data tests](https://github.com/kotest/kotest/pull/5574/commits/2b54eeeb66d1a326a6f68491f6bcbab38cfd0b4a)
- ~update the docs to inform users how to use the all data test env variable (might be useful for someone)~ [update docs](https://github.com/kotest/kotest/pull/5574/commits/e112fa4d87091c12cad09476c203a17e004a42b8)
- once https://github.com/kotest/kotest/pull/5613 is merged the same logic should be applied for data test within regular containers - same sort of logic should be handled for when regular container of data test is disabled


@kotest/kotest-developers this is an idea on a feature that I really would like to have 😄 - as you can see there are a few TODOs left, but I won't do them until we agree that this is the way forward for us (cause it's a lot of boring copy/paste left only now 😄) - let me know your thoughts!

alfonsoristorato:
@kotest/kotest-developers marking this as ready to review as it is feature complete (minus some cleanups and methods merging as I can see in the plugin I can combine a few methods) - if we agree that this is the way to go, I will add support to the other specs

sksamuel:
I'm excited to review this, will get to it very soon (and will be part of 6.2)

sksamuel:
Fixes https://github.com/kotest/kotest/issues/5389

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/core/source/sourceRef.kt
```diff
@@ -56,7 +60,7 @@ object SourceRefUtils {
             it.className.startsWith("kotlinx.") ||
             it.className.startsWith("io.kotest.core.") ||
             it.className.startsWith("io.kotest.engine.") ||
-            it.className.startsWith("io.kotest.datatest.")
-      }.firstOrNull()
+            (excludeDataTest && it.className.startsWith("io.kotest.datatest."))
```
alfonsoristorato, line 63: I have done this to ensure that we don't break changes made in https://github.com/kotest/kotest/pull/5663

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/BehaviorSpecContextContainerScope.kt
```diff
@@ -17,6 +19,14 @@ class BehaviorSpecContextContainerScope(
    suspend fun given(name: String, test: suspend BehaviorSpecGivenContainerScope.() -> Unit) =
       given(name, xmethod = TestXMethod.NONE, test)
 
+   /**
+    * Adds a test case with config passed as a param.
+    * Marked as internal as it should be used only by the data test registrars.
+    */
+   @KotestInternal
+   suspend fun given(name: String, config: TestConfig, test: suspend BehaviorSpecGivenContainerScope.() -> Unit) =
```
sksamuel: can the data tests not use the standard way of adding a test with config ?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/BehaviorSpecContextContainerScope.kt
sksamuel, line 32: do we need this, if the data tests can do the normal given().config() ?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/BehaviorSpecGivenContainerScope.kt
```diff
@@ -133,10 +153,24 @@ class BehaviorSpecGivenContainerScope(
 
    suspend fun Then(name: String, test: suspend TestScope.() -> Unit) = addThen(name, test, xmethod = TestXMethod.NONE)
    suspend fun then(name: String, test: suspend TestScope.() -> Unit) = addThen(name, test, xmethod = TestXMethod.NONE)
+
+   /**
+    * Adds a test with config passed as a param.
+    * Marked as internal as it should be used only by the data test registrars.
+    */
+   @KotestInternal
+   suspend fun then(name: String, config: TestConfig, test: suspend TestScope.() -> Unit) =
```
sksamuel: same question as before, can the data tests use the public registration with config methods (I won't add this comment to every spec now :)

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/run/gradle/GradleMultiplatformJvmTestTaskRunProducer.kt
```diff
@@ -183,8 +196,33 @@ class GradleMultiplatformJvmTestTaskRunProducer : GradleTestRunConfigurationProd
...
+    * Have to rely on env vars here because Gradle system properties (-D) do not propagate to the test JVM.
+    *
+    * @param runConfiguration The Gradle run configuration to modify.
+    * @param dataTestInfoMaybe The optional [DataTestInfo] for the test.
+    */
+   private fun setOrRemoveDataTestEnvVarIfNeeded(
+      runConfiguration: GradleRunConfiguration,
+      dataTestInfoMaybe: DataTestInfo?
+   ) {
+      dataTestInfoMaybe.takeIf { it != null }
```
sksamuel: can't we just do ?.let and ignore the take if ?

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/run/gradle/GradleMultiplatformJvmTestTaskRunProducer.kt
```diff
@@ -171,9 +175,18 @@ class GradleMultiplatformJvmTestTaskRunProducer : GradleTestRunConfigurationProd
       ) { tasks ->
          logger.info("MultiplatformChooseTasks: $tasks")
 
+         val (spec, test) = testContext
```
sksamuel, line 178: what's the advantage of splitting this out here ?

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/psi/offsets.kt
```diff
@@ -35,3 +35,15 @@ fun PsiElement.findElementInRange(offsets: IntRange): PsiElement? {
...
 fun PsiFile.elementAtLine(line: Int): PsiElement? =
    offsetForLine(line)?.let { findElementInRange(it) }
+
+/**
+ * Returns the 1-based line number where this element is defined in the source file.
+ * Returns null if the line number cannot be determined.
+ */
+fun PsiElement.lineNumber(): Int? {
+   val file = containingFile ?: return null
+   val document = file.viewProvider.document ?: return null
```
sksamuel: I don't know for sure, but I think the best way is PsiDocumentManager.getInstance(element.project)).getDocument(file)

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/DataTestTag.kt
```diff
@@ -0,0 +1,37 @@
...
+ * Returns a [TestConfig] that applies data test tags.
+ */
+@KotestInternal
+fun getDataTestTagConfig(): TestConfig = TestConfig(
+   tags = setOf(
+      /**
+       * A general tag for all data tests.
+       * Can be used to run all data tests in a spec or project with env variable `KOTEST_TAGS="kotest.data"`
+       */
+      NamedTag("kotest.data"),
```
sksamuel, line 17: maybe kotest.data.all might be better to be very clear it's different from kotest.data.1 ?

sksamuel:
Based on your reply to my "do we need extra config methods rather than use existing ones" question, it might be worth creating a PR just with that boilerplate (as its the majority of the files) and then having another PR with the real meat.

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/run/gradle/GradleMultiplatformJvmTestTaskRunProducer.kt
alfonsoristorato: we can, pushed [address nullability checks and line number comments](https://github.com/kotest/kotest/pull/5574/commits/bd94754cbbf1077c386dceba979b7b8a4db701e0)

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/run/gradle/GradleMultiplatformJvmTestTaskRunProducer.kt
alfonsoristorato: Before a refactor I used `spec` and `test` a few times in this method, whereas now I only use `spec` once and `test` twice - at this point this is more a de-structuring for readability, but happy to do `testContext.{prop}` if you prefer

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/psi/offsets.kt
alfonsoristorato: going to trust you here 😄 pushed [address nullability checks and line number comments](https://github.com/kotest/kotest/pull/5574/commits/bd94754cbbf1077c386dceba979b7b8a4db701e0)

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/DataTestTag.kt
alfonsoristorato: I am happy to change it if it makes more sense to you, I personally did not want to use `all` cause this tag is often used to handle exclusions, i.e. doing `!kotest.data.all` in the final tag might sound counter-intuitive. That being said, the same argument could be made for `!kotest.data`. Happy to change it if you like

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/BehaviorSpecContextContainerScope.kt
alfonsoristorato: the thing with the `.config` approach is that:
- for `FreeSpec` and `StringSpec` it's defined as extension method of `String` but the registrar of these tests returns `Unit`
- for other specs it's defined as `context(testName:String)` - which means that I would have to add some methods to handle this and pass the tests in too

Happy to explore it, but I feel it would end up with some more repetitive code than what we have, is the worry that we don't want users to use these new methods?

alfonsoristorato: wait - I think I see what you mean and yeah, I can potentially make that work - leave it with me 😄

documentation/docs/framework/datatesting/data_driven_testing.md
```diff
@@ -168,3 +168,14 @@ you are defining a container or a leaf test.
 :::
 
 Full examples of how each of these can be used can be found in these kotest [tests](https://github.com/kotest/kotest/tree/master/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/datatest/styles)
+
+### Running the data tests
+The IntelliJ Plugin provides full support to run data driven tests. You can run individual data tests, from whenever in the Spec.
```
Kantis:
```suggestion
The IntelliJ Plugin provides full support to run data driven tests. You can run individual data tests, from wherever in the Spec.
```

Kantis (commented):
This is an awesome feature, looking forward to it! 

I had a hard time finding examples where siblings are split by line. Would this be covered already, or do you have thoughts about adding it in the future?

```kotlin
class DataTestTagsStringSpec : FunSpec({

   // StringSpec only supports root-level tests, no nesting
   withData("test1", "test2") { // line 10 -> kotest.data.10
         withData(
            "test3",  // line 12 -> kotest.data.12
            "test4"   // line 13 -> kotest.data.13
          ) {
            1 + 1 shouldBe 2
         }
   }
})
```

alfonsoristorato:
> This is an awesome feature, looking forward to it!
> 
> I had a hard time finding examples where siblings are split by line. Would this be covered already, or do you have thoughts about adding it in the future?
> 
> ```kotlin
> class DataTestTagsStringSpec : FunSpec({
> 
>    // StringSpec only supports root-level tests, no nesting
>    withData("test1", "test2") { // line 10 -> kotest.data.10
>          withData(
>             "test3",  // line 12 -> kotest.data.12
>             "test4"   // line 13 -> kotest.data.13
>           ) {
>             1 + 1 shouldBe 2
>          }
>    }
> })
> ```

Hey 😄 thanks!
At the moment a singular test cannot be picked and run with these changes. The changes are for the ability to run a data test block from wherever, rather than a single data test within a block. And for the example above, the tag would be `// line 11 -> kotest.data.11` (plus whatever negation to avoid running all siblings at the same nesting level).

In terms of finding ways to implement that, I supposed it could be possible (with a lot of work, as at the moment we do not have a way to know the exact data test names from the plugin side, hence the tags hack), but it would become very dependant on whether the users choose to declare these on different lines or not. Definitely something to explore as a next step (if there is a want for it 😄)

documentation/docs/framework/datatesting/data_driven_testing.md
alfonsoristorato: thanks - addressed 😄 [docs typo](https://github.com/kotest/kotest/pull/5574/commits/eda8433e6a80e49a56bd234b5eefe200b22618f0)

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/FreeSpecRootScope.kt
```diff
@@ -145,7 +145,12 @@ interface FreeSpecRootScope : RootScope {
...
 
-   fun String.config(config: TestConfig, test: suspend TestScope.() -> Unit): FreeSpecContextConfigBuilder {
-      return FreeSpecContextConfigBuilder(this, config)
+   fun String.config(config: TestConfig, test: suspend TestScope.() -> Unit) {
+      addTest(
+         testName = TestNameBuilder.builder(this).build(),
+         xmethod = TestXMethod.NONE,
+         config = config,
+         test = test
+      )
```
alfonsoristorato, line 154: I think it was a mistake to have this return a `FreeSpecContextConfigBuilder`, as the only use-case for it to return `FreeSpecContextConfigBuilder` is if anyone would have called a `-` operation after, but that is covered in [this method](https://github.com/kotest/kotest/blob/b38a70b84e096a9232f2061731e24277184606b8/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/FreeSpecRootScope.kt#L83-L87)- when a name and a test are both passed, we can safely establish that we are doing a leaf test, and therefore we need to call `addTest` (like the method straight above this one) instead of returning  `FreeSpecContextConfigBuilder` - I noticed this as some tests were failing for `freeSpec`

PR #5583, opened 2026-01-19
deprecate-shouldContainAllInAnyOrder

AlexCue987:
shouldContainAllInAnyOrder does exactly the same thing as shouldContainAll, no need to keep them both


sksamuel (approved):

sksamuel:
The plugin is being released now, and 6.1 then immediately after, will sneak this in.

PR #5584, opened 2026-01-19
6.1 docs

alfonsoristorato:


documentation/sidebars.js
```diff
@@ -35,7 +35,6 @@ module.exports = {
      "extensions/decoroutinator",
      "extensions/instant",
      "extensions/junit_xml",
-     // "extensions/koin",
```
alfonsoristorato, line 38: this was causing the script to fail jq parsin, as obvs can't have comments in `json` - hopefully no-one minds - if we care about this being commented out I will revert

sksamuel: We should re-add it as koin is back in now

sksamuel:
Site still showing 6.1 as WIP, did you do this bit:
```
After running this script, update the current version to the next value in docusaurus.config.js
```
It's hard to tell from the commit as its large :)

documentation/sidebars.js
alfonsoristorato: Will re-add It in next pr, where I fix docusaurus 😅

alfonsoristorato:
> Site still showing 6.1 as WIP, did you do this bit:
> ```
> After running this script, update the current version to the next value in docusaurus.config.js
> ```
> It's hard to tell from the commit as its large :)

Just now realised it was 15k lines add, TBF I struggled with this pr so I ended up doing another push after this to fix it. Will raise another pr for this and koin (although somehow  koin is there in the 6.1 docs, and I don't know how 😂)

Issue #5601, opened 2026-01-21
CustomEq fails when part of a data class in 6.1.0


scottdfedorov:
When registering a custom eq, it passes when comparing the class directly, however fails when comparing a data class with the class as a field.
Stepped through the debugger, the `DataClassEq` is correctly using the registered custom eq, but the data class comparator is still failing. I believe because the test method is doing a simple compares
`private fun test(a: Any?, b: Any?): Boolean = makeComparable(a) == makeComparable(b)`

Example:
```kotlin
object InstantEq : Eq<Instant> {
    override fun equals(actual: Instant, expected: Instant, context: EqContext): EqResult {
        return if (actual.toEpochMilliseconds() == expected.toEpochMilliseconds())
            EqResult.Success
        else EqResult.Failure {
            AssertionErrorBuilder.create().withMessage("Expected $actual to equal $expected").build()
        }
    }
}


class CustomEqTest : StringSpec({
    beforeSpec {
        DefaultEqResolver.register(Instant::class, InstantEq)
    }
    
    "Instant equality test" {
        val i1 = Instant.fromEpochMilliseconds(1640995200000).plus(200.nanoseconds)
        val i2 = i1.plus(400.nanoseconds)
        i1 shouldBe i2
    }

    "Instant as field equality test" {
        data class Foo(val instant: Instant)

        val f1 = Foo(Instant.fromEpochMilliseconds(1640995200000).plus(200.nanoseconds))
        val f2 = Foo(Instant.fromEpochMilliseconds(1640995200000).plus(400.nanoseconds))

        f1 shouldBe f2
    }
})
```

sksamuel:
Fixed by #5602

PR #5602, opened 2026-01-21
Fix DataClassEq comparator

scottdfedorov:
Fix for https://github.com/kotest/kotest/issues/5601

The DataClassEq was failing to use the Eq methods to compare the data classes if the simple check by equals method failed. It was only using the custom check on the data classes to get the differences to print.

This change keeps the existing behavior for checking by equals method, and if that fails then attempts to check via the existing differences method (unchanged) and returns a failure only if there are differences.


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/DataClassEq.kt
```diff
@@ -41,26 +41,32 @@ internal object DataClassEq : Eq<Any> {
...
-               dataClassDiff(actual, expected, context = context)?.let { diff -> formatDifferences(diff) + "\n\n" } ?: ""
-            }.getOrElse { "" }
+            val differences = dataClassDiff(makeComparable(actual), makeComparable(expected), context = context)
 
-            EqResult.Failure {
-               AssertionErrorBuilder.create()
-                  .withMessage(detailedDiffMsg)
-                  .withValues(Expected(expected.print()), Actual(actual.print()))
-                  .build()
+            if (differences == null || differences.differences.isEmpty()) {
```
AlexCue987: can you elaborate on the problem you are trying to solve? maybe it could be a unit test that fails without the fix, and passes with it

scottdfedorov: Sorry, this wasnt meant to be ready for review, just was opening a PR to show @sksamuel  what I was talking about and let CI run the tests for me. (A productivity trick he taught me when we worked together.)

It's in reference to this: https://github.com/kotest/kotest/issues/5601

The custom eq was my req initially so that when comparing data classes, we can set Instants to not care about nanos to make tests with like a postgres Testcontainer easy.

AlexCue987: makes sense. can you convert this PR to draft?

sksamuel:
@scottdfedorov anything outstanding on this ?

scottdfedorov:
Yeah still need to write tests just haven't gotten around to it yet.

scottdfedorov:
@AlexCue987 @sksamuel Got the fix and tests up for this.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/DataClassEqTest.kt
```diff
@@ -35,12 +43,54 @@ data class IntRatio(
...
       IntRatio(1, 2) shouldBe IntRatio(2, 4)
    }
 
+   "respects custom registered Eq function" {
+      val i1 = Instant.fromEpochMilliseconds(1640995200000).plus(200.nanoseconds)
+      val i2 = i1.plus(400.nanoseconds)
+
+      i1 shouldNotBe i2 // Custom eq not yet registered
+
+      DefaultEqResolver.register(Instant::class, InstantWithoutNanosEq)
```
AlexCue987, line 72: `DefaultEqResolver.register` is not thread safe. Is that a concern?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/DataClassEqTest.kt
```diff
@@ -35,12 +43,54 @@ data class IntRatio(
...
+      i1 shouldNotBe i2 // Custom eq not yet registered
+
+      val f1 = Foo(i1)
+      val f2 = Foo(i2)
+      f1 shouldNotBe f2 // Custom eq not yet registered
+
+      DefaultEqResolver.register(Instant::class, InstantWithoutNanosEq)
+      f1 shouldBe f2
+      DefaultEqResolver.unregister(Instant::class)
+   }
```
AlexCue987, line 92: do we need to unit test the following scenarios:

* two fields in a data class with the same custom Eq function
* two fields in a data class with two different custom Eq functions
* field in a nested data class
* two fields in two different nested data classes

AlexCue987 (commented):
IMO this is highly useful and can be applied to several similar cases, such as BigDecimals as in `BigDecimal.ONE vs BigDecimal("1.00")` . as such, do we need to add/modify any docs?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/DataClassEqTest.kt
scottdfedorov: @sksamuel Has [added to the docs already](https://kotest.io/docs/assertions/shouldbe.html#custom-eq-instances) some details.

and I followed a similar pattern in the [unit tests for the CustomEq](https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/CustomEqTest.kt) but didn't wanna register it for the entire spec. 

I suppose technically it another test running in parallel also checking an instant with differences in the nanosecond precision would ultimately fail to fail because of this being added to the default, but I figured adding and immediately removing makes this overlap unlikely. We could create a custom class, but given the instant is really the thing that was the driver for my need on this, I figured I'd test with that. Happy to create a new data class to compare with if you think it worthwhile.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/DataClassEqTest.kt
scottdfedorov: Sure, I can add those, though I think I'll add those to the custom eq test file rather than the data class eq file? Might be better scoped there?
I could move these tests also but figured even if a bit duplicative worth having it here. Let me know and I'll make some other examples. 

https://github.com/kotest/kotest/blob/master/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/CustomEqTest.kt

scottdfedorov:
Going to convert this to a draft for a bit. Nice catch, the new tests are showing that nested custom eq functions aren't being respected when comparing data classes. 

Not sure I'll have time to get to this this month but we'll see.

scottdfedorov:
Seemed a bit easier than I expected. We weren't using the EqResolve to check for equality, so I updated the flow a bit, lmk your thoughts.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/DataClassEq.kt
```diff
@@ -81,19 +96,20 @@ internal object DataClassEq : Eq<Any> {
...
          val expectedPropertyValue = prop.call(expected)
-         if (isDataClassInstance(actualPropertyValue) && isDataClassInstance(expectedPropertyValue))
-            dataClassDiff(actualPropertyValue, expectedPropertyValue, depth + 1, context)?.let { diff ->
-               Pair(prop, diff)
-            }
-         else {
-            val result = EqCompare.compare(actualPropertyValue, expectedPropertyValue, context)
-            when (result) {
-               is EqResult.Failure -> {
+         val result = EqCompare.compare(actualPropertyValue, expectedPropertyValue, context)
```
scottdfedorov, line 99: Checking first using the comparator for all types, including data classes. Only if the comparator fails, then we fall back to recursing though the data class again if it is a data class, rather than only using the EqCompare for non-data classes

sksamuel:
Anything outstanding on this I can help with ?

AlexCue987:
> Anything outstanding on this I can help with ?

oops dropped the ball on this thing. I'll re-review tomorrow, hope that's not too late.

sksamuel:
Failing on an unrelated flakely property test test which I thought I had finally fixed.
SequencesTest[jvm] > Arb.sequence should[jvm] > allow repeated elements in edge cases for given sequence size[jvm] FAILED
Looking at it again now.

AlexCue987 (approved):
LGTM

PR #5606, opened 2026-01-22
Auto-add the engine dep for KMP projects

sksamuel:
Will simplify setup, just add this plugin


kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestGradleExtension.kt
```diff
@@ -11,4 +11,9 @@ open class KotestGradleExtension {
     * Set to true and the Kotest engine will propagate ignore reasons to Gradle.
     */
    var showIgnoreReasons = false
+
+   /**
+    * Set to false, and the Gradle plugin will not automatically add the engine dependency
+    */
+   var autoAddEngineDependency = true
```
alfonsoristorato: this is worth some doc updates.
as mainly a jvm dev I actually like the control and intention I need to put to use both a plugin and a dependency (as kotest can work without the plugin in jvm) - I assume in KMP you need both the plugin and the deps for kotest to work, but maybe people like me won't like this - so defo add some docs on how to disable this feature :)

alfonsoristorato (approved):
as mainly a jvm dev I actually like the control and intention I need to put to use both a plugin and a dependency - I assume in KMP you need both the plugin and the deps for kotest to work, but maybe adding kotest to all

kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestGradleExtension.kt
sksamuel: I thought if you're using JVM only you won't have commonTest and so it won't do anything. I'll check.

alfonsoristorato: I think you are correct in that, I meant more - if we are now forcing this dependency to users via the plugin only, let's also provide them with a way to disable such automation - assuming we can

sksamuel: It wasn't meant to be like, you have to do it this way, more that we can update the docs to be like "just add the plugin, the engine bit is done for you". But if you think that's not really useful, happy to close this PR. I'm just looking for ways to make the plugin more useful :)

alfonsoristorato: My only comment was about remembering to update the docs, in case users prefer to do this manually 😅 I am sure plenty of people will find this useful

PR #5608, opened 2026-01-22
reintroduce string spec tests with config to be root tests

alfonsoristorato:
As per PR title - we had accidentally changed the logic to register container tests instead of Root test when using `.config`


sksamuel (approved):

sksamuel:
The test failure seems to be a docker rate limit issue, unrelated.

PR #5612, opened 2026-01-24
introduce full support for multiline and trimming tests

alfonsoristorato:
With the new `GradleMultiplatformJvmTestTaskRunProducer` we had lost the possibility to run tests whose names
- span over multiple lines
- have extra spaces

this should fix it


sksamuel (approved):
awesome.

at some point I think we need to make a module that is shared between the engine and the intellij plugin that contains the common contract stuff.

Issue #5605, opened 2026-01-22
Disable child tests when parent is interpolated


alfonsoristorato:
Writing this here as I will forget.

When a test is interpolated the plugin informs the user that such test cannot be started through it, as it can't figure out the test name.

Ideally all children of this test (if any, and obviously if this is a container) should also have the same logic as their test path will also, inherently,  impossible to be figured out by the plugin.

PR #5613, opened 2026-01-24
make sure all children of an interpolated container test are marked as interpolated too

alfonsoristorato:
For Issue #5605
This ensures that if a container test name is considered interpolated, all its children are considered as such too (with different tooltips accordingly)

### DIFF
--- a/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/InterpolatedTestLineMarker.kt
+++ b/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/InterpolatedTestLineMarker.kt
@@ -18,13 +18,15 @@ import org.jetbrains.kotlin.psi.KtDeclarationModifierList
 import org.jetbrains.kotlin.psi.KtImportDirective
 import org.jetbrains.kotlin.psi.KtImportList
 import org.jetbrains.kotlin.psi.KtPackageDirective
+import io.kotest.plugin.intellij.Test
 
 /**
  * Adds an icon to the gutter for tests which have an interpolated name.
  */
 class InterpolatedTestLineMarker : LineMarkerProvider {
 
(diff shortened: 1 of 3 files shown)


alfonsoristorato:
will add tests once https://github.com/kotest/kotest/pull/5612 is merged

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/InterpolatedTestLineMarker.kt
```diff
@@ -18,13 +18,15 @@ import org.jetbrains.kotlin.psi.KtDeclarationModifierList
...
+import io.kotest.plugin.intellij.Test
 
 /**
  * Adds an icon to the gutter for tests which have an interpolated name.
  */
 class InterpolatedTestLineMarker : LineMarkerProvider {
 
-   private val text = "Tests with an interpolated name cannot be run using the plugin."
+   private val thisTestIsInterpolatedText = "Tests with an interpolated name cannot be run using the plugin."
+   private val parentOfTestIsInterpolatedText = "Tests contained within other that have an interpolated name cannot be run using the plugin."
```
sksamuel: typo: Tests contained within [an]other...

sksamuel (approved):

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/InterpolatedTestLineMarker.kt
alfonsoristorato: I'd like to use plural, cause the chain can go on and on. Maybe 'tests contained within other tests...'?

sksamuel: Ok sounds good

PR #5618, opened 2026-01-24
deleted:    documentation/docs/assertions/sql.md

AlexCue987:


sksamuel:
did the docs move somewhere else ?

AlexCue987:
> did the docs move somewhere else ?

this empty file is not referenced in 6.0 or 6.1 docs

sksamuel:
But the matchers still exist so I wonder if I moved them or deleted the docs by accident?

AlexCue987:
> But the matchers still exist so I wonder if I moved them or deleted the docs by accident?

according to git history, this file has never ever been populated, it's got nothing but the header from day one.

should we document `io.kotest.matchers.sql` instead? I did know it existed when I opened this PR.

PR #5643, opened 2026-01-29
Flatten nested test output in TeamCity for Native

### LINKED ISSUE #5101: JunitXmlReporter not generating XML report files Kotest 6.0.2
We did the the configuration based on the official Kotest documentation and the JunitXmlReporter no longer generating the xml report files with Kotest 6.0.2.
We upgraded from Kotest 5.9.1 and it worked with that version.

### LINKED ISSUE #5181: Test name starting with the same word as the last word of the context name causes leading dash in test name output
Here's a weird one.... When the context is named "parse", and the test name starts with "parse", the runner inserts a `-` at the beginning of the test name:

```kotlin

class NameTests : FunSpec(
    {
        context("parse") {
            test("parse me") {
                "test" shouldBe "test"
            }
        }
    }
)
```

Output:
```
2.  NameTests
+ parse
    - - parse me OK
```

If we add anything to the context name (like, "parse it"), or change the first word of the test name, then the test name is reported correctly:

```kotlin
        context("parse it") {
            test("parse me") {
                "test" shouldBe "test"
            }
        }
```

Output:
```
2.  NameTests
+ parse it
    - parse me OK
```

Obviously, this is a fairly minor corner case. But it's just odd.

**Kotest: `6.0.4`**
**Kotlin/Native: `2.2.20`**

sksamuel:
Native output doesn't support nested tests. This doesn't matter for kotlin.test because they're not nested.
So for Kotest we need to flatten output so each test "looks like its a root test".

Fixes #5181
Fixes #5101


kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/listener/TeamCityTestEngineListenerEmbeddedLocationsTest.kt
```diff
@@ -1,99 +1,98 @@
...
 import io.kotest.core.test.TestType
-import io.kotest.engine.teamcity.TeamCityWriter
+import io.kotest.engine.listener.TeamCityTestEngineListener
 import io.kotest.engine.test.TestResult
-import io.kotest.engine.test.names.DisplayNameFormatting
 import io.kotest.extensions.system.captureStandardOut
 import io.kotest.matchers.shouldBe
 
-class TeamCityWriterJvmTest : FunSpec() {
+class TeamCityTestEngineListenerEmbeddedLocationsTest : FunSpec() {
```
alfonsoristorato, line 14: Both contexts in this test are disabled, are we going to circle back to them?

sksamuel: Yes, they were disabled before this too - in 6.2 or later we'll make "jump to source" work on native too, but there's some work still to be done.

alfonsoristorato (approved):

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/listener/TeamCityTestEngineListenerEmbeddedLocationsTest.kt
alfonsoristorato: nice ty

PR #5657, opened 2026-02-03
Add option to gradle plugin to configure power assert

### LINKED ISSUE #5656: Update the Kotest Gradle plugin to configure power assert (optionally)

sksamuel:
Closes #5656


kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
```diff
@@ -98,6 +100,35 @@ abstract class KotestPlugin : Plugin<Project> {
...
+   private fun configurePowerAssert(project: Project) {
+
+      // apply the power assert plugin, won't matter if already applied
+      project.pluginManager.apply("org.jetbrains.kotlin.plugin.power-assert")
+
+      // then configure it
+      project.pluginManager.withPlugin("org.jetbrains.kotlin.plugin.power-assert") {
+         project.extensions.configure(PowerAssertGradleExtension::class.java) {
+            // we can add new functions to this later without requiring users to make changes
+            functions.set(listOf("io.kotest.matchers.shouldBe"))
```
alfonsoristorato, line 119: would it be too much to make this list configurable?

alfonsoristorato (approved):

kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt
sksamuel: nope we can do that, but the point of this plugin is that we align it with the assertions library so you don't have to think about it.
What's the use case you're thinking of ?

alfonsoristorato: If I read this power assert correctly, users would have to do 
```
powerAssert {
  functions = listOf("io.kotest.matchers.shouldBe")
}
```
but with these changes they no longer have a need to declare the above, as (at least `shouldBe`) will come by default. I was simply proposing that the same way they can declare a list of functions in a `powerAssert` block, we might want to allow them to do the same in the `kotest` block

sksamuel: Yes, we can allow them to do that, but the reason they have to do it in the power assert block currently is because power assert doesn't know Kotest natively. So there's no reason for them to set the setting to anything other than shouldBe if they are using Kotest. So I don't know why we'd want to allow people to override it? We absolutely can, just curious if there's a use case for it.

alfonsoristorato: Sorry, I think I misunderstood this powerAssert stuff. I had assumed we could pass whatever matcher from kotest and it would decorate it somehow. If it's only shouldBe that we need to pass, then ignore me :)

sksamuel: Yeah it needs special support in the assertion function and it's power assert we are configuring. Should be is the only supported function

PR #5658, opened 2026-02-04
Refactor Gradle plugin to use `Property<Boolean>` for extension values

### LINKED ISSUE #5634: kotest and jvmKotest gradle task not created by gradle plugin
Tried gradle plugins
id("io.kotest") version "6.1.2"
and
id("io.kotest") version "6.1.1"

The issue is here
https://github.com/kotest/kotest/blob/9f4bfcc5e307da49c5ffd8e3ecf4f113a1fc552a/kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt#L74-L84

at the time the plugin is applied `customGradleTask` is false

so if one has config like 
`plugins {
    id("io.kotest") version "6.1.2"
}
kotest {
    customGradleTask = true
}
`

its too late and kotest and jvmKotest tasks are never created, This violates Gradle's standard DSL pattern where extensions are typically configured after plugin application.

ive worked with gradle plugins before so top of my head potential solutions might be

1. supporting a Gradle property: Allow configuration via -Pkotest.customGradleTask=true or similar
2. Or using a provider: Make customGradleTask a Property<Boolean> that can be configured lazily

hope this helps!

sksamuel:
Fixes #5634


alfonsoristorato (approved):

kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestGradleExtension.kt
```diff
@@ -1,27 +1,29 @@
...
    /**
     * Set to true, and the Kotest engine will propagate ignore reasons to Gradle.
     */
-   var showIgnoreReasons = false
+   @ExperimentalKotest
+   abstract val showIgnoreReasons: Property<Boolean>
 
    @ExperimentalKotest
-   var alwaysRerunTests = false
+   abstract var alwaysRerunTests: Property<Boolean>
```
alfonsoristorato: this needs to be set as val for gradle to work :)

sksamuel: oh is that right, will change.

sksamuel: done

sksamuel:
not gonna merge until I have a chat with @aSemy if this makes sense

Issue #5666, opened 2026-02-05
SpringExtension no longer usable via property override in base class


rei-paypay:
Hello there.

I'm not sure if this is a bug, but I'm experiencing some breaking changes when migrating to 6.x and would like to get some help here.

# Context
I'm using
- Spring Boot 4
- Kotest 6.1.2
- Kotlin 2.3.0

I have a `BaseIntegrationTestSpec` that looks like this
```
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
@ActiveProfiles("test")
@EnableAutoConfiguration
@DirtiesContext
open class BaseIntegrationTestSpec(body: DescribeSpec.() -> Unit = {}) : DescribeSpec(body) {
  override val extensions: List<Extension> = listOf(SpringExtension)
  ...
}
```
and I can subclass this to create test classes, e.g.
```
@ContextConfiguration(...)
@TestPropertySource(...)
class SomeIntegrationTest(...) : BaseIntegrationTestSpec({...})
```

# Problem
Before 6.x, this works perfectly fine, with each subclass running as expected.
Now after 6.x, it seems like to get autowiring to work, either
- `@ApplyExtension(SpringExtension::class)` must be used, or
- `SpringExtension` must be registered in project-level config

The latter is not feasible as I have other non-Spring related tests within the same project, and the former means that every single subclass (and we have a lot) must be annotated or the zero-arg error will be thrown.

# Attempts
I tried to use `@ApplyExtension(SpringExtension::class)` on `BaseIntegrationTestSpec` itself but this seems to do nothing. Overriding `extensions` in the same class does nothing either.

# Question
Is this expected and is there a way I can get this to work?

sksamuel:
Looks like ApplyExtension isn't being picked up from parents, that seems like the bug we need to fix.

rei-paypay:
@sksamuel 
Thanks for the prompt response.
I submitted #5667 to fix this issue and it'd be great if you can have a look at it.
Thanks again.

sksamuel:
Fixed by #5667

sksamuel:
6.1.3 being released now

PR #5667, opened 2026-02-05
SpringExtension now supports test class inheritance

rei-paypay:
# Context
https://github.com/kotest/kotest/issues/5666

# Changes
- `applyExtensions` now searches for both annotations and superclasses
- Added small Spring test to test for `SpringExtension` being applied to the superclass


sksamuel:
Awesome

PR #5671, opened 2026-02-05
ensure alwaysRerunTests is evaluated before deciding if to apply

alfonsoristorato:
The new `alwaysRerunTests` property was not being read during the apply - this solution works, and I suspect we'll have to do the same for `powerAssert`


alfonsoristorato:
ah - nvm - this was already addressed 🤣 https://github.com/kotest/kotest/pull/5658

PR #5681, opened 2026-02-10
support for config on all methods and data test changes for all specs

alfonsoristorato:
Adds support to call any spec method with a config attached - used this in data tests for all specs.
Will be supporting the use-case of adding tags for data tests in https://github.com/kotest/kotest/pull/5574

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/RootTestWithConfigBuilder.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/RootTestWithConfigBuilder.kt
@@ -18,6 +18,13 @@ class RootTestWithConfigBuilder(
 ) {
 
    fun config(
+      config: TestConfig,
+      test: suspend TestScope.() -> Unit,
+   ) {
+      context.addTest(testName = name, xmethod = xmethod, config = config, test = test)
+   }
+
+   fun config(
       enabled: Boolean? = null,
       invocations: Int? = null,
       tags: Set<Tag>? = null,
@@ -50,6 +57,6 @@ class RootTestWithConfigBuilder(
          retries = retries,
          retryDelay = retryDelay,
       )
-      context.addTest(testName = name, xmethod = xmethod, config = config, test = test)
+      config(config, test)
    }
 }
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/behaviorSpecContextContainerScope.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/behaviorSpecContextContainerScope.kt
@@ -229,7 +229,7 @@ suspend fun <T> BehaviorSpecContextContainerScope.withContexts(
    test: suspend BehaviorSpecContextContainerScope.(T) -> Unit
 ) {
    ts.forEach { t ->
-      context(nameFn(t), xmethod = TestXMethod.NONE) { this.test(t) }
+      context(nameFn(t)).config() { this.test(t) }
    }
 }
 
@@ -244,7 +244,7 @@ suspend fun <T> BehaviorSpecContextContainerScope.withGivens(
    test: suspend BehaviorSpecGivenContainerScope.(T) -> Unit
 ) {
    ts.forEach { t ->
-      given(nameFn(t), xmethod = TestXMethod.NONE) { this.test(t) }
+      given(nameFn(t)).config() { this.test(t) }
    }
 }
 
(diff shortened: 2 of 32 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/RootTestWithConfigBuilder.kt
```diff
@@ -17,6 +17,13 @@ class RootTestWithConfigBuilder(
    private val xmethod: TestXMethod,
 ) {
 
+   fun config(
```
sksamuel, line 20: this is a good idea, instead of all the parameters, allow people to just pass in a test config object.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/behaviorSpecContextContainerScope.kt
```diff
@@ -284,5 +284,5 @@ suspend fun <T> BehaviorSpecContextContainerScope.withGivens(
    data: Map<String, T>,
    test: suspend BehaviorSpecGivenContainerScope.(T) -> Unit
 ) {
-   data.forEach { (name, t) -> given(name, xmethod = TestXMethod.NONE) { this.test(t) } }
+   data.forEach { (name, t) -> given(name).config() { this.test(t) } }
```
sksamuel, line 287: why do we need to call .config() at all? Isn't' the default given() that just uses null config sufficient?

like

 data.forEach { (name, t) -> given(name)) { this.test(t) } }

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/wordSpecWhenContainerScope.kt
```diff
@@ -210,7 +210,7 @@ suspend fun <T> WordSpecWhenContainerScope.withWhens(
    test: suspend WordSpecWhenContainerScope.(T) -> Unit
 ) {
    ts.forEach { t ->
-      nameFn(t) `when` { this.test(t) }
+      nameFn(t).config() `when` { WordSpecWhenContainerScope(this).test(t) }
```
sksamuel: Do we need to re-wrap the this in another container scope instead of using the the receiver one ?

sksamuel:
Added a couple of questions

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/behaviorSpecContextContainerScope.kt
alfonsoristorato: You are correct, we do not need to pass a null config for now. This is in preparation for the final pr where I will be passing the tags in that config, and I wanted to be sure that by doing this change I was not breaking any existing test.

sksamuel (approved):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/behaviorSpecContextContainerScope.kt
sksamuel: Gotcha

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/datatest/wordSpecWhenContainerScope.kt
alfonsoristorato: good catch, I have had to do this for `StringSpec` and `FreeSpec` cause their equivalent `.config` methods take `TestScope` as parameter, as opposed to `StringSpecScope` (for example). I have wrapped them all first and then added the equivalent `config` where was missing and never removed it in some. This fixes it [avoid rewrapping in containers and use receivers directly](https://github.com/kotest/kotest/pull/5681/commits/ae0cefa35389082378cea5aefcc3241ab5a5a5d0)

`StringSpec` and `FreeSpec` do still require this wrapping, and I would be happy to consider a better API for those, but I don't want to break the api for anyone that is already using these methods - i.e. https://github.com/kotest/kotest/blob/a9b5ba7c1474032c7f2b96e6ed06a2c6fd8d522a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/FreeSpecRootScope.kt#L148

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/style/scopes/RootTestWithConfigBuilder.kt
alfonsoristorato: this was already implemented in `RootContainerWithConfigBuilder` this way, and it made sense to copy. Sadly this was not mine, but a copy of your idea 🤣

PR #5684, opened 2026-02-14
Test Result Gutters

### LINKED ISSUE #5675: Feature request: indicate pass/fail status using gutter icons
When running a junit test you can see individual test method status and easily run the failed test:

Image

When running kotest it's not easy to run failed test because there's no distinction between failed and passed:

Image

It would be nice to have different gutter icons depending on the last result of test execution

alfonsoristorato:
Closes https://github.com/kotest/kotest/issues/5675
This introduces different gutters according to test result, in tests, containers and specs.

### Implementation screenshot - some failures
image

### Implementation screenshot - all pass
image


kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/LineMarkerUtils.kt
```diff
@@ -0,0 +1,47 @@
...
+import io.kotest.plugin.intellij.existingEditor
+import io.kotest.plugin.intellij.gradle.GradleUtils
+import io.kotest.plugin.intellij.psi.enclosingKtClassOrObject
+import io.kotest.plugin.intellij.psi.isTestFile
+import io.kotest.plugin.intellij.psi.specStyle
+import io.kotest.plugin.intellij.testMode
+
+object LineMarkerUtils {
+
+   fun validateElementAndReturnTestDescriptorPath(
```
alfonsoristorato: I have combined a few of the optimisations already present when scanning the file - by the end of this PR I'd like to re-use this in all the Markers at least

sksamuel:
Before we review, can you confirm this doesn't just work if you use the 6.1+ Kotest gradle plugin on JVM targets?

alfonsoristorato:
> Before we review, can you confirm this doesn't just work if you use the 6.1+ Kotest gradle plugin on JVM targets?

Ah good shout. The spec gutter will work with or without the plugin, the tests gutter will require the plugin with this impl. Happy to find an impl to make it work without the plugin too, unless we want to use this as a push for people to use the plugin 🤣 

For before 6.1, same as above, I am sure I can find a way to make it work, but wonder if we want to do that or if we can use it as a push for people to upgrade.

alfonsoristorato:
> > Before we review, can you confirm this doesn't just work if you use the 6.1+ Kotest gradle plugin on JVM targets?
> 
> Ah good shout. The spec gutter will work with or without the plugin, the tests gutter will require the plugin with this impl. Happy to find an impl to make it work without the plugin too, unless we want to use this as a push for people to use the plugin 🤣
> 
> For before 6.1, same as above, I am sure I can find a way to make it work, but wonder if we want to do that or if we can use it as a push for people to upgrade.

I have now introduced support for both plugin users and non plugin users - only on kotest 6.1+.

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/LineMarkerUtils.kt
```diff
@@ -0,0 +1,158 @@
...
+    * Note: We check for Kotest 6.1+ to be present and return [TestStatus.UNKNOWN] if not, to avoid wasting time
+    * looking up keys in the storage for older versions of Kotest that might not follow the same pattern.
+    */
+   private fun determineTestStatus(
+      test: Test,
+      allKeys: Collection<String>,
+      storage: TestStateStorage
+   ) =
+      determineTestStatusWithKotestTags(test, allKeys, storage)
+         ?: determineTestStatusWithoutKotestTags(test, allKeys, storage) ?: TestStatus.UNKNOWN
```
alfonsoristorato: I wish I could have used the util `hasKotestGradlePlugin`, but that util relies on the `kotest` gradle task being present, which is no longer auto-enabled. This approach checks first as if the user had the plugin, then moves onto checking as if the user does not have the plugin, and if none matches, falls back to to a default

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/LineMarkerUtils.kt
```diff
@@ -0,0 +1,158 @@
...
+      return state?.let { getTestStatus(it.magnitude) } ?: TestStatus.UNKNOWN
+   }
+
+   /**
+    * Determines the test status for individual tests.
+    *
+    * The URL format varies whether the Kotest Gradle plugin is present:
+    *
+    * 1. Kotest 6.1+ with Kotest Gradle plugin:
+    *    java:test://<kotest>fqn/testName -- innerTestMaybe</kotest>TestDisplayName
```
sksamuel: I want to ditch this <kotest> thing and move to something like

`java:test://fqn/testName » innerTest` as we don't need to support display names when running tests in the IDE, just in reports. And reports can work out that it's not running in IDE. This simplifies everything, and means if you don't have the ide support enabled, you don't get the ugly kotest tags all over everything.

So that being said, I think what we should do in this class is just parse out the kotest bit, and make the keys always be the "test path", so maybe add a parser that just looks for the presence of kotest tags, and if it exists, take what's inside, if it doesn't exist, take whatever is after java:test.

Then if we change the format latter to be `java:test://fqn/testName » innerTest` it'll just continue to work in all cases. Wdyt?

As an aside, is intellij generating java:suites for nested parents, or only for the class?

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/LineMarkerUtils.kt
sksamuel: If you see my comment, we might not need to care.

sksamuel (approved):
This is fking incredible btw.

sksamuel:
I disabled auto merge as I added comments. Would love to get this into the 6.2 release.

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/LineMarkerUtils.kt
```diff
@@ -0,0 +1,158 @@
...
+                  key.startsWith("java:test://$pathMarker") || key.startsWith("java:suite://$pathMarker"))
+         }
+         matchingKey?.let { storage.getState(it) }
+      }
+
+      return state?.let { getTestStatus(it.magnitude) }
+
+   }
+
+   private fun getTestStatus(magnitude: Int): TestStatus {
```
sksamuel: I noticed there is support for this in RunLineMarkerContributor.getTestStateIcon so we might not need to recreate (or if there's a reason we have, then maybe add a comment?)

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/LineMarkerUtils.kt
alfonsoristorato: >As an aside, is intellij generating java:suites for nested parents, or only for the class?

this applies to nested parents

As for the rest of your comment, I think you are describing what is in my PR already, but suggesting to remove the `isKotest61OrAbove` (which I did) - let me know if I misunderstood :) 
[simplify by reusing RunLineMarkerContributor.getTestStateIcon and r…](https://github.com/kotest/kotest/pull/5684/commits/83b4d218394fca00a1d92314635067bd99bd7d62)

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/LineMarkerUtils.kt
alfonsoristorato: absolutely right! [simplify by reusing RunLineMarkerContributor.getTestStateIcon and r…](https://github.com/kotest/kotest/pull/5684/commits/83b4d218394fca00a1d92314635067bd99bd7d62) now supports 5.9.1 too (for example)

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/linemarker/LineMarkerUtils.kt
alfonsoristorato: good shout - done :) [simplify by reusing RunLineMarkerContributor.getTestStateIcon and r…](https://github.com/kotest/kotest/pull/5684/commits/83b4d218394fca00a1d92314635067bd99bd7d62)

PR #5686, opened 2026-02-17
Fix ErrorCollector implementation for non-JVM and add JS tests

### LINKED ISSUE #5685: assertSoftly appears broken on JS targets since 6.0.0
**Kotest Assertions, versions tested: 6.0.0, 6.1.3**

`assertSoftly` no longer produces a `MultiAssertionError` on JS targets, like it did in Kotest 5.9.1.

**Reproducer:**
```kotlin
import io.kotest.assertions.MultiAssertionError
import io.kotest.assertions.throwables.shouldThrow
import io.kotest.matchers.string.shouldContain
import io.kotest.matchers.string.shouldNotEndWith
import kotlin.test.Test

class CommonTests {
    @Test
    fun testAssertSoftly() {
        shouldThrow<MultiAssertionError> {
            assertSoftly {
                "Expect failure 1!" shouldNotEndWith "!"
                "Expect failure 2!" shouldNotEndWith "!"
            }
        }.message shouldContain "The following 2 assertions failed:"
    }
}
```

**Output for `:jsNodeTest`:**
```
AssertionError: "Expect failure 1!" should not end with "!"
```

sksamuel:
Fixes #5685


kotest-assertions/kotest-assertions-shared/src/nonjvmMain/kotlin/io/kotest/matchers/errorCollector.nonjvm.kt
```diff
@@ -1,29 +1,3 @@
...
-   override fun clear() {   }
-
-   override fun pushClue(clue: Clue) {   }
-
-   override fun popClue() {   }
-
-   override fun clueContext(): List<Clue> = emptyList()
-}
-
-
```
Kantis, line 29: Seems a breaking change, should we keep it around as `@Deprecated` for a bit?

Kantis (approved):

kotest-assertions/kotest-assertions-shared/src/nonjvmMain/kotlin/io/kotest/matchers/errorCollector.nonjvm.kt
sksamuel: Hmm interesting point, I figured this was just an internal thing, but I suppose since it's public, people could be referencing it themselves. I'll deprecate it as you suggest.

sksamuel: done

Kantis (approved):

PR #5691, opened 2026-02-18
Fix multiple `--tests` issue

### LINKED ISSUE #5678: multiple `--tests` cannot find any tests from Kotest 6.1.x
6.1.0 -- 6.1.3

in Gradle specs, we can specify multiple `--tests` for test command.

> It is also possible to supply multiple --tests options, all of whose patterns will take effect

https://docs.gradle.org/current/userguide/java_testing.html

it was working well until Kotest 6.0.7, but from Kotest 6.1.0, it started to fail to find any tests.

reproducer:
[6.0.7](https://github.com/T45K/kotest-sandbox/tree/01361cbd08347e8f1765e0c50f92a399b3f15e5c)
[6.1.3](https://github.com/T45K/kotest-sandbox/tree/4ecb314a578620ee6281e5ad0169c522e2d114cd)

is there any regression?

T45K:
close #5678

**NOT TEST** (i'd like maintainers to do that 🙏 )

- ClassMethodNameFilterAdapter.adapt() was creating a separate DescriptorFilter for each extracted --tests pattern
- Since the engine applies DescriptorFilters with AND semantics, using multiple --tests arguments (e.g., --tests TestA --tests TestB) caused no tests to be found because each test had to match ALL patterns simultaneously
- Group all patterns into a single filter per type (regex-based and nested-test-based) so they are combined with OR logic within the filter

### DIFF
--- a/kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/gradle/ClassMethodNameFilterAdapter.kt
+++ b/kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/gradle/ClassMethodNameFilterAdapter.kt
@@ -54,7 +54,7 @@ internal object ClassMethodNameFilterAdapter {
          ClassMethodNameFilterUtils.reset(request.postFilters())
       }
 
-      val descriptionFilter = when {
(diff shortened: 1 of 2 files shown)


sksamuel (approved):

sksamuel:
nice!

PR #5692, opened 2026-02-19
Restore 5.x package names for error and assertion counters.

sksamuel:
They shouldn't have been renamed as they were public API. This restores the original names and adds type aliases for any library authors who flipped.

@OliverO2


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/aliases.kt
```diff
@@ -0,0 +1,21 @@
+package io.kotest.matchers
+
+import io.kotest.assertions.BasicErrorCollector
+
+@Deprecated("Use io.kotest.assertions.AssertionCounter. Will be removed in a future release.")
+typealias AssertionCounter = io.kotest.assertions.AssertionCounter
+
+@Deprecated("Use io.kotest.assertions.assertionCounter. Will be removed in a future release.")
+val assertionCounter = io.kotest.assertions.assertionCounter
```
OliverO2:
```suggestion
val assertionCounter: AssertionCounter get() = io.kotest.assertions.assertionCounter
```

Value may change per invocation as it depends on a `ThreadLocal`.

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/aliases.kt
```diff
@@ -0,0 +1,21 @@
...
+
+@Deprecated("Use io.kotest.assertions.ErrorCollectionMode. Will be removed in a future release.")
+typealias ErrorCollectionMode = io.kotest.assertions.ErrorCollectionMode
+
+@Deprecated("Use io.kotest.assertions.ErrorCollector. Will be removed in a future release.")
+typealias ErrorCollector = io.kotest.assertions.ErrorCollector
+
+@Deprecated("Use io.kotest.assertions.errorCollector. Will be removed in a future release.")
+val errorCollector = io.kotest.assertions.errorCollector
```
OliverO2:
```suggestion
val errorCollector: ErrorCollector get() = io.kotest.assertions.errorCollector
```

Value may change per invocation as it depends on a `ThreadLocal`.

kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/matchers/aliases.jvm.kt
```diff
@@ -0,0 +1,4 @@
+package io.kotest.matchers
+
+@Deprecated("Use io.kotest.assertions.errorCollectorContextElement. Will be removed in a future release.")
+val errorCollectorContextElement = io.kotest.assertions.errorCollectorContextElement
```
OliverO2:
```suggestion
val errorCollectorContextElement: CoroutineContext.Element get() = io.kotest.assertions.errorCollectorContextElement
```

Value may change per invocation as it depends on a `ThreadLocal`.

OliverO2 (changes requested):
Small changes to aliases for correctness. Otherwise looks good!

sksamuel:
thanks for feedback @OliverO2

OliverO2 (approved):

PR #5705, opened 2026-02-19
Refactor eventually to ensure wall-clock time usage

### LINKED ISSUE #5147: Ugrading Kotest Assertsions from 6.0.3 to 6.0.4 Failed My Tests

sksamuel:
Now eventually will detect virtual time and flip onto a wall-clock time instead.

Fixes #5147


kotest-common/src/commonMain/kotlin/io/kotest/common/withNonVirtualTimeout.kt
```diff
@@ -0,0 +1,71 @@
...
+   return if (currentCoroutineContext()[TestCoroutineScheduler] != null) {
+      withWallClockTimeout(timeout, block)
+   } else {
+      withTimeout(timeout, block)
+   }
+}
+
+// The implementation is adapted from the Turbine library
+// https://github.com/cashapp/turbine/blob/1.1.0/src/commonMain/kotlin/app/cash/turbine/channel.kt#L93
+private suspend fun <T> withWallClockTimeout(
```
AlexCue987: do we want to unit test this function explicitly?

sksamuel: Yes i can do that

sksamuel: test added

sksamuel:
@OliverO2 I want to be conscious of your users who might be using Kotest's assertions library. Is there anything in this that would give you concern? Essentially, updating eventually to check for TestDispatcher and if virtual time is present, then using a wall clock implementation (that I think you derived actually), since eventually using virtual time makes no sense.

OliverO2:
Thanks for pinging! Ensuring real time for `eventually`, even when on the `TestDispatcher`, makes sense. [I did that a while ago in a simpler implementation](https://github.com/OliverO2/kmp-e2e-stress-testing-example/blob/c674559f97d541686ee9ee47ebedcfba56b8b431/application/library-testing/src/commonMain/kotlin/library/testing/eventuallyTimestamped.kt), which did not care to interrupt a single block invocation. Your implementation offers more interruption points.

So all is good! Another question is whether it would make sense at all to use real-time `eventually` and a virtual time test dispatcher at the same time. But at least this PR makes it safer if you do.

AlexCue987 (approved):
LGTM

PR #5707, opened 2026-02-20
Fix failfast in one context incorrectly skipping sibling contexts 

### LINKED ISSUE #4944: Do not skip sibling context with failfast=true
Kotest: 6.0.0.M5

if a test fails inside the container, the next sibling containers are also skipped
```kotlin
list.forEach { item ->
  context("when adding a new $item").config(failfast = true) {
    it("shows modal")         { /* … */ }
    it("closes the modal")    { /* … */ }   // ← fails here and skips all remained items in the list, marks context as skipped
    it("shows the new item")  { /* … */ }
  }
}
```
```kotlin
// Expected flow:
┌ context A (failfast = true)
│ ├ it 1   ✅
│ ├ it 2   ❌ → stop only context A
│ └ it 3   ⏭
└ context B                 ✅ still runs
```

sksamuel:
Fixes #4944 (allegedly) 

When failfast was configured at the context level (context.config(failfast=true)), a failure inside that context set a global specContext.testFailed flag, causing any subsequent sibling context that also had failfast=true to be skipped.

Fix: replace the single testFailed boolean with scoped tracking. FailFastInterceptor now walks up each test's parent chain to find the nearest ancestor with config.failfast=true (the "scope"). Failures are recorded per-scope in SpecContext.failedScopes; only tests within the same scope are skipped. Spec- and project-level failfast continue to use the existing global testFailed flag.


Kantis (approved):
I'm guessing the entire failfast scope is canceled, no matter the nestedness of the error? i.e.:

```kotlin
context("context with fail fast enabled").config(failfast = true) {
         test("a") {} // pass
         test("b") {} // pass
         context("c") { 
            test("d") { error("boom") } 
         }
         test("e") { } // skipped
      }
```

Worth adding another test to cover that?

sksamuel:
> I'm guessing the entire failfast scope is canceled, no matter the nestedness of the error? i.e.:
> 
> ```kotlin
> context("context with fail fast enabled").config(failfast = true) {
>          test("a") {} // pass
>          test("b") {} // pass
>          context("c") { 
>             test("d") { error("boom") } 
>          }
>          test("e") { } // skipped
>       }
> ```
> 
> Worth adding another test to cover that?

Good idea, I'll add

sksamuel:
Closing in favor of #5774

PR #5713, opened 2026-02-21
Add withError builder to MatcherResultBuilder to rethrow a supplied Throwable

### LINKED ISSUE #5592: Ability to provide error in custom matcher
I am maintainer of [JsonUnit](https://github.com/lukas-krecan/JsonUnit) project that also has support for Kotest using custom [matcher](https://github.com/lukas-krecan/JsonUnit/blob/master/json-unit-kotest/src/main/kotlin/net/javacrumbs/jsonunit/kotest/Matchers.kt#L29).

I have noticed that in IntelliJ the difference report is broken, it contains

```
Expected :null
Actual   :null
```
due to this [branch](https://github.com/kotest/kotest/blob/62b9b44675269a8ee4d2c1145db91efcdee4003c/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/InvokeMatcher.kt#L29) being taken and the expected and actual value being null.

I tried to use `MatcherResultWithError` but it leads to duplicate intellijFormattedComparison string (one from JsonUnit  and one from [Kotest](https://github.com/kotest/kotest/blob/8f9403df5c2b02057787bf9d0502169bdef5a7e7/kotest-assertions/kotest-assertions-shared/src/jvmMain/kotlin/io/kotest/assertions/AssertionErrorBuilder.jvm.kt#L60))

As JsonUnit is generating its own exceptions, the easiest way to integrate would be to use something like [MatcherResultWithError](https://github.com/kotest/kotest/blob/62b9b44675269a8ee4d2c1145db91efcdee4003c/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/InvokeMatcher.kt#L36) so JsonUnit could propagate its own exception to Kotest. Would it be possible to make `MatcherResultWithError` public or expose an equivalent feature?

sksamuel:
## Summary

- Adds `ThrowableMatcherResult` to `Matcher.kt` alongside `SimpleMatcherResult` and `DiffableMatcherResult`. It carries a pre-built `Throwable` that should be rethrown as-is when an assertion fails.
- Adds a `withError(error: Throwable)` builder method to `MatcherResultBuilder`. When called, `build()` returns a `ThrowableMatcherResult` instead of a `SimpleMatcherResult` or `DiffableMatcherResult`.
- Updates `invokeMatcher` to handle `ThrowableMatcherResult` before the `AssertionErrorBuilder` path: the supplied error is passed directly to `errorCollector.collectOrThrow`, preserving its original type and message.
- 
Closes #5592

### DIFF
--- a/.github/workflows/master-eap.yml
+++ b/.github/workflows/master-eap.yml
@@ -3,12 +3,14 @@ name: master-eap
 on:
    schedule:
       -  cron: '0 10 * * *'
+   workflow_dispatch:
+
 
 permissions:
    contents: read
 
 env:
-   KOTLIN_OVERRIDE_VERSION: "2.3.20-Beta2"
+   KOTLIN_VERSION: "2.3.20-Beta2"
 
 jobs:
    check:
@@ -16,7 +18,6 @@ jobs:
       runs-on: ${{ matrix.os }}
       continue-on-error: true
       strategy:
-         fail-fast: false
          matrix:
             include:
                # JVM-only
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -98,12 +98,10 @@ dependencies {
    findProject(":kotest-extensions:kotest-extensions-wiremock")?.let { nmcpAggregation(it) }
 
    // Runners
-   findProject(":kotest-runner:kotest-runner-junit4")?.let { nmcpAggregation(it) }
-   findProject(":kotest-runner:kotest-runner-junit6")?.let { nmcpAggregation(it) }
-
-   // Runners
    nmcpAggregation(projects.kotestRunner.kotestRunnerJunitPlatform)
+   nmcpAggregation(projects.kotestRunner.kotestRunnerJunit4)
    nmcpAggregation(projects.kotestRunner.kotestRunnerJunit5)
+   nmcpAggregation(projects.kotestRunner.kotestRunnerJunit6)
 
 }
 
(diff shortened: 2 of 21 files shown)


Kantis (approved):
Breaking ABI for `MatcherResultBuilder`.. but I don't think it makes sense to use that as a library consumer? Should we mark MatcherResultBuilder as Deprecated and say that we intend to make it internal?

sksamuel:
> Breaking ABI for `MatcherResultBuilder`.. but I don't think it makes sense to use that as a library consumer? Should we mark MatcherResultBuilder as Deprecated and say that we intend to make it internal?

I think MatcherResultBuilder is the public entry point, so we want people to use it when integrating with Kotest. The ABI change is adding a new method, so should be fine, nothing was removed. What do you think ?

Kantis:
> > Breaking ABI for `MatcherResultBuilder`.. but I don't think it makes sense to use that as a library consumer? Should we mark MatcherResultBuilder as Deprecated and say that we intend to make it internal?
> 
> I think MatcherResultBuilder is the public entry point, so we want people to use it when integrating with Kotest. The ABI change is adding a new method, so should be fine, nothing was removed. What do you think ?

Ah, gotcha. A bit hard to know if the compiler will generate a method that maintains binary compatibility or not.

PR #5723, opened 2026-02-24
Add MultipleMatcherResult for combining multiple matcher results

### LINKED ISSUE #3343: idea: composite matchers, MatcherResultBuilder

sksamuel:
Introduces MultipleMatcherResult, a MatcherResult that aggregates a list of underlying results, passing only when all pass and surfacing each result's message (indexed) on failure. Adds MatcherResult.multiple() factory, uses it in areEqualRegexMatcher, and adds tests covering pass, single-failure, all-failure, and message format cases.

Also fixes bug with regex options matcher showing wrong negative error message.

Closes #3343


kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/MultipleMatcherResult.kt
```diff
@@ -0,0 +1,34 @@
...
+      return results.all { it.passed() }
+   }
+
+   override fun failureMessage(): String {
+      // when there's only one, just output it directly
+      if (results.count { !it.passed() } == 1) return results.first { !it.passed() }.failureMessage()
+      return buildString {
+         appendLine("Matcher failed due to:")
+         results.filterNot { it.passed() }.forEachIndexed { index, value ->
+            appendLine("$index) ${value.failureMessage()}")
```
AlexCue987, line 19: can `MultipleMatcherResult` be nested? if yes, do we want some indentation and/or appropriate indexing such as 2.1 or 2.3.1?

AlexCue987 (commented):
LGTM, one question

kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/MultipleMatcherResult.kt
sksamuel: Good question.

sksamuel: Do you think its worth going with this as it is for now and revisiting if we find the need to do complex nesting?

AlexCue987: > Do you think its worth going with this as it is for now and revisiting if we find the need to do complex nesting?

we can change it later if/when needed

AlexCue987 (approved):

Issue #5732, opened 2026-03-02
Compatibility broken in 6.1.4


orange-buffalo:
In kotest 6.1.4, `AssertionCounter` was moved from `io.kotest.matchers` ([6.1.3](https://github.com/kotest/kotest/blob/v6.1.3/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/matchers/AssertionCounter.kt)) to `io.kotest.assertions` ([6.1.4](https://github.com/kotest/kotest/blob/v6.1.4/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/AssertionCounter.kt)). This change is breaking compatibility of public API (at least, the interface is public and the docs do not say anything opposite) with a patch version increment. Consumers do not expect this. Just as one particular example, [lukas-krecan/JsonUnit ](https://github.com/lukas-krecan/JsonUnit) is broken due to this issue.

sksamuel:
We were trying to revert the backwards incompatible change introduced with 6.1. I thought the typealias would be enough to keep it happy from both sides, but nope. Let us rush out 6.1.5 with something that fixes this for everyone.

sksamuel:
Do you have an example as I cannot see AssertionCounter referenced in JsonUnit so is probably a transient issue. I guess you use JsonJunit and upgrade to 6.1.4 yourself ?

orange-buffalo:
No, unfortunately I do not have a simple reproducer. Your are right, I have the latest JsonUnit and explicit Kotest dependencies, which started to fail after Kotest upgrade to 6.1.4. The root cause analysis was performed by AI (https://github.com/orange-buffalo/simple-accounting/pull/2365) and, to be honest, I did not assess it deeply, just checked that the outcome was plausible.

Here is the failing build: https://scans.gradle.com/s/5i2wqztvum5u2 
The stacktrace is:
```
java.lang.NoClassDefFoundError: io/kotest/matchers/AssertionCounter_jvmKt

at net.javacrumbs.jsonunit.kotest.MatchersKt.shouldBeJsonNumber(Matchers.kt:130) |  
-- | --
  | at io.orangebuffalo.simpleaccounting.tests.infra.utils.JsonUnitExtensionsKt.shouldBeJsonInteger(JsonUnitExtensions.kt:7) |  
[...]

PR #5738, opened 2026-03-06
reintroduce io.kotest.matchers classes and interfaces

alfonsoristorato:
Unable to re-introduce:
- ErrorCollectionMode: this is an enum and it cannot implement from another ENUM 

tries to address https://github.com/kotest/kotest/issues/5732

### DIFF
--- a/kotest-assertions/kotest-assertions-shared/api/kotest-assertions-shared.api
+++ b/kotest-assertions/kotest-assertions-shared/api/kotest-assertions-shared.api
@@ -186,6 +186,18 @@ public final class io/kotest/matchers/Aliases_jvmKt {
 	public static final fun getErrorCollectorContextElement ()Lkotlin/coroutines/CoroutineContext$Element;
 }
 
+public abstract interface class io/kotest/matchers/AssertionCounter : io/kotest/assertions/AssertionCounter {
+}
+
+public final class io/kotest/matchers/AssertionCounter$DefaultImpls {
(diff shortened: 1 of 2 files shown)


sksamuel:
I was thinking of something similar myself but wasn't sure what would happen to consumers of this.
The val's don't implement the new interfaces, so not sure what would happen if code links to the wrong one?

alfonsoristorato:
> I was thinking of something similar myself but wasn't sure what would happen to consumers of this. The val's don't implement the new interfaces, so not sure what would happen if code links to the wrong one?

Do you mean `val assertionCounter` and `val errorCollector`?
The interface and classes seem to be usable withing a Java class:
image
image

sksamuel:
I mean if the vals in the same file are pointing to implementations that only implement the old interface...

I wonder if we should just bite the bullet and accept I fucked up here, introduce an official way to integrate with assertion counter and error collector and release 6.2.0 ? No user should ever see this, it's only for anyone who wants to extend Kotest assertions. Thoughts?

alfonsoristorato:
> I mean if the vals in the same file are pointing to implementations that only implement the old interface...
> 
> I wonder if we should just bite the bullet and accept I fucked up here, introduce an official way to integrate with assertion counter and error collector and release 6.2.0 ? No user should ever see this, it's only for anyone who wants to extend Kotest assertions. Thoughts?

sounds good to me! closing the PR and until 6.2.0 we can at least still keep the aliases as they are

PR #5743, opened 2026-03-07
re-enable location hints for TeamCity

### LINKED ISSUE #5676: Feature request: run selected test from test results window
With junit I can click on some test method and run just that one test:

Image

With kotest I don't see that possibility:

Image

Image

alfonsoristorato:
fixes https://github.com/kotest/kotest/issues/5676
Re-enables locationHint in TeamCity - this is not needed by the `gradle` producers, but still needed by the `idea`'s ones


alfonsoristorato:
~WIP - I am still figuring out if this location hint is needed somewhere else~
confirmed that locationHint was present only in:
- specStarted
- testStarted
- testIgnored

this PR re-introduces it for those use cases

sksamuel (approved):

sksamuel:
Probably should check we still have the hint parser in the intellij plugin

alfonsoristorato:
> Probably should check we still have the hint parser in the intellij plugin

We still do, I built this locally and was able to see the menu in our maven example

sksamuel:
Shall we do 6.1.5 ?

alfonsoristorato:
> Shall we do 6.1.5 ?

Let's go!

atomcat1978:
Hi guys,

Should it work all right in the intelliJ plugin now? I still get the same result on 6.1.5.

alfonsoristorato:
> Hi guys,
> 
> Should it work all right in the intelliJ plugin now? I still get the same result on 6.1.5.

Hey there, it should work with the plugin and version 6.1.5, but if it doesn't are you able to provide in the old issue a reproducer and also let us know if you can get the maven examples to work properly?

PR #5747, opened 2026-03-13
Fix power assert applied unconditionally with wrong dependency scope

### LINKED ISSUE #5745: Unsupported API dependency types in test source sets
Issue started in 6.1.5, not present in 6.1.4

Issue can be seen in my project build logs here: https://github.com/teobaranga/monica/actions/runs/22831461930/job/66220291199#step:9:33

Full message:
> w: Unsupported API dependency types in test source sets
API dependency types are used in test source sets
Dependencies:
    - io.kotest:kotest-assertions-core:6.1.5 (source sets: iosArm64Test, iosSimulatorArm64Test)
>
> Adding API dependency types to test source sets is not supported and will removed in a future version of Kotlin.
>
>API dependencies are transitively exposed to consumers, but test source sets should not be consumable.
Solutions:
 • Replace API dependency types in test source sets with implementation dependencies.
 • For Kotlin/JVM projects, consider using Test Fixtures https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures
 • For Test Fixtures support in non-Kotlin/JVM projects, please add your use-case to https://youtrack.jetbrains.com/issue/KT-63142

Just digging a bit through the source code, it seems like:
- the power assert library is added to the `commonTestApi` and the fix suggests it should be in `commonTestImplementation` instead: https://github.com/kotest/kotest/blob/e37a1ffa1300b5261a3904aae74486a4ae6b666e/kotest-framework/kotest-framework-plugin-gradle/src/main/kotlin/io/kotest/framework/gradle/KotestPlugin.kt#L124
- the power assert library is added unconditionally as the `enablePowerAssert` property is not used

sksamuel:
## Summary

Fixes #5745 — `kotest-assertions-core` was being added as an `api` dependency in test source sets, causing a Kotlin warning about unsupported API dependency types in test source sets.

Three bugs were fixed in `configurePowerAssert`:
- The `enablePowerAssert` property was passed in but never checked — power assert was always applied. Added an early return when `enablePowerAssert` is `false` (the default).
- The call to `configurePowerAssert` was moved into `project.afterEvaluate` so the property value is finalized before it is read.
- Dependency scope changed from `commonTestApi` to `commonTestImplementation` to avoid the Kotlin warning about API dependencies in test source sets.

## Test plan
- [x] Apply the Kotest Gradle plugin without setting `enablePowerAssert` — verify the power assert plugin is NOT applied and `kotest-assertions-core` is NOT added as a dependency.
- [x] Set `kotest { enablePowerAssert = true }` — verify the power assert plugin is applied and `kotest-assertions-core` is added as a `commonTestImplementation` dependency (no warning about API dependency types in test source sets).


alfonsoristorato (approved):
We go 6.1.7?

sksamuel:
> We go 6.1.7?

Yep

PR #5748, opened 2026-03-14
Precise Gradle configuration needed for pitest extension

szpak:
This is the gradle-pitest-plugin author here. Analyzing this [issue](https://github.com/szpak/gradle-pitest-plugin/issues/396), I've spotted a sub optimal kotest pitest extension configuration. The tests itself don't need it, only the Pitest execution triggered by the gradle-pitest-plugin (and `pitest()` is enought - as described in the [plugin documentation](https://github.com/szpak/gradle-pitest-plugin#generic-plugin-support-also-junit-5-in-gradle-pitest-plugin-147)).

Btw, creating a draft, as I create a PR from a browser with "one file edit" and I cannot easily change also all of the documentation versions. It would be great to have it covered by you. Alternatively, I can "fix" it after the weekend.


sksamuel:
I can merge this then apply it to all other doc versions.
Thanks for the pr!

sksamuel (approved):

PR #5754, opened 2026-03-17
AI Agent kotest skill

alfonsoristorato:
This PR is a baseline and is to be used by all kotest devs to push whatever they think should be in this skill.
Once agreed upon, we will propose the skill to this repo https://github.com/Kotlin/kotlin-agent-skills
For reference:
- we are trying to reach this level of detail: https://github.com/Kotlin/kotlin-agent-skills/tree/main/skills/kotlin-tooling-agp9-migration
- we must follow this specification: https://agentskills.io/specification

Guide:
- if pushing to this pr, create your own dir within `skill` dir, so that we can combine all dumps and decide what to use


sksamuel:
What do you think about merging this in, and the rest of us making PRs to add more, and you act as the gatekeeper of those PRs, and once we're all happy we submit to Jetbrains? Might be easier than creating separate skills?

alfonsoristorato:
Let's go, bearing in mind that this has not received human review whatsoever. But at least we have a starting point 😃

sksamuel (approved):

PR #5756, opened 2026-03-18
Make shouldHaveSingleElement and shouldNotHaveSingleElement chainable

### LINKED ISSUE #5755: Make shouldHaveSingleElement and friends chainable
shouldHaveSingleElement and friends return Unit rather than matcher, so I can't do things like:
```
myCollection.shouldhaveSize(2)
    .shouldHaveSingleElement{ it is OneType }
    .shouldHaveSingleElement{ it is OtherType }
```

This seems unnecessary, as the underlying singleElement function does return its matcher.

I think there are a few other assertions with similar limitations lying around, but can' recall which ones right now.
Might be worth scanning for other candidates too if this is fixed. 
Changing away from Unit return is thankfully always backward-compatible.

PreAgile:
Closes #5755

## What

Some overloads of `shouldHaveSingleElement` and `shouldNotHaveSingleElement`
returned `Unit`, which prevented fluent chaining. The `Iterable` overloads already
returned the receiver, and this change aligns the remaining `Collection`, `Array`,
and `Sequence` overloads with that behavior.

## Scope

This PR is intentionally limited to the single-element assertion family:
- `singleElement.kt`
- the sequence equivalent in `matchers.kt`

A broader sweep of other assertion functions is intentionally left out to keep
the change small and reviewable.

## Changes

- `singleElement.kt`: 6 functions now return the receiver instead of `Unit`
- `matchers.kt` (sequences): 2 functions now return the receiver instead of `Unit`
- API dump updated accordingly
- Added chaining regression tests for `Collection` and `Sequence`

## Compatibility

Existing call sites that ignore the return value continue to work.


sksamuel:
Looks great. Can we remove the research files, otherwise ready to merge

sksamuel:
think we need to run apiDump too and commit the change

AlexCue987 (changes requested):
LGTM, almost ready to go.

let's remove all *.md files and explicitly test filtering

PreAgile:
I removed the contributor research docs from this PR. Thanks for the review and the quick feedback.

sksamuel:
Needs gradlew apiDump

PreAgile:
Thanks for the heads-up. I followed up on the apiDump request and pushed a fix in `4790c14ad`.

  While running `./gradlew apiDump`, I found a related compile issue in the `kotest-assertions-arrow`
  `NonEmptyList` wrappers caused by the new chainable return type. I aligned those wrappers, updated
  the API dump, and added a small regression test for chaining.

sksamuel (approved):

sksamuel:
> LGTM, almost ready to go.
> 
> let's remove all *.md files and explicitly test filtering

Your requests have been met so will free this up.

PR #5757, opened 2026-03-18
ci: add workflow to auto-update API dumps on PR failures

Kantis:
When the pr-main validate-api job fails due to outdated .api dump files, this workflow automatically runs apiDump and commits the updated files back to the PR branch, saving contributors from having to do this manually.


sksamuel:
Awesome

sksamuel (approved):

PR #5759, opened 2026-03-19
Add LazyJvmAssertion to avoid stacktraces in inspectors

### LINKED ISSUE #5728: Performance issues with "shouldForNone" matcher
Which version of Kotest are you using: **6.1.4**

When debugging some performance issues in my algorithm, I've noticed that the Kotest code was the biggest performance bottleneck itself. This was because I was using the "shouldForNone" matcher, which performs way worse than the "shouldForAll" one. I believe this is closely related to #5245. In my case, I was able to reduce my test time from over 1s to 100ms just by replacing this single operator.

How to reproduce this:
```kotlin
val listOfSets = listOf(
   setOf(1, 2, 3),
   setOf(4, 5, 6),
   setOf(7, 8, 9),
   setOf(99)
)
```

This works fine:
```kotlin
listOfSets.shouldForAll { it.shouldNotContain(0) }
```

This has a big performance penalty:
```kotlin
listOfSets.shouldForNone { it.shouldContain(0) }
```

sksamuel:
Fixes #5728


alfonsoristorato (approved):
nice

PR #5761, opened 2026-03-19
Fix Slow Arb.generate for large objects

### LINKED ISSUE #4932: Property Testing: Slow Arb.generate for large objects
        <dependency>
            <groupId>io.kotest</groupId>
            <artifactId>kotest-property</artifactId>
            <scope>test</scope>
            <version>6.0.0.M5</version>
        </dependency>

In my use case I'm generating large and deep objects representing deserialized Avro Records (thanks for #4907 btw).
Generating 1000 of these like with the default settings of checkAll() takes up to a minute on my M2 Max MacbookPro.

Running multiple property tests against many scenarios is not viable at the moment. I could reduce the sample size but rapid execution of large samplesizes is the big appeal of kotest-property for me.

I've done some quick and dirty profiling and found SingleShotArbContinuation in builders.kt as the likely culprit. As I read it this implements some kind of lazy continuation of a stream of coroutine results - sounds a little like a Flow<T> before there was a Flow<T>.

My gut feeling tells me this part of the code could benefit hugely from a migration to actual Flows but I might be mistaken. Happy to provide more information and test any suggestions.

sksamuel:
Fixes #4932


kotest-property/src/jvmTest/kotlin/com/sksamuel/kotest/property/arbitrary/ArbitraryPerformanceTest.kt
```diff
@@ -0,0 +1,44 @@
...
+   data class L2(val m: L3, val n: L3, val o: L3)
+   data class L1(val i: L2, val j: L2)
+
+   val l5Arb = arbitrary { L5(Arb.int().bind(), Arb.string(1..8).bind()) }
+   val l4Arb = arbitrary { L4(l5Arb.bind(), l5Arb.bind(), l5Arb.bind()) }
+   val l3Arb = arbitrary { L3(l4Arb.bind(), l4Arb.bind()) }
+   val l2Arb = arbitrary { L2(l3Arb.bind(), l3Arb.bind(), l3Arb.bind()) }
+   val l1Arb = arbitrary { L1(l2Arb.bind(), l2Arb.bind()) }
+
+   test("generating 1000 samples of a 5-level nested arbitrary completes in under 10 seconds") {
```
AlexCue987: this might be fragile, should we wrap it in eventually?

sksamuel: good spot, I've added 5 retries which will do the trick

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/builders.kt
```diff
@@ -218,40 +219,29 @@ fun <A> arbitraryBuilder(
...
-   override fun edgecase(rs: RandomSource): Sample<A>? = singleShotArb(SingleShotGenerationMode.Edgecase, rs).edgecase(rs)
-   override fun sample(rs: RandomSource): Sample<A> = singleShotArb(SingleShotGenerationMode.Sample, rs).sample(rs)
    override val classifier: Classifier<out A>? = classifier
 
+   override fun sample(rs: RandomSource): Sample<A> {
+      val value = runBuilderFn(SingleShotGenerationMode.Sample, rs)
+      return if (shrinker == null) Sample(value) else sampleOf(value, shrinker)
+   }
+
+   override fun edgecase(rs: RandomSource): Sample<A>? {
```
alfonsoristorato: we should not need the ? here

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/builders.kt
```diff
@@ -268,40 +258,21 @@ suspend fun <A> suspendArbitraryBuilder(
...
-
-         return with(suspendableContinuation) {
-            this@with.createSingleShotArb()
-         }
+      override fun sample(rs: RandomSource): Sample<A> {
+         val value = runBuilderFn(SingleShotGenerationMode.Sample, rs)
+         return if (shrinker == null) Sample(value) else sampleOf(value, shrinker)
+      }
+
+      override fun edgecase(rs: RandomSource): Sample<A>? {
```
alfonsoristorato: nor here

alfonsoristorato (commented):
looks good - happy to approve once we remove those ?

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/builders.kt
sksamuel: done

kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/builders.kt
sksamuel: done

alfonsoristorato (approved):

PR #5762, opened 2026-03-19
Add KOTEST_TEST_ENABLED_OVERRIDE env var to bypass disabled tests

### LINKED ISSUE #4882: intellij-plugin: run disabled test manually
With JUnit, I could tag a test class as disabled, and it would
- not be run at build
- not be run when running all tests in a folder from IntelliJ
- still run if I triggered that specific test from IntelliJ
- still let me run single test cases explicitly from IntelliJ
- Similarly, a test case tagged as disabled would be still be run if triggered explicitly from IntelliJ, but not when running the whole test class 

This is very useful for exploratory or system-integration tests that I only want to run manually,
and in general for running work-in-progress tests without fear of accidentally checking in any temporary reenabling.

With kotest, all types of disabling seem to apply to all types of manual runs from within intellij.

I could resort to a custom Condition requiring an env-variable or something, which I would have to add to my run-configs, but it seems needlessly cumbersome.
I see there is a bundled "NotCICondition", but this would pass for all test runs from inside IntelliJ, and to me it is fairly important that the disabled tests are NOT run when I use "All in folder"

### Proposal:
For me the ideal solution would be to mimic the JUnit behavior as closely as possible out-of-the-box, with no further overhead for users.

If that is too surprising behavior for existing users, a passable second choice would be if the plugin could set an environment variable or something *only* when triggering a specific spec or case from IntelliJ, so I could use a custom (or preferably bundled) condition

This would lose som granularity, though: running a whole spec manually would apply any disabled sub-sections, which may itself be too unexpected for users.

Possibly a more granular Condition-based solution is feasible - I don't know the innards of kotest or the plugin well enough.

sksamuel:
When the system property or env var KOTEST_TEST_ENABLED_OVERRIDE is set to "true", relevant TestEnabledExtension and SpecRefEnabledExtension implementations now return enabled immediately, bypassing any annotations (@Ignored, @DisabledIf, @EnabledIf, @RequiresTag), tags, or config flags that would otherwise disable the test or spec.

The IntelliJ plugin Gradle run configuration producers now set this env var when launching a specific test from the IDE, ensuring that tests marked as disabled can still be executed when explicitly invoked.

Closes #4882


sksamuel:
@Kantis it worked :)

alfonsoristorato (approved):
this works for the use-cases described - do we want to keep the issue open to iterate in a second PR that handles this?
Also, nice work, I was looking forward to have this in :) 
Image

sksamuel:
> this works for the use-cases described - do we want to keep the issue open to iterate in a second PR that handles this? Also, nice work, I was looking forward to have this in :) Image

I'll update this pr to drop the disabled gutter annotator

sksamuel:
disabled gutter annotator has gone

jlous:
Looking forward to trying this out in 6.2.
A couple of questions:

I see gradle is mentioned. My project uses maven.  
- Will it still work there?
- Will it work without idea delegating test execution?

sksamuel:
> Looking forward to trying this out in 6.2. A couple of questions:
> 
> I see gradle is mentioned. My project uses maven.
> 
> * Will it still work there?
> * Will it work without idea delegating test execution?

This is for the gradle kotest plugin.
If you are using maven, then the way you invoke tests wouldn't be affected.

PR #5763, opened 2026-03-19
Publish IntelliJ plugin distribution zip to GitHub Packages on master

### LINKED ISSUE #5354: Feature request: Add releases page with binaries
Sometimes it's not possible to install plugin using IntelliJ plugin marketplace, but the releases section to download binary and install from it is also missing.
Could you please add Releases section?

sksamuel:
## Summary

- Adds `buildPlugin` step to the master IntelliJ plugin workflow so the distribution ZIP is built on every push to master
- Publishes each product-specific ZIP (`IC-251`, `IC-252`, `IC-253`, `IC-261`) to the GitHub Packages Maven registry as `io.kotest:kotest-intellij-plugin-<PRODUCT>:<version>`
- Adds `packages: write` permission to the job
- Configures `actions/setup-java` with `server-id: github` so Maven authenticates to GitHub Packages via `GITHUB_TOKEN`

Packages are published to `https://maven.pkg.github.com/kotest/kotest` and versioned as `6.1.<GITHUB_RUN_NUMBER>`.

Closes #5354 


.github/workflows/master_intellij_plugin.yml
```diff
@@ -36,6 +43,24 @@ jobs:
...
+            env:
+               PRODUCT_NAME: ${{ matrix.product }}
+
+         -  name: Publish distribution zip to GitHub Packages
+            run: |
+               mvn --batch-mode deploy:deploy-file \
+                  -Dfile=$(ls kotest-intellij-plugin/build/distributions/*.zip) \
+                  -DgroupId=io.kotest \
+                  -DartifactId=kotest-intellij-plugin-${{ matrix.product }} \
+                  -Dversion=6.1.${{ github.run_number }} \
```
alfonsoristorato: wonder if we can get claude to come up with a different strategy here to get the 6.1, as otherwise we'd have to remember every time to update this. Happy to approve if it's not worth it

sksamuel: I've added an org wide var called KOTEST_RELEASE_MAJOR_MINOR which is set to 6.1
Anyone should be able to edit it.
We can then make further PRs to use this in all the workflows / CI builds and bump it when we move to 6.2 etc.

alfonsoristorato (approved):

.github/workflows/master_intellij_plugin.yml
sksamuel: I updated the other workflows now

PR #5769, opened 2026-03-19
Support @Suppress for duplicated test name and focus-in-nested-test warnings

### LINKED ISSUE #5348: Suppress "Duplicated test name"
I have a use case where I create tests when some conditions are satisfied. And for different disjoint conditions, I create tests with same names. I know that it is risky, but the conditions are disjoint so there won't be any two tests with the same name in the same test context. So I would like to suppress this warning. With `@Suppress` annotation or in any other way.

sksamuel:
## Summary

- Adds a `ProblemGroup` with a stable ID to the `DuplicatedTestNameAnnotator` and `FocusInNestedTestAnnotator` warnings
- IntelliJ's `KotlinInspectionSuppressor` (registered by the Kotlin plugin) recognizes these IDs and honors `@Suppress` annotations

After this change, users can suppress warnings at test, function, class, or file level:

```kotlin
@Suppress("DuplicatedTestName")
test("my duplicate test name") { ... }
```

```kotlin
@Suppress("FocusInNestedTest")
test("f: focus in nested") { ... }
```

Fixes https://github.com/kotest/kotest/issues/5348

## Test plan
- [ ] Manually verify that `@Suppress("DuplicatedTestName")` suppresses the "Duplicated test name" warning in the IDE
- [ ] Manually verify that `@Suppress("FocusInNestedTest")` suppresses the "Focus only works on top level tests" warning

### DIFF
--- a/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/AnnotationHolderUtil.kt
+++ b/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/AnnotationHolderUtil.kt
@@ -1,9 +1,13 @@
 package io.kotest.plugin.intellij
 
+import com.intellij.codeInspection.ProblemGroup
 import com.intellij.lang.annotation.AnnotationHolder
 import com.intellij.lang.annotation.HighlightSeverity
 import com.intellij.psi.PsiElement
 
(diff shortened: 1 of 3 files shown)


alfonsoristorato (approved):
we'll need some docs for the plugin - I am sure you have some notes somewhere for it :)

PR #5777, opened 2026-03-20
Add primitive array support for containExactly matcher (#4354)

### LINKED ISSUE #4354: Collection matchers should work on primitive arrays

sksamuel:
## Summary

- Adds `shouldContainExactly` and `shouldNotContainExactly` extension functions for all 8 primitive array types: `BooleanArray`, `ByteArray`, `ShortArray`, `CharArray`, `IntArray`, `LongArray`, `FloatArray`, `DoubleArray`
- Each type gets both an infix array-to-array form and a vararg form, delegating to the existing `List`-based `containExactly` matcher via `asList()`
- Adds tests in `ShouldContainExactlyTest` for each primitive type covering: correct order passes, wrong order fails, `shouldNotContainExactly` passes when order differs

Fixes #4354 

## Test plan

- [ ] `intArrayOf(1,2,3).shouldContainExactly(intArrayOf(1,2,3))` passes
- [ ] `intArrayOf(1,2,3).shouldContainExactly(1, 2, 3)` passes
- [ ] `intArrayOf(3,2,1).shouldContainExactly(1,2,3)` throws `AssertionError`
- [ ] `intArrayOf(1,2,3).shouldNotContainExactly(3, 2, 1)` passes
- [ ] Same pattern verified for all 8 primitive array types
- [ ] Run `./gradlew :kotest-assertions:kotest-assertions-core:jvmTest --tests "com.sksamuel.kotest.matchers.collections.ShouldContainExactlyTest"`

### DIFF
--- a/kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
+++ b/kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
@@ -1266,15 +1266,47 @@ public final class io/kotest/matchers/collections/ContainExactlyKt {
 	public static final fun shouldContainExactly (Ljava/lang/Iterable;[Ljava/lang/Object;)V
 	public static final fun shouldContainExactly (Ljava/util/Collection;Ljava/util/Collection;)V
 	public static final fun shouldContainExactly (Ljava/util/Collection;[Ljava/lang/Object;)V
+	public static final fun shouldContainExactly ([B[B)[B
+	public static final fun shouldContainExactly ([C[C)[C
+	public static final fun shouldContainExactly ([D[D)[D
+	public static final fun shouldContainExactly ([F[F)[F
+	public static final fun shouldContainExactly ([I[I)[I
+	public static final fun shouldContainExactly ([J[J)[J
 	public static final fun shouldContainExactly ([Ljava/lang/Object;[Ljava/lang/Object;)V
+	public static final fun shouldContainExactly ([S[S)[S
+	public static final fun shouldContainExactly ([Z[Z)[Z
 	public static final fun shouldContainExactly_array ([Ljava/lang/Object;[Ljava/lang/Object;)V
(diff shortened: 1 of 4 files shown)


Kantis (approved):

sksamuel:
Wonder why the API dump isn't triggering

PR #5789, opened 2026-03-23
Add type-safe shouldEq / shouldNotEq assertions

### LINKED ISSUE #5589: add shouldEq with only input types
https://github.com/kotest/kotest/discussions/5279

PreAgile:
Closes #5589

## Summary

- Add `shouldEq` and `shouldNotEq` as type-safe alternatives to `shouldBe`
- Uses `@OnlyInputTypes` shadowing to restrict type inference, catching type mismatches at compile time
- Follows the gradual approach discussed in the issue — just `shouldEq` / `shouldNotEq` for now, derivatives can come later

## Changes

- Shadow `kotlin.internal.OnlyInputTypes` annotation in `kotest-assertions-core`
- Add `-Xallow-kotlin-package` compiler flag to `kotest-assertions-core` build
- Implement `shouldEq` / `shouldNotEq` in `io.kotest.matchers.equals` (commonMain)
- Add tests covering equality, inequality, chaining, custom equals, nullable types, and error messages

## Example

```kotlin
val count: Int = 5
count shouldEq 5       // compiles — same type
count shouldEq "5"     // compile error — Int vs String
```

## Design decisions

- **Shadowing over error suppression**: as discussed in the issue, shadowing degrades gracefully if a future Kotlin version stops recognising the annotation
- **Reuses `beEqual()` matcher**: consistent error messages with `shouldBeEqual`
- **Scope**: only `shouldEq` / `shouldNotEq` — no changes to `shouldBe` or other matchers


sksamuel:
Is  compilerOptions {
      freeCompilerArgs.add("-Xallow-kotlin-package")
   }
required by end users too, or just the kotest project ?

sksamuel:
can you run ./gradlew apiDump and commit ?

PreAgile:
> Is compilerOptions { freeCompilerArgs.add("-Xallow-kotlin-package") } required by end users too, or just the kotest project ?

No, it's only required for building the kotest project itself. The @OnlyInputTypes annotation is   
compiled into the bytecode and shipped as part of the artifact, so end users don't need to add any   
compiler flags 
they just use shouldEq / shouldNotEq as normal

> can you run ./gradlew apiDump and commit ?

Done! Ran ./gradlew apiDump and committed the updated .api file.

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/equals/shouldEq.kt
```diff
@@ -0,0 +1,36 @@
...
+ * Unlike [io.kotest.matchers.shouldBe], this function restricts the type parameter
+ * so that only values of the same type can be compared.
+ *
+ * ```
+ * 1 shouldEq 1       // compiles
+ * 1 shouldEq "1"     // compile error — Int vs String
+ * ```
+ */
+infix fun <@kotlin.internal.OnlyInputTypes T> T.shouldEq(expected: T): T {
+   this should beEqual(expected)
```
sksamuel: Do we want to use equality instead of the Eq typeclass ?
This PR adds the type safety we want, but given the name, it would indicate it's using the Eq functionality, and not equals.  Perhaps we need another name?
I want to be sure we get this right before we merge
@AlexCue987 any thoughts?

PreAgile: My intent with `shouldEq` was "should equal", not the `Eq` typeclass, but I agree the name is ambiguous in Kotest.

The important part for me is the type-safe equality check, not the exact naming, so I’m happy to rework it. Since `shouldBeEqual` / `shouldNotBeEqual` already represent `equals`-based checks, I can align this PR around those instead of introducing `shouldEq` / `shouldNotEq`.

AlexCue987: > Do we want to use equality instead of the Eq typeclass ? This PR adds the type safety we want, but given the name, it would indicate it's using the Eq functionality, and not equals. Perhaps we need another name? I want to be sure we get this right before we merge @AlexCue987 any thoughts?

still thinking about it

sksamuel: > My intent with `shouldEq` was "should equal", not the `Eq` typeclass, but I agree the name is ambiguous in Kotest.
> 
> The important part for me is the type-safe equality check, not the exact naming, so I’m happy to rework it. Since `shouldBeEqual` / `shouldNotBeEqual` already represent `equals`-based checks, I can align this PR around those instead of introducing `shouldEq` / `shouldNotEq`.

Yes that makes sense. You want the type safety, not the Eq typeclass, so you don't care too much what it's called as long as the compiler is doing what you expect.

In that case, my vote would be that we _don't_ call it shouldEq because that should be reserved for our Eq machinery. shouldBeEqual is a good name, but then we can't just break that for existing users so that's out.

I'm not precious about the name personally, as long as its clear to the user and isn't confusing with Eq vs equals. These things of things are sticky so we have to be sure we get it right first time !

So... the hardest thing in software, a naming conundrum. Perhaps shouldEqual (not shouldBeEqual). Once @AlexCue987 's has had time to weigh in, we can make a consensus agreement. I think we can target this for 6.2 anyway, which I'd like to release in the next 2 weeks (we have some amazing new IDE work coming for that).

alfonsoristorato: If I can chip in, when I made similar work for shouldBeEqualUsingFields (if that is the name I used, pretty sure) we used the 'usingTypes'  suffix
 Maybe we can reuse that and call this 'shouldBeEqualUsingType'?

Kantis:
I still think it would make more sense to add this to `shouldBe`

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/equals/shouldEq.kt
AlexCue987: > Do we want to use equality instead of the Eq typeclass ? This PR adds the type safety we want, but given the name, it would indicate it's using the Eq functionality, and not equals. Perhaps we need another name? I want to be sure we get this right before we merge @AlexCue987 any thoughts?

I concur we should come up with a different name. Not sure which name describes this awesome improvement well enough.

AlexCue987:
> I still think it would make more sense to add this to `shouldBe`

we are all different. personally I feel it's way too difficult to change that Swiss Army knife entitled `shouldBe`. It is complicated, it has so many edge cases etc. this is why I'm usually building specific tools instead - they do just one thing but try to do it better than `shouldBe`. but that's just me

sksamuel:
> I still think it would make more sense to add this to `shouldBe`

It would make more sense to have developed shouldBe that way in the first place, absolutely. But I don't think it's feasible to change now, 10 years worth of tests would be broken across everyone's projects.

Best we can do is a new name. We could introduce the same functionality to a matcher called shouldEq and promote that as the successor to shouldBe (not deprecate shouldBe, but just promote shouldEq as a better alternative)

For this PR, some ideas:

- shouldEqual
- shouldEquals
- shouldBeEqualTo
- shouldEqualAs

PreAgile:
I've been thinking about the naming options, and I'm leaning toward `shouldEqual` / `shouldNotEqual` myself.
                                                                                                                   
It avoids the Eq typeclass confusion, and it's short enough that people would actually want to use it daily — something like `shouldBeEqualUsingType` feels a bit heavy for that purpose. Plus it sits nicely next to the existing `shouldBeEqual` without colliding.                                                                      
                                                                                                                   
Happy to rename and push once we're aligned

AlexCue987:
> It would make more sense to have developed shouldBe that way in the first place, absolutely. But I don't think it's feasible to change now, 10 years worth of tests would be broken across everyone's projects.

this +100

AlexCue987:
> * shouldEqual

the shortest and straight to the point.

sksamuel:
I think we have a consensus, @Kantis you all good if we go with shouldEqual ?

Kantis:
Of course, go ahead. Can't please everyone 😁 

Promoting it as a replacement over time sounds like a good idea. 

I was mostly concerned with the "main assertion" not delivering what people expect, and introducing a growing number of "A should be the same thing as B" type of assertions.

sksamuel (approved):

PR #5794, opened 2026-03-24
Add robolectric tests module

sksamuel:


sksamuel:
Pushed an implementation of `RobolectricFunSpec`/`RobolectricExtension`/`KotestRobolectricSandbox` that gets the failing `compileAndroidHostTest` step compiling and both the JUnit4 reference test and the new Kotest spec passing in the same JVM. CI didn't appear to auto-trigger on the push — could you re-run the workflow if needed?

sksamuel:
Moving to https://github.com/kotest/kotest/pull/6011

PR #5795, opened 2026-03-24
Add custom Json parser support to JSON matchers

### LINKED ISSUE #4601: Json Matchers Supporting Custom `Json`
It would be nice if there were additional json matchers allowing a kotlinx.serialization `Json` to be passed to them.

For example:
```kotlin
fun beValidJson(parser: Json) = object : Matcher<String?> {
   override fun test(value: String?): MatcherResult {
      return try {
         value?.let(parser::parseToJsonElement)
         MatcherResult(
            true,
            { "expected: actual json to be valid json: $value" },
            { "expected: actual json to be invalid json: $value" }
         )
      } catch (ex: Exception) {
         MatcherResult(
            false,
            { "expected: actual json to be valid json: $value" },
            { "expected: actual json to be invalid json: $value" }
         )
      }
   }
}

fun String.shouldBeValidJson(parser: Json): String {
   this should beValidJson(parser)
   return this
}

fun String.shouldNotBeValidJson(parser: Json): String {
   this shouldNot beValidJson()
   return this
}
```

An example usecase of this is if I want to validate that the specified json is valid when `isLenient` is set to `true`.
Another usecase would be providing a custom `serializersModule`.

PreAgile:
Closes #4601

## Summary

All JSON matchers in `kotest-assertions-json` use a hardcoded `pretty` Json instance, making it impossible to validate JSON with custom parsing configurations (lenient mode, comments, trailing commas, custom `SerializersModule`).

This PR adds overloaded variants of all JSON matcher functions that accept a custom `kotlinx.serialization.json.Json` instance:

- `beValidJson(parser)`, `matchJson(expected, parser)`, `beJsonType(kClass, parser)`
- `beJsonArray(parser)`, `beJsonObject(parser)`
- `equalJson(expected, options, parser)`
- All `String`, `Path`, `File` extension functions (`shouldBeValidJson(parser)`, `shouldEqualJson(expected, parser)`, etc.)
- `matchJsonResource(resource, parser)`, `shouldMatchJsonResource(resource, parser)`

## Design

Existing no-arg signatures are preserved and new overloads are added alongside them. This ensures **binary compatibility** — consumers compiled against older versions will not encounter `NoSuchMethodError` on upgrade.

```kotlin
// Existing signature preserved
fun beValidJson(): Matcher<String?> = beValidJson(pretty)

// New overload
fun beValidJson(parser: Json): Matcher<String?> = ...
```

## Usage

```kotlin
val lenient = Json {
    isLenient = true
    allowComments = true
    allowTrailingComma = true
}

// Validates lenient JSON that would fail with the default strict parser
"""{ name: "sam", age: 25 }""".shouldBeValidJson(lenient)

// Compare JSON parsed with a custom parser
"""{ name: "sam" }""".shouldEqualJson("""{ "name": "sam" }""", lenient)
```

## Test plan

- [x] `CustomJsonParserTest` — unquoted keys, comments, trailing commas, failure messages, backward compatibility
- [x] `CustomJsonParserFileTest` — File, Path, and resource matchers with custom parser
- [x] All existing tests pass (`./gradlew :kotest-assertions:kotest-assertions-json:jvmTest`)
- [x] API dump updated (`./gradlew apiDump` + `apiCheck`)


sksamuel:
I think this looks great. We should wait for @AlexCue987 to sign off before we merge.

kotest-assertions/kotest-assertions-json/src/jvmMain/kotlin/io/kotest/assertions/json/paths/jsonMatchers.kt
```diff
@@ -26,27 +27,57 @@ fun Path.shouldBeJsonArray(): Path {
...
+   return this
+}
+
 fun Path.shouldBeJsonObject(): Path {
    this should (exist() and aFile() and beJsonObject().contramap { it.readText() })
    return this
 }
 
+fun Path.shouldBeJsonObject(parser: Json): Path {
+   this should (exist() and aFile() and beJsonObject(parser).contramap { it.readText() })
```
AlexCue987, line 51: this is minor and pre-existing, and should be resolved in a separate PR, but: 
`this should (exist() and aFile()` is repeated how many times in this file? should we extract one common function and unit test is separately instead of copying and pasting it a few more times?

definitely not a blocker for this PR

AlexCue987 (approved):
very nice. one minor suggestion to improve in a later PR

AlexCue987:
> I think this looks great. We should wait for @AlexCue987 to sign off before we merge.

I liked it a lot.

PR #5796, opened 2026-03-24
Fix @transactional support with afterEach

### LINKED ISSUE #5621: Spring lifecycle hooks and kotest hooks
As part of this PR https://github.com/kotest/kotest/pull/5317 which was addressing this issue https://github.com/kotest/kotest/issues/5283 we have fixed the wrong order we had in kotest for Spring lifecycle hooks.
[Sam has also ensured to document](https://github.com/kotest/kotest/blob/ac0ddac91f58eb457d210a4b03e901a9961c95b5/documentation/docs/extensions/spring.md#order-of-events) how these hooks are supposed to run in a Spring + Kotest test lifecycle.

This works quite well, however have identified a potential issue on how these hooks are supposed to run.
If we look at[ this test ](https://github.com/kotest/kotest-examples/blob/master/kotest-spring-data/src/test/kotlin/io/kotest/exaples/spring/data/TransactionalTest.kt), which uses `@Transactional` (some Spring test annotations make use of `@Transactional` , for example `@DataJpaTest`), the order of the invocations is as follows:

1. **Kotest Spring Extension:** `testContextManager().beforeTestClass()`
2. **Kotest Spring Extension:** `execute(spec)`
3. **Kotest Spring Extension:** `testContextManager().beforeTestMethod(testCase.spec, methodName)`
4. **Kotest Spring Extension:** `execute(testCase)`
5. **User Test:** `beforeEach`
6. **Kotest Spring Extension:** `testContextManager().beforeTestExecution(testCase.spec, methodName)`
7. **User Test:** `the actual test`
8. **Kotest Spring Extension:** `testContextManager().afterTestExecution(testCase.spec, methodName, null as Throwable?)`
9. **User Test:** `afterEach`
10. **Kotest Spring Extension:** `testContextManager().afterTestMethod(testCase.spec, methodName, null as Throwable?)`
11. repeat until the end and finally call **Kotest Spring Extension:** `testContextManager().afterTestClass()`

This to me seems the right order, and for more context, according to Spring docs:
- `beforeTestMethod` opens a transaction
- `afterTestMethod` closes a transaction
And yet, the *User Test:** `afterEach` does not seem to be registered in the transaction, as an example in the test linked the `pkId` sequence does not seem to be reset as part of the `afterEach` hook.

Is this a Kotest issue with regards to how the hooks are registered? 
Should we have used different listeners in `SpringExtension` to register the Spring hooks in.
I am not sure, just writing this issue with the initial findings, will be sure to upate this once I get around this (unless someone beats me to it 😄 )

sksamuel:
Fixes #5621

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -33,8 +33,8 @@ plugin-publish = "1.3.1"
 rgxgen = "2.0"
 ksp = "2.2.0-2.0.2"
 shadowjar = "8.1.1"
-spring-boot = "2.7.18"
-spring = "5.3.39"
+spring-boot = "4.0.4" # version 4 requires JVM17+
+spring = "7.0.6"
 
 allure-commons = "2.30.0"
 allure-plugin = "3.0.2"
@@ -49,7 +49,7 @@ mockserver = "5.15.0"
 fuel = "2.3.1"
 konform = "0.10.0"
 arrow = "2.1.2"
-testcontainers = "2.0.3"
+testcontainers = "2.0.4"
 asm = "9.8"
 nmcp = "1.0.2"
 kotlin-poet = "2.2.0"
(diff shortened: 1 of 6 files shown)


AlexCue987 (commented):
I'm skeptical about Spring's transaction management. It is flaky to put it mildly. Is there a plan to use something lower level like JDBI or JDBC and have robust transactions?

sksamuel:
> I'm skeptical about Spring's transaction management. It is flaky to put it mildly. Is there a plan to use something lower level like JDBI or JDBC and have robust transactions?

I'm not even sure this is an issue yet.
I can see that spring is applying the transaction to the afterEach properly.
So I think this might be a red herring. Continuing to investigate.

PR #5798, opened 2026-03-25
Escape dots in test names in --tests filter to prevent Gradle FQN misparse

### LINKED ISSUE #5797: Periods in a test name break --tests filter
Gradle is interpreting the periods as part of the FQN when running an individual test from the IDE

Example failing test:

```kotlin

class MySpec : FunSpec() {
  init {
    test("1.2.3 my test") {
    }
  }
}
```

sksamuel:
## Summary

- Periods in test names (e.g. `test("1.2.3 my test")`) were passed literally in the Gradle `--tests` filter, causing Gradle to misinterpret them as package/class-name separators
- Gradle could not discover the test class and no tests ran
- Fix: replace `.` with `?` (Gradle's single-character wildcard) in the test name portion of the filter only — the spec FQN dots are preserved

## Details

`GradleTestFilterBuilder` previously generated:
```
--tests 'io.example.MySpec.1.2.3 my test'
```
Gradle splits on `.` to find the class boundary and would misparse this as class `io.example.MySpec.1.2.3` (non-existent), so no tests were discovered.

After the fix it generates:
```
--tests 'io.example.MySpec.1?2?3 my test'
```
Gradle correctly identifies `io.example.MySpec` as the class, and the `?` wildcards (each matching any single character) still match the literal `.` in the test name.

Fixes #5797

## Test plan

- [x] Added `testPeriodInTestNameIsReplacedWithQuestionMark` — verifies a root test with dots in its name produces `?`-escaped filter
- [x] Added `testPeriodInNestedTestNameIsReplacedWithQuestionMark` — verifies the same for nested tests


sschuberth:
For reference, I believe this was superseded by https://github.com/kotest/kotest/pull/5842.

PR #5805, opened 2026-03-26
Fix edge cases not being shrunk in property tests (#2973)

### LINKED ISSUE #2973: Property based testing edge cases are not shrinked
**Example scenario:**

1. A function takes an argument of type `Filters`:
```kotlin
data class Filters(val olderThan: Instant?, val cheaperThan: Money?, val hasTags: Set<Tag>?, ...)
```

2. We are using property based testing to test it. The `Arb<Filters>` is generated with `bind` from several other `Arb` instances (i.e. `Arb<Instant?>`, `Arb<Money?>`, `Arb<Set<Tag>?>`, etc.). Each of these base `Arb` instances shrink to null (and may shrink to other values).
3. There is a bug in the implementation of the `olderThan` part of the filter.

**Result:**

Kotest tries to shrink the failing example, but now and then we get a failure that is not shrinked. Such example gives no clue as what fails and is hard to debug. e.g. An instance of Filters with many fields having not null values. Debugging such cases I find that Kotest does not even try to shrink the value.

**Expected:**

I'd expect Kotest to always try to shrink the values, even in the case of edge cases.

**Further information:**

This is an issue that has been difficult to diagnose but that I think I understand now.

Whenever an Arb chooses to generate an edge case, it does so with:
```kotlin
this.edgecase(rs)?.asSample() ?: samples.next()
```

That means `null` can't be an edge case, which would be convenient. Furthermore, edgecases samples won't benefit of shrinking.  This may seem appropiate for simple `Arb`s, but it is causing some complex examples to not be shrinked when they should be.

In particular, the family of `bind` functions in `io.kotest.property.arbitrary` does an amazing job at automatically calculating shrinks given the shrinks of the `Arb` instances provided to `bind`. Unfortunately, like I said before, when Kotest chooses to generate a sample value, those shrinks are not generated in the `Sample`.

IMHO, edgecases should generate Samples as well. That way, `Arb` combinators may generate `Arb` instances capable of shrinking for non trivial edge cases.

I'm aware that may not be a low-hanging fruit. Let me know if I can help with it in case you maintainers are ok with giving it any priority. I have a couple more ideas on shrinking I'd create issues for, if there is any interest.

sksamuel:
## Summary

Fixes #2973

When `Arb.generate()` decides to emit an edge case, the resulting `Sample` was created with `.asSample()` — which produces `RTree({ value })` with **no shrink children**. This caused `doShrinking()` to immediately bail out (`initial.isEmpty()` = true), so any property failure on an edge case was reported as-is, completely unshrunk.

The bug was most visible with `Arb.bind`: when a composite Arb randomly selected edge cases from its component Arbs, the combined `Sample` had no shrink tree at all. Even though `bind.sample()` correctly uses `combineShrinks()`, `bind.edgecase()` discarded all component shrink information.

### Changes

- **`bind.kt` (`bindN.edgecase`)** — the key fix: instead of extracting raw values from component edge cases and calling `.asSample()`, we now take each component's full `Sample` (including its shrink tree) and use `combineShrinks` to compose them, exactly as `bindN.sample()` does
- **`map.kt` (`Arb.map` edgecase)** — propagate `sampleA.shrinks.map(fn)` through the mapped function rather than discarding them with `.asSample()`
- **`builders.kt`** — wherever a `Shrinker` is available, use `sampleOf(value, shrinker)` for edge cases instead of `value.asSample()` (affects `ArbitraryBuilder.withEdgecases`, `arbitrary(edgecases, shrinker, fn)`, `generateArbitrary(edgecases, shrinker, fn)`, and the `arbitraryBuilder`/`suspendArbitraryBuilder` fallback path)
- **`collections.kt` (`Arb.list` edgecase)** — use `sampleOf(list, ListShrinker(range))` so list edge cases carry structural shrinks

### Test plan

New test file `EdgeCaseShrinkTest`:
- [ ] `Arb.int` edge case carries a non-empty shrink tree
- [ ] `Arb.list` edge case for multi-element lists carries shrinks
- [ ] `Arb.bind` edge case carries shrinks from component Arbs
- [ ] `Arb.map` edge case preserves shrinks from underlying Arb
- [ ] Integration test: `forAll` with `Arb.bind` and 100% edge case probability — at least one of 50 seed runs shows `"shrunk from"` in the failure message


kotest-property/src/commonMain/kotlin/io/kotest/property/arbitrary/ints.kt
```diff
@@ -41,8 +41,8 @@ fun Arb.Companion.int(min: Int = Int.MIN_VALUE, max: Int = Int.MAX_VALUE) = int(
 fun Arb.Companion.int(range: IntRange = Int.MIN_VALUE..Int.MAX_VALUE): Arb<Int> {
    val edgeCases = intArrayOf(range.first, -1, 0, 1, range.last).filter { it in range }.distinct()
    return ArbitraryBuilder.create { it.random.nextInt(range) }
-      .withEdgecases(edgeCases)
       .withShrinker(IntShrinker(range))
+      .withEdgecases(edgeCases)
```
AlexCue987, line 45: is there a reason `withEdgecases` has to be invoked after `withShrinker`?

AlexCue987 (approved):
LGTM one thing is not clear

PR #5771, opened 2026-03-20
Add JSON Schema prefixItems (tuple validation) and enum support

sksamuel:
## Summary

- Adds `prefixItems` field to `JsonSchema.JsonArray` for tuple validation — each array position is validated against its own specific schema. When combined with `typeBuilder`, elements beyond the prefix are validated against the additional items schema.
- Adds `JsonSchema.JsonEnum` and an `enum()` DSL builder function for constraining values to a fixed set of allowed strings, numbers, booleans, or null.
- Updates the schema validator and JSON Schema parser (`parseSchema`) to support both new features.

Addresses #2980

### Usage

**Tuple validation with `prefixItems`:**
```kotlin
val addressSchema = jsonSchema {
    array(prefixItems = listOf(number(), string(), enum("Avenue", "Street", "Boulevard")))
}

// Validates each position: [number, string, enum]
"[1600, \"Pennsylvania Avenue NW\", \"Avenue\"]" shouldMatchSchema addressSchema
```

**Enum validation:**
```kotlin
val streetTypeSchema = jsonSchema { enum("Avenue", "Street", "Boulevard") }
"\"Avenue\"" shouldMatchSchema streetTypeSchema
```

**Parsing from JSON Schema documents:**
```kotlin
val schema = parseSchema("""
    {
        "type": "array",
        "prefixItems": [
            {"type": "number"},
            {"enum": ["Avenue", "Street"]}
        ]
    }
""")
```

## Test plan

- [ ] `ArraySchemaTest` — added tests for `prefixItems` validation, combining with `elementType`, and parser support
- [ ] `EnumSchemaTest` — new test class covering string/number/boolean/null enum values, type mismatch, nested usage in `prefixItems`, and parser support

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5807, opened 2026-03-26
Add anyOf/oneOf support to JsonSchema DSL and parser (#4463)

### LINKED ISSUE #4463: Add `JsonSchema` Support for Multiple Types
Basically just JSON Schema's [`anyOf`](https://json-schema.org/understanding-json-schema/reference/combining#anyOf) and/or [`oneOf`](https://json-schema.org/understanding-json-schema/reference/combining#oneOf)

This feature is similar, although not identical, to #2980. (in certain cases #2980 is a replacement for this, although it does not fit all cases)

an example syntax would look like this:
```kotlin
val stringOrIntegerListSchema = jsonSchema {
    array {
        anyOf(string(), integer())
    }
}

"""[ "foo", "bar", "baz" ]""" shouldMatchSchema stringOrIntegerListSchema // passes
"""[ 1, 2, 3 ]""" shouldMatchSchema stringOrIntegerListSchema // passes
"""[ true, false ]""" shouldMatchSchema stringOrIntegerListSchema // fails
```

in order to do this, it would either be necessary to add an additional `JsonSchemaElement` called something like `JsonCompositeElement`, which composes multiple elements (imo the less desirable choice)
or to change the `JsonSchema.JsonObjectBuilder.properties` field from a `MutableMap<String, JsonSchemaElement>` to a `MutableMap<String, Matcher<JsonSchemaElement>>` (imo the better choice)

PreAgile:
Closes #4463

## Summary

- Add `JsonSchema.JsonAnyOf` and `JsonSchema.JsonOneOf` as nested subtypes of `JsonSchema` (same pattern as `JsonEnum` in #5771)
- Support `anyOf()` / `oneOf()` builder functions in the DSL
- Add validation logic: `anyOf` passes if any candidate matches, `oneOf` passes if exactly one matches
- Add parser support with `selectDeserializer` priority over `type`-based dispatch (since `anyOf`/`oneOf` can appear without a `type` field)
- Empty `anyOf`/`oneOf` schemas are rejected with `require` in both DSL and parser

## Changes

- `builder.kt` — `JsonAnyOf`, `JsonOneOf` data classes + `anyOf()`, `oneOf()` extension functions
- `matchSchema.kt` — Validation branches for anyOf (any match) and oneOf (exactly one match) with path-preserving error messages
- `parse.kt` — `JsonSchemaAnyOfSerializer`, `JsonSchemaOneOfSerializer` + `selectDeserializer` pre-check
- `AnyOfOneOfSchemaTest.kt` — Tests covering DSL, parser, nested schemas, oneOf overlap semantics, error messages, and empty schema validation
- `.api` — Updated via `./gradlew apiDump`


kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/builder.kt
```diff
@@ -203,6 +203,14 @@ data class JsonSchema(
    object Null : JsonSchemaElement {
       override fun typeName() = "null"
    }
+
```
AlexCue987, line 206: do we need Javadocs for new classes?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/parse.kt
```diff
@@ -57,6 +57,8 @@ fun parseSchema(@Language("json") jsonSchema: String): JsonSchema =
 @ExperimentalKotest
 internal object SchemaDeserializer : JsonContentPolymorphicSerializer<JsonSchemaElement>(JsonSchemaElement::class) {
    override fun selectDeserializer(element: JsonElement): DeserializationStrategy<JsonSchemaElement> {
+      if (element.jsonObject.containsKey("anyOf")) return JsonSchemaAnyOfSerializer
```
AlexCue987: should we fold these three similar `if .. return ...` into a `when`?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/parse.kt
```diff
@@ -239,3 +241,45 @@ internal object JsonSchemaNumberSerializer : KSerializer<JsonSchema.JsonDecimal>
...
+   }
+
+   override fun serialize(encoder: Encoder, value: JsonSchema.JsonAnyOf) {
+      TODO("Serialization of JsonSchema not supported atm")
+   }
+}
+
+@ExperimentalKotest
+internal object JsonSchemaOneOfSerializer : KSerializer<JsonSchema.JsonOneOf> {
+   override fun deserialize(decoder: Decoder): JsonSchema.JsonOneOf {
```
AlexCue987: these two `deserialize` implementations are strikinly similar - is there a way to reuse one implementation?

kotest-assertions/kotest-assertions-json/src/commonTest/kotlin/io.kotest.assertions.json/schema/AnyOfOneOfSchemaTest.kt
```diff
@@ -0,0 +1,255 @@
...
+
+import io.kotest.assertions.shouldFail
+import io.kotest.assertions.throwables.shouldThrow
+import io.kotest.common.ExperimentalKotest
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.string.shouldContain
+
+@OptIn(ExperimentalKotest::class)
+class AnyOfOneOfSchemaTest : FunSpec(
```
AlexCue987: would the tests be easier to comprehend if we had one test suite for `anyOf` and a separate one for `oneOf`?

AlexCue987 (commented):
LGTM, some non-blocking suggestions. WDYT @sksamuel ?

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/builder.kt
sksamuel: some documentation is always good

sksamuel:
> LGTM, some non-blocking suggestions. WDYT @sksamuel ?

seems good, should wait for your comments to be addressed. I'll leave the approval  to you

kotest-assertions/kotest-assertions-json/src/commonMain/kotlin/io/kotest/assertions/json/schema/parse.kt
PreAgile: Extracted the shared combinator deserialization into a small helper to avoid duplicating the `anyOf` / `oneOf` decoding logic in `f4a64a028`.

AlexCue987 (approved):
very nice!

PR #5814, opened 2026-03-28
Add public API for creating custom styles

### LINKED ISSUE #2702: Custom spec style
Hi, i want to add custom spec style, nothing serious just rename FunSpec Methods like

container -> migration
test or preparation or validation 

e.g. 

```
class MyMigration: MigrationSpec(
{    
    migration("some migration name"){
       preparation("some name"){
           // test code
       }

      validation("some name"){
           // test code
       }
    }
}
)
```

is it possible to add support to IntelliJ IDEA plugin to get possibility to run this tests using right click?

### LINKED ISSUE #4342: Expose internal API used for defining testing styles
It would be really nice if some of the internal api used for designing testing styles could be exposed.

Specifically:
- the `build()` extension function on `TestFactoryConfiguration`
- the `tests` property on `TestFactoryConfiguration`
- mark the `include` method on `TestFactoryConfiguration` as `open`

My current workaround for this has been to implement `TestFactoryConfiguration`, and then override the `add` method, and also copy the `build` extension method into my codebase. (`_tags` must be replaced with `appliedTags()` and `_extensions` must be replaced with `registeredExtensions()`)

### LINKED ISSUE #5353: Support running custom spec styles
Currently, the kotest idea plugin doesn't support custom test styles. It would be nice if at the very least, a run arrow was shown for the custom test style classes.

sksamuel:
Formalizes the public API for creating custom spec styles by subclassing `AbstractSpec`.

Adds @TestRunnable annotation which instructs the IDE that the function is a runnable test. This allows any custom method that defines a test to be picked up in the plugin. Over time, we can migrate the existing Specs to use this infrastructure so they become as generic as any other custom spec style.

Updated some nomenclature for some previously internal structures which are now part of the public API. Most notably, the Kotest internal RootTest and NestedTest data classes have been deprecated and a new TestDefinition added which is now used in both places.

This gives users an official way to do what some projects are doing already, eg https://github.com/solo-studios/nyx/blob/master/src/testFixtures/kotlin/ca/solostudios/nyx/kotest/spec/NyxSpec.kt and https://prepared.opensavvy.dev/tutorials/index#__tabbed_1_3

Fixes https://github.com/kotest/kotest/issues/4342
Fixes https://github.com/kotest/kotest/issues/5353
Fixes https://github.com/kotest/kotest/issues/2702


sksamuel:
Gonna merge this in as its a huge PR but mostly consists of renames.

PR #5815, opened 2026-03-28
IJ Plugin N times setter for test invocation

### LINKED ISSUE #5361: KoTest run configuration does not have options to run test repeatedly
The JUnit run configuration has options to run a test repeatedly: given number of executions, until failure, ...

image

KoTest run configuration do not have this option.

alfonsoristorato:
Fixes https://github.com/kotest/kotest/issues/5361

1. Press `Run with repetitions...`
image

2. Select number of runs
image

3. Go to the target picker (or straight into tests if only one target) - this is unchanged, simply adds an env variable
image

4. Tests are run N times
image


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/TestConfigResolver.kt
```diff
@@ -156,9 +158,16 @@ class TestConfigResolver(
       return testConfigs(testCase).firstNotNullOfOrNull { it.invocations }
          ?: testCase.spec.defaultTestConfig?.invocations
          ?: projectConfig?.invocations
+         ?: invocationCountFromExtensions()
          ?: Defaults.INVOCATIONS
    }
 
+   private fun invocationCountFromExtensions(): Int? {
```
alfonsoristorato: unfortunately doing this on `invocationCount` brings us into states where users might try to set N times on container tests, and kotest does not allow it - still thinking about that :)

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
```diff
@@ -82,6 +81,7 @@ internal class TestCaseExecutor(
             context.specConfigResolver,
             context.testConfigResolver
          ),
+         InvocationCountCheckInterceptor(context.testConfigResolver),
```
alfonsoristorato: moved this after the enable check, as this env var sets invocation for each test, and this would throw for every other test within the spec as long as it is a container - I think in general, it's better to do this check after the enabled check even without this feature

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/TestConfigResolver.kt
alfonsoristorato: I think, similar to how we inform users via an error that `invocations` cannot be used in container tests, this will now do the same after they click, so I am inclined to leave this as is.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/TestConfigResolver.kt
```diff
@@ -156,9 +159,23 @@ class TestConfigResolver(
       return testConfigs(testCase).firstNotNullOfOrNull { it.invocations }
          ?: testCase.spec.defaultTestConfig?.invocations
          ?: projectConfig?.invocations
+         /**
```
alfonsoristorato: This is to support setting invocations on leaf tests within containers, but I'd like to understand what was the rationale to disallow `invocations` in containers @kotest/kotest-admins ?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/TestConfigResolver.kt
sksamuel: Can we just not offer the option if it's a container test ?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
sksamuel: How about we go one better - and just error out as the test is being defined? I can make a PR for that.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/TestConfigResolver.kt
sksamuel: It was because the isolation mode logic combined with multiple parents becomes super complex (its already complex). If you have instance per leaf, and then you run the parent multiple times....

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/TestConfigResolver.kt
alfonsoristorato: We can, now that I got the reason why we don't do this for containers. I think there is some reusable cose in the data test pr that can be combined to do this check, once we are ready to merge 6.2 prs I will merge that and re-use some logic

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/TestCaseExecutor.kt
alfonsoristorato: Sounds good to me, in hindsight applying this invocation only on non-containers does not require me to move this check's position anymore, so will revert

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/TestConfigResolver.kt
alfonsoristorato: Got it, thanks 🙏

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/TestConfigResolver.kt
alfonsoristorato: Or, and this is up to you.
If we leave things as they are (the above comment is outdated), users will not see errors when running n times a container test, but rather the inner tests (that classify as non-containers) will run n times. This is probably a bit counter intuitive, and could do with a different configuration name, but it would allow to run multiple inner tests n times. Wdyt?

sksamuel: I think it'll lead to questions "why isn't my container running 5 times".
I think best to just offer it on leafs.

sksamuel (approved):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/extensions/InvocationCountExtension.kt
```diff
@@ -0,0 +1,5 @@
+package io.kotest.core.extensions
+
+interface InvocationCountExtension: Extension {
```
sksamuel: Should we make this internal for the first release so we can change things if we find out later we missed something.

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/actions/RunRepeatAction.kt
```diff
@@ -0,0 +1,46 @@
...
+         }
+      ) ?: return
+
+      val count = countStr.toIntOrNull() ?: return
+      project.service<KotestRunState>().pendingInvocationCount = count
+
+      // Delegate to the standard Run executor action, this triggers the full normal
+      // flow: doSetupConfigurationFromContext (which reads and clears pendingInvocationCount),
+      // then onFirstRun (multiplatform task chooser), then execution.
+      ExecutorAction.getActions(1).firstOrNull()?.actionPerformed(e)
```
sksamuel, line 44: I know we do this in other places, but I wonder what significance the 1 has

alfonsoristorato: I had assumed it was to take the first possible action, which normally is Run (if we somehow decide to consider that array as non 0 based) - but realistically it does not make sense :)

sksamuel: That's why I was curious, because 0 should be run, and 1 debug or whatever.
I guess I'll just dump it out if I'm that interested.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/extensions/InvocationCountExtension.kt
alfonsoristorato: good shout - done

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/TestConfigResolver.kt
alfonsoristorato: done :) [ensure RunRepeatAction is only enabled for leaf tests](https://github.com/kotest/kotest/pull/5815/commits/cee28aeb26c43118abde938038d169129c062b6c)[ensure RunRepeatAction is only enabled for leaf tests](https://github.com/kotest/kotest/pull/5815/commits/cee28aeb26c43118abde938038d169129c062b6c)

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/actions/RunRepeatAction.kt
alfonsoristorato: I have also tried to set it to 0 and there is no change in how the gutter shows...

PR #5820, opened 2026-03-30
Align Kotlin version usage with best practice.

sksamuel:
Mainly, compile with latest kotlin tooling, mark language as earliest we support - in this case 2.2.0

Also removed three targets that are no longer supported in Kotlin 2.3

See https://mbonnin.net/2026-02-22-kotlin-versions


alfonsoristorato (approved):
this is really cool - thanks for the link too :)

PR #5835, opened 2026-04-01
feat: show per-element data class diffs in collection comparisons

### LINKED ISSUE #2545: Make diff for data classes work with collections
Kotlin 4.1 introduced [diffs for data classes](https://github.com/kotest/kotest/issues/826). However, that currently only works for asserting single data classes via `shouldBe`. As an extension, it would be nice if data class diffs would also work for the non-matching entries in collections.

This might be hard to do for unordered comparisons like `containExactlyInAnyOrder` as it's not clear which data classes to diff if there are multiple similar mismatches. However, for ordered comparisons like `containExactly` this should be doable.

PreAgile:
## Summary

- Extend `CollectionEq.checkEquality()` to preserve data class diff messages from per-element comparisons
- Previously, only differing indices were reported without field-level detail
- Covers both `shouldBe` and `shouldContainExactly` paths since `CollectionEq` is the single source of truth for ordered collection comparison errors

### Before
```
Element differ at index: [0, 2]
expected:<[DataClass1(a=2, b=3.5), ...]> but was:<[DataClass1(a=1, b=3.4), ...]>
```

### After
```
Element differ at index: [0, 2]

The following elements differ:
index 0: data class diff for com.sksamuel.kotest.eq.DataClass1
├ a: expected:<2> but was:<1>
└ b: expected:<3.5f> but was:<3.4f>

index 2: data class diff for com.sksamuel.kotest.eq.DataClass1
└ a: expected:<99> but was:<2>

expected:<[DataClass1(a=2, b=3.5), ...]> but was:<[DataClass1(a=1, b=3.4), ...]>
```

## Scope

- Ordered collection comparisons only (`shouldBe`, `shouldContainExactly`)
- Unordered comparisons (`shouldContainExactlyInAnyOrder`) are excluded as agreed in #2545

## Test plan

- [x] Data class list with differing elements at multiple indices
- [x] Nested data class diff (e.g. `DataClass2` containing `DataClass1`)
- [x] Non-data-class elements produce unchanged output (no regression)
- [x] `shouldBe` path with collection of data classes
- [x] `apiCheck` passes (no public API changes)

Closes #2545


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/CollectionEq.kt
```diff
@@ -121,11 +121,14 @@ object CollectionEq : Eq<Collection<*>> {
 
    private const val DISALLOWED = "$TRIGGER nesting iterator"
 
+   private const val MAX_DETAILED_DIFFS = 10
```
AlexCue987: should we make this configurable?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/CollectionEq.kt
```diff
@@ -177,7 +180,14 @@ object CollectionEq : Eq<Collection<*>> {
                else -> equalXorDisallowed(EqCompare.compare(a, b, context))
             }
             if (!accrueDetails) break
-            if (t != null) elementDifferAtIndex.add(index)
+            if (t != null) {
+               elementDifferAtIndex.add(index)
+               val msg = t.message
+               if (msg != null && msg.startsWith("data class diff")) {
```
AlexCue987: it will be safer to verify that both `a` and `b` are data classes

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/CollectionEq.kt
```diff
@@ -194,6 +204,15 @@ object CollectionEq : Eq<Collection<*>> {
          if (missingElementAt != null) {
             append("Missing elements from index $missingElementAt\n")
          }
+         if (elementDiffDetails.isNotEmpty()) {
+            append("\nThe following elements differ:\n")
```
AlexCue987: because there might be one element, let's change to `The following element(s) differ`

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/CollectionEqTest.kt
```diff
@@ -208,4 +210,56 @@ class CollectionEqTest : FunSpec({
...
    }
+
+   test("should show data class diff for lists of data classes with differing elements") {
+      val result = CollectionEq.equals(
+         listOf(DataClass1(1, 3.4F), DataClass1(1, 3.4F), DataClass1(2, 7.6F)),
+         listOf(DataClass1(2, 3.5F), DataClass1(1, 3.4F), DataClass1(99, 7.6F)),
+         EqContext()
+      ) as EqResult.Failure
+      val error = result.error()
+      error.message shouldStartWith """Element differ at index: [0, 2]
```
AlexCue987: long term, such wide-brush tests are a bit fragile. how about we make these tests more specific, just verifying that the message contains in order the following substrings:

* "index 0: data class diff for com.sksamuel.kotest.eq.DataClass1"
* "├ a: expected:<2> but was:<1>"
* "└ b: expected:<3.5f> but was:<3.4f>"

and so on?

AlexCue987 (commented):
LGTM, nice improvement, a few minor suggestions

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/CollectionEq.kt
PreAgile: Thanks for the suggestion! I looked into AssertionsConfig.kt to understand the existing patterns before deciding.

I noticed there are two patterns in the codebase:

- The older approach uses sysprop() directly (e.g. mapDiffLimit, maxErrorsOutput)
- Newer additions use EnvironmentConfigValue (e.g. maxCollectionPrintSize, 
  maxSimilarityPrintSize), which reads from both system property and env var 
  via syspropOrEnv(), and also provides a withValue() helper for safe 
  test-time overrides without manual teardown.

Since MAX_DETAILED_DIFFS is a new addition, I'd lean toward the EnvironmentConfigValue pattern to stay consistent with the newer style —  something like:

    val maxCollectionDataClassDiffs: EnvironmentConfigValue<Int> =
        EnvironmentConfigValue(
            "kotest.assertions.collection.max-data-class-diffs", 
            10, 
            String::toInt
        )

That said, I want to make sure I'm aligned with your preference before changing AssertionsConfig.kt. Would you prefer:
 (a) EnvironmentConfigValue in AssertionsConfig (newer pattern)
(b) Direct sysprop() in CollectionEq (simpler, no extra file change)
(c) Something else entirely?

Happy to go either way — just want to make the right call for the codebase.

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/CollectionEqTest.kt
PreAgile: Good point — the existing tests were asserting the full error message as a literal string, which makes them brittle against any formatting change (whitespace, punctuation, reordering of clauses).

Refactored all affected tests to use shouldContainInOrder, which verifies that the key pieces of information are present in the right order without locking down the exact format:

    message.shouldContainInOrder(
        "Element differ at index: [0]",
        "index 0: data class diff",
        "a: expected:<2> but was:<1>",
        "b: expected:<3.5f> but was:<3.4f>",
    )

Also added a negative assertion for non-data-class elements to confirm the diff section is not emitted when it shouldn't be:

    message shouldNotContain "The following element"

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/CollectionEq.kt
PreAgile: Fixed — updated to use conditional singular/plural:

    val label = if (elementDiffDetails.size == 1) "element" else "elements"
    append("\nThe following $label differ:\n")

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/CollectionEq.kt
PreAgile: Good point. I see the issue — checking the error message string ties CollectionEq to DataClassEq's internal formatting, which is fragile.

I noticed that isDataClassInstance() is already available as an internal function in the same package (DataClassEq.kt), and it's used in DefaultEqResolver the same way to decide whether to delegate to DataClassEq.

So instead of checking msg.startsWith("data class diff") after the fact, 
I'll check isDataClassInstance(a) && isDataClassInstance(b) before the comparison — which removes the dependency on the message format entirely.

Does that match what you had in mind?

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/CollectionEq.kt
AlexCue987: the following commit is an example on adding config params and using them:

https://github.com/kotest/kotest/commit/36a7a1411d8562ca23c0aaac91c16f05cb821665

search for new config params such as `maxSubstringCount`. good luck!

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/CollectionEq.kt
AlexCue987: > Does that match what you had in mind?

yes exaclty

AlexCue987 (approved):
very nice!

PR #5843, opened 2026-04-05
support IJ plugin singular data test run on Kotest producer

alfonsoristorato:
support IJ plugin singular data test run on Kotest producer

image

image


sksamuel (approved):
Awesome

PR #5894, opened 2026-04-06
Fix coroutines probe dump with background coroutines

### LINKED ISSUE #4058: Foreground coroutines are not waited for by the test when debug probes are enabled, in the root spec
5.9.0

Normally, kotlinx-coroutines-test awaits all foreground coroutines before ending the test.

### Example 1

```kotlin
class ForegroundTest1 : FunSpec({
	this.coroutineTestScope = true
	this.coroutineDebugProbes = false

	test("Start foreground task") {
		launch {
			delay(1000)
			println("In a foreground scope")
		}
		println("In the test body")
	}
})
```

Output:
```
In the test body
In a foreground scope
```

This is the expected behavior. The test body finished, then awaited foreground coroutines.

### Example 2

```kotlin
class ForegroundTest2 : FunSpec({
	this.coroutineTestScope = true
	this.coroutineDebugProbes = true

	test("Start foreground task") {
		launch {
			delay(1000)
			println("In a foreground scope")
		}
		println("In the test body")
	}
})
```

Output:
```
In the test body
Coroutines dump 2024/06/03 20:33:44

Coroutine "coroutine#15":StandaloneCoroutine{Active}@6d071712, state: CREATED
	at _COROUTINE._CREATION._(CoroutineDebugging.kt:30)
	at kotlin.coroutines.intrinsics.IntrinsicsKt__IntrinsicsJvmKt.createCoroutineUnintercepted(IntrinsicsJvm.kt:161)
	at kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable(Cancellable.kt:26)
	at kotlinx.coroutines.BuildersKt__Builders_commonKt.launch$default(Builders.common.kt:43)
	at kotlinx.coroutines.BuildersKt.launch$default(Unknown Source)
```

This is incorrect behavior: the test should have waited for all foreground tasks to run before terminating. The output should be exactly the same as the first example.

The same problem is observed with `StringSpec`. I did not test the other specs.

### Example 3

Interestingly enough, this bug does not appear in nested tests.

```kotlin
class ForegroundTest3 : FunSpec({

	context("Without debug probes") {
		coroutineTestScope = true
		coroutineDebugProbes = false

		test("Start foreground task") {
			launch {
				delay(1000)
				println("In a foreground scope")
			}
			println("In the test body")
		}
	}

	context("With debug probes") {
		coroutineTestScope = true
		coroutineDebugProbes = true

		test("Start foreground task") {
			launch {
				delay(1000)
				println("In a foreground scope")
			}
			println("In the test body")
		}
	}
})
```

Output:
```
In the test body
In a foreground scope

In the test body
In a foreground scope
```

This made it particularly difficult to track down :sweat_smile: I hope that helps you find the cause, somehow.

sksamuel:
Fixes #4058 

Coroutine probe dump should happen after the tests and launched coroutines have completed. We can do this by moving the probe interceptor to be invoked sooner.

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/CoroutineDebugTest.kt
+++ b/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/CoroutineDebugTest.kt
@@ -1,6 +1,7 @@
 package com.sksamuel.kotest.engine.coroutines
 
 import io.kotest.core.annotation.EnabledIf
+import io.kotest.core.annotation.Isolate
 import io.kotest.core.annotation.LinuxOnlyGithubCondition
 import io.kotest.core.config.AbstractProjectConfig
 import io.kotest.core.spec.SpecRef
@@ -10,9 +11,12 @@ import io.kotest.engine.listener.NoopTestEngineListener
 import io.kotest.extensions.system.captureStandardOut
 import io.kotest.matchers.collections.shouldBeEmpty
 import io.kotest.matchers.string.shouldContain
+import io.kotest.matchers.string.shouldNotContain
 import kotlinx.coroutines.async
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
 
+@Isolate
 @EnabledIf(LinuxOnlyGithubCondition::class)
 class CoroutineDebugTest : FunSpec() {
    init {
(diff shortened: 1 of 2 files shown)


LeoColman (approved):

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/coroutines/CoroutineDebugTest.kt
```diff
@@ -27,10 +31,54 @@ class CoroutineDebugTest : FunSpec() {
...
                .withProjectConfig(p)
+               .withoutEnvFilters()
                .execute()
                .errors.shouldBeEmpty()
          }
-         output shouldContain "DeferredCoroutine"
+         output shouldContain "Coroutines dump"
+      }
+
+      // https://github.com/kotest/kotest/issues/4058
```
LeoColman, line 41: I like tests for specific issues

sksamuel: Yeah me too

Issue #5826, opened 2026-03-30
Re-use the same runner when tests are run from the kotest window


alfonsoristorato:
When tests are run via the kotest window (see screenshot)

<img width="525" height="549" alt="Image" src="https://github.com/user-attachments/assets/d64f6949-7cd8-4a9f-91b1-8fcf007be802" />

The runner of choice is always kotest producer, even though gradle producer would have been picked.
Re-use the same underlying producer when tests are run via the kotest window

PR #5902, opened 2026-04-10
use the correct producer when running tests from within kotest custom IJ window

alfonsoristorato:
For https://github.com/kotest/kotest/issues/5826

**Using Gradle:**
image

**Using Kotest:**
image


kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/actions/RunAction.kt
```diff
@@ -50,62 +64,175 @@ fun runNode(tree: JTree, project: Project, executorId: String, executeBranch: Bo
...
+   executor: Executor,
+   producer: GradleRunConfigurationProducer,
+   manager: RunManager,
+   module: Module,
+   configName: String
+) {
+   val config = manager.createConfiguration(configName, producer.configurationFactory)
+   val gradleConfig = config.configuration as GradleRunConfiguration
+   gradleConfig.settings.externalProjectPath = ExternalSystemApiUtil.getExternalProjectPath(module) ?: return
+   gradleConfig.settings.taskNames = listOf("test")
```
alfonsoristorato, line 234: I just have a doubt about this, as maybe we still want to run the `jvmKotest` task in cases of `GRADLE_KOTEST_TASK`?

sksamuel: Consider that deprecated and dead.

sksamuel (approved):
approved but I wonder if claude can add some tests for the RunAction ?

alfonsoristorato:
> approved but I wonder if claude can add some tests for the RunAction ?

Claude obliged 😄

PR #5905, opened 2026-04-14
feat: add type-safe test metadata API (#5103)

### LINKED ISSUE #5103: Store arbitrary test metadata
It would be really nice if there was a way to store arbitrary test metadata.

This would be used for something like the allure extension to allow storing things like the relevant issue, links, epic, feature, severity, story, owner, or description for a test instead of having to use annotations (which cannot be reflectively accessed on all platforms, only on the jvm platform)

here's an example of how it might look:

let's say you have some class `MetadataKey`. you could define a key like:

```kotlin
val Issue = MetadataKey<String>("Issue")
```

then you could have a spec like:
```kotlin
class MyTests : FunSpec({
    metadata[Issue] = "https://github.com/foo/bar/issue/123" // sets the issue for all the tests in this spec

    test("whatever") {
        metadata[Issue] = "https://github.com/foo/bar/issue/456" // overrides the issue for this test

        // ...
    }

    // ...
})
```

then in your extension you could have something like
```kotlin
fun TestCase.issue(): String? = metadata[Issue] // gets the issue of the first parent or the spec
```

this allows for things like the allure extension to be ported to other platforms (pending a rewrite of the allure stuff, though)

PreAgile:
## Summary

Adds a type-safe `MetadataKey<T>` / `TestMetadata` API for attaching arbitrary key-value metadata to tests and specs.

- **`MetadataKey<T>`** — type-safe key with `name + type` equality (prevents `ClassCastException` when two modules use the same name with different types)
- **`TestMetadata`** — mutable container used during registration (spec init, `test.config()`)
- **`ResolvedTestMetadata`** — immutable snapshot for thread-safe reads from extensions/listeners
- **`TestConfigResolver.metadata(testCase)`** — resolves metadata from all levels (spec defaults → spec DSL → test/container config chain), child wins per-key

Closes #5103

## Design Decisions

- **Mirrors the existing `tags` pattern**: metadata lives in the same three places as tags — `TestConfiguration._metadata` (spec/factory), `TestConfig.metadata` (test/container), and resolved via `TestConfigResolver` using the existing `testConfigs(testCase)` parent-chain traversal
- **Two types, not one**: mutable `TestMetadata` for registration-time, immutable `ResolvedTestMetadata` for execution-time — `TestConfiguration._tags` is a plain `Set`, not a concurrent collection, so we don't expose mutable state to parallel extensions
- **KMP compatible**: zero reflection, zero annotations — compiles on all Kotest targets (JVM, JS, Native, Wasm)
- **Non-goal**: migrating existing `tags`/`severity` onto metadata. This PR adds the infrastructure; migration can happen later

## Usage

```kotlin
val Issue = MetadataKey<String>("Issue")
val Owner = MetadataKey<String>("Owner")

class PaymentTests : FunSpec({
    metadata[Issue] = "JIRA-123"           // spec-level

    context("refunds") {
        test("partial refund").config(
            metadata = TestMetadata().also { it[Owner] = "payments-team" }
        ) {
            // ...
        }
    }
})

// From extensions:
val resolved = testConfigResolver.metadata(testCase)
val issue = resolved[Issue]  // "JIRA-123"
```

## Test plan
- [x] MetadataKey equality (name + type)
- [x] MetadataKey inequality (different name, same name different type)
- [x] TestMetadata get/set/isEmpty
- [x] TestMetadata mergeWith (parent inheritance, child wins)
- [x] TestMetadata snapshot → ResolvedTestMetadata
- [x] Spec-level metadata via `appliedMetadata()`
- [x] TestConfig metadata storage
- [x] apiDump passes
- [x] All existing engine jvmTests pass

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/TestConfiguration.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/TestConfiguration.kt
@@ -52,13 +52,13 @@ abstract class TestConfiguration : Extendable() {
    internal var _tags: Set<Tag> = emptySet()
 
    @JsName("_metadata")
-   internal var _metadata: TestMetadata = TestMetadata()
+   internal val _metadata: TestMetadata = TestMetadata()
 
    /**
     * Returns the metadata container for this spec or factory.
     * Use operator syntax to set values: `metadata[key] = value`
     */
-   open val metadata: TestMetadata get() = _metadata
+   val metadata: TestMetadata get() = _metadata
 
    private var _autoCloseables = emptyList<Lazy<AutoCloseable>>()
 
@@ -86,8 +86,6 @@ abstract class TestConfiguration : Extendable() {
    }
 
    internal fun appliedTags() = _tags
-
-   internal fun appliedMetadata() = _metadata
 
    /**
     * Registers an [AutoCloseable] to be closed when the spec is completed.
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/TestConfiguration.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/TestConfiguration.kt
@@ -36,6 +36,7 @@ import io.kotest.core.spec.Spec
 import io.kotest.core.spec.TestCaseExtensionFn
 import io.kotest.core.test.AssertionMode
 import io.kotest.core.test.TestCase
+import io.kotest.core.test.TestMetadata
 import io.kotest.engine.test.TestResult
 import io.kotest.core.test.TestType
 import kotlin.js.JsName
@@ -49,6 +50,15 @@ abstract class TestConfiguration : Extendable() {
 
    @JsName("_tags")
    internal var _tags: Set<Tag> = emptySet()
+
+   @JsName("_metadata")
+   internal var _metadata: TestMetadata = TestMetadata()
+
+   /**
+    * Returns the metadata container for this spec or factory.
+    * Use operator syntax to set values: `metadata[key] = value`
+    */
+   open val metadata: TestMetadata get() = _metadata
 
    private var _autoCloseables = emptyList<Lazy<AutoCloseable>>()
 
@@ -76,6 +86,8 @@ abstract class TestConfiguration : Extendable() {
    }
 
    internal fun appliedTags() = _tags
+
+   internal fun appliedMetadata() = _metadata
 
    /**
     * Registers an [AutoCloseable] to be closed when the spec is completed.
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/DslDrivenSpec.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/DslDrivenSpec.kt
@@ -6,8 +6,6 @@ import io.kotest.core.factory.TestFactory
 import io.kotest.core.listeners.AfterProjectListener
 import io.kotest.core.listeners.ContextAwareListener
 import io.kotest.core.spec.style.scopes.RootScope
-import io.kotest.core.test.MetadataKey
-import io.kotest.core.test.TestMetadata
 import kotlin.js.JsName
 
 /**
@@ -45,12 +43,6 @@ abstract class DslDrivenSpec : Spec(), RootScope {
       if (sealed) throw InvalidDslException("Cannot add a tag after the spec has been instantiated")
       super.tags(*tags)
    }
-
-   override val metadata: TestMetadata
-      get() {
-         if (sealed) throw InvalidDslException("Cannot access metadata for mutation after the spec has been instantiated")
-         return super.metadata
-      }
 
    /**
     * Include the tests and extensions from the given [TestFactory] in this spec.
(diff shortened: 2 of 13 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/TestConfigResolver.kt
```diff
@@ -185,6 +187,28 @@ class TestConfigResolver(
...
+
+      // spec-level DSL: metadata[key] = value in spec init block
+      merged.putAll(testCase.spec.appliedMetadata().entriesForMerge)
+
+      // test/container config chain: outermost first, innermost last (child wins)
+      testConfigs(testCase)
+         .asReversed()
+         .forEach { merged.putAll(it.metadata.entriesForMerge) }
+
+      return if (merged.isEmpty()) ResolvedTestMetadata.EMPTY
```
sksamuel: Is there a difference between the already isEmpty one and the const EMPTY ?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/TestConfiguration.kt
```diff
@@ -77,6 +87,8 @@ abstract class TestConfiguration : Extendable() {
 
    internal fun appliedTags() = _tags
 
+   internal fun appliedMetadata() = _metadata
```
sksamuel: If this is internal and so is the var, can internal APIs just use the var ?

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/DslDrivenSpec.kt
```diff
@@ -44,6 +46,12 @@ abstract class DslDrivenSpec : Spec(), RootScope {
       super.tags(*tags)
    }
 
+   override val metadata: TestMetadata
+      get() {
+         if (sealed) throw InvalidDslException("Cannot access metadata for mutation after the spec has been instantiated")
+         return super.metadata
+      }
```
sksamuel: when would the metadata be accessible?
As the spec is sealed as soon as its instantiated

sksamuel:
In your PR docs, you do this:

```
metadata = TestMetadata().also { it[Owner] = "payments-team" }
```

I would imagine the metadata object is fixed, and you just add to it at runtime?

Otherwise here you are overriding the instance set by the framework which might lead to unpredictable behaviour.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/TestConfiguration.kt
```diff
@@ -50,6 +51,15 @@ abstract class TestConfiguration : Extendable() {
    @JsName("_tags")
    internal var _tags: Set<Tag> = emptySet()
 
+   @JsName("_metadata")
+   internal var _metadata: TestMetadata = TestMetadata()
```
sksamuel: Would we want this to be a var rather than a val?
The TestMetadata obj is already mutable right ?

sksamuel:
Nice idea, some comments added

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/TestConfiguration.kt
PreAgile: Yeah you're right, the metadata container is already mutable so there's no reason for the field itself to be var. Changing to val.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/TestConfiguration.kt
PreAgile: Fair point — they're both internal so the wrapper is just noise. Removed it, resolver now hits _metadata directly.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/spec/DslDrivenSpec.kt
PreAgile: It's accessible during the spec's init block (before sealing), so in practice the check always passes anyway.

But the bigger issue is that metadata returns a mutable reference — if someone grabs it during init they can still mutate it after sealing. tags() doesn't have this problem since it's a function call you can fully block. So this override is giving false safety. Removing it — with _metadata as val the container can't be swapped out, and mutating entries during init is the intended use.

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/config/TestConfigResolver.kt
PreAgile: Nah, no real difference — just a minor allocation optimization. Dropped it, resolver now always returns ResolvedTestMetadata(merged.toMap()).

PreAgile:
Yeah that's a good call — passing a new `TestMetadata` into `.config()` feels like you're replacing something rather than adding to it, which is weird compared to the spec-level `metadata[key] = value` pattern.

I looked at how the other TestConfig fields work — `tags` is `Set<Tag>`, `extensions` is `List<Extension>` — they're all immutable values, not mutable containers. So I've switched metadata to match:

```kotlin
data class TestConfig(
   ...
   val metadata: Map<MetadataKey<*>, Any> = emptyMap(),
)

// usage now mirrors tags:
test("refund").config(metadata = mapOf(Owner to "payments-team")) { ... }
```

`TestMetadata` (mutable) stays for spec-level DSL in `TestConfiguration`, and the resolver merges both into `ResolvedTestMetadata` as before. Pushed the update.

sksamuel (approved):

PR #5906, opened 2026-04-14
AI: add Kotest debugging path and reference guide

sksamuel:
Adds Path G with a quick checklist and a detailed KOTEST-DEBUGGING.md covering KOTEST_DEBUG=true, version collection, test isolation, ProjectConfig resolution, and common debug scenarios (discovery, lifecycle, timeouts, property seeds).

### DIFF
--- a/skills/alfonsos-ai-dump/references/KOTEST-DEBUGGING.md
+++ b/skills/alfonsos-ai-dump/references/KOTEST-DEBUGGING.md
@@ -1,0 +1,155 @@
+# Kotest Debugging Guide
+
+When a user reports a Kotest issue, gather the following information before diagnosing.
+
+## Step 1: Enable Debug Logging
+
+Set the `KOTEST_DEBUG` environment variable to `true` to enable verbose internal logging from the
+Kotest engine. This prints lifecycle events, spec discovery, extension callbacks, and configuration
+resolution to stderr.
+
+**Gradle (one-off):**
+```bash
+KOTEST_DEBUG=true ./gradlew test
+```
+
+**Gradle (persistent in build file):**
+```kotlin
+tasks.withType<Test>().configureEach {
+    environment("KOTEST_DEBUG", "true")
+}
+```
+
+**IntelliJ Run Configuration:**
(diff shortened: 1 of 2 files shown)


skills/alfonsos-ai-dump/references/KOTEST-DEBUGGING.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+The package `io.kotest.provided` is the default scan location. If a custom location is used, it
+must be set via:
+
+```bash
+./gradlew test -Dkotest.framework.config.fqn=com.example.MyProjectConfig
+```
+
+## Step 7: Reproduce with a Minimal Example
```
alfonsoristorato: is this more in case users want to report this as an issue to us?

alfonsoristorato (approved):

skills/alfonsos-ai-dump/references/KOTEST-DEBUGGING.md
sksamuel: That's exactly what it is. Do you think we should change any wording?

alfonsoristorato: not so much the wording, but we can make it explicit and say something along the lines of
```
if you can't figure out what the issue is, and think that it is worth reporting to the kotest team, please ask the user to create an issue via https://github.com/kotest/kotest/issues/new/choose
```

alfonsoristorato: I have removed from the queue in case you want to push this now, otherwise just merge as is (as I plan to remove my name from this folder ahah)

sksamuel: Why don't you edit this with whatever you think is best (you probably know more about skills than me since i know zero), then just add it back to merge when you're done ?

alfonsoristorato: cool :)

PR #5908, opened 2026-04-14
IJ plugin and Framework shared module

alfonsoristorato:
This is an initial attempt at creating a shared module of code that is used by both the plugin and framework.
It starts very simple by exposing a `normalizeTestName` method and some data test utils constants to test the waters, if this is the way to go, we can slowly add more shared code onto it


alfonsoristorato:
I am only playing around for now :) this has not yet been tested, nor do I know whether implementing our own modules is something that could work for a library

sksamuel:
> I am only playing around for now :) this has not yet been tested, nor do I know whether implementing our own modules is something that could work for a library

I'm not sure if the intellij plugin will include it inside the zip or if it expects it to be in maven central already?

alfonsoristorato:
> > I am only playing around for now :) this has not yet been tested, nor do I know whether implementing our own modules is something that could work for a library
> 
> I'm not sure if the intellij plugin will include it inside the zip or if it expects it to be in maven central already?

ah good point, no it won't find it - let me see what I can do :)

alfonsoristorato:
What I have (import the `srcDir` and compile) works, however, should any module that imports this bridge (framework and plugin at the moment) ever be imported in another module that would also import this bridge, we would end up with re-declarations - very big ifs here, but it kinda bothers me - still looking for a better way 😄

PR #5913, opened 2026-04-16
Alex/should contain exact copies for lists (#5912)

AlexCue987:
add new matcher:

```
           listOf(1, 2, 3).shouldContainExactCopies(t = 2, copies = 1)
           listOf(3, 1, 2, 3).shouldContainExactCopies(t = 3, copies = 2)
```

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/ContainExactCopies.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/ContainExactCopies.kt
@@ -1,0 +1,107 @@
+package io.kotest.matchers.collections
+
+import io.kotest.assertions.equals.Equality
+import io.kotest.assertions.print.print
+import io.kotest.assertions.similarity.possibleMatchesDescription
+import io.kotest.matchers.Matcher
+import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldNot
+
+fun BooleanArray.shouldContainExactCopies(t: Boolean, copies: Int): BooleanArray = apply { asList().shouldContainExactCopies(t, copies) }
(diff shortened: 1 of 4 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/ContainExactCopies.kt
```diff
@@ -0,0 +1,107 @@
+package io.kotest.matchers.collections
```
alfonsoristorato, line 1: looks fine to me, just two questions 😄 
1. do we want to create `infix` fun or are you happy without?
2. why the param is called `t`? (this is curiosity)

AlexCue987: > looks fine to me, just two questions 😄
> 
> 1. do we want to create `infix` fun or are you happy without?
> 2. why the param is called `t`? (this is curiosity)

1. in kotin, infix functions can have only one parameter, so it's not an option.
2. I copied/pasted/tweaked the existing `shouldContain`. I concur `t` is not the most intuitive name. What name would you suggest?

TY!

alfonsoristorato: ah bummer for `infix` - guess it's not worth then to pass a data class as param just to be able to have infix.
As for the name, and this is up to you, but another name could be `ofElement` - approving anyway

alfonsoristorato (approved):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/ContainExactCopies.kt
AlexCue987: > ah bummer for `infix` - guess it's not worth then to pass a data class as param just to be able to have infix. As for the name, and this is up to you, but another name could be `ofElement` - approving anyway

I'm AFK so will rename on Mon

PR #5921, opened 2026-04-21
Add character matchers

JordanLongstaff:
Added matchers and assertions:
- `Char.shouldBeUpperCaseChar`
- `Char.shouldBeLowerCaseChar`
- `Char.shouldBeTitleCaseChar`
- `Char.shouldBeDigit`
- `Char.shouldBeLetter`
- `Char.shouldBeLetterOrDigit`
- `Char.shouldBeWhitespace`


sksamuel (approved):
v nice

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/char/case.kt
```diff
@@ -0,0 +1,129 @@
...
+ */
+fun beUpperCaseChar(): Matcher<Char?> = charCaseMatcher("upper") { it.uppercaseChar() }
+
+/**
+ * Asserts that this [Char] is lowercase.
+ * @see [shouldNotBeLowerCaseChar]
+ * @see [beLowerCaseChar]
+ */
+@OptIn(ExperimentalContracts::class)
+fun Char?.shouldBeLowerCaseChar(): Char {
```
AlexCue987: is there a reason we extend a nullable char and then have a contract to make sure it's not null? would it be simpler to just `fun Char.shouldBeLowerCaseChar(): Char`?

JordanLongstaff: Same reason I made #4807 and #5088. Except now I just noticed that there is an inconsistency with the negative matchers, which also enforce a non-null contract. I was looking at the string case matchers when I opted to use `neverNullMatcher`.

sksamuel: should we get this merged in?
I think we'll look to release 6.2 next week

JordanLongstaff: Not until I know whether to keep these assertions as having nullable receivers. As is, `null.shouldNotBeUpperCaseChar` would fail, which is wrong in my point of view.

JordanLongstaff: I decided to make the case matchers' receivers non-nullable instead of using `neverNullMatcher`, but I kept the value matches (digit, letter, whitespace) the same. I think this is ready to merge, barring any objections.

JordanLongstaff: On secont thought, I'm going to make the rest non-nullable too. I'm basing this off the non-nullable receivers of the Kotlin functions, e.g. `Char.isDigit`, `Char.isUpperCase`.

sksamuel: Once green I think we're good - @AlexCue987  ?

JordanLongstaff: Apparently there's a compiler error I have to fix.

sksamuel: gotcha

AlexCue987: > Once green I think we're good - @AlexCue987 ?

LGTM

AlexCue987 (approved):
Nice

Issue #5923, opened 2026-04-22
shouldBeFailure with both type and block


jlous:
I like to maintain a strict separation between the act and assert phases, so I avoid `shouldThrow` and prefer `runCatching` + `shouldBeFailure`

This works well - shouldBeFailure() is excellent, offering both this:

```
result.shouldBeFailure<MyExceptionType>()
```
and this: 
```
result.shouldBeFailure { e ->
    e.shouldBeInstanceOf<MyExceptionType> {
        it.myCustomExceptionField shouldBe expectedValue
     }
}
```

But I am an ungrateful and greedy child, so I would like to combine them and do both at once:
```
result.shouldBeFailure<MyExceptionType>{ e ->
    e.myCustomExceptionField shouldBe expectedValue
}

alfonsoristorato:
very greedy, especially cause you could have called `apply` - let's see if the PR flies 😄

alfonsoristorato:
Closing it, it will come in a next release :)

PR #5927, opened 2026-04-24
allow passing a block asserting on T when calling shouldBeFailure

alfonsoristorato:
- allow passing a block asserting on T when calling shouldBeFailure

For https://github.com/kotest/kotest/issues/5923


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/result/FailureMatchersTest.kt
```diff
@@ -117,11 +117,26 @@ class FailureMatchersTest : FunSpec({
          shouldThrowAny { failure<String>(DummyException2).shouldBeFailure<DummyException3>() }
          shouldThrowAny { success("abc").shouldBeFailure<DummyException2>() }
       }
+
+      test("Should be failure with reified type and test block") {
```
AlexCue987, line 121: I think this duplicates already existing
`inline fun <reified T : Throwable> shouldThrow( ....): T`

alfonsoristorato: This is needed as an extension of 'Result' as per the issue in the description

AlexCue987 (approved):

Issue #5924, opened 2026-04-23
False negative: Intellij test runner shows test as green if it fails at container level


jlous:
kotest plugin 6.1.13
Intellij 2026.1

IntelliJ test runner shows test as green when failing at container level:
```
class FooTest : FreeSpec({
    "root" - {
        1 shouldBe 2
        "leaf 1a" {
        }
    }
})
```

<img width="293" height="181" alt="Image" src="https://github.com/user-attachments/assets/d2dbcf21-523b-4f8c-bae4-e30d98c8845e" />

There is an "AssertionFailedError" label buried there, but the entire test looks green.
This is extremely misleading.

Notes:
- Varying the Isolation mode does not affect the behavior.
- Other spec styles have the same problem.
- From maven/surefire the test fails like it should.

alfonsoristorato:
very interesting that this happens only for containers that assert on stuff before inner tests - thanks 😄

alfonsoristorato:
Closing this, will be fixed in next release

jlous:
would that mean 6.1.12?

alfonsoristorato:
It'll probably be 6.2.0, as we are getting close to release it

sksamuel:
6.2 released

jlous:
Confirmed fix

PR #5928, opened 2026-04-24
pass exception to TEST_FAILED placeholder to ensure that TeamCity can represent the test as a failure

alfonsoristorato:
For https://github.com/kotest/kotest/issues/5924

image

image

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/TeamCityTestEngineListener.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/TeamCityTestEngineListener.kt
@@ -143,17 +143,15 @@ class TeamCityTestEngineListener(
    // a dummy "test" to tag the error against that
    private fun insertPlaceholderTest(testPath: String, t: Throwable) {
 
-      val (name, cause) = ExtensionExceptionExtractor.resolve(t)
+      val (_, cause) = ExtensionExceptionExtractor.resolve(t)
 
       TeamCityMessage(prefix, TeamCityMessage.Types.TEST_STARTED) {
          name(testPath)
       }.output()
 
-      // we must print out the stack trace in between the dummy, so it appears when you click on the test name
-      cause.printStackTrace()
-
       TeamCityMessage(prefix, TeamCityMessage.Types.TEST_FAILED) {
          name(testPath)
+         exception(cause, details)
       }.output()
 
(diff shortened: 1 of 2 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/TeamCityTestEngineListener.kt
```diff
@@ -143,17 +143,15 @@ class TeamCityTestEngineListener(
...
 
-      val (name, cause) = ExtensionExceptionExtractor.resolve(t)
+      val (_, cause) = ExtensionExceptionExtractor.resolve(t)
 
       TeamCityMessage(prefix, TeamCityMessage.Types.TEST_STARTED) {
          name(testPath)
       }.output()
 
-      // we must print out the stack trace in between the dummy, so it appears when you click on the test name
-      cause.printStackTrace()
```
alfonsoristorato, line 153: because we now pass the exception, we no longer need this hack - unless it was needed for other targets?

sksamuel:
This is for non junit platforms right?

alfonsoristorato:
> This is for non junit platforms right? 

I think junit suffers from the same? At least the issue was replicable in the maven module of our examples

sksamuel:
The team city listener is not used by junit only for kmp

alfonsoristorato:
> The team city listener is not used by junit only for kmp

Is our `kotest-maven` setup wrong then you reckon? Cause we clearly import junit, and the fact that this fix makes it work makes me believe that TeamCity is still used?

sksamuel:
If you are running from intellij then it's using the java fork exec and not maven

alfonsoristorato:
> If you are running from intellij then it's using the java fork exec and not maven 

Yeah, which is how the kotest producer runs tests, so that still uses TeamCity I gather

sksamuel:
Yes. But mvn clean test etc won't use this producer. I guess if the bug is limited to only running single tests in intellij then yeah this is the fix

alfonsoristorato:
> Yes. But mvn clean test etc won't use this producer. I guess if the bug is limited to only running single tests in intellij then yeah this is the fix 

Ah yes, even in the issue description it says that the test runs as expected when using mvn.

sksamuel:
> > Yes. But mvn clean test etc won't use this producer. I guess if the bug is limited to only running single tests in intellij then yeah this is the fix
> 
> Ah yes, even in the issue description it says that the test runs as expected when using mvn.

I should read the fucking descriptions then shouldn't I 😂

sksamuel (approved):

PR #5929, opened 2026-04-28
Drive IntelliJ jump-to-source via locationUrl instead of displayName tag

sksamuel:
## Summary

- The engine previously prefixed every nested test's JUnit displayName with `<kotest>path/to/test</kotest>`. The Kotest IntelliJ plugin stripped the tag on the IDE side, but anyone running tests in IntelliJ **without** the plugin saw the raw mangled name in the test tree.
- The same path is already carried by the `MethodSource` attached to each leaf descriptor (`className=fqn`, `methodName=seg/seg/...`), which IntelliJ's JUnit5 launcher exposes as `proxy.locationUrl = "java:test://<fqn>/<seg>/..."`. The plugin now navigates from that URL, so the displayName can stay clean for everyone.
- Backward compatibility is preserved: the new plugin still strips and acts on the legacy `<kotest>` tag when running against older engines.

### Engine
- `descriptors.kt` no longer injects the `<kotest>...</kotest>` tag into the displayName.
- `LocationEmbedder` is kept (now `@Deprecated`) so older callers still compile.

### Plugin
- `EmbeddedLocationSMTRunnerEventsAdapter` checks the legacy displayName tag first (so the new plugin still gives navigation against older engine versions), then falls back to parsing the locationUrl.
- New `EmbeddedLocationParser.parseLocationUrl` handles `java:test://` and `java:suite://` URLs, gating on the methodName containing a `/` so we don't trample the default locator on regular JUnit `@Test` methods.

### Test
- Added `NestedTestMethodSourceTest` which runs a spec with deeply nested tests through JUnit Platform's `EngineTestKit` and pins the contract that every leaf descriptor carries a `MethodSource(<fqn>, <seg>/<seg>/...)`. If JUnit ever changes how it propagates `TestSource` for dynamic tests, this will fail loudly and we'll revisit the navigation strategy.

## Test plan

- [x] `./gradlew :kotest-runner:kotest-runner-junit-platform:jvmTest` — green
- [x] `./gradlew :kotest-intellij-plugin:test` — green (legacy tag-parsing tests + new locationUrl tests)
- [x] `./gradlew :kotest-framework:kotest-framework-engine:apiCheck` — green
- [ ] Manually verify in IntelliJ:
  - [ ] new engine + new plugin: clean names + jump-to-source for nested tests
  - [ ] new engine + no plugin: clean names (no jump for nested, lands on spec class)
  - [ ] old engine + new plugin: tag stripped + jump-to-source still works


alfonsoristorato (approved):
nice - tested manually and nothing is broken 😄

PR #5985, opened 2026-05-01
Fix WordSpec config-form addWhen/addShould missing withDefaultAffixes

sksamuel:
## Summary

The config-form code paths in both `WordSpecRootScope` and `WordSpecWhenContainerScope` (`addWhen` and `addShould`) built the container name without `.withDefaultAffixes()`:

```kotlin
testName = TestNameBuilder.builder(name).withSuffix(" when").build(),  // missing .withDefaultAffixes()
```

The parallel non-config form (lines 30/78 of root, 42/70 of when-container) does apply it:

```kotlin
testName = TestNameBuilder.builder(name).withSuffix(" should").withDefaultAffixes().build(),
```

Net effect: `"a context".config(...) When { ... }` and `"a context".config(...) should { ... }` ended up with `defaultAffixes = false` on their `TestName`, so reporters that respect default affixes rendered them differently from their non-config siblings.

Same family as the previously-fixed [BehaviorSpec missing-affixes issue (#5975)](https://github.com/kotest/kotest/pull/5975).

Note: the `addShould` lines still pass `" when"` as the suffix on master — that separate bug is fixed by [#5964](https://github.com/kotest/kotest/pull/5964) and is left alone in this PR (the affix change is orthogonal and composes cleanly).

## Test plan
- [ ] Compiles. Existing WordSpec tests still pass.


sksamuel:
Recreating from fresh master.

PR #5999, opened 2026-05-03
Add Amper-aware run configuration producer to the IntelliJ plugin

### LINKED ISSUE #5893: Amper green “play” icon in intellij error
```
Exception in thread "main" java.lang.ClassNotFoundException: MyTestClass
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:580)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:490)
	at java.base/java.lang.Class.forName0(Native Method)
	at java.base/java.lang.Class.forName(Class.java:467)
	at java.base/java.lang.Class.forName(Class.java:458)
	at io.kotest.engine.launcher.MainKt.main(main.kt:77)
```

[amper-kotest-issue-example.zip](https://github.com/user-attachments/files/26492441/amper-kotest-issue-example.zip)

sksamuel:
Closes #5893.

## Summary

Adds a `KotlinTestFrameworkProvider` so the **Amper IntelliJ plugin's existing `AmperKotlinTestConfigurationProducer`** recognises Kotest specs. With that one extension in place, the Amper plugin handles all of the run-config plumbing — discovery, building `amper test --include-classes <fqn>`, attaching to the Amper SMTRunner console — and we don't need to ship our own configuration type, factory, run-config, run state, editor, or producer.

## How it works

`AmperKotlinTestConfigurationProducer` extends `BaseAmperJvmTestConfigurationProducer`, whose `findTestClass` / `findTestMethod` / `findTestClassesInFile` iterate `org.jetbrains.kotlin.idea.testFrameworkProvider` extensions and ask each one whether a given PSI element identifies a test class. We register a provider that:

- Returns `true` from `isTestJavaClass` for any `KtLightClass` whose `KtClassOrObject` origin `isRunnableSpec()`.
- Always returns `false` from `isTestJavaMethod` (Kotest tests are DSL invocations, not methods, and `amper test` only exposes class-level filtering anyway).
- Scopes itself to Amper-managed modules via `AmperUtils.isAmperModule(...)` (detects `module.yaml`) so it does not influence Kotlin test discovery in non-Amper projects.

## What's also added

- `RunnerMode.AMPER` in `mode.kt`. The existing IDEA-flavour producers (`SpecPlatformRunConfigurationProducer`, `TestPlatformRunConfigurationProducer`, `PackageRunConfigurationProducer`) already condition on `mode == LEGACY`, so they cleanly skip Amper modules — the broken `ClassNotFoundException` config no longer gets created.
- `AmperUtilsTest` for the module-detection logic.

## Diff size

5 files, +207 lines.

## Known limitation

The Amper CLI exposes `--include-classes` but no Kotest test-path or JUnit method filter, so a single-test gutter click runs the whole enclosing spec. Tracked as a follow-up.

## Test plan

- [x] `./gradlew :kotest-intellij-plugin:test` — 232 tests, all green (228 existing + 4 new).
- [ ] Manual smoke test in a sandboxed IDE against [kotest/kotest-examples#105](https://github.com/kotest/kotest-examples/pull/105) (the kotest-amper example).

### DIFF
--- a/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/amper/AmperUtils.kt
+++ b/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/amper/AmperUtils.kt
@@ -20,9 +20,6 @@ import com.intellij.openapi.vfs.VirtualFile
 internal object AmperUtils {
 
    private const val MODULE_FILE = "module.yaml"
-   private const val PROJECT_FILE = "project.yaml"
-   private const val WRAPPER_UNIX = "amper"
-   private const val WRAPPER_WIN = "amper.bat"
 
    /**
     * Returns true if [module] is an Amper-managed module — that is, any of its content roots
@@ -47,22 +44,6 @@ internal object AmperUtils {
          if (found != null) return found
       }
       return null
-   }
-
-   /**
-    * Returns the directory containing the Amper project root (the directory holding
-    * `project.yaml`, or the module root if there's no `project.yaml`). The wrapper script
-    * (`./amper`) lives at this location.
-    */
-   fun amperProjectRoot(module: Module?): VirtualFile? {
-      val moduleRoot = amperModuleRoot(module) ?: return null
-      var current: VirtualFile? = moduleRoot
-      while (current != null) {
-         if (current.findChild(PROJECT_FILE) != null) return current
-         current = current.parent
-      }
-      // No project.yaml found — for single-module projects, the module root IS the project root.
-      return moduleRoot
    }
 
    private fun findModuleYamlFrom(start: VirtualFile): VirtualFile? {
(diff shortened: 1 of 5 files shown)


kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/amper/AmperUtils.kt
```diff
@@ -0,0 +1,87 @@
...
+      }
+      // No project.yaml found — for single-module projects, the module root IS the project root.
+      return moduleRoot
+   }
+
+   /**
+    * Returns the `amper` wrapper script (the platform-appropriate one) for the project root,
+    * or null if neither variant is present.
+    */
+   fun amperWrapper(module: Module?): VirtualFile? {
```
alfonsoristorato: this is not used :) if I approve it will auto-merge so will approve later (unless you think it has a future use)

alfonsoristorato (approved):

kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/amper/AmperUtils.kt
alfonsoristorato: removed :)

PR #6008, opened 2026-05-09
Make Arb.stringPattern multiplatform

wilmveel:
## What

Make `Arb.Companion.stringPattern(pattern: String): Arb<String>` callable from `commonMain` of `kotest-property`, so it works on every Kotest target — JVM, JS, Wasm-JS, Wasm-WASI, and all Native targets including the four `androidnative*` variants.

## Why

`Arb.stringPattern` previously lived only in `jvmMain`, backed by `com.github.curious-odd-man:rgxgen:2.0`. The recent `community.flock.kotlinx.rgxgen:kotlin-rgxgen` port now publishes for every Kotlin/Native target Kotest builds, which closes the gap.

## How

- Adds `community.flock.kotlinx.rgxgen:kotlin-rgxgen:0.0.5` to the version catalog.
- Introduces `expect fun Arb.Companion.stringPattern` in `commonMain`.
- Keeps the existing `jvmMain` actual using the Java `rgxgen` (no behaviour change for JVM users; `@file:JvmName("StringsjvmKt")` preserves the existing JVM facade class name, so binary compat is intact).
- Adds a single `nonjvmMain` actual backed by `kotlin-rgxgen` covering every non-JVM target.
- Adds a `commonTest` smoke test exercising the matcher on all platforms.

## Verification

- `./gradlew :kotest-property:jsNodeTest --refresh-dependencies` — passes against `kotlin-rgxgen 0.0.5`.


sksamuel:
Can I request a change - lets deprecate the original stringPattern, and add a new Arb.pattern or Arb.regex that exclusively uses the new rxg-gen library. Then in a future release we can remove the java lib entirely.

wilmveel:
`kotlin-rgxgen 0.0.6` is published — the upstream blocker (flock-community/kotlin-rgxgen#8) is closed. PR is updated to the clean-break shape:

- `kotlin-rgxgen` `0.0.5` → `0.0.6` in the catalog.
- Moved `kotlin-rgxgen` from `nonjvmMain` to `commonMain` and dropped the `findByName("nonjvmMain")` workaround.
- Added `Arb.Companion.pattern(pattern: String)` in `commonMain`, backed by `community.flock.kotlinx.rgxgen.RgxGen` — works on every target, JVM included.
- Reverted `Arb.stringPattern` to JVM-only and annotated it `@Deprecated(WARNING, replaceWith = ReplaceWith("Arb.pattern(pattern)"))`. The Java `rgxgen` stays in `jvmMain` only as long as the deprecated function lives, then both go in the release after.
- Renamed the common smoke test `StringPatternCommonTest` → `PatternCommonTest`, pointed at `Arb.pattern`.
- Refreshed `kotest-property` API dump (adds `PatternKt`, keeps `StringsjvmKt` for binary compat).

`./gradlew :kotest-property:jvmTest -PjvmOnly=true` passes locally — 810 tests, 0 failures, 0 errors. `apiCheck` clean.

One scoped-out follow-up: `kotest-property/src/jvmMain/kotlin/io/kotest/property/arbitrary/uuid.kt` still backs `Arb.uuid()` with the deprecated `Arb.stringPattern` (6 internal call sites, all `@Suppress`-able). Migrating to `Arb.pattern` swaps RgxGen implementations on JVM, which is a subtle behavior change worth its own commit — happy to follow up if you'd like it in this PR or as a separate one.

sksamuel (approved):
Approved. Needs the conflict resolving though.

sksamuel (approved):

PR #6009, opened 2026-05-13
Propagate expected/actual diff values to property test errors (#5909)

### LINKED ISSUE #5909: property tests don't propose nice diff in IntelliJ

6.1.10

The following test :

```
class IntelliJ :
    FreeSpec({
        "checkAll doesn't display nice diff in IntelliJ" {
            checkAll(Arb.int()) { input ->
                input.shouldBeEqual(input + 1)
            }
        }
    })
```

doesn't display a nice message, it makes property tests harder than necessary to use.

sksamuel:
## Summary

- When a matcher with expected/actual values (e.g. `shouldBeEqual`, `shouldBe`) failed inside `checkAll` / `forAll`, the outer `AssertionError` built by the property runner discarded those values. IntelliJ's OpenTest4J integration uses `AssertionFailedError.expected/actual` to render the "click to see difference" link, so the diff link was missing on property test failures.
- The outer error is now constructed directly via `AssertionFailedError(message, expected, actual, cause)` on the JVM (and `KotestAssertionFailedError` on non-JVM), preserving the diff values from the cause without modifying the message text.

Fixes #5909

## Test plan
- [x] Added `PropertyAssertionFailedErrorTest` covering both the diff-bearing and non-diff cases.
- [x] `:kotest-property:jvmTest` passes — existing `ForAll2Test` message-format expectations still match.
- [x] `compileKotlinJs`, `compileKotlinLinuxX64`, `compileKotlinWasmJs` succeed for the non-JVM actual.
- [ ] Verify the diff link renders in IntelliJ on the repro snippet from the issue.

### DIFF
--- a/kotest-property/src/nonjvmMain/kotlin/io/kotest/property/internal/errors.nonjvm.kt
+++ b/kotest-property/src/nonjvmMain/kotlin/io/kotest/property/internal/errors.nonjvm.kt
@@ -1,0 +1,19 @@
+package io.kotest.property.internal
+
+import io.kotest.assertions.KotestAssertionFailedError
+
+internal actual fun createPropertyAssertionError(message: String, cause: Throwable): AssertionError {
+   // if the underlying failure carries expected/actual values, propagate them to the outer error
+   // so IDEs and TeamCity service messages can render a diff for the outer error
+   val kotestCause = cause as? KotestAssertionFailedError
+   return if (kotestCause != null) {
+      KotestAssertionFailedError(
+         message = message,
+         cause = cause,
+         expected = kotestCause.expected,
+         actual = kotestCause.actual,
+      )
+   } else {
+      AssertionError(message, cause)
+   }
+}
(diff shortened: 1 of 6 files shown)


kotest-property/src/nonjvmMain/kotlin/io/kotest/property/internal/errors.nonjvm.kt
```diff
@@ -0,0 +1,19 @@
+package io.kotest.property.internal
+
+import io.kotest.assertions.KotestAssertionFailedError
+
+internal actual fun createPropertyAssertionError(message: String, cause: Throwable): AssertionError {
+   // if the underlying failure carries expected/actual values, propagate them to the outer error
+   // so IDEs and TeamCity service messages can render a diff for the outer error
+   val kotestCause = cause as? KotestAssertionFailedError
```
AlexCue987, line 8: would the following accomplish the same and be simpler and more idiomatic Kotlin:

```
return when(cause) {
  is KotestAssertionFailedError -> KotestAssertionFailedError(
         message = message,
         cause = cause,
         expected = kotestCause.expected,
         actual = kotestCause.actual,
      )
   else -> AssertionError(message, cause)
}
```

AlexCue987 (approved):
LGTM, suggested one improvement

PR #6010, opened 2026-05-13
Add per-call Eq overrides via withEqs DSL

### LINKED ISSUE #5910: Custom Eq should be defined at the spec and/or test level

I would like to use Custom Eq Instances as documented here https://kotest.io/docs/assertions/shouldbe.html#custom-eq-instances for a specific test suite.

But it states that:

> We could do this at the spec level too, but bear in mind if you are running tests in parallel then the registration will be non-deterministic.

I guess Eq being a typeclass, it doesn't really make sense for it to be defined globally.

By the way, I certainly don't want my test suite to have any impact on existing tests.

So maybe we could pass a list of `Eq` instances in a variant of `shouldBe`?

```
actual shouldBeUsingEq(This::class to ThisEq, That::class to ThatEq) expected
```

PreAgile:
## Summary

`DefaultEqResolver.register(...)` mutates a shared map under the hood, so any per-spec custom `Eq` registration races as soon as specs run in parallel — the existing docs already call this out:

> *We could do this at the spec level too, but bear in mind if you are running tests in parallel then the registration will be non-deterministic.*

This adds a per-call `withEqs { ... } shouldBe expected` DSL that scopes overrides to a single comparison and leaves the global resolver alone.

```kotlin
val a = mapOf("USD" to BigDecimal("3.14"), "EUR" to BigDecimal("2.99"))
val b = mapOf("USD" to BigDecimal("3.140"), "EUR" to BigDecimal("2.990"))

a withEqs {
   register<BigDecimal>(BigDecimalIgnoreScaleEq)
} shouldBe b
```

## How it works

`EqContext` gains a `resolver: EqResolver` field (default `DefaultEqResolver`, plus a backward-compat constructor with the old single-arg shape). `EqCompare.compare` now goes through `context.resolver.resolve(...)` instead of calling `DefaultEqResolver` directly. `withEqs` wraps the actual value in a `WithEqs<T>` together with the override map, and the matching `shouldBe` builds an `EqContext` carrying a `LayeredEqResolver(overrides, fallback = DefaultEqResolver)`.

Because the context is already threaded through every recursive compare in `MapEq`, `CollectionEq`, `DataClassEq`, `ArrayEq`, and `MapEntryEq`, overrides propagate into nested values automatically — no extra plumbing in those classes.

`LayeredEqResolver` only applies an override when `actual::class == expected::class`, mirroring the type-match guard `DefaultEqResolver` uses for its own custom registrations. Existing cross-type dispatch (e.g. `null` vs non-null, `Int` vs `Long`) keeps working.

This is the infix-builder variant of shape #2 from the issue discussion — same infix surface, but a reified `register<T>(eq)` builder for type safety instead of vararg `Pair<KClass<*>, Eq<*>>`. Spec-level scoping can layer on top in a follow-up PR if larger suites need it.

## Test plan

- [x] New `WithEqsTest` covers: scoping (override doesn't leak to subsequent comparisons), nested data-class / Map / Collection propagation, deep nesting, failure path, precedence over a global registration, and the runtime-type-mismatch fallback.
- [x] Existing `CustomEqTest` (6 tests) and the rest of `:kotest-assertions:kotest-assertions-core:jvmTest` stay green — backward-compat for `DefaultEqResolver.register` / `unregister` is preserved.
- [x] `./gradlew :kotest-assertions:kotest-assertions-core:apiDump` — added entries are the new public surface only (`EqContext` extra constructor + `resolver` getter, `EqOverrides`, `LayeredEqResolver`, `WithEqs`, top-level `withEqs` / `shouldBe`).

Fixes https://github.com/kotest/kotest/issues/5910

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/WithEqsTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/WithEqsTest.kt
@@ -159,6 +159,16 @@ class WithEqsTest : FunSpec() {
             DefaultEqResolver.unregister(BigDecimal::class)
          }
       }
+
+      test("register can be called multiple times for the same type — last call wins") {
+         val a = BigDecimal("1.00")
+         val b = BigDecimal("1")
+
+         a withEqs {
+            register<BigDecimal>(RejectAllBigDecimalsEq)
+            register<BigDecimal>(BigDecimalIgnoreScaleEq)
+         } shouldBe b
+      }
    }
 }
 
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/WithEqsTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/WithEqsTest.kt
@@ -6,6 +6,7 @@ import io.kotest.assertions.eq.Eq
 import io.kotest.assertions.eq.EqContext
 import io.kotest.assertions.eq.EqResult
 import io.kotest.assertions.eq.shouldBe
+import io.kotest.assertions.eq.shouldNotBe
 import io.kotest.assertions.eq.withEqs
 import io.kotest.assertions.throwables.shouldThrowAny
 import io.kotest.core.spec.style.FunSpec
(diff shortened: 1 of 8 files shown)


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/WithEqsTest.kt
```diff
@@ -0,0 +1,172 @@
...
+
+      test("withEqs shouldNotBe override wins over a globally registered Eq for the same type") {
+         try {
+            DefaultEqResolver.register(BigDecimal::class, RejectAllBigDecimalsEq)
+
+            BigDecimal("1") shouldNotBe BigDecimal("1")
+
+            shouldThrowAny {
+               BigDecimal("1.00") withEqs {
+                  register<BigDecimal>(BigDecimalIgnoreScaleEq)
```
AlexCue987, line 155: what happens if we accidentally register for the same type twice:

```
register<BigDecimal>(BigDecimalIgnoreScaleEq)
register<BigDecimal>(SomeOtherEq)
```

do we need a test to verify this edge case?

kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/eq/WithEqsTest.kt
```diff
@@ -0,0 +1,172 @@
...
+               } shouldNotBe BigDecimal("1")
+            }
+         } finally {
+            DefaultEqResolver.unregister(BigDecimal::class)
+         }
+      }
+   }
```
AlexCue987, line 172: the whole point of this PR is thread safety. do we need any tests that demonstrate it? this could be easily done via `runInParallel`, examples here 

https://github.com/kotest/kotest/blob/fa86b5f7fa700287541e9541000a483db9b82446/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/io/kotest/raceConditions/ParallelRunnerTest.kt#L8

AlexCue987 (approved):
great idea and implementation, suggested a few tests, which could be added in this or some next PR. I can add test(s) using `runInParallel` as well

PreAgile:
Thanks for the review.

I went with last-wins for duplicate registrations, documented it, and added a regression test in this PR.

For the `runInParallel` test, I left that for follow-up since you suggested it as well.
Happy to send a follow-up PR for it if you'd like.

AlexCue987:
> Thanks for the review.
> 
> I went with last-wins for duplicate registrations, documented it, and added a regression test in this PR.
> 
> For the `runInParallel` test, I left that for follow-up since you suggested it as well. Happy to send a follow-up PR for it if you'd like.

Yeah let's add a test leveraging `runInParallel` in the next PR.

PR #6011, opened 2026-05-18
Add Robolectric integration

sksamuel:
- Introduce `:kotest-tests:kotest-tests-robolectric` for testing Robolectric integration.
- Add new dependencies for Robolectric and AndroidX in `libs.versions.toml`.
- Implement example `WelcomeActivity` and its corresponding `WelcomeActivityTest`.


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/contain.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("RETURN_VALUE_NOT_USED_COERCION")
```
alfonsoristorato, line 1: while you move things around, I have a qq.
What do these solve? I tried removing them and I don't get any IDE error nor compilation

sksamuel: Intellij shows me a warning. Do you not get them?

alfonsoristorato: I do now that I patiently waited for IJ to load the project :) soz

LeoColman:
Could you publish this on a SNAPSHOT version? I'd like to test on my android projects before approving for production

PR #6028, opened 2026-05-27
Add shouldBeSingle matcher

### LINKED ISSUE #6027: `shouldBeSingleton()` return the singleton
```
Any opinions about making .shouldBeSingleton() actually return the singleton instead of the original (single-valued) collection? That would allow for nicer chaining of assertions, like .shouldBeSingleton().shouldBeSuccess().
```

https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1779894068165469

sksamuel:
## Summary

Fixes #6027 

- Adds a `shouldBeSingle` matcher on `Collection<T>`, `Iterable<T>`, and `Array<T>` that asserts the receiver contains exactly one element and **returns** that element (mirroring stdlib `Collection.single()`).
- Unlike `shouldBeSingleton`, which returns the receiver to allow chaining, `shouldBeSingle` returns the sole element so it can be captured at the call site:

  ```kotlin
  val onlyUser = users.shouldBeSingle()
  onlyUser.name shouldBe "Alice"
  ```

- On failure (0 or 2+ elements) the matcher raises an `AssertionError` with the same message produced by `shouldHaveSize 1`, rather than the `NoSuchElementException` / `IllegalArgumentException` thrown by the stdlib counterpart.

## Test plan

- [x] `./gradlew :kotest-assertions:kotest-assertions-core:jvmTest --tests "*CollectionMatchersTest"` passes locally, including the new `shouldBeSingle` block.
- [x] `./gradlew :kotest-assertions:kotest-assertions-core:apiDump` regenerated; updated `.api` committed.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/singleton.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/singleton.kt
@@ -89,6 +89,37 @@ fun <T> Array<T>.shouldBeSingle(): T = asList().shouldBeSingle()
 
 
 /**
+ * Verifies this collection does not contain exactly one element.
+ *
+ * This will pass for empty collections and for collections with two or more elements.
+ * Mirrors the negated form of [shouldBeSingle]; returns the receiver to allow chaining.
+ *
+ * ```
+ * listOf<Int>().shouldNotBeSingle()   // Assertion passes
+ * listOf(1, 2).shouldNotBeSingle()    // Assertion passes
+ * listOf(1).shouldNotBeSingle()       // Assertion fails
+ * ```
+ *
+ * @see [shouldBeSingle]
+ * @see [shouldNotBeSingleton]
+ */
+fun <T, C : Collection<T>> C.shouldNotBeSingle(): C {
+   this shouldNotHaveSize 1
+   return this
+}
+
+fun <T, I : Iterable<T>> I.shouldNotBeSingle(): I {
+   toList().shouldNotBeSingle()
+   return this
+}
+
+fun <T> Array<T>.shouldNotBeSingle(): Array<T> {
+   asList().shouldNotBeSingle()
+   return this
+}
+
+
+/**
  * Verifies this collection contains only one element and executes the given lambda against that element.
  */
 inline fun <T, C : Collection<T>> C.shouldBeSingleton(fn: (T) -> Unit): C {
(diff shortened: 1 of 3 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/singleton.kt
```diff
@@ -64,6 +64,30 @@ fun <T> Array<T>.shouldNotBeSingleton(): Array<T> {
...
+ * @see [shouldBeSingleton]
+ */
+fun <T, C : Collection<T>> C.shouldBeSingle(): T {
+   this shouldHaveSize 1
+   return this.first()
+}
+
+fun <T, I : Iterable<T>> I.shouldBeSingle(): T = toList().shouldBeSingle()
+
+fun <T> Array<T>.shouldBeSingle(): T = asList().shouldBeSingle()
```
alfonsoristorato, line 88: is it worth adding the `not` matchers, considering we have them for the `singleton` variant?

sksamuel: good spot!
done.

alfonsoristorato (approved):

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/singleton.kt
```diff
@@ -64,6 +64,61 @@ fun <T> Array<T>.shouldNotBeSingleton(): Array<T> {
...
+ * ```
+ * val only = listOf(1).shouldBeSingle() // returns 1
+ * listOf(1, 2).shouldBeSingle()         // Assertion fails
+ * ```
+ *
+ * @see [shouldBeSingleton]
+ */
+fun <T, C : Collection<T>> C.shouldBeSingle(): T {
+   this shouldHaveSize 1
+   return this.first()
```
sschuberth, line 83: Nit: Would it have been slightly more natural to use `single()` here?

sksamuel: Instead of first ?

sschuberth: Yes.

sksamuel: Updated https://github.com/kotest/kotest/commit/17f6979a005d7aab8bcae30a472353222faa8cf3#diff-46b96334b31adb7044c0bc825219ad7b21894735dc16d4d407a31e8fc0e7524aR83

PR #6033, opened 2026-05-28
 Fix data test run including sibling regular containers                                                                                                 

alfonsoristorato:
<!-- 
If this PR updates documentation, please update all relevant versions of the docs, see: https://github.com/kotest/kotest/tree/master/documentation/versioned_docs
The documentation at https://github.com/kotest/kotest/tree/master/documentation/docs is the documentation for the next minor or major version _TO BE RELEASED_
-->
**Summary**                                                                                                                                              

  When running a specific data test nested inside regular containers (e.g. `context("parent") { context("child") { withData(...) } }`), the IntelliJ plugin  generates a tag expression like `(kotest.data.LINE) | kotest.data.nonJvm`. The enclosing regular containers have no `kotest.data` tag, so without  intervention the tag filter would prevent them from running and the data test would never be discovered.                                               
                                                                                                                                                       
The previous approach appended `| !kotest.data` to the expression so untagged containers could pass the filter — but this matched every untagged test and container at those levels, causing sibling tests and sibling containers (e.g. a `describe("whatever test bame")` next to the `withXXX`) to run alongside the intended target.                                                                                                                                                
                                                                                                                                                       
**Changes:**                                                                                                                                               
   
  - Introduces `KOTEST_DATA_TEST_ANCESTOR_PATH`, an env var set by the plugin run producers to the -- separated path of the regular containers enclosing  the target data test (e.g. `"parent context -- child context"`). Only set when the data test is actually inside regular containers, omitted for root-level data tests.                                                                                                                                 
  - In `TagsEnabledExtension`, when a container fails tag filtering, its path-from-root is built by walking `TestCase.parent` and joining raw `name.name` values. The container is allowed through only if its path equals or is a strict prefix of the ancestor path. Meaning only the direct ancestors of the target data test are let through, while sibling containers at every level remain excluded.
  - Removes the `| !kotest.data` component from tag expressions generated by the plugin for data tests inside regular containers, as it is no longer  needed.

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/enabled/TagsEnabledExtension.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/enabled/TagsEnabledExtension.kt
@@ -4,6 +4,7 @@ import io.kotest.common.syspropOrEnv
 import io.kotest.core.Logger
 import io.kotest.core.test.Enabled
 import io.kotest.core.test.TestCase
+import io.kotest.core.test.TestType
 import io.kotest.engine.config.KotestEngineProperties
 import io.kotest.engine.config.TestConfigResolver
 import io.kotest.engine.tags.TagExpression
@@ -19,7 +20,13 @@ import io.kotest.engine.tags.parse
  * - Included tags have been specified, and this test either has no tags
  *   or does not have any of the specified inclusion tags.
  *
- *  Note: tags are attached to tests either through test config or at the spec level.
+ * Note: tags are attached to tests either through test config or at the spec level.
+ *
+ * **Data test ancestor path bypass**: when [KotestEngineProperties.KOTEST_DATA_TEST_ANCESTOR_PATH] is set
+ * (by the IntelliJ plugin when running a specific data test nested inside regular containers), containers
+ * with no explicit tags whose path-from-root is a prefix of — or exactly equals — that value are allowed
+ * through despite failing the tag filter. This ensures the engine traverses the ancestor containers needed
+ * to discover the target data test, without also running sibling containers at any level.
  */
 internal class TagsEnabledExtension(
    private val tags: TagExpression,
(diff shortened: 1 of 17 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/enabled/TagsEnabledExtension.kt
```diff
@@ -32,10 +39,44 @@ internal class TagsEnabledExtension(
...
+         val ancestorPath = syspropOrEnv(KotestEngineProperties.KOTEST_DATA_TEST_ANCESTOR_PATH)
+         if (ancestorPath != null
+            && testCase.type == TestType.Container
+            && testCase.config?.tags.isNullOrEmpty()
+         ) {
+            val containerPath = buildContainerPath(testCase)
+            if (ancestorPath == containerPath || ancestorPath.startsWith("$containerPath -- ")) {
+               return Enabled.enabled
+            }
+         }
```
alfonsoristorato, line 55: I am not a big fan of combining this check here, but I can't think of another place without duplicating some Extension code

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/enabled/TagsEnabledExtension.kt
```diff
@@ -32,10 +39,44 @@ internal class TagsEnabledExtension(
...
+    * [io.kotest.core.names.TestName.name] values with `" -- "`.
+    *
+    * Raw names (no spec-style prefix/suffix such as `"Given: "`) are used intentionally so the result
+    * matches the path the IntelliJ plugin builds from PSI source — which also extracts only the string
+    * literal passed to the container function.
+    *
+    * Example: `context("child context")` nested inside `context("parent context")` → `"parent context -- child context"`.
+    * TODO: this is a good candidate for the shared module between IJ plugin and Framework - as this code is done times and times again
+    */
+   private fun buildContainerPath(testCase: TestCase): String {
```
sksamuel: does this not exist in the descriptor classes ?

sksamuel (approved):
one comment to check before merge

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/test/enabled/TagsEnabledExtension.kt
alfonsoristorato: not a method per se, but the descriptor exposes parts, which I can then joinToString with the delimiter - thanks

Issue #5996, opened 2026-05-02
Support class discovery


sksamuel:
From slack:

I'm trying to use https://docs.junit.org/6.0.3/running-tests/console-launcher.html#options-executing-tests with the Kotest engine. But while Kotest as the engine is discovered, and also a "container" (whatever that is), no tests are discovered or executed when I use options execute --scan-classpath --include-engine=kotest. Any idea why that is? (More output in thread.)

https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1775840873011489

sschuberth:
Any chance to get this n for this week's 6.2 release, @sksamuel?

sksamuel:
yes, its one of the last couple of things before 6.2 is cut

zackman0010:
> yes, its one of the last couple of things before 6.2 is cut

@sksamuel Hi! Sorry for pinging you about this on a closed issue, but I do not have access to the Slack yet to ask there. Do you have any news on when 6.2 will be released after this was merged? I was searching for information on it and found this comment.

sksamuel:
6.2 is going out tomorrow.

PR #6034, opened 2026-05-28
Implement SpecScanner using ClassGraph

sksamuel:
## Overview

Closes #https://github.com/kotest/kotest/issues/5996

Implements `SpecScanner`, which uses the [ClassGraph](https://github.com/classgraph/classgraph) library to discover all **concrete, non-private** subclasses of `Spec` on the classpath, returning them as `KClass<out Spec>` instances.

This backs the launcher's `--specs scan` argument (already wired in `main.kt`), allowing the set of specs to execute to be discovered at runtime rather than passed explicitly.

## Changes

- **`SpecScanner.kt`** — scans the classpath via ClassGraph, filtering to concrete (`isStandardClass`, not `isAbstract`) and non-private classes, excluding noise packages (jdk/kotlin/etc.) via `rejectPackages`.
- **`libs.versions.toml`** — adds the `classgraph` version (`4.8.179`) and library entry.
- **`kotest-framework-engine/build.gradle.kts`** — adds `implementation(libs.classgraph)` to `jvmMain`.
- **`SpecScannerTest.kt`** — verifies concrete public specs are discovered, while abstract specs (e.g. `FunSpec`, `Spec`) and private specs are not.

## Testing

`./gradlew :kotest-framework:kotest-framework-engine:jvmTest --tests "io.kotest.engine.launcher.SpecScannerTest"` passes.

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/engine/launcher/SpecScannerTest.kt
+++ b/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/engine/launcher/SpecScannerTest.kt
@@ -18,14 +18,17 @@ private class PrivateConcreteSpecForScannerTest : FunSpec()
 
 @EnabledIf(LinuxOnlyGithubCondition::class)
 class SpecScannerTest : FunSpec() {
+
+   // scan the classpath once and reuse the result across all tests
+   private val specs = SpecScanner.scan()
+
    init {
 
       test("scan should discover concrete, public subclasses of Spec") {
-         SpecScanner.scan() shouldContain PublicConcreteSpecForScannerTest::class
+         specs shouldContain PublicConcreteSpecForScannerTest::class
       }
 
       test("scan should not return abstract subclasses of Spec") {
-         val specs = SpecScanner.scan()
          specs shouldNotContain AbstractSpecForScannerTest::class
          // the built in spec styles are themselves abstract subclasses of Spec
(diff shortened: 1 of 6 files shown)


kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/io/kotest/engine/launcher/SpecScannerTest.kt
```diff
@@ -0,0 +1,39 @@
...
+class PublicConcreteSpecForScannerTest : FunSpec()
+
+// an abstract subclass of Spec - should not be discovered as it cannot be instantiated
+abstract class AbstractSpecForScannerTest : FunSpec()
+
+// a private, concrete subclass of Spec - should not be discovered as the engine cannot instantiate it
+private class PrivateConcreteSpecForScannerTest : FunSpec()
+
+@EnabledIf(LinuxOnlyGithubCondition::class)
+class SpecScannerTest : FunSpec() {
```
alfonsoristorato, line 20: I'd say we can avoid doing `SpecScanner.scan()` for every test and set it as a val at the top?

sksamuel: Good call — done in 15ff785f5: `SpecScanner.scan()` is now hoisted into a single `private val specs` and reused across all three tests.

PR #6046, opened 2026-05-29
Parenthesize operands in TagExpression.combine to preserve precedence

sksamuel:
## Problem

`TagExpression.combine(other)` built the combined expression as `this.expression & other.expression` without parentheses. Because the `&` operator binds more tightly than `|`, combining an OR expression with another tag produced an incorrectly grouped result. For example, combining `"linux | mac"` with `"mysql"` yielded `"linux | mac & mysql"`, which parses as `"linux | (mac & mysql)"` instead of the intended `"(linux | mac) & mysql"`. This silently changed which tests were considered active/inactive at runtime.

## Fix

In `TagExpression.combine`, wrap each non-empty operand in parentheses, producing `"(this) & (other)"`. The empty-expression short-circuits are preserved unchanged: if one side is empty, the other side is returned as-is with no added parentheses. Only the `combine` function was modified.

```kotlin
else -> TagExpression("(" + this.expression + ") & (" + other.expression + ")")
```

## Verification

Verified by reading the code: `combine` is the single funnel used by `include`/`exclude` helpers and the precedence semantics now match the documented `(linux | mac) & mysql` example in the class KDoc. The empty-side branches are untouched, so simple single-tag chains are unaffected. Compilation is verified centrally by the author (no build/test was run in this worktree).

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/tags/TagExpression.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/tags/TagExpression.kt
@@ -32,7 +32,7 @@ data class TagExpression(val expression: String) {
    fun combine(other: TagExpression): TagExpression = when {
       this.expression == "" -> other
       other.expression == "" -> this
-      else -> TagExpression(this.expression + " & " + other.expression)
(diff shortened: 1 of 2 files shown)


alfonsoristorato (approved):
nice

PR #6063, opened 2026-05-29
Register spec-body afterProject listeners once per project, not once per instance

sksamuel:
## Problem

`SpecRefInflator.inflate()` in `SpecRef.kt` ran `spec.afterProjectListeners().forEach { registry.add(it) }` on **every** inflate. The `ExtensionRegistry` is engine-wide and its `add()` does not dedupe.

Under `InstancePerRoot` / `InstancePerLeaf` / `InstancePerTest` isolation, `inflate()` is called once per fresh spec instance, and each instance builds **new** `AfterProjectListener` objects from its `DslDrivenSpec.afterProject { }` blocks. As a result, a spec-body `afterProject { }` block was registered (and therefore executed) **once per instance** instead of once per project. `SingleInstance` mode (the default) was correct because the spec is inflated once.

## Fix

Dedupe afterProject registration by spec class, using state held on the engine-wide registry so it works across the multiple `SpecRefInflator` instances created by the isolation executors:

- Added `markAfterProjectListenersRegistered(kclass: KClass<*>): Boolean` to the `ExtensionRegistry` interface. It atomically records the spec class and returns `true` only the first time it is called for that class.
  - `DefaultExtensionRegistry` backs it with a `MutableSet<KClass<*>>` (also cleared in `clear()`).
  - `EmptyExtensionRegistry` returns `true` (it never actually stores extensions; behaviour unchanged).
- In `SpecRefInflator.inflate()`, only register the spec's afterProject listeners when `registry.markAfterProjectListenersRegistered(ref.kclass)` returns `true`.

This guarantees spec-body afterProject listeners are registered at most once per spec class (exactly once across all isolation modes, including SingleInstance). The sealing logic and all other behaviour are untouched. Listeners remain registered as global extensions so they are still picked up by `registry.all()` in `ProjectConfigResolver.extensions()`.

## Verification

- `ExtensionRegistry` has only two implementations (`DefaultExtensionRegistry`, `EmptyExtensionRegistry`); both updated. No anonymous/`by`-delegated implementations exist.
- `ref.kclass` is part of the sealed `SpecRef` interface, available for both `Reference` and `Function` refs.
- Compilation is verified centrally by the author. Note: `ExtensionRegistry` is public API, so `./gradlew apiDump` may need to be run to record the new interface method.

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/ExtensionRegistry.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/ExtensionRegistry.kt
@@ -34,6 +34,16 @@ interface ExtensionRegistry {
    fun remove(extension: Extension)
    fun remove(extension: Extension, kclass: KClass<*>)
 
+   /**
+    * Atomically marks the given spec [kclass] as having had its spec-level afterProject listeners
+    * registered, returning true only the first time it is called for a given class.
+    *
+    * This allows spec-level afterProject listeners to be registered globally exactly once per spec
+    * class, even when (under InstancePerRoot/InstancePerLeaf/InstancePerTest isolation) many fresh
+    * instances of the same spec are inflated, each building new afterProject listener objects.
+    */
+   fun markAfterProjectListenersRegistered(kclass: KClass<*>): Boolean
+
    fun clear()
    fun isEmpty(): Boolean
    fun isNotEmpty(): Boolean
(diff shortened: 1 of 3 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/ExtensionRegistry.kt
```diff
@@ -34,6 +34,16 @@ interface ExtensionRegistry {
...
 
+   /**
+    * Atomically marks the given spec [kclass] as having had its spec-level afterProject listeners
+    * registered, returning true only the first time it is called for a given class.
+    *
+    * This allows spec-level afterProject listeners to be registered globally exactly once per spec
+    * class, even when (under InstancePerRoot/InstancePerLeaf/InstancePerTest isolation) many fresh
+    * instances of the same spec are inflated, each building new afterProject listener objects.
+    */
+   fun markAfterProjectListenersRegistered(kclass: KClass<*>): Boolean
```
alfonsoristorato, line 45: seeing how `afterProjectRegisteredClasses` is a set, will the whole logic still hold and work as expected when we have concurrency in those isolation modes?

sksamuel: Gonna close this anyway. Claude is not wrong but this behavior is already documented as a limitation of per spec after project

PR #6064, opened 2026-05-29
Honor WithDataTestName.dataTestName() on all platforms, not just JVM

sksamuel:
## Problem

In `StableIdents.getStableIdentifier(t)`, the `when` had this branch ordering:

```kotlin
t == null -> "<null>"
t::class.hasAnnotation<IsStableType>() || platform != Platform.JVM -> t.toString()
t is WithDataTestName -> t.dataTestName()
```

Because `platform != Platform.JVM` was OR-ed into the second branch, on JS/Wasm/Native any value implementing `WithDataTestName` short-circuited to `t.toString()` and the `t is WithDataTestName -> t.dataTestName()` branch was never reached. `WithDataTestName.dataTestName()` is a plain interface method that needs no reflection and should work identically on every platform, so the custom data-test name was silently dropped off-JVM.

## Fix

Split the combined condition and reorder so `WithDataTestName` is handled before the non-JVM short-circuit, preserving the documented precedence (`IsStableType` over `WithDataTestName`):

```kotlin
t == null -> "<null>"
t::class.hasAnnotation<IsStableType>() -> t.toString()
t is WithDataTestName -> t.dataTestName()
platform != Platform.JVM -> t.toString()
else -> { /* JVM-only reflection-based stability analysis */ }
```

- `IsStableType` still short-circuits to `toString()` first (takes precedence over `WithDataTestName`, matching the `WithDataTestName` KDoc).
- `WithDataTestName.dataTestName()` is now invoked on all platforms.
- The genuinely JVM-only, reflection-based data-class stability analysis remains gated behind `platform != Platform.JVM -> t.toString()` for the non-`WithDataTestName`, non-`IsStableType` case.

Minimal change: 2 insertions, 1 deletion, no reformatting.

## Verification

Compilation is verified centrally by the author. The change is a pure branch reordering within an existing `when`; no new symbols or imports are introduced, and `dataTestName()` is already a common-source interface method, so it is available on all targets.

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonTest/kotlin/io/kotest/engine/stable/StableIdentsCommonTest.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonTest/kotlin/io/kotest/engine/stable/StableIdentsCommonTest.kt
@@ -34,7 +34,20 @@ class StableIdentsCommonTest : FunSpec({
       StableIdents.getStableIdentifier(element) shouldBe "stable-name"
    }
 
+   test("getStableIdentifier should use toString() for an @IsStableType-annotated type") {
+      // @IsStableType signals the type's toString() is stable and may be used as the identifier.
+      // On the JVM this is resolved via the IsStableType annotation branch; the resulting identifier
+      // (the toString()) is the same on every platform.
+      StableIdents.getStableIdentifier(StableTypeForTest("abc")) shouldBe "stable:abc"
+      StableIdents.getStableIdentifier(StableTypeForTest("xyz")) shouldBe "stable:xyz"
+   }
+
    test("null should be stable on all platforms") {
       StableIdents.getStableIdentifier(null) shouldBe "<null>"
    }
 })
+
+@IsStableType
+private data class StableTypeForTest(val value: String) {
+   override fun toString(): String = "stable:$value"
+}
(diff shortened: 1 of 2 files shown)


kotest-framework/kotest-framework-engine/src/commonTest/kotlin/io/kotest/engine/stable/StableIdentsCommonTest.kt
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Multiplatform coverage for [StableIdents.getStableIdentifier].
+ *
+ * These tests run on every target (JVM, JS, Wasm, Native). They guard against the regression
+ * where the non-JVM short-circuit (`platform != Platform.JVM -> toString()`) was evaluated
+ * before the [WithDataTestName] branch, causing a custom [WithDataTestName.dataTestName] to be
+ * silently ignored on all platforms except the JVM.
+ */
+class StableIdentsCommonTest : FunSpec({
```
alfonsoristorato, line 15: how about a further test with `@IsStableType` here?

sksamuel: Added in the latest commit — a multiplatform test asserting an `@IsStableType`-annotated type resolves to its `toString()`. Note: on non-JVM targets `hasAnnotation` is a no-op (BasicReflection returns no annotations), so the precedence of `@IsStableType` *over* `WithDataTestName` is JVM-only behaviour; the test asserts the toString() outcome, which holds on every platform.

sksamuel: Good idea - done.

alfonsoristorato (approved):

PR #6077, opened 2026-05-29
Path containNFiles should fail gracefully on a non-directory instead of throwing

sksamuel:
## Problem

The `Path`-based `containNFiles(n)` matcher in `kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/paths/nfiles.kt` unconditionally called `Files.newDirectoryStream(value).use { it.count() }`. When `value` is a regular file or does not exist, `newDirectoryStream` throws `NotDirectoryException` / `NoSuchFileException`. As a result, assertions such as `regularFile.shouldNotContainNFiles(3)` threw an exception instead of passing cleanly.

## Fix

Only open the directory stream when `value.isDirectory()`. Otherwise the file count defaults to `0` and the matcher returns a normal failed/negated `MatcherResult` (no exception). This mirrors the graceful, null-safe approach used by the sibling `File`-based `containNFiles` in `io/kotest/matchers/file/matchers.kt`, which checks `isDirectory` and never throws. The `isDirectory()` result is now computed once and reused in both the predicate and the failure message (which already referenced `isDir`).

## Verification

Compilation is verified centrally by the parent build. The change is minimal and JVM-only, preserving existing message wording.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/paths/nfiles.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/paths/nfiles.kt
@@ -13,10 +13,11 @@ infix fun Path.shouldNotContainNFiles(n: Int) = this shouldNot containNFiles(n)
 
 fun containNFiles(n: Int): Matcher<Path> = object : Matcher<Path> {
    override fun test(value: Path): MatcherResult {
-      val count = Files.newDirectoryStream(value).use { it.count() }
+      val isDir = value.isDirectory()
+      val count = if (isDir) Files.newDirectoryStream(value).use { it.count() } else 0
       return MatcherResult(
-         value.isDirectory() && count == n,
-         { "$value should be a directory and contain $n files (isDir = ${value.isDirectory()}; file count = $count)" },
+         isDir && count == n,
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/paths/nfiles.kt
```diff
@@ -13,10 +13,11 @@ infix fun Path.shouldNotContainNFiles(n: Int) = this shouldNot containNFiles(n)
 
 fun containNFiles(n: Int): Matcher<Path> = object : Matcher<Path> {
    override fun test(value: Path): MatcherResult {
-      val count = Files.newDirectoryStream(value).use { it.count() }
+      val isDir = value.isDirectory()
+      val count = if (isDir) Files.newDirectoryStream(value).use { it.count() } else 0
```
alfonsoristorato, line 17: worth adding a test for the case where `isDir` is false.

alfonsoristorato (approved):

kotest-assertions/kotest-assertions-core/src/jvmMain/kotlin/io/kotest/matchers/paths/nfiles.kt
sksamuel: done

PR #6082, opened 2026-05-29
Gradle test-filter spec-prefix match must respect a path boundary

sksamuel:
## Problem

In `GradleClassMethodRegexTestFilter`, the `isSpecPrefix` check matched a `SpecDescriptor` against a `--tests` pattern using a raw `startsWith`:

```kotlin
pattern.removePrefix("\\Q").removeSuffix("\\E").startsWith(descriptor.id.value)
```

This has no path boundary, so filtering with `--tests io.pkg.FooBar.someContext` also matched a sibling spec `io.pkg.Foo`, because the cleaned pattern string starts with `io.pkg.Foo`. Sibling specs whose id is a string prefix of an unrelated, longer pattern were incorrectly included.

## Fix

In `kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/gradle/GradleClassMethodRegexTestFilter.kt`, the `isSpecPrefix` computation now requires either an exact match or a match up to a path separator:

```kotlin
val p = pattern.removePrefix("\\Q").removeSuffix("\\E")
val id = descriptor.id.value
p == id || p.startsWith("$id.")
```

`io.pkg.Foo` no longer matches `io.pkg.FooBar.someContext` (since `io.pkg.FooBar.someContext` neither equals `io.pkg.Foo` nor starts with `io.pkg.Foo.`), while genuine spec+method patterns like `io.pkg.Foo.someContext` still match. The rest of the `match()` logic is unchanged.

## Verification

Compilation is verified centrally by the parent build. The change is a minimal, behavior-preserving narrowing of one boolean condition; no other code paths were touched.

### DIFF
--- a/kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/gradle/GradleClassMethodRegexTestFilter.kt
+++ b/kotest-runner/kotest-runner-junit-platform/src/jvmMain/kotlin/io/kotest/runner/junit/platform/gradle/GradleClassMethodRegexTestFilter.kt
@@ -68,7 +68,11 @@ internal class GradleClassMethodRegexTestFilter(private val patterns: Set<String
       // matches a spec descriptor when we have a spec + method pattern
       val isSpecPrefix by lazy {
          if (descriptor !is Descriptor.SpecDescriptor) false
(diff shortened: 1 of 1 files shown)


alfonsoristorato (approved):
worth a test for this

PR #6085, opened 2026-05-29
Keep the ISO-8601 Z designator in the JUnit XML timestamp

sksamuel:
**Problem**

In `JUnitXmlReportGenerator`, the `<testsuite>` `timestamp` was computed as `clock.now().toString().substringBeforeLast(".")`. The intent was to drop fractional seconds, but `substringBeforeLast(".")` also strips the trailing `Z` UTC designator whenever a sub-second component is present (e.g. `2023-01-01T12:00:00.123Z` becomes `2023-01-01T12:00:00`). The result is a non-ISO-8601 timestamp in the generated XML report. `kotlin.time.Instant.toString()` is always UTC (ends in `Z`) and only contains a `.` when there is a sub-second component, so the existing whole-second test clock never exercised the buggy path.

**Fix**

Compute the timestamp once and only split on `.` when present, re-appending `Z`:

```kotlin
val now = clock.now().toString()
val timestamp = if (now.contains('.')) now.substringBefore('.') + "Z" else now
```

Whole-second instants are passed through unchanged (still ending in `Z`); sub-second instants drop the fractional part while preserving `Z`. The XML structure is otherwise unchanged.

**Verification**

Added a focused test `should drop fractional seconds but keep the ISO-8601 Z designator in the timestamp` using a `FixedClock` seeded with `2023-01-01T12:00:00.123Z`. It asserts the output contains `timestamp="2023-01-01T12:00:00Z"` and does not contain the Z-less `timestamp="2023-01-01T12:00:00"`. The four existing tests already assert the `...Z` form for whole-second clocks and remain valid. Tests were not run locally (parent compiles and runs centrally).

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/reports/JUnitXmlReportGenerator.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/reports/JUnitXmlReportGenerator.kt
@@ -7,6 +7,7 @@ import nl.adaptivity.xmlutil.core.XmlVersion
 import nl.adaptivity.xmlutil.serialization.XML
 import kotlin.reflect.KClass
 import kotlin.time.Clock
+import kotlin.time.Instant
 
 /**
  * @target the Kotlin KMP target this report is generated for. If set will be prefixed to each test name.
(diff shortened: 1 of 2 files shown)


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/reports/JUnitXmlReportGenerator.kt
```diff
@@ -29,13 +29,17 @@ class JUnitXmlReportGenerator(
    }
 
    private fun generate(spec: KClass<*>, tests: Map<TestCase, TestResult>): TestSuite {
+      // Instant.toString() is ISO-8601 in UTC, always ending in 'Z', and only contains a '.' when there
+      // is a sub-second component. Drop fractional seconds while preserving the 'Z' UTC designator.
+      val now = clock.now().toString()
+      val timestamp = if (now.contains('.')) now.substringBefore('.') + "Z" else now
```
alfonsoristorato: java has a very handy `truncatedTo(ChronoUnit.SECONDS)`, I guess for Kotlin we can do `Instant.fromEpochSeconds(now.epochSeconds).toString()` if we don't want to deal with substrings

sksamuel: Very nice, updated to use that :)

alfonsoristorato (approved):

PR #6102, opened 2026-06-01
SpringExtension: forward test exception to the test context manager (#6094)

### LINKED ISSUE #6094: SpringExtension does not forward thrown exception to the test context manager
I am using Kotest version 6.1.11

When I test a Spring application using Kotest SpringExtension, and a Camunda exception is thrown during the test, Camunda listener does not display information about this exception in its report.

The cause is as follows. After the test is finished, SpringExtensiton calls `afterTestMethod` of the test context manager, but if the test result contains an exception, the test context manager remains unaware of this and thus Camunda listener does not display information about it, because SpringExtension always forwards `null` as the argument for the `exeption` parameter ([line 95](https://github.com/kotest/kotest/blob/master/kotest-extensions/kotest-extensions-spring/src/jvmMain/kotlin/io/kotest/extensions/spring/SpringExtension.kt#L95)):
**Existing implementation**
```
testContextManager().afterTestMethod(testCase.spec, methodName, null as Throwable?)
```
Instead, the exception should be extracted from the test result and forwarded as the argument, i.e. the line should be as follows:
**Requested implementation**
```
testContextManager().afterTestMethod(testCase.spec, methodName, result.errorOrNull)
```
Then the listener gets the thrown exception and generates an according report.

sksamuel:
Fixes #6094

## Problem

When a test fails under `SpringExtension`, the thrown exception was **never forwarded** to Spring's `TestContextManager`. Both after-callbacks hard-coded `null`:

```kotlin
testContextManager().afterTestMethod(testCase.spec, methodName, null as Throwable?)
...
testContextManager().afterTestExecution(testCase.spec, methodName, null as Throwable?)
```

As a result, Spring `TestExecutionListener`s (e.g. Camunda's) read `testContext.getTestException()` as `null` and couldn't report the failure.

## Fix

Forward the exception from the test result into both callbacks:

```kotlin
testContextManager().afterTestMethod(testCase.spec, methodName, result.errorOrNull)
...
testContextManager().afterTestExecution(testCase.spec, methodName, result.errorOrNull)
```

(`afterTestExecution` lives in `afterAny(testCase, result)`, which already has the result.)

## Test

Adds `SpringExtensionTestExceptionTest`, which runs a deliberately-failing Spring spec **in-process** via `TestEngineLauncher` (so its failure is collected, not propagated) and registers a `TestExecutionListener` that records `testContext.testException`. It asserts the thrown `TestMethodBoomException` reaches both `afterTestMethod` and `afterTestExecution`.

## Verification

- New test passes with the fix; **fails without it** (verified by reverting both lines → `AssertionFailedError`, exception was `null`).
- `:kotest-extensions-spring:apiCheck` passes.

### DIFF
--- a/kotest-extensions/kotest-extensions-spring/src/jvmMain/kotlin/io/kotest/extensions/spring/SpringExtension.kt
+++ b/kotest-extensions/kotest-extensions-spring/src/jvmMain/kotlin/io/kotest/extensions/spring/SpringExtension.kt
@@ -92,7 +92,8 @@ open class SpringExtension(
          testContextManager().beforeTestMethod(testCase.spec, methodName)
          val result = execute(testCase)
          // the spring docs state that afterTestMethod must be called immediately after framework-specific after lifecycle callbacks
-         testContextManager().afterTestMethod(testCase.spec, methodName, null as Throwable?)
+         // forward any exception from the test result so the test context manager (and its listeners) are aware of it
+         testContextManager().afterTestMethod(testCase.spec, methodName, result.errorOrNull)
          result
       } else {
          execute(testCase)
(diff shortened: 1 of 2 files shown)


alfonsoristorato (approved):
wonder if this is the solution for that `afterAny `sequence resetter we wondered about last time

PR #6110, opened 2026-06-08
fix(kotest-assertions-core): Make clear what is the actual string value

sschuberth:
Maybe the current message was supposed to be read like "It is expected for <actual> to match <expected>", but the way it was written is misleading WRT to the usual "expected: / "but was:" syntax. So clarify that the first string is not the expected but the actual string.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/maps/MapMatchers.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/maps/MapMatchers.kt
@@ -209,7 +209,7 @@ class MapContainsMatcher<K, V>(
       }
       val failureMsg = """
       |
-      |Expected:
+      |Actual:
       |  ${stringify(value)}
       |$expectMsg:
       |  ${stringify(expected)}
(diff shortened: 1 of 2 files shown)


kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/maps/MapMatchers.kt
```diff
@@ -209,7 +209,7 @@ class MapContainsMatcher<K, V>(
       }
       val failureMsg = """
       |
-      |Expected:
+      |Actual:
```
alfonsoristorato, line 212: Isn't expected correct here? As we assert whether the expected contains the actual values?

sschuberth: Well, what we're currently showing after "Expected:" is not the expected string in the sense of `actual shouldBe expected`. The current wording always confused me. Also see the commit message.

alfonsoristorato: I see what you mean now - I always get confused by the order of actual and expected - makes sense :) ty

alfonsoristorato (approved):

Issue #4015, opened 2024-05-11
add wasmWasi build target


jillesvangurp:
Expand the wasm support by supporting wasmWasi target in addition to the wasmJs target that was added in #3598

OliverO2:
This will have to wait until support for the `wasmWasi` target lands in **kotlinx.coroutines**:
https://github.com/kotest/kotest/blob/13b7fbcebc1ff34c4d24b009972652799f9d38da/buildSrc/src/main/kotlin/kotest-js-conventions.gradle.kts#L19

vdshb:
FYI: **kotlinx.coroutines 1.9.0-RC** has wasmWasi support :
https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-core-wasm-wasi/

OliverO2:
Thanks for the heads up. Leaving it as blocked for now until kotlinx.coroutines version 1.9.0 is released.

vdshb:
FYI: kotlinx.coroutines 1.9.0 with Wasm/WASI support is released:
https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.9.0

JesusMcCloud:
could we get a test build?

OliverO2:
There is no wasmWasi implementation yet.

I tried to add a wasmWasi back in August (locally). Thought it could be done quickly, since I already did it for the wasmJs target. Unfortunately, it did not work out for me. After changing 48 files in assertions, framework, engine, compiler and Gradle plugin, it required further changes to the compiler plugin and potentially a different reporting infrastructure. It took more time than I had available to figure out the details and the interplay between Kotest components.

JesusMcCloud:
I see, the comment in the sources made it seem like it was only blocked by coroutines. Thanks for the elaborate explanation!

sksamuel:
Supported in 6.0

EdricChan03:
It seems that the Wasi target still isn't published to Maven Central? [Trying to search for "kotest wasm wasi" yields no results](https://central.sonatype.com/search?q=kotest+wasm+wasi), while ["kotest wasm" only shows the wasmJs variant](https://central.sonatype.com/search?q=kotest+wasm)

vdshb:
Works fine in my projects - 6.0.7

sksamuel:
Closing as this was added in 6.0.

PR #6129, opened 2026-06-11
feat(kotest-property): Add Wasm WASI target to kotest-property and kotest-property-permutations

jsoizo:
Adds the `wasmWasi` target to `kotest-property` and `kotest-property-permutations`, so property-based testing can be used on Wasm WASI projects.

Wasm WASI support was added to the framework, assertions and engine modules in 6.0 (#4015), but the property modules were left out of that rollout. There a no technical blocker anymore:

- Both modules already support `wasmJs`, and all non-JVM `actual` implementations  live in `nonjvmMain` (seed persistence no-ops, `kotlin.uuid.Uuid`,  `KotestAssertionFailedError`, `EmptyArbResolver`), which the `wasmWasi` target
  inherits via the `nonjvm` hierarchy group configured by `kotest-wasi-conventions`.
- All dependencies are wasmWasi-ready: kotest-common, kotest-assertions-core,  kotest-framework-engine, kotlinx-coroutines (since 1.9.0), and `kotlin  0.0.6 (which backs the common `Arb.pattern`) publishes a `wasm-wasi` artifact.

The change is therefore just applying the existing `kotest-wasi-conventions`plugin to both modules — no source changes required.

I considered using kotest pbt when adding Wasm WASI support to the kotlin-csv library, but due to this issue, I have implemented the following workaround.

```
listOf("compileTestKotlinWasmWasi", "wasmWasiTest", "wasmWasiNodeTest").forEach { taskName ->
    tasks.matching { it.name == taskName }.configureEach { enabled = false }
}
```

### DIFF
--- a/kotest-property/build.gradle.kts
+++ b/kotest-property/build.gradle.kts
@@ -4,6 +4,7 @@ plugins {
    id("kotest-watchos-device-conventions")
    id("kotest-native-conventions")
    id("kotest-js-conventions")
+   id("kotest-wasi-conventions")
    id("kotest-publishing-conventions")
 }
 
--- a/kotest-property/kotest-property-permutations/build.gradle.kts
+++ b/kotest-property/kotest-property-permutations/build.gradle.kts
@@ -1,6 +1,7 @@
 plugins {
    id("kotest-jvm-conventions")
    id("kotest-js-conventions")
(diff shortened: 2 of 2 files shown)


sksamuel:
Amazing

sksamuel (approved):

PR #6134, opened 2026-06-14
align compile toolchain with minimum supported Kotlin version for KLIB ABI compatibility

alfonsoristorato:
Kotest's KMP/Native KLIBs were being compiled with the Kotlin 2.3.20 compiler, stamping KLIB ABI version 2.3.x into the artifacts. Consumers on Kotlin 2.2 would see the resolver skip these KLIBs with:
```
  w: KLIB resolver: Skipping '...' having incompatible ABI version '2.3.0'.
  The library was produced by '2.3.20' compiler.
```
KLIB ABI compatibility is one-directional: newer compilers can consume older ABI versions, but not vice versa. To support users on any Kotlin version from 2.2 onwards, KLIBs must be produced by a 2.2 compiler.

This project already uses the Gradle plugin / compiler version split (kotlin-gradle-plugin stays at 2.3.20 via compilerVersion in kotlin-conventions).
This PR takes advantage of that: drop kotlin-compile-version and kotlin-core-libraries-version to 2.2.0 while leaving the Gradle plugin unchanged. 
Tooling and IDE support stay on 2.3.20; only the compiler binary and stdlib version used to produce artifacts changes.

Note: kotlin-language-version was already set to 2.2.0 and wired up via languageVersion/apiVersion in `kotlin-conventions.gradle.kts`, so the source-level guardrail was already in place.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/InvokeMatcher.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/InvokeMatcher.kt
@@ -45,12 +45,12 @@ fun <T> invokeMatcher(t: T, matcher: Matcher<T>): T {
 }
 
 internal class MatcherResultWithError(
-   val passed: Boolean,
+   val passedResult: Boolean,
    val error: () -> Throwable?,
    val failureMessageFn: (error: Throwable?) -> String,
    val negatedFailureMessageFn: (error: Throwable?) -> String,
 ) : MatcherResult {
-   override fun passed(): Boolean = passed
+   override fun passed(): Boolean = passedResult
    override fun failureMessage(): String = failureMessageFn(error())
    override fun negatedFailureMessage(): String = negatedFailureMessageFn(error())
 }
--- a/.github/workflows/master_intellij_plugin.yml
+++ b/.github/workflows/master_intellij_plugin.yml
@@ -36,7 +36,10 @@ jobs:
 
          -  name: Override Kotlin version for IC-261
             if: matrix.product == 'IC-261'
-            run: sed -i 's/^kotlin = .*/kotlin = "2.3.10"/' gradle/libs.versions.toml
+            run: |
+               sed -i 's/^kotlin-gradle-plugin = "[0-9][^"]*"/kotlin-gradle-plugin = "2.3.10"/' gradle/libs.versions.toml
+               sed -i 's/^kotlin-compile-version = .*/kotlin-compile-version = "2.3.10"/' gradle/libs.versions.toml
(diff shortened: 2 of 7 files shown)


alfonsoristorato:
I don't know how to do the fancy `generated by claude` stuff - but I am trying to get around these errors
https://github.com/kotest/kotest/actions/runs/27492125635

Now that we can run specific modules via `test-kotest-examples` action it should be easy to figure out what this breaks

alfonsoristorato:
The ripple down effect to support 2.2.x for `multiplatform` and `native` seems quite excessive, I think we can make this next 6.2 release the one that supports it and increase the baseline from there onwards?
Alternatively we can create kotlin version specific modules. I am sure that @sksamuel will have a better idea, so leaving this decision 😄

kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/InvokeMatcher.kt
```diff
@@ -45,12 +45,12 @@ fun <T> invokeMatcher(t: T, matcher: Matcher<T>): T {
 }
 
 internal class MatcherResultWithError(
-   val passed: Boolean,
+   val passedResult: Boolean,
```
alfonsoristorato, line 48: apparently passed was reserved in JS

alfonsoristorato:
Had to remove targets in kotest-examples (pr https://github.com/kotest/kotest-examples/pull/107 )to align with https://github.com/kotest/kotest/pull/5820

alfonsoristorato:
kotest examples green with `2.2.10` compilation target https://github.com/kotest/kotest/actions/runs/27498762080

alfonsoristorato:
kotest examples green with `2.2.21` compilation target https://github.com/kotest/kotest/actions/runs/27499465600

sksamuel:
> The ripple down effect to support 2.2.x for `multiplatform` and `native` seems quite excessive, I think we can make this next 6.2 release the one that supports it and increase the baseline from there onwards? Alternatively we can create kotlin version specific modules. I am sure that @sksamuel will have a better idea, so leaving this decision 😄

Weren't we on 2.2 anyway before I decided to try 2.4 ?

sksamuel (approved):

sksamuel:
We should be able to use the 2.4 toolchain and just compile to 2.2, at least according to that article I read. 🤷🏻

alfonsoristorato:
> > The ripple down effect to support 2.2.x for `multiplatform` and `native` seems quite excessive, I think we can make this next 6.2 release the one that supports it and increase the baseline from there onwards? Alternatively we can create kotlin version specific modules. I am sure that @sksamuel will have a better idea, so leaving this decision 😄
> 
> Weren't we on 2.2 anyway before I decided to try 2.4 ?

I think we went 2.2 - 2.4 - 2.3 😅
Let's get this in and then we try to see how to make that thing work

PR #6135, opened 2026-06-14
Make fibonacci interval functions public

### LINKED ISSUE #6130: Duration.fibonacci() is @KotestInternal but the eventually docs recommend it
## Which version of Kotest are you using
6.1.10

## Summary

The [eventually documentation](https://kotest.io/docs/assertions/eventually.html) and the API surface disagree. The documentation recommends configuring a fibonacci interval via:

```kotlin
intervalFn = 100.milliseconds.fibonacci()
```

As of 6.1.10, `Duration.fibonacci()` and `FibonacciIntervalFn` in `io.kotest.assertions.nondeterministic` are annotated `@KotestInternal`, so following the documentation produces:

```
This declaration needs opt-in. Its usage should be marked with '@io.kotest.common.KotestInternal' or '@OptIn(io.kotest.common.KotestInternal::class)'
```

It is important to note that for builds compiled with `allWarningsAsErrors = true` this is not a warning but a compile failure. We encountered it migrating a service from Kotest 5.3.1 to 6.1.10, where the equivalent `interval = 100.milliseconds.fibonacci()` was public API in `io.kotest.assertions.until`.

## Expected behaviour

If it matches the direction you want to take, I would suggest making `Duration.fibonacci()` and `FibonacciIntervalFn` public. `intervalFn` and `DurationFn` are already the designed public extension point, and a fibonacci interval is a sensible growth curve to offer out of the box, as Kotest 5 did. If the intent is instead for users to supply their own `DurationFn`, then the documentation should show a hand-rolled implementation and stop recommending the internal helper. Either resolution is workable but the current doc ambiguity should be resolved.

Possibly related: #6104, which proposes more expressive `EventuallyConfiguration` helpers and could subsume the public-API question??

mvanhorn:
## Summary

The `eventually` docs recommend configuring a fibonacci backoff with
`eventuallyConfig { intervalFn = 100.milliseconds.fibonacci() }`, but as of 6.1.x the
`FibonacciIntervalFn` class in `io.kotest.assertions.nondeterministic` is annotated
`@KotestInternal`. `Duration.fibonacci()` is already public, yet the type it returns is
not, so following the docs requires a `@OptIn(KotestInternal::class)`. For consumers that
compile with `allWarningsAsErrors = true` this is a hard build failure. In Kotest 5 the
equivalent usage was public API.

This removes the `@KotestInternal` annotation (and the now-unused
`io.kotest.common.KotestInternal` import) from `FibonacciIntervalFn`, bringing it in line
with the sibling `ExponentialIntervalFn`, which is already public and implements the same
public `DurationFn` interface. The internal `fibonacci(n: Int)` helper stays `internal`.

The binary-compatibility signature file `kotest-assertions-core.api` was regenerated with
`./gradlew :kotest-assertions:kotest-assertions-core:apiDump` (not hand-edited) to add the
now-public `FibonacciIntervalFn` entry, mirroring the existing `ExponentialIntervalFn`
entry.

## Why this matters

Following the published `eventually` documentation should not require an opt-in to an
internal API, and should not break builds that treat warnings as errors. The documented
extension point (`intervalFn` / `DurationFn`) is already the designed public surface, and a
fibonacci growth curve is a sensible built-in alongside the already-public exponential one.

https://github.com/kotest/kotest/issues/6130

## Testing

- `./gradlew :kotest-assertions:kotest-assertions-core:compileKotlinJvm -PjvmOnly=true` - passes.
- `./gradlew :kotest-assertions:kotest-assertions-core:jvmTest --tests "io.kotest.assertions.nondeterministic.FibonacciIntervalTest" -PjvmOnly=true` - passes, including a new case asserting that `eventuallyConfig { intervalFn = 100.milliseconds.fibonacci() }` compiles and runs with no `@OptIn` and yields a `FibonacciIntervalFn`.
- `./gradlew :kotest-assertions:kotest-assertions-core:apiCheck -PjvmOnly=true` - passes after the `apiDump` regeneration, so binary-compatibility validation stays green.

Fixes #6130

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/io/kotest/assertions/nondeterministic/FibonacciIntervalTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/io/kotest/assertions/nondeterministic/FibonacciIntervalTest.kt
@@ -6,7 +6,9 @@ import io.kotest.core.spec.style.FunSpec
 import io.kotest.matchers.comparables.shouldBeGreaterThan
 import io.kotest.matchers.comparables.shouldBeLessThan
 import io.kotest.matchers.shouldBe
+import io.kotest.matchers.types.shouldBeInstanceOf
 import kotlin.time.Duration.Companion.hours
+import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.minutes
 
 @EnabledIf(LinuxOnlyGithubCondition::class)
@@ -70,6 +72,17 @@ class FibonacciIntervalTest : FunSpec() {
                u shouldBeGreaterThan max
             }
          }
+      }
+
+      // https://github.com/kotest/kotest/issues/6130
+      // The eventually docs recommend `intervalFn = 100.milliseconds.fibonacci()`. This must
+      // compile and run without any @OptIn, i.e. fibonacci() and the type it returns
+      // (FibonacciIntervalFn) are public, non-opt-in API.
+      test("fibonacci() is usable as a public eventuallyConfig intervalFn without opt-in") {
+         val config = eventuallyConfig {
+            intervalFn = 100.milliseconds.fibonacci()
+         }
+         config.intervalFn.shouldBeInstanceOf<FibonacciIntervalFn>()
       }
    }
 }
--- a/kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
+++ b/kotest-assertions/kotest-assertions-core/api/kotest-assertions-core.api
@@ -469,6 +469,13 @@ public final class io/kotest/assertions/nondeterministic/ExponentialIntervalFnKt
 	public static synthetic fun exponential-4_FzEXg$default (JDLkotlin/time/Duration;ILjava/lang/Object;)Lio/kotest/assertions/nondeterministic/ExponentialIntervalFn;
 }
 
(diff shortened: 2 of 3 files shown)


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/io/kotest/assertions/nondeterministic/FibonacciIntervalTest.kt
```diff
@@ -71,5 +73,16 @@ class FibonacciIntervalTest : FunSpec() {
...
+      // https://github.com/kotest/kotest/issues/6130
+      // The eventually docs recommend `intervalFn = 100.milliseconds.fibonacci()`. This must
+      // compile and run without any @OptIn, i.e. fibonacci() and the type it returns
+      // (FibonacciIntervalFn) are public, non-opt-in API.
+      test("fibonacci() is usable as a public eventuallyConfig intervalFn without opt-in") {
+         val config = eventuallyConfig {
+            intervalFn = 100.milliseconds.fibonacci()
+         }
+         config.intervalFn.shouldBeInstanceOf<FibonacciIntervalFn>()
+      }
```
alfonsoristorato, line 86: thanks - unfortunately we don't require `OptIn` within kotest, but we'll eventually write a test to check this contract is not changed

alfonsoristorato (approved):
thanks a lot 😄

mvanhorn:
Appreciate you landing this, @alfonsoristorato. Making the fibonacci interval functions public matches what the eventually docs already recommend.

PR #6133, opened 2026-06-14
allow test kotest examples action dispatch

alfonsoristorato:

PR #6136, opened 2026-06-14
allow building all targets via dispatch of kotest-test-examples

alfonsoristorato:
- allow building all targets via dispatch of kotest-test-examples


alfonsoristorato:
same as https://github.com/kotest/kotest/pull/6133 will merge straight without approval - soz again

PR #6137, opened 2026-06-16
Add builder-style DSL for eventually configuration (#6104)

### LINKED ISSUE #6104: Feature: more expressive EventuallyConfiguration helpers
This reads beautifully:
```
eventually { item(newId) shouldNotBe null }
```

This, not so much:
```
eventually(eventuallyConfig { interval = 5.seconds }){
   item(newId) shouldNotBe null
}
```

I find myself making additional shorthand helpers to keep it both inline and readable, so I can do things like:
```
eventually(checkingEvery(5.seconds)){
   item(newId) shouldNotBe null
}
```

A fuller set of well-thought-out config-helpers like this would make a nice addition to the dsl.

The next level is probably deranged overengineering, but just a thought: 
If eventually() could receive an EventuallyConfigurationBuilder, one could make composable helpers and say things like:
```
eventually(within(2.seconds).checkingEvery(100.milliseconds)){
   item(newId) shouldNotBe null
}
```
Or taking it really far:
```
within(2.seconds) checkingEvery 100.milliseconds {
   item(newId) shouldNotBe null
}
```

sksamuel:
Closes #6104.

## What

Adds a readable, chainable, builder-style DSL for configuring `eventually`, as requested in #6104 (level 2 of the suggestions):

```kotlin
eventually(within(2.seconds).checkingEvery(100.milliseconds)) {
   item(newId) shouldNotBe null
}
```

## How

- Each new function on `EventuallyConfigurationBuilder` updates a single config var and returns the same builder so calls can be chained:
  - `within(duration)`
  - `withInitialDelay(initialDelay)`
  - `checkingEvery(interval)` and `checkingEvery(intervalFn)`
  - `withRetries(retries)`
  - `retryingOn(vararg KClass)` and `retryingOn(predicate)`
  - `withListener(listener)`
  - `shortCircuitOn(predicate)`
  - `includingFirstError(includeFirst)`
- The verb-style names deliberately differ from the raw property names to read well in a chain and to avoid shadowing the public `var`s.
- A top-level `within(duration)` function is the entry point — it returns a builder pre-configured with that duration.
- A new `eventually(builder, test)` overload accepts the builder directly. It resolves unambiguously against the existing lambda-`configure` overload (instance vs function literal).

## Notes

This PR also includes the previously staged `eventually(configure: EventuallyConfigurationBuilder.() -> Unit, test)` lambda overload (and the small accompanying `EventuallyTest.kt` tidy-ups), which were already in the working tree as part of this same issue.

## Tests

New `EventuallyDslTest` spec covers each function by asserting the builder state it sets, plus two behavioural tests that run `eventually(within(...).checkingEvery(...)) { }` to prove the new overload is wired and respects the configured duration.

`./gradlew :kotest-assertions:kotest-assertions-core:apiDump` was run and the updated `.api` file is committed; `apiCheck` passes.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/io/kotest/assertions/nondeterministic/EventuallyDslTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/io/kotest/assertions/nondeterministic/EventuallyDslTest.kt
@@ -1,0 +1,116 @@
+@file:Suppress("RETURN_VALUE_NOT_USED_COERCION", "RETURN_VALUE_NOT_USED")
+
+package io.kotest.assertions.nondeterministic
+
+import io.kotest.assertions.throwables.shouldThrow
+import io.kotest.core.annotation.EnabledIf
+import io.kotest.core.annotation.LinuxOnlyGithubCondition
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.matchers.ints.shouldBeLessThan
+import io.kotest.matchers.shouldBe
+import java.io.IOException
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
+
+@EnabledIf(LinuxOnlyGithubCondition::class)
+class EventuallyDslTest : FunSpec() {
+
+   init {
+
+      test("within sets the duration and returns a builder") {
+         val builder = within(2.seconds)
+         builder.duration shouldBe 2.seconds
+      }
+
+      test("within is the entry point for chaining") {
+         val builder = within(2.seconds).checkingEvery(100.milliseconds)
+         builder.duration shouldBe 2.seconds
+         builder.interval shouldBe 100.milliseconds
+      }
+
+      test("checkingEvery sets the interval") {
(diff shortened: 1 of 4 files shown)


kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/io/kotest/assertions/nondeterministic/EventuallyDslTest.kt
```diff
@@ -0,0 +1,116 @@
...
+      test("eventually with a builder respects the configured duration") {
+         var count = 0
+         shouldThrow<AssertionError> {
+            eventually(within(100.milliseconds).checkingEvery(20.milliseconds)) {
+               count += 1
+               error("never passes")
+            }
+         }
+         count shouldBeLessThan 100
+      }
```
alfonsoristorato, line 114: do you think it's worth expanding these with every builder param - or are you happy with the fact that we tested they give the right config and `EventuallyTest` inherently tests those?

alfonsoristorato (approved):
feel free to merge when you like, I removed auto-merge so you have a chance to read the comment 😄

Issue #6006, opened 2026-05-08
Koin bug is more likely to trigger in Kotest's KoinExtension after Kotest v6 upgrade


glasser:
If you use Kotest's KoinExtension, and you use `KoinTest.declare` in a particular test function to modify the Koin context after already using the binding in question, state may not be fully cleared between test functions. This behavior was generally masked in Kotest v5 because `fun extensions()` got re-evaluated on more frequently than Kotest v6's `val extensions` will (assuming you don't define your extensions with `get()` or something).

I think this is a Koin bug and I filed it at https://github.com/InsertKoinIO/koin/issues/2412 but since the Kotest v6 upgrade makes it easier to trigger, I wanted to file something here as well.

**Which version of Kotest are you using**

v6.1.11

sksamuel:
Thanks for the detailed report — I dug into this and can confirm your analysis.

**Confirmed root cause (upstream Koin).** As you say, this is fundamentally the Koin defect tracked in InsertKoinIO/koin#2412: `declare()` displaces a factory in the running `InstanceRegistry`, but the original `Module.mappings` still references the orphaned `SingleInstanceFactory` holding its cached (mutated) singleton. `stopKoin()` doesn't drop that orphaned factory, so a subsequent `startKoin()` **with the same `Module` instance** resurrects the stale instance.

**Why Kotest v6 exposes it.** In v6 a spec's extensions are a property — `open val extensions: List<Extension> = emptyList()` (`Spec.kt`) — whereas v5 used `fun extensions()`. `KoinExtension` captures the `Module` *instances* at construction (`private val modules: List<Module>`) and feeds the same instances to `startKoin { modules(modules) }` on every test. Under v5 people typically rebuilt their modules each time the function was evaluated, so each test got a fresh `Module`; under v6 the same instances are reused for the spec's lifetime, which is exactly the precondition the cache leak rides on.

**Version note.** We're currently pinned to Koin 4.1.0 (4.2.x needs Kotlin 2.3.x KLIB ABI, which we can't adopt yet), and the upstream issue reports the bug across 3.x–4.2.1, so a version bump alone neither fixes it nor is feasible right now.

[...]

PR #6138, opened 2026-06-16
Add module-factory support to KoinExtension (#6006)

sksamuel:
Mitigates #6006.

## Background

`KoinExtension` captured `Module` *instances* at construction (`private val modules: List<Module>`) and fed the same instances to `startKoin { modules(modules) }` on every test. Since Kotest v6 exposes a spec's extensions as a property (`open val extensions: List<Extension>`) rather than v5's re-evaluated `fun extensions()`, those module instances now live for the spec's whole lifetime.

When a test mutates the Koin context via `KoinTest.declare`, reusing the same `Module` instance can leak singleton state into later tests: Koin caches instances on the `SingleInstanceFactory`s held by the module, and `declare()` orphans a factory whose cached singleton survives `stopKoin()`/`startKoin()`. This is fundamentally the upstream Koin bug [InsertKoinIO/koin#2412](https://github.com/InsertKoinIO/koin/issues/2412); Kotest v6 just makes it far easier to hit.

We're pinned to Koin 4.1.0 (4.2.x needs Kotlin 2.3.x KLIB ABI), and the upstream bug spans 3.x–4.2.1, so a version bump isn't a fix.

## Change

Add an opt-in `KoinExtension` constructor that accepts a module **factory** (`() -> List<Module>`), invoked on each Koin start so every test gets fresh module instances (and therefore fresh singletons), restoring v5-like isolation regardless of the upstream bug:

```kotlin
// myModule() builds a fresh Module each call
extension(KoinExtension { listOf(myModule()) })
```

- The previous primary constructor `(List<Module>, Provider?, KoinLifecycleMode)` is preserved as a public secondary constructor, so all existing instance-based usage and binary compatibility are unaffected (API diff is purely additive).
- New regression test `KoinExtensionModuleFactoryTest` reproduces the `declare()` scenario and asserts the next test sees a clean singleton.
- Documents the workaround in `documentation/docs/extensions/koin.md`.

## Verification

- `:kotest-extensions-koin:jvmTest` — passes (new test + existing Koin tests)
- `apiDump` regenerated and committed; `apiCheck` passes

The underlying defect should ultimately be fixed in Koin (koin#2412); this is a pragmatic mitigation for v6 users in the meantime.


sksamuel:
@LeoColman you know Koin well, would appreciate your thoughts

alfonsoristorato (approved):

PR #6143, opened 2026-06-16
Remove usages of K1's KtUltraLightClass from kotest plugin sources

sksamuel:
Removes usages of K1's `KtUltraLightClass` from the in-repo IntelliJ plugin sources.

Otherwise a `ClassNotFoundException` would happen with IJ 2026.2+ where K1 classes are not bundled anymore.

`KtUltraLightClass` inherits `KtLightClass`, so the change is compatible with the K1 plugin for old versions of IJ (only 2024.X should be affected, because starting with 2025.1 K2 was the default mode).

This mirrors the same change made upstream in the standalone plugin repo: https://github.com/kotest/kotest-intellij-plugin/pull/407


alfonsoristorato (approved):
An easy approval

PR #6155, opened 2026-06-22
Fix typo in proptest config docs

axiom:
Change the header for the `maxFailure` configuration property from "Min Failure" to "Max Failure" to mirror the configuration property name.

image


AlexCue987 (approved):
nice!

PR #6160, opened 2026-06-26
fix: attach Fail fast scope tracker for each root test case in instance per spec mode

### LINKED ISSUE #6157: Failfast is shared by all root tests in spec
Using InstancePerRoot isolation mode with Kotest 6.2.1 makes this test to don't execute the second root test.

I expect `"Test root with failure test"` to run and skip the `"Test that should not be executed"` but the `"Second root test"` should execute complete. 

```kotlin
class TestSample : FreeSpec({
    "Test root with failure test" - {
        "Test a" {
            println("a")
        }
        "Test that will fail" {
            false.shouldBeTrue()
        }
        "Test that should not be executed" {
            println("c")
        }
    }

    "Second root test" - {
        "Test a" {
            println("a")
        }
        "Test b" {
            println("b")
        }
        "Test c" {
            println("c")
        }
    }
})
```

running same test with same config using Kotest 6.1.11 works as expected.

ashdawngary:
fixes #6157 

In 6.1.11 (pre 6.2) it seems like FailFastInterceptor was directly tied to SpecContext and the state was stored on the context itself. Within The Instance per root spec inspector, a new spec context was created hence tracking was clean.

In the new system the tracking was moved to its own coroutine level context FailFastScopeTracker, and that is not being instantiated fresh in between root spec executions. The errors then share the same fail fast execution tracker (at the seed spec level ), and fail fast gets conflated between root specs.

Wrote two regression tests (the majority of this PR) that verify that both spec level and project level fail fast  do not have the same state conflation issue.

```kt
// verify that a failure in first root spec does not affect the second root spec

private class InstancePerRootProjectFailFastFreeSpec : FreeSpec({
   "Test root with failure test" - {
      "Test a" { }
      "Test b" { error("fail") }
      "Test c" {}
   }

   "Second root test" - {
      "second test" { }
      "third test" { }
      "fourth test" { }
   }
})
```

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/FailFastInstancePerRootTest.kt
+++ b/kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/FailFastInstancePerRootTest.kt
@@ -1,0 +1,91 @@
+package com.sksamuel.kotest.engine.test
+
+import io.kotest.core.annotation.EnabledIf
+import io.kotest.core.annotation.LinuxOnlyGithubCondition
+import io.kotest.core.config.AbstractProjectConfig
+import io.kotest.core.spec.IsolationMode
+import io.kotest.core.spec.SpecRef
+import io.kotest.core.spec.style.FreeSpec
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.engine.TestEngineLauncher
+import io.kotest.engine.listener.CollectingTestEngineListener
+import io.kotest.matchers.shouldBe
+
+@EnabledIf(LinuxOnlyGithubCondition::class)
+class FailFastInstancePerRootTest : FunSpec({
+
+   test("project failfast should not skip sibling roots in InstancePerRoot mode") {
+      val config = object : AbstractProjectConfig() {
+         override val failfast = true
+         override val isolationMode = IsolationMode.InstancePerRoot
+      }
+
+      val listener = CollectingTestEngineListener()
+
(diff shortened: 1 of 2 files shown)


kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/FailFastInstancePerRootTest.kt
```diff
@@ -0,0 +1,91 @@
...
+import io.kotest.core.config.AbstractProjectConfig
+import io.kotest.core.spec.IsolationMode
+import io.kotest.core.spec.SpecRef
+import io.kotest.core.spec.style.FreeSpec
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.engine.TestEngineLauncher
+import io.kotest.engine.listener.CollectingTestEngineListener
+import io.kotest.matchers.shouldBe
+
+@EnabledIf(LinuxOnlyGithubCondition::class)
```
ashdawngary, line 14: im not sure when to add this, but looking around at other tests it seemed reasonable.

sksamuel (approved):

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/test/FailFastInstancePerRootTest.kt
sksamuel: yep that's fine

sksamuel:
Thanks for the pr :)

Issue #6154, opened 2026-06-22
Annotate matchers with @IgnorableReturnValue?


spheroid:
After enabling the return value checker introduced in Kotlin 2.3 and updating to most recent `kotest-assertions-core` (6.2.1), most of my unit tests now produce warnings due to unused return values from `shouldBe` assertions. Should these assertion functions be annotated with `@IgnorableReturnValue`? Adding a `val _ = foo shouldBe bar` to every assertion feels very verbose.

Satook:
Yes please. I have mine set to `check`, which creates warnings but `full` will cause compile errors without that boilerplate.

RaySmith-ttc:
You can temporarily exclude the `return-value-checker` argument in Gradle. For example, here is how to do it for KMP:
``` kts
tasks {
    withType<KotlinCompile> {
        compilerOptions {
            // other freeCompilerArgs

            if (name != "compileTestKotlinJvm") {
                freeCompilerArgs.add("-Xreturn-value-checker=full")
            }
        }
    }
}
```

PR #6165, opened 2026-07-07
disable return-value-checker for releases

alfonsoristorato:
for https://github.com/kotest/kotest/issues/6154
I went with `disable` instead of annotating every method cause the feature is still experimental and I don't want to waste tokens on `@IgnorableReturnValue` if kotlin changes things or comes up with something better.
Before:
image

After (with local release of kotest):
image


alfonsoristorato:
I guess at this point we might as well remove the flag, but seeing all those warnings will remind us to do something with it once the feature gets into stable

sksamuel (approved):

PR #6167, opened 2026-07-09
Use a data class's own equals() even when it implements Map or Collection (#5202)

### LINKED ISSUE #5202: Data classes that inherit from map or collection will use wrong comparison

Develop-KIM:
Fixes #5202.

### Problem
`DefaultEqResolver` selects `MapEq`/`CollectionEq` for any value that is a `Map`/`Collection`. A data class that delegates to a map or list is *also* a `Map`/`Collection`, so it was compared structurally (by entries/elements) and its other properties were ignored:

```kotlin
data class NamedMap(val name: String, val map: Map<String, String>) : Map<String, String> by map

val map1 = NamedMap("map1", mapOf("foo" to "bar"))
val map2 = NamedMap("map2", mapOf("foo" to "bar"))
map1 shouldNotBe map2   // failed — treated as equal maps, `name` ignored
```

The failure message also used the delegate map's `toString()` instead of the data class's.

### Fix
Skip the structural `MapEq`/`CollectionEq` branches when the actual value is a data class instance, so it falls through to the existing data-class / default equality path and its own generated `equals()` is used. Plain maps and collections are unaffected.

### Tests
Added cases in `DataClassEqTest` for a data class delegating to a `Map` and to a `Collection` (resolver selection + `shouldBe`/`shouldNotBe`), plus a guard that plain maps/collections still resolve to `MapEq`/`CollectionEq`. The full `com.sksamuel.kotest.eq.*` suite passes. No public-API change, so no `apiDump` update is needed.

_Drafted with AI assistance (Claude) and reviewed/verified by me._

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/DefaultEqResolver.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/assertions/eq/DefaultEqResolver.kt
@@ -33,12 +33,16 @@ object DefaultEqResolver : EqResolver {
       return when {
          actual == null || expected == null -> NullEq
          actual::class == expected::class && customEqInstances.contains(actual::class) -> customEqInstances[actual::class]!!
-         actual is Map<*, *> && expected is Map<*, *> -> MapEq
+         // A data class that also implements a structural type such as Map or Collection
+         // (e.g. `data class Foo(val name: String, val m: Map<K, V>) : Map<K, V> by m`) must be
+         // compared using its own equals(), not by structural (entry/element) equality. Otherwise
+         // its non-delegated properties would be silently ignored. See #5202.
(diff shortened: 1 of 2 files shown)


AlexCue987:
"DefaultEqResolver selects MapEq/CollectionEq for any value that is a Map/Collection"
this feature was built by design, and changing it might break lots of existing test suites.
so let's not do this. if we want to match data classes, we can use `shouldBeEqual` or `shouldBeEqualToComparingFields` which allows a lot of control over matching.

PR #6170, opened 2026-07-14
Annotate matchers with @IgnorableReturnValue and re-enable the return value checker for releases

LeoColman:
## Why

Kotest is compiled with `-Xreturn-value-checker=full`, which marks every declaration in the module as `@MustUseReturnValue`. Because matchers return their receiver to allow chaining, consumers who enable the checker get a warning for every assertion written as a statement:

```kotlin
test("doc-index txt exists and lists files") {
   docIndex shouldHaveAtLeastSize 1
   // warning: unused return value of 'shouldHaveAtLeastSize'
}
```

This was reported on Slack after upgrading 6.1.1 → 6.2.x ("my project's warning list literally explodes"). 6.2.2 worked around it by disabling the checker for release builds. This PR does the real fix, so the checker can be turned back on.

## What

`kotlin.IgnorableReturnValue` opts a declaration out of the check. It targets `FUNCTION` only — there is no file- or class-level escape hatch — so each declaration is annotated individually (2,514 of them).

Annotated:

- every `should*` / `shouldNot*` matcher across all `kotest-assertions-*` modules
- the inspectors (`forAll`, `forNone`, `forExactly`, `forAtLeastOne`, the map aliases, …)
- `assertSoftly`, `withClue`, `asClue`, `invokeMatcher`
- the nondeterministic helpers (`eventually`, `continually`, `until`, `retry`)
- the property entry points (`checkAll`, `forAll`, `proptest`, `permutations`, `checkCoverage`)
- the spec DSL helpers that are written as bare statements (`extension`, `autoClose`, `install`, `register`)
- the scoped-resource helpers (`withSystemProperty`, `withEnvironment`)

**Matcher factories (`beEmpty()`, `haveSize()`, …) are deliberately left unannotated** — their result is the `Matcher` itself, and discarding it is always a mistake, so the checker should keep flagging those.

The annotation is a no-op on a `Unit`-returning function, so matchers are annotated uniformly rather than depending on whether a given overload happens to return its receiver. That keeps the rule auditable ("every matcher has it") and stays correct if a matcher later gains a chaining return.

The second commit re-enables the checker for release builds, which was the point of the exercise.

## Verification

Compiling an **external consumer** against the built jar with `-Xreturn-value-checker=full`:

```kotlin
docIndex shouldHaveAtLeastSize 1   // no warning
content shouldStartWith "# Alpha"  // no warning
beEmpty<String>()                  // still warns — factory result must be used
```

The control line matters: it confirms the checker really is active in that compile, and that factories were not over-annotated.

- `Unused return value` warnings across kotest's own build: **10,909 → 493**. Every remaining one is Kotlin stdlib (`apply`/`let`/`invoke`), a test-local helper, or `internal` engine code — none reach consumers.
- `apiCheck` passes; force-regenerated `.api` dumps are **byte-identical** (BCV does not record annotations), so this is not a binary-compatibility change.
- Full `check`: **22,301 tests** pass across JVM, JVM-max-JDK, wasmJs (browser + node), wasmWasi, linuxX64 and Android-host.

## Note for reviewers

Re-enabling the checker for releases (commit 2) means *all* public non-Unit kotest API becomes must-use for consumers who opt in, not just matchers. I covered the statement-style APIs that kotest's own test suite actually discards, but a consumer could use one that kotest never does. If that feels too broad, commit 2 can be dropped and landed separately — commit 1 stands on its own.

### DIFF
--- a/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
@@ -33,7 +33,7 @@ kotlin {
    compilerOptions {
 
       freeCompilerArgs.add("-Xexpect-actual-classes")
-      freeCompilerArgs.add("-Xreturn-value-checker=${if (Ci.isRelease) "disable" else "full"}")
+      freeCompilerArgs.add("-Xreturn-value-checker=full")
       freeCompilerArgs.add("-XXLanguage:+UnnamedLocalVariables")
 
       // See https://mbonnin.net/2026-02-22-kotlin-versions
--- a/kotest-assertions/kotest-assertions-arrow-fx-coroutines/src/commonMain/kotlin/io/kotest/assertions/arrow/fx/coroutines/ExitCase.kt
+++ b/kotest-assertions/kotest-assertions-arrow-fx-coroutines/src/commonMain/kotlin/io/kotest/assertions/arrow/fx/coroutines/ExitCase.kt
@@ -7,6 +7,7 @@ import kotlin.contracts.contract
 import kotlinx.coroutines.CancellationException
 
 @OptIn(ExperimentalContracts::class)
+@IgnorableReturnValue
 public fun ExitCase.shouldBeCancelled(
   failureMessage: (ExitCase) -> String = { "Expected ExitCase.Cancelled, but found $it" }
 ): ExitCase.Cancelled {
@@ -21,6 +22,7 @@ public fun ExitCase.shouldBeCancelled(
 }
 
 @OptIn(ExperimentalContracts::class)
+@IgnorableReturnValue
 public fun ExitCase.shouldBeCancelled(cancelled: CancellationException): ExitCase.Cancelled {
   contract {
     returns() implies (this@shouldBeCancelled is ExitCase.Cancelled)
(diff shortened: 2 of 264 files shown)


alfonsoristorato:
Thanks :) I think we can have a discussion as to whether to keep the flag. We did have an initial chat and because it's still in 'preview' we felt it was too soon to annotate all the marchers, just in case the API evolves. Wdyt?

LeoColman:
> Thanks :) I think we can have a discussion as to whether to keep the flag. We did have an initial chat and because it's still in 'preview' we felt it was too soon to annotate all the marchers, just in case the API evolves. Wdyt?

I agree. I think we can keep the flag disabled BUT keep the annotations. For one main reason:

- We keep verifying with FULL in our own test suite and that will make warnings disappear

alfonsoristorato:
> > Thanks :) I think we can have a discussion as to whether to keep the flag. We did have an initial chat and because it's still in 'preview' we felt it was too soon to annotate all the marchers, just in case the API evolves. Wdyt?
> 
> I agree. I think we can keep the flag disabled BUT keep the annotations. For one main reason:
> 
> - We keep verifying with FULL in our own test suite and that will make warnings disappear
> 

Or we remove the flag entirely, but I guess we can always do this and ask Claude to remove the annotations if it comes to that (I am hoping that Kotlin provides overrides for specific use cases, he c'è the hesitation). Approving from my perspective but I don't think we should continue with commit 2 just yet, at least wait for this to become stable.

alfonsoristorato (approved):

LeoColman:
I wonder if this is an use case that we should mention to kotlinlang. This will happen across many codebases if the feature comes forth

AlexCue987:
> I wonder if this is an use case that we should mention to kotlinlang. This will happen across many codebases if the feature comes forth

this exactly

PR #6187, opened 2026-07-23
Force commons-beanutils to 1.11.0 (CVE-2025-48734)

### LINKED ISSUE #6139: Almost perfect ... but exists one single vulnerability
Image
----------------------------------------------------------------------------------------------------------------
Dependency maven:commons-beanutils:commons-beanutils:1.9.4 is vulnerable

Update to unaffected version 1.11.0

CVE-2025-48734,  Score: 8.8

Improper Access Control vulnerability in Apache Commons.
A special BeanIntrospector class was added in version 1.9.2. This can be used to stop attackers from using the declared class property of Java enum objects to get access to the classloader. However this protection was not enabled by default. PropertyUtilsBean (and consequently BeanUtilsBean) now disallows declared class level property access by default.
Releases 1.11.0 and 2.0.0-M2 address a potential security issue when accessing enum properties in an uncontrolled way. If an application using Commons BeanUtils passes property paths from an external source directly to the getProperty() method of PropertyUtilsBean, an attacker can access the enum’s class loader via the “declaredClass” property available on all Java “enum” objects. Accessing the enum’s “declaredClass” allows remote attackers to access the ClassLoader and execute arbitrary code. The same issue exists with PropertyUtilsBean.getNestedProperty().
Starting in versions 1.11.0 and 2.0.0-M2 a special BeanIntrospector suppresses the “declaredClass” property. Note that this new BeanIntrospector is enabled by default, but you can disable it to regain the old behavior; see section 2.5 of the user's guide and the unit tests.
This issue affects Apache Commons BeanUtils 1.x before 1.11.0, and 2.x before 2.0.0-M2.Users of the artifact commons-beanutils:commons-beanutils
1.x are recommended to upgrade to version 1.11.0, which fixes the issue.
Users of the artifact org.apache.commons:commons-beanutils2
2.x are recommended to upgrade to version 2.0.0-M2, which fixes the issue.

Read More: https://www.mend.io/vulnerability-database/CVE-2025-48734?utm_source=Jetbrains

Results powered by Mend.io

sksamuel:
Addresses the vulnerable `commons-beanutils` dependency reported in #6139 (CVE-2025-48734, CVSS 8.8).

### Root cause

The `kotest-extensions-mockserver` module pulls **`commons-beanutils:1.9.4`** transitively:

```
mockserver-netty -> mockserver-core -> json-path:2.10.0 -> commons-beanutils:1.9.4
```

### Why not just bump mockserver?

The fixed `commons-beanutils:1.11.0` only arrives transitively via `json-path:3.0.0`, which is only pulled by mockserver **7.2.0+**. However `json-path:3.0.0` requires **Java 17**, and even mockserver 7.0.0 (`json-path:3.0.0`) still resolves `commons-beanutils:1.9.4`. Kotest's baseline is `jvmMinTarget = 11`, so upgrading mockserver would break the Java 11 build without cleanly fixing the CVE.

`commons-beanutils:1.11.0` is itself still Java 8+ compatible, so the fix here pulls it explicitly in the mockserver extension, overriding the vulnerable transitive `1.9.4` (Gradle's highest-version-wins) while keeping mockserver on its current, Java-11-compatible release.

### Result

```
commons-beanutils:commons-beanutils:1.9.4 -> 1.11.0
```

The declared `api` dependency also propagates the fixed floor to downstream consumers of the extension.

### Verification

- `jvmTest` for `kotest-extensions-mockserver` passes on the Java 11 target.
- `dependencies` confirms `commons-beanutils` now resolves to `1.11.0`.

Fixes #6139

### DIFF
--- a/kotest-extensions/kotest-extensions-mockserver/build.gradle.kts
+++ b/kotest-extensions/kotest-extensions-mockserver/build.gradle.kts
@@ -10,6 +10,13 @@ kotlin {
             implementation(projects.kotestFramework.kotestFrameworkEngine)
             api(libs.mockserver.netty)
             api(libs.mockserver.client.java)
+
+            // mockserver pulls commons-beanutils 1.9.4 transitively (via json-path), which is
+            // affected by CVE-2025-48734. Pull 1.11.0 explicitly to override it: it fixes the
+            // vulnerability and is still Java 8+ compatible. (The fixed version otherwise only
+            // arrives via json-path 3.0.0, which requires Java 17 and would break Kotest's Java 11
+            // baseline.) See #6139.
+            api(libs.apache.commons.beanutils)
          }
       }
       jvmTest {
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -27,6 +27,7 @@ byte-buddy = "1.18.8"
 classgraph = "4.8.184"
 
 commons-lang3 = "3.20.0"
+commons-beanutils = "1.11.0" # CVE-2025-48734 fix; pulled transitively by mockserver via json-path
 gradle-test-logger-plugin = "4.0.0"
 
 jaxbApi = "2.3.1"
@@ -126,6 +127,7 @@ arrow-functions = { module = "io.arrow-kt:arrow-functions", version.ref = "arrow
 
 asm = { module = "org.ow2.asm:asm", version.ref = "asm" }
 apache-commons-lang = { module = "org.apache.commons:commons-lang3", version.ref = "commons-lang3" }
+apache-commons-beanutils = { module = "commons-beanutils:commons-beanutils", version.ref = "commons-beanutils" }
 
 blockhound = { module = "io.projectreactor.tools:blockhound", version.ref = "blockhound" }
 byte-buddy = { group = "net.bytebuddy", name = "byte-buddy", version.ref = "byte-buddy" }


kotest-extensions/kotest-extensions-mockserver/build.gradle.kts
```diff
@@ -10,6 +10,13 @@ kotlin {
             implementation(projects.kotestFramework.kotestFrameworkEngine)
             api(libs.mockserver.netty)
             api(libs.mockserver.client.java)
+
+            // mockserver pulls commons-beanutils 1.9.4 transitively (via json-path), which is
+            // affected by CVE-2025-48734. Pull 1.11.0 explicitly to override it: it fixes the
+            // vulnerability and is still Java 8+ compatible. (The fixed version otherwise only
+            // arrives via json-path 3.0.0, which requires Java 17 and would break Kotest's Java 11
+            // baseline.) See #6139.
+            api(libs.apache.commons.beanutils)
```
alfonsoristorato, line 19: I wonder if this should be our responsiblity - on a more general approach that I thought just now, what if we expose extension modules where the libraries are only used for compilation but we rely on users to have the library declared in their repo - this way we would not have to worry about this in future

alfonsoristorato (approved):

kotest-extensions/kotest-extensions-mockserver/build.gradle.kts
sksamuel: That's interesting, probably would work well for simple libraries

PR #6189, opened 2026-07-24
chore: route a few hardcoded dependency/plugin versions through the version catalog

stlpine:
## What

A few modules were declaring dependency/plugin versions as hardcoded string literals instead of going through `gradle/libs.versions.toml`:

- `kotest-runner-junit-platform`: `dev.gradleplugins:gradle-api:8.11.1`
- `kotest-intellij-plugin`: `org.jetbrains:annotations:26.1.0` (a catalog
  entry for this already exists and is used elsewhere, e.g.
  `kotest-assertions-json`)
- `kotest-intellij-plugin`: the `org.jetbrains.intellij.platform` plugin version
- `kotest-framework-plugin-gradle`: the `com.github.node-gradle.node` plugin version

This PR points all four at the catalog instead, adding entries for the two plugin versions that weren't catalogued yet.

## Why I think this is worth doing

I couldn't find a structural reason these needed to be managed separately from the rest of the catalog. There's no comment, doc, or CI logic tying them to a different lifecycle, and sibling modules already use the catalog for the same dependencies/plugins (e.g. `kotest-runner-junit5` already uses `libs.gradle.api`). It looks like a simple leftover mistake rather than something intentional.

If there is a reason to keep some of these on an independent update cadence (I could easily be missing context here, especially around the IntelliJ plugin's release process), I'd guess the better fix would still be a separate version catalog for that group rather than plain string literals. Happy to change approach if that's the case.

This seemed small enough that I skipped opening a separate issue first. Apologies if I should have opened one, and happy to do things differently for future contributions.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -29,6 +29,8 @@ classgraph = "4.8.184"
 commons-lang3 = "3.20.0"
 gradle-test-logger-plugin = "4.0.0"
 
+intellij-platform-gradle-plugin = "2.13.1"
+
 jaxbApi = "2.3.1"
 java-diff-utils = "4.16"
 jedis = "7.4.1"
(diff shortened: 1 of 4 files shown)


alfonsoristorato (approved):
> If there is a reason to keep some of these on an independent update cadence (I could easily be missing context here, especially around the IntelliJ plugin's release process), I'd guess the better fix would still be a separate version catalog for that group rather than plain string literals. Happy to change approach if that's the case.

no reason, we probably just missed these when we moved the IJ-plugin into the main repo

> This seemed small enough that I skipped opening a separate issue first. Apologies if I should have opened one, and happy to do things differently for future contributions.

no probs 😄 

thanks 👍

PR #6190, opened 2026-07-24
General improvements - framework

alfonsoristorato:
[avoid checking if log is enabled on every call and inline log methods](https://github.com/kotest/kotest/commit/f550ed5efd1591ba1c5bd9eefeb6d928fa437a5b)

made the Logger.log(...) overloads inline (no lambda allocation when logging is off) and cached isLoggingEnabled behind a by lazy instead of re-reading KOTEST_DEBUG via sysprop/env on every single call.

[cache per-spec constructor and extension lookups in the engine](https://github.com/kotest/kotest/pull/6190/commits/4129a7246e1b9d4df73461f5baf070609c63eda0)

instantiateOrObject resolved the zero-arg constructor via `kclass.constructors.find { ... }` on every call, with no cache across calls for the same class. `DefaultExtensionRegistry.get(kClass)` had the same shape: an O(n) filter + allocation over every registered extension, run twice per spec instantiation (once for constructor extensions, once for post-instantiation extensions).

Under `IsolationMode.SingleInstance` (the default) each spec class is only instantiated once, so neither of these costs repeats. But under `IsolationMode.InstancePerTest`, `InstancePerLeaf`, and `InstancePerRoot`, the engine deliberately creates a fresh instance of the same spec class per test/leaf/root to isolate mutable state between tests,
which means the same class's constructor and extension list were being re-resolved from scratch on every single test in that spec.

### DIFF
--- a/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/logger.kt
+++ b/kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/logger.kt
@@ -11,8 +11,10 @@ import kotlin.time.TimeSource
 @KotestInternal
 val start by lazy { TimeSource.Monotonic.markNow() }
 
+// Cached once per process: KOTEST_DEBUG is not expected to change mid-run, and this check
+// otherwise re-read a sysprop/env var on every single log call across the whole engine.
 @PublishedApi
-internal fun isLoggingEnabled(): Boolean = syspropOrEnv("KOTEST_DEBUG")?.uppercase() == "TRUE"
+internal val isLoggingEnabled: Boolean by lazy { syspropOrEnv("KOTEST_DEBUG")?.uppercase() == "TRUE" }
 
 @KotestInternal
 /**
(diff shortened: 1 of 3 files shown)


LeoColman (approved):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/ExtensionRegistry.kt
```diff
@@ -43,30 +44,50 @@ class DefaultExtensionRegistry : ExtensionRegistry {
 
    private val extensions = mutableListOf<Pair<Extension, KClass<*>?>>()
 
+   // get(kClass) is invoked twice per spec instantiation (once for constructor extensions, once
```
AlexCue987, line 47: presumably it doesn't need to be thread safe, right? because it is not

AlexCue987 (approved):

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/ExtensionRegistry.kt
alfonsoristorato: it does, or at least it would be good for it to be, which is what `@Volatile` is supposed to help with. This is because for every mutation we do a rebuild, so a reader of this either finds the full map, or creates it.
I thought this would be good cause normally extensions are registered on instantiation, although we allow them to be injected mid project.
The scenarios I thought of:
**Extensions registered via `ProjectListener` only**
Project starts, map is built and not touched anymore

**Extensions registered via `ProjectListener` but also in specs (different extensions I hope 😄 )**
Project starts, map is built, spec with extension starts at some point, map is built again 

**Extensions in specs only**
Same as above

I might have gotten this wrong tho, and there may be more use-cases

AlexCue987: I'm a bit skeptical, as two extensions might register at the same time and one might overwrite another's change. IOW it looks like a typical race condition. We migh want to improve it later.

alfonsoristorato: that's fair - i have not tested this yet anyway, I plan to do a manual release of this and the plugin changes to see if it actually improves things.

AlexCue987: I can easily make it thread safe, lmk if you wanf

alfonsoristorato: sure - feel free - I have built this locally and ran a few scenarios, so anything better is an improvement at this point - ty

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/core/logger.kt
```diff
@@ -11,8 +11,10 @@ import kotlin.time.TimeSource
 @KotestInternal
 val start by lazy { TimeSource.Monotonic.markNow() }
 
+// Cached once per process: KOTEST_DEBUG is not expected to change mid-run, and this check
+// otherwise re-read a sysprop/env var on every single log call across the whole engine.
 @PublishedApi
-internal fun isLoggingEnabled(): Boolean = syspropOrEnv("KOTEST_DEBUG")?.uppercase() == "TRUE"
+internal val isLoggingEnabled: Boolean by lazy { syspropOrEnv("KOTEST_DEBUG")?.uppercase() == "TRUE" }
```
sksamuel, line 17: brilliant

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/instantiate.kt
```diff
@@ -2,9 +2,18 @@
...
 import kotlin.reflect.KClass
+import kotlin.reflect.KFunction
 import kotlin.reflect.jvm.isAccessible
 
+// resolving the zero-arg constructor via kclass.constructors.find { ... } is a reflective scan.
+// Under IsolationMode.SingleInstance each spec class is only instantiated once, so this cache
+// doesn't matter there, but under InstancePerTest/InstancePerLeaf/InstancePerRoot the same spec
+// class is instantiated fresh per test/leaf/root, so without this cache the same class's
+// constructor would be re-resolved from scratch on every single test in that spec.
+private val constructorCache = ConcurrentHashMap<KClass<*>, KFunction<*>>()
```
sksamuel, line 15: love

PR #6193, opened 2026-07-25
fix(assertions): state the containExactCopies lower bound as at least 1

### LINKED ISSUE #6182: `shouldContainExactCopies`/`shouldNotContainExactCopies` throw `IllegalArgumentException` for `copies = 0` instead of evaluating the assertion
6.2.3 (also reproduces on current `master`)

**Description**

Both `io.kotest.matchers.collections.containExactCopies` (`ContainExactCopies.kt:85`) and `io.kotest.matchers.string.containExactCopies` (`containExactCopies.kt:67`) guard their `copies` parameter with:

```kotlin
require(copies > 0) { "Copies should be positive, was $copies" }
```

This runs before any pass/fail evaluation, so calling either matcher (via `shouldContainExactCopies` or `shouldNotContainExactCopies`) with `copies = 0` throws `IllegalArgumentException` instead of returning a normal `MatcherResult`.

**Repro**

```kotlin
// Expected: passes, the list contains exactly 0 copies of 4
// Actual: throws IllegalArgumentException: Copies should be positive, was 0
listOf(1, 2, 3).shouldContainExactCopies(element = 4, copies = 0)

// Expected: fails with AssertionError, the list contains 1 copy of 2, not 0
// Actual: also throws IllegalArgumentException
listOf(1, 2, 3).shouldContainExactCopies(element = 2, copies = 0)
```

**Why this looks like a bug, not intentional validation**

Elsewhere in the same matcher package, an analogous "occurrence count" concept treats zero as a perfectly valid value and only rejects genuinely impossible (negative) counts:

```kotlin
// containExactlyInAnyOrder.kt
internal data class CountMismatch<T>(val key: T, val expectedCount: Int, val actualCount: Int) {
   init {
       require(expectedCount >= 0 && actualCount >= 0) { ... }
   }
}
```

That's `>= 0`, not `> 0`. Based on that precedent, `copies = 0` might not need to be rejected as invalid input here either.

**Open question before I send a PR**

It's worth discussing whether `copies = 0` is even a case worth supporting, rather than assuming the fix is "just relax `> 0` to `>= 0`". To assert absence, `shouldNotContain(element)` is the natural call, and `shouldContain(element)` already covers presence, so nobody would typically reach for `shouldContainExactCopies(x, 0)` on purpose. Two reasonable directions:

1. **Allow `copies = 0`**, consistent with `CountMismatch`'s `>= 0`. `copies` is a plain, unrestricted `Int` that could arrive as a computed/parameterized value (loop variable, generated count) rather than a hand-typed literal, so crashing with an unrelated exception type at an ordinary `Int` boundary is a rough edge regardless of whether `0` is the idiomatic way to write the assertion by hand.
2. **Keep rejecting `copies = 0`, on purpose**, e.g. steering people toward `shouldContain`/`shouldNotContain`, in which case the fix is a clearer message/docs, not a behavior change.

Happy to send a PR either way.

mvanhorn:
## Summary

Fixes #6182.

`shouldContainExactCopies` / `shouldNotContainExactCopies` reject `copies = 0`
with `IllegalArgumentException: Copies should be positive, was 0`. Per the
discussion on the issue the rejection itself is intentional - `copies` was
designed so that zero is impossible, and a zero-count assertion belongs in
`shouldContain` / `shouldNotContain`. So this keeps the behaviour and only makes
the boundary explicit.

## Why this matters

The current message says "should be positive", which does not tell the caller
what the accepted range actually is. On the issue, `copies = 0` was read as a
matcher bug rather than an out-of-range argument, precisely because the message
describes a property instead of a bound - the reporter's first assumption was
that `listOf(1, 2, 3).shouldContainExactCopies(element = 4, copies = 0)` should
pass, since the list does contain exactly zero copies of `4`.

Nothing in the test suite pinned the boundary either: the existing specs cover
`copies = 1` and mismatched counts, so the `require` could have been loosened or
tightened without a failing test. Both halves are addressed here - the message
names the bound, and a test asserts it - so the intended contract ("the edge is
1, not 0") is documented in the code and enforced by CI rather than living only
in the issue thread.

## Changes

- `require(copies > 0)` in both
  `kotest-assertions-core/.../matchers/collections/ContainExactCopies.kt` and
  `.../matchers/string/containExactCopies.kt` now reads
  `"Copies must be at least 1, was $copies."`
- Both existing `ContainExactCopiesTest` specs gain a case asserting that exact
  message for `copies = 0`.

No behaviour change: the same inputs are accepted and rejected as before, and
the matchers' pass/fail logic is untouched.

## Testing

```
./gradlew :kotest-assertions:kotest-assertions-core:jvmTest \
  --tests "com.sksamuel.kotest.matchers.collections.ContainExactCopiesTest" \
  --tests "com.sksamuel.kotest.matchers.string.ContainExactCopiesTest"

BUILD SUCCESSFUL
collections ContainExactCopiesTest: tests=5 failures=0 errors=0 skipped=0
string      ContainExactCopiesTest: tests=6 failures=0 errors=0 skipped=0
```

Both new cases fail on `master` (the message assertion does not match) and pass
with this change.

## Notes

The issue also raised documenting the boundary. That is left out here to keep
the diff to code + test; happy to follow up with a docs pass across
`documentation/docs` and the versioned docs if you would like it folded in.

AI was used for assistance.

### DIFF
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/ContainExactCopies.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/collections/ContainExactCopies.kt
@@ -82,7 +82,7 @@ fun <T> Array<T>.shouldContainExactCopies(element: T, copies: Int): Array<T> = a
 }
 fun <T, C : Collection<T>> containExactCopies(element: T, copies: Int) = object : Matcher<C> {
    override fun test(value: C) : MatcherResult {
-      require(copies > 0) { "Copies should be positive, was $copies" }
+      require(copies > 0) { "Copies must be at least 1, was $copies." }
       val passedAtIndexes = value.mapIndexedNotNull {
             index, it -> if(it == element) index else null
       }
--- a/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/string/containExactCopies.kt
+++ b/kotest-assertions/kotest-assertions-core/src/commonMain/kotlin/io/kotest/matchers/string/containExactCopies.kt
@@ -64,7 +64,7 @@ fun String.containExactCopies(
    ) = object : Matcher<String> {
    override fun test(value: String) : MatcherResult {
       require(substring.isNotEmpty()) { "Element should not be empty" }
-      require(copies > 0) { "Copies should be positive, was $copies" }
+      require(copies > 0) { "Copies must be at least 1, was $copies." }
       val containsAtIndexes = substringFoundAtIndexes(
          value,
          substring,
--- a/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/ContainExactCopiesTest.kt
+++ b/kotest-assertions/kotest-assertions-core/src/jvmTest/kotlin/com/sksamuel/kotest/matchers/collections/ContainExactCopiesTest.kt
@@ -4,6 +4,7 @@ import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.collections.shouldContainExactCopies
 import io.kotest.matchers.collections.shouldNotContainExactCopies
+import io.kotest.matchers.shouldBe
 import io.kotest.matchers.string.shouldContain
 import io.kotest.matchers.string.shouldContainInOrder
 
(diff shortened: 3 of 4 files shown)


alfonsoristorato (approved):
Thanks 😄. Do update the docs too tho, as you mentioned. Up to you if you prefer another pr or chuck it in here.
Since this is going in 6.2.4 it's safe to update the versioned and the core of the docs

LeoColman (approved):

mvanhorn:
Appreciate the merge. Stating the containExactCopies lower bound as at least 1 should save the next person the same confusion, thanks.

Issue #6194, opened 2026-07-25
Jump To Source no longer working for root and inner tests


alfonsoristorato:
As per title

PR #6195, opened 2026-07-25
fix jump to source

alfonsoristorato:
Addresses: https://github.com/kotest/kotest/issues/6194

Problem

"Jump to Source" on test results in the IntelliJ test tree had regressed in several cases:
- Top-level (non-nested) tests, e.g. test("a test") { } directly in a spec.
- Nested tests whose literal source name has irregular whitespace (double spaces, tabs, etc.).
- Nested tests run via the Gradle test runner specifically — e.g. context("outer") { test("inner") { } } — while the top-level container navigated fine.

Root causes

1. Top-level tests bailing to the default locator (EmbeddedLocationSMTRunnerEventsAdapter.kt). EmbeddedLocationParser.parseLocationUrl treated any single-segment methodName (no /) as "must be a real JVM method" and left it to IntelliJ's default locator. But a top-level Kotest test also produces a single-segment methodName, and it isn't a real JVM method, so the default locator can't resolve it. Fixed by resolving the FQN against the stub index and only deferring to the default locator when the class genuinely isn't a Kotest spec (isKotestSpec).
2. Whitespace mismatch between engine and PSI names (EmbeddedLocationTestLocator.kt). The engine always whitespace-normalizes test names (TestNameBuilder.removeAllExtraWhitespaces()) before they reach locationUrl, but findTest's PSI-side comparison used the raw, un-normalized literal source string. Any test with irregular whitespace in source failed the exact-string match. Fixed by reusing the existing flattenTestName() normalizer (now internal, was private) on the PSI side.
3. Gradle's test events drop the nested path entirely (EmbeddedLocationTestLocator.kt). Confirmed via IDE log: IntelliJ's Gradle test integration only encodes a leaf test's own name on locationUrl (java:test://Spec/an inner test), unlike the JUnit Platform launcher's MethodSource, which joins every ancestor container into the path. A nested test therefore arrives as a single-segment path indistinguishable in shape from a top-level test — and findTest only ever searched the top-level test list for a single segment, so nested tests never resolved even though the container above them did. Fixed by searching the whole test tree (findByNameAtAnyDepth) when only one segment is available. This can't disambiguate same-named tests under different containers (Gradle's protocol has already lost that information), but resolves to some matching test instead of nothing.

image

image

image

### DIFF
--- a/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/Test.kt
+++ b/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/Test.kt
@@ -49,7 +49,7 @@ data class TestName(
    }
 }
 
-private fun String.flattenTestName() = this.trim().replace("\\s+".toRegex(), " ")
+internal fun String.flattenTestName() = this.trim().replace("\\s+".toRegex(), " ")
 // components for the path should not include prefixes
 data class TestPathEntry(val name: String)
 
--- a/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/locations/EmbeddedLocationSMTRunnerEventsAdapter.kt
+++ b/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/locations/EmbeddedLocationSMTRunnerEventsAdapter.kt
@@ -139,7 +139,7 @@ internal object EmbeddedLocationParser {
       // Only bail to the default locator when the FQN genuinely isn't a Kotest spec - Kotest's
       // own single-level tests aren't real JVM methods, so the default locator can't resolve them.
       if (!segments.contains('/') && !isKotestSpec(fqn)) return null
-      val allSegments = if (segments.contains('/')) segments.split("/") else listOf(segments)
+      val allSegments = if (segments.contains('/')) segments.split("/") else ancestorNames + segments
       val path = "$fqn${DescriptorPaths.SPEC_DELIMITER}${allSegments.joinToString(DescriptorPaths.TEST_DELIMITER)}"
       return EmbeddedLocation(path, displayName)
    }
--- a/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/locations/EmbeddedLocationSMTRunnerEventsAdapter.kt
+++ b/kotest-intellij-plugin/src/main/kotlin/io/kotest/plugin/intellij/locations/EmbeddedLocationSMTRunnerEventsAdapter.kt
@@ -8,6 +8,7 @@ import com.intellij.openapi.project.Project
 import com.intellij.psi.search.GlobalSearchScope
 import io.kotest.plugin.intellij.psi.kotestStyleSyntactic
 import org.jetbrains.kotlin.idea.stubindex.KotlinFullClassNameIndex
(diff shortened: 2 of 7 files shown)


alfonsoristorato:
now if the inner test happens to have the same name as any other inner test, only the first is found - so I am still keeping this in draft, as while this works I don't understand why it was not an issue before. So will keep this PR as a baseline until I figure it out

LeoColman:
> now if the inner test happens to have the same name as any other inner test, only the first is found

I think this is acceptable as a partial fix while you figure out the rest. I believe it's kinda corner-case to have the test and the inner test with the same name.


I suppose you can't deduplicate by test ID, can you?

LeoColman (approved):

alfonsoristorato:
> > now if the inner test happens to have the same name as any other inner test, only the first is found
> 
> I think this is acceptable as a partial fix while you figure out the rest. I believe it's kinda corner-case to have the test and the inner test with the same name.
> 
> I suppose you can't deduplicate by test ID, can you?

So `Gradle` exposes some sort of parent hierarchy, both in the form of `parent` (this does not come from gradle, but SMT, anyway) and `parentId`.
`Parent` is a reference to its own type, until null, however I have found this to be inconsistent.
`ParentId` could actually help finding the exact location of the test.
For test:
```kotlin
class Random : FreeSpec({
    "base" - {
        "inner root" - {
            "child1" {

            }
            "child2" {

            }
        }
    }

    "base 2" - {
        "inner root" - {
            "child1" {

            }
            "child2" {

            }
        }
    }
})
```
We get the following `parentId`.
```
[-1862748639:723354132] > [Test suite 'inner root'] > [Test suite 'base'] > [Test class io.github.alfonsoristorato.jpaspeckotlindsl.archunit.Random] > [Task :jpa-spec-kotlin-dsl:test]
[-1862748639:723354132] > [Test suite 'base'] > [Test class io.github.alfonsoristorato.jpaspeckotlindsl.archunit.Random] > [Task :jpa-spec-kotlin-dsl:test]
[-1862748639:723354132] > [Test class io.github.alfonsoristorato.jpaspeckotlindsl.archunit.Random] > [Task :jpa-spec-kotlin-dsl:test]
[-1862748639:723354132] > [Test suite 'inner root'] > [Test suite 'base'] > [Test class io.github.alfonsoristorato.jpaspeckotlindsl.archunit.Random] > [Task :jpa-spec-kotlin-dsl:test]
[-1862748639:723354132] > [Test suite 'inner root'] > [Test suite 'base 2'] > [Test class io.github.alfonsoristorato.jpaspeckotlindsl.archunit.Random] > [Task :jpa-spec-kotlin-dsl:test]
[-1862748639:723354132] > [Test suite 'base 2'] > [Test class io.github.alfonsoristorato.jpaspeckotlindsl.archunit.Random] > [Task :jpa-spec-kotlin-dsl:test]
[-1862748639:723354132] > [Test class io.github.alfonsoristorato.jpaspeckotlindsl.archunit.Random] > [Task :jpa-spec-kotlin-dsl:test]
[-1862748639:723354132] > [Test suite 'inner root'] > [Test suite 'base 2'] > [Test class io.github.alfonsoristorato.jpaspeckotlindsl.archunit.Random] > [Task :jpa-spec-kotlin-dsl:test]
```
`ParentId` also worked well for `child1`, but `child2` somehow only had one parent, so I will skip that and build the hierarchy from the `parentId` `String` - if I manage

PR #6197, opened 2026-07-27
make-DefaultExtensionRegistry-thread-safe (#6196)

AlexCue987:
because `Volatile` does not guarantee atomicity, `synchronized` does.


AlexCue987:
oops it only compiles on JVM

AlexCue987:
no good options for non-JVM platforms, so closing

PR #6199, opened 2026-07-28
Alex/make default extension registry thread safe

AlexCue987:
make DefaultExtensionRegistry thread safe, at least on JVM.


kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/ExtensionRegistry.kt
```diff
@@ -92,6 +108,8 @@ class DefaultExtensionRegistry : ExtensionRegistry {
 
    override fun isEmpty(): Boolean = extensions.isEmpty()
    override fun isNotEmpty(): Boolean = extensions.isNotEmpty()
+
+   private class Lock
```
AlexCue987: this is not needed, cleaning it up

AlexCue987: later

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/ExtensionRegistry.kt
```diff
@@ -57,33 +59,47 @@ class DefaultExtensionRegistry : ExtensionRegistry {
...
 
-   override fun all(): List<Extension> = extensions.map { it.first }
+   override fun all(): List<Extension> = synchronizer.synchronized {
+      extensions.map { it.first }
+   }
 
-   override fun get(kClass: KClass<*>): List<Extension> = byClass[kClass] ?: emptyList()
+   override fun get(kClass: KClass<*>): List<Extension> = synchronizer.synchronized {
+      byClass[kClass] ?: emptyList()
+   }
```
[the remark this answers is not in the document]
AlexCue987: nope

kotest-common/src/nonjvmMain/kotlin/io/kotest/common/Synchronizer.kt
```diff
@@ -0,0 +1,7 @@
+package io.kotest.common
+
+actual class Synchronizer {
+   actual fun<T> synchronized(block: () -> T): T {
+      return block()
+   }
+}
```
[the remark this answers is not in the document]
AlexCue987: no

kotest-common/src/commonMain/kotlin/io/kotest/common/Synchronizer.kt
```diff
@@ -0,0 +1,5 @@
+package io.kotest.common
+
+expect class Synchronizer() {
```
alfonsoristorato, line 3: I assume you declared this in `common` instead of `framework` because of re-usability, but if the re-usability will only ever be needed in framework, what do you think of moving it there, so it can also be made internal and does not leak into the api?

If instead you do see the value with it in `common`, how about annotating with `@KotestInternal` (which sadly requires opt-in cross module, but it's not the end of the world)

kotest-common/src/nonjvmMain/kotlin/io/kotest/common/Synchronizer.kt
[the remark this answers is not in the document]
alfonsoristorato: love the AI fight here

alfonsoristorato (commented):
really cool - little nit on it making its way into the api

kotest-common/src/commonMain/kotlin/io/kotest/common/Synchronizer.kt
AlexCue987: > I assume you declared this in `common` instead of `framework` because of re-usability, but if the re-usability will only ever be needed in framework, what do you think of moving it there, so it can also be made internal and does not leak into the api?
> 
> If instead you do see the value with it in `common`, how about annotating with `@KotestInternal` (which sadly requires opt-in cross module, but it's not the end of the world)

good call, it's internal now

sksamuel:
Take a look at 
```
val mutex = Mutex()
mutex.withLock {
}
```
Which already exists on all platforms I believe.

AlexCue987:
> Take a look at
> 
> ```
> val mutex = Mutex()
> mutex.withLock {
> }
> ```
> 
> Which already exists on all platforms I believe.

it does and I did. but it is a suspend function and as such cannot be used where we need it. am I missing something?

sksamuel:
> > Take a look at
> > ```
> > val mutex = Mutex()
> > mutex.withLock {
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Which already exists on all platforms I believe.
> 
> it does and I did. but it is a suspend function and as such cannot be used where we need it. am I missing something?

ah right

sksamuel (approved):

alfonsoristorato (approved):
before merging let's remove the `Synchronizer` from common please :)

PR #6203, opened 2026-08-02
sourceRef improvements

alfonsoristorato:
## Faster jump-to-source / failure reporting

`sourceRef()` runs once per registered test to work out the source location (used for jump-to-source and failure reports). It was walking the *entire* call stack with `Thread.currentThread().stackTrace` and resolving the enclosing spec with kotlin-reflect's `isSubclassOf`, which drags in kotlin-reflect's metadata deserialization on every single call.

Swapped that for `StackWalker` (lazy, stops at the first matching frame, gives you the `Class` directly via `RETAIN_CLASS_REFERENCE`) and plain `Class.isAssignableFrom` instead of kotlin-reflect. Same fix applied to `getDataTestCallSiteLineNumber` in the datatest module, which had the same pattern (and was even worse, doing `Class.forName` + reflection per candidate frame).

No behavior change, same `SourceRef` output, jump-to-source and failure locations still work exactly the same, just without the reflection tax.

Locally this took a synthetic 200k-test benchmark from ~23s to ~12s.


LeoColman (approved):

sksamuel:
Amazing

PR #6207, opened 2026-08-03
update docs to clarify that `copies` in shouldContainExactCopies must be 1 or higher

alfonsoristorato:
addressing this https://github.com/kotest/kotest/pull/6193#issuecomment-5120464584


alfonsoristorato:
@mvanhorn just in case you were following this up :)

LeoColman (approved):

Issue #6214, opened 2026-08-12
Intellij gutter buttons create duplicate run configs


designed4device:
Library version: 6.2.4
Plugin version: 6.1.20-2026.1.1
IntelliJ version: 2026.2.1 Build #IU-262.9437.185

Running Kotest tests from any gutter buttons causes a new, duplicate run config to be created every time. Running JUnit tests from gutter buttons does not create duplicates.

Minimum reproducer: https://github.com/designed4device/kotest-gutter-runconfig-repro

<img width="1003" height="435" alt="Image" src="https://github.com/user-attachments/assets/38034206-80eb-4c34-b322-92fc2e0ba875" />

alfonsoristorato:
Thanks for that :) I should have spotted this sooner - will get a fix in soon

alfonsoristorato:
For transparency, this happens with gradle too

alfonsoristorato:
Closing this as the latest plugin release fixes it

PR #6216, opened 2026-08-13
use a singleton strategy for getConfigurationFactory of Kotest legacy producers + ensure the same config is reused for gradle producers

alfonsoristorato:
for https://github.com/kotest/kotest/issues/6214

Running tests via the gutter icon was creating a new config every time it was run, both on Legacy runners and on Gradle runners


sksamuel (approved):
Great

Issue #6188, opened 2026-07-24
With SpecExecutionMode.Concurrent all tests run on the runner's single runBlocking thread — one blocking test freezes eventually()/withTimeout in every other test


MaciekHar:
## Which version of Kotest are you using
6.1.7 (kotest-runner-junit5, JVM), kotlinx-coroutines 1.10.2, JDK 21 (Corretto), Gradle 9.6, macOS — but the mechanism is platform-independent.

## Describe the bug

`KotestJunitPlatformTestEngine.execute` runs the whole engine inside a single `runBlocking { ... }`, and `TestSuiteScheduler.schedule` launches spec coroutines with a plain `launch { }` (no dispatcher), so with `SpecExecutionMode.Concurrent` / `TestExecutionMode.Concurrent` **all spec and test coroutines multiplex cooperatively on the one thread of that `runBlocking` event loop**.

A test body that performs consecutive blocking calls (Selenium page flows, JDBC, a blocking SDK, `Thread.sleep`) has no suspension points, so it holds that thread from start to finish. While it does, every other concurrently running test is frozen — in particular:

- resumption after `EventuallyControl.step()`'s `withContext(Dispatchers.Default) { delay(...) }` — the delay fires on time on `Dispatchers.Default`, but the continuation has to resume back on the frozen event loop;
- delivery of `withTimeout` / `withNonVirtualTimeout` cancellations (we observed a 20 s `eventually` window whose timeout fired after 200 s);
- returns from any `withContext(Dispatchers.IO)` block back into the test.

The practical result: in a suite where a few tests do minutes of blocking work, `eventually(duration = 180s, interval = 1s)` in unrelated tests performed **2 attempts instead of ~90** and failed with `Block failed after 3m 1.5s; attempted 2 time(s)` — with near-identical elapsed times across five unrelated tests, while a thread dump showed 100+ idle `DefaultDispatcher` workers. Tests pass when their spec runs alone, so this surfaces as "flaky only in full runs".

Related counter nit that misleads diagnosis: `step()` increments `iterations` *before* delaying, and the abort path in `runIterations` increments it again, so the failure message counts an attempt that never invoked the block — it reads like a "slow second poll" when the second poll never ran.

## Steps to reproduce

```kotlin
// io.kotest:kotest-runner-junit5:6.1.7
@OptIn(ExperimentalKotest::class)
class ProjectConfig : AbstractProjectConfig() {
    override val specExecutionMode = SpecExecutionMode.Concurrent
}

[...]

PR #6227, opened 2026-09-09
enable concurrent launching of tests and specs

alfonsoristorato:
For https://github.com/kotest/kotest/issues/6188


alfonsoristorato:
This works in JVM, but fails in nonJvm - still addressing it and might have to go back on a real `multiplatform` approach to handle the registry - effectively going against [this](https://github.com/kotest/kotest/pull/6199) 

UPDATE: the above should be handled by [handle concurrent modifications to the ExtensionRegistry](https://github.com/kotest/kotest/pull/6227/commits/77966ec549947de8551de8752c382e3e7461550e) - I will let the pipeline confirm or deny 😄

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/extensions/ExtensionRegistry.kt
```diff
@@ -40,76 +41,67 @@ interface ExtensionRegistry {
...
 
-   private fun rebuild() {
-      byClass = extensions.groupBy({ it.second }, { it.first })
+   private tailrec fun update(transform: (List<Pair<Extension, KClass<*>?>>) -> List<Pair<Extension, KClass<*>?>>) {
+      val current = snapshot.load()
+      val extensions = transform(current.extensions)
+      val updated = Snapshot(extensions, extensions.groupBy({ it.second }, { it.first }))
+      if (!snapshot.compareAndSet(current, updated)) {
+         update(transform)
+      }
```
alfonsoristorato, line 99: AI insisted on declaring this with a `while(true) ...` cause it would allow the method to be inline - saving some lambda allocation when called - I never liked `while(true)` and can reason with this better (also I never had a real reason to use `tailrec`, so got excited).
My excitement aside, if anyone prefers this to be `inline` with a `while(true)` let me know

alfonsoristorato:
TODO:
running native tests concurrently succeeds at kotest level, but gradle reports an error
```
./gradlew :kotest-tests:kotest-tests-blocking-concurrent-spec:macosArm64Test --rerun
```
```
FAILED: Failed to close some test reporters / succeeded(...)/skipped(...)/failed(...) must be called before close() if started(...) was called 
```
this is cause TeamCity-service-message corrupts from two tests running concurrently on separate Native threads and racing to write status messages to stdout.

sksamuel:
> TODO: running native tests concurrently succeeds at kotest level, but gradle reports an error
> 
> ```
> ./gradlew :kotest-tests:kotest-tests-blocking-concurrent-spec:macosArm64Test --rerun
> ```
> 
> ```
> FAILED: Failed to close some test reporters / succeeded(...)/skipped(...)/failed(...) must be called before close() if started(...) was called 
> ```
> 
> this is cause TeamCity-service-message corrupts from two tests running concurrently on separate Native threads and racing to write status messages to stdout.

If that's our team city service message output, then we can wrap in the engine listener that just outputs them sequentially regardless of how interleaved they are when they arrive. The JUnit module is already using it.

team city also has a flow id that's used for this, but we don't use it currently, but we could. Probably a separate PR though.

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/concurrency/engineDispatcher.jvm.kt
```diff
@@ -0,0 +1,6 @@
+package io.kotest.engine.concurrency
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+
+internal actual fun concurrentExecutionDispatcher(): CoroutineDispatcher = Dispatchers.IO
```
sksamuel: Do you think it's worth exposing a project conf option to allow users to set their own dispatcher, and only use dispatchers io if its null?

sksamuel: Also, do you think IO is the right dispatcher here and not Default? IO is used mostly for blocking calls, and we already have the blocking options on the tests which launch on a dedicated thread.

alfonsoristorato: that is a very good idea - I am still drafting this cause there are explicit tests stating `this runs on a single thread` - so I am not sure if I am going against some design decision

alfonsoristorato:
> > TODO: running native tests concurrently succeeds at kotest level, but gradle reports an error
> > ```
> > ./gradlew :kotest-tests:kotest-tests-blocking-concurrent-spec:macosArm64Test --rerun
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > FAILED: Failed to close some test reporters / succeeded(...)/skipped(...)/failed(...) must be called before close() if started(...) was called 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this is cause TeamCity-service-message corrupts from two tests running concurrently on separate Native threads and racing to write status messages to stdout.
> 
> If that's our team city service message output, then we can wrap in the engine listener that just outputs them sequentially regardless of how interleaved they are when they arrive. The JUnit module is already using it.
> 
> team city also has a flow id that's used for this, but we don't use it currently, but we could. Probably a separate PR though.

That is currently my blocker yes, and my plan was to use `flowId` too - I will tackle that in the same release cause otherwise we break stuff :)

kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/concurrency/engineDispatcher.jvm.kt
sksamuel: I don't remember deliberating wanting a single thread so I think you're good to change it.

alfonsoristorato: My coroutines understanding is very much on the simple side, I will do a quick dive in and get back on this :)

alfonsoristorato: went with `Default` after all - the kdoc itself was enough to understand where you are coming from :)

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/concurrency/engineDispatcher.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.coroutineContext
+
+/**
+ * Returns a dispatcher that can be used to launch spec/test coroutines concurrently,
+ * so that a blocking test occupies only its own thread instead of starving every other concurrently scheduled spec/test,
+ * which would otherwise all share the single-threaded event loop of the runner's top level runBlocking.
+ */
+internal expect fun concurrentExecutionDispatcher(): CoroutineDispatcher
```
alfonsoristorato, line 15: no real need for expect-actual at the moment - still I'd keep this here if we want to iterate and decide to use different dispatchers for specific targets for whatever reason

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/PinnedTestEngineListener.kt
```diff
@@ -0,0 +1,121 @@
...
+ * concurrently under TestExecutionMode.Concurrent. That's fine for consumers that track
+ * hierarchy structurally (eg JUnit Platform's EngineExecutionListener, keyed by TestDescriptor),
+ * but breaks consumers that infer hierarchy purely from message ORDER (eg TeamCity service
+ * messages, parsed by Gradle's Kotlin/Native test runner or IntelliJ's non-Gradle JVM launcher),
+ * which assume strict LIFO nesting of start/finish pairs.
+ *
+ * TODO: this currently serializes concurrent siblings in "whichever test's testStarted call wins
+ * the race to arrive first" order, not necessarily declaration order. If callers need output to
+ * always match declaration order, this needs to hold each test's notification until its declared
+ * turn rather than releasing on first-arrival.
```
alfonsoristorato, line 29: left this as a TODO - I'd be inclined in leaving things as they are for now - and potentially try and address this with `flowId` - although that, if it's the solution, will require a bigger change all around

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/PinnedTestEngineListener.kt
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * TODO: holding a slower test's node open for its whole real duration means a faster sibling's
+ * *unrelated* console output (eg from a plain [ConsoleTestEngineListener], which this class does
+ * not wrap) can still print while the slower test's node is the one open at the TC layer, and
+ * consumers that attribute stdout to "whichever TC node is currently open" will show it nested under the wrong
+ * test. Pass/fail/timing results are unaffected; this is a console-output-attribution artifact.
+ *
```
alfonsoristorato, line 36: same as above, I think I am happy to leave things as they are here, with the potential fix coming from the `flowId` implementation - if that works. If not, I think we can add another wrapper around the console and see how to play around that

kotest-framework/kotest-framework-engine/src/commonMain/kotlin/io/kotest/engine/listener/PinnedTestEngineListener.kt
```diff
@@ -0,0 +1,121 @@
...
+
+ * Note: This class is not thread safe. It is up to the caller to ensure that calls
+ * to the methods of this listener are strictly sequential, for example, by using
+ * an instance of [ThreadSafeTestEngineListener].
+ */
+@KotestInternal
+class PinnedTestEngineListener(val listener: TestEngineListener) : TestEngineListener {
```
alfonsoristorato, line 43: This is the minimum implementation I could think of, in order to leave `TeamCityTestEngineListener` alone as much as possible, so that when we decide to try and address this differently (see comments above) we simply have to refactor/remove this class

kotest-framework/kotest-framework-engine/src/jvmTest/kotlin/com/sksamuel/kotest/engine/concurrency/SpecInstancePerRootTest.kt
```diff
@@ -26,7 +27,7 @@ class SpecInstancePerRootTest : FunSpec({
 
    afterProject {
       assertSoftly {
-         externalMultipleThreadCounter.map shouldHaveSize 1
+         externalMultipleThreadCounter.map.size shouldBeLessThanOrEqual 3
```
alfonsoristorato, line 30: before we had one thread running 3 tests, now we may end up with up to 1 thread per test, but never more

kotest-framework/kotest-framework-engine/src/nativeMain/kotlin/io/kotest/engine/launcher/invokeTestEngine.native.kt
```diff
@@ -19,6 +19,14 @@ actual suspend fun invokeTestEngine(specs: List<SpecRef>, config: AbstractProjec
...
       // a test-finished event, so this TCSM listener must come after the console listener, otherwise, some console
       // output will be swallowed
+      //
+      // TODO under TestExecutionMode.Concurrent/SpecExecutionMode.Concurrent, withTeamCityListener()'s
+      // PinnedTestEngineListener (see kotest#6188) can hold one test's TC node open for its whole real
+      // duration while a faster sibling finishes underneath it. Console output printed by that faster
+      // sibling in the meantime still gets attributed to whichever TC node is currently open (this test),
+      // per the "swallowed" comment above -- so a fast test's own console progress line can visibly show
+      // up nested under a slower, still-open sibling's node in Gradle/IntelliJ's output pane. Results
+      // (pass/fail/timing) are still correct; this is a console-output-attribution artifact only.
```
alfonsoristorato, line 29: same as comments around TODO above, for the purpose of this PR I am happy with this - we can try and mitigate this with either `flowId` or a console wrapper

kotest-tests/kotest-tests-blocking-concurrent-spec/src/commonTest/kotlin/com/sksamuel/kotest/tests/concurrency/ConcurrentContainersWithNestedTestsTest.kt
```diff
@@ -0,0 +1,61 @@
...
+// than a plain root test: the container's own open/close pair must behave exactly like a root
+// test's, while its nested child opens/closes safely inside it without contending against the
+// unrelated sibling container.
+//
+// This also reproduces, one level deeper, the console-output-attribution artifact noted in
+// TODO of PinnedTestEngineListener/invokeTestEngine.native.kt: whichever container's nested test wins the
+// race holds its TC node open for its whole real duration, so the faster sibling's own console
+// completion line -- printed for real the moment it actually finishes -- visibly shows up
+// attributed to the slower, still-open sibling in Gradle/IntelliJ's per-test output pane.
+// pass/fail/timing results are unaffected.
```
alfonsoristorato, line 29: left this test here to explicitly prove that nested tests are not concurrent - if we want to make those concurrent too we can discuss it but I am happy with leaving them alone for now

kotest-tests/kotest-tests-concurrency-tests/src/jvmTest/kotlin/com/sksamuel/kotest/tests/concurrency/ConcurrentTestsSingleInstanceTest.kt
```diff
@@ -44,9 +54,11 @@ class ConcurrentTestsSingleInstanceTest : FunSpec() {
...
       // The sum all delays is 1500 ms, but tests should run concurrently.
       start.elapsedNow() shouldBeLessThan 1499.milliseconds
-      befores.shouldHaveLength(4)
-      // beforeTest should be called in declaration order
-      befores shouldBe "abcd"
+      // beforeTest's contract is that it runs before its own test, not that hooks across
+      // different tests fire in declaration order. Once tests genuinely execute on separate
+      // threads, which test's beforeTest wins the race to run first is not deterministic -- we
+      // only assert that every hook ran exactly once.
+      befores.toList().sorted().joinToString("") shouldBe "abcd"
```
alfonsoristorato, line 61: because `beforeTest` is guaranteed to run before its test, the order with which letters are recorded can fundamentally change according to which test is started (not delays involved here yet) at which time. 
We are never going to have `abcd` intentionally anymore - this is actually **important** as we are no longer promising that tests are registered sequentially - which was the case before due to the fake concurrency we assumed wasn't fake

alfonsoristorato:
@kotest/kotest-developers please have a look (if you want obvs 😄) as there are a lot of fundamental points this PR is changing - so if you have concerns please lmk (in the threads I started for each change - or start a new one if you think it's necessary).
To recap what this (plan to release as 6.3 or more) changes from before:
- concurrency (both spec level and tests within spec level) is now multithreaded
- because of the above, tests (when concurrency is enabled) may start randomly and no longer be sequential - nor will their hooks be - see here https://github.com/kotest/kotest/pull/6227#discussion_r4056751100
- Potentially fixable: when concurrent, tests (either with Junit or TC) are now going to be reported as they are registered, not in the sequence they are declared - see here https://github.com/kotest/kotest/pull/6227#discussion_r4056726670
- Potentially fixable:when concurrent, tests (either with Junit or TC) test logs might be attached to another sibling - see here https://github.com/kotest/kotest/pull/6227#discussion_r4056739103
- Potentially addressable in another PR: nested tests, even when their parents are concurrent, are not affected by this change - see here https://github.com/kotest/kotest/pull/6227#discussion_r4056741666

PR #6230, opened 2026-09-14
fix(kotest-tests-timeout-project): Format timeout in exception message

marcphilipp:
Prior to this commit, a timeout of `20m` was formatted as `1.2s seconds` (sic!).


AlexCue987:
LGTM. why are some build steps failing?

marcphilipp:
Looks like an unrelated failure of `SourceRefTest > source ref should be performant`

AlexCue987 (approved):
very nice!

PR #6233, opened 2026-09-17
Fix outdated doc reference to kotlinx-datetime proptest artifact

### LINKED ISSUE #6232: Kotlinx DateTime Gens page of the Property Testing section references the old Maven GAV
Currently, the [Kotlinx Date-Time Gens page](https://kotest.io/docs/proptest/kotlinx-datetime-gens.html) of the Property Testing section references the old Maven GAV coordinates:

Screenshot of Kotlinx Date-Time Gens page with the old artifact coordinate

This confused me when I noticed that there was no corresponding 6.2.2 version for the `io.kotest.extensions:kotest-property-datetime` artifact...

The reference page should be updated accordingly so that future readers are not confused as to which artifact exactly to use now that v6.x.x is out and the extension artifacts have been moved to the core io.kotest group

EdricChan03:
This artifact was moved to the core io.kotest group, but the doc was not updated

Renamed io.kotest.extensions to io.kotest

Fixes #6232


AlexCue987 (approved):
TY

documentation/versioned_docs/version-6.1/proptest/date_gens.md
```diff
@@ -8,10 +8,10 @@ sidebar_label: Kotlinx DateTime
 Kotest provides an optional module that provides generators for [KotlinX DateTime](https://github.com/Kotlin/kotlinx-datetime).
 
 :::note
-To use, add `io.kotest.extensions:kotest-property-datetime:version` to your build.
```
alfonsoristorato, line 11: Sorry I was too late in commenting here :) thanks for the update but this should also be applied in the same file under the 'docs' dir, otherwise any new release will still apply the wrong info.
The docs get cut from that dir and copy/pasted into new versions

AlexCue987: > Sorry I was too late in commenting here :) thanks for the update but this should also be applied in the same file under the 'docs' dir, otherwise any new release will still apply the wrong info. The docs get cut from that dir and copy/pasted into new versions

oops missed that, my bad. can you make the suggested change @EdricChan03

EdricChan03: > > Sorry I was too late in commenting here :) thanks for the update but this should also be applied in the same file under the 'docs' dir, otherwise any new release will still apply the wrong info. The docs get cut from that dir and copy/pasted into new versions
> 
> oops missed that, my bad. can you make the suggested change @EdricChan03

@AlexCue987 Thanks for the heads up, I'll take a look.

I've noticed that there are quite a few other documentation files that still reference the old `io.kotest.<whatever>` group...
~~And they also mention that they have since moved to `io.kotest` in 6.0, but the oldest version I could find for most of the moved artifacts is 6.1.0-RC2 - should this be updated as well?~~ - apologies, I've just realised that Maven Central doesn't show earlier versions in its version dropdown picker, you may disregard this

AlexCue987: > I've noticed that there are quite a few other documentation files that still reference the old `io.kotest.<whatever>` group

feel free to fix older versions as well. thanks!

